PR #111, opened 2018-12-01
feat(#98): move to AirBnb linter

Kurdty:


Kurdty:
✖ 1840 problems (1840 errors, 0 warnings)

We still have some work but it will be fixed along the way :)

PR #284, opened 2020-05-31
[GRUNT TASKS] Clean & fix all grunt tasks

Clm-Roig:
It seems to work locally but a production test might be mandatory in my opinion...

What are the changes ? Mainly using npm grunt task modules. This seems to be mandatory when using node v10+ (see this thread : https://github.com/balderdashy/sails/issues/4513)


Clm-Roig:
I don't much about JST to be honest... 
These 3 lines, from 52 to 24, are not JST ? 
https://github.com/GrottoCenter/Grottocenter3/blob/d506406d0285d01a1369055d34ae2f36e3960b9f/views/layout.ejs#L52

Clm-Roig:
Ok I saw some things related to templates I think in the Sail's error pages (in /views folder). There is also something in `/config/sails-linker.js` about lodash and templates...

bsoufflet:
Dans `layout.ejs` : 
```
    <!--
        Client-side Templates
        ========================
        HTML templates are important prerequisites of modern, rich client applications.
        To work their magic, frameworks like Backbone, Angular, Ember, and Knockout require
        that you load these templates client-side.
        By default, your Gruntfile is configured to automatically load and precompile
        client-side JST templates in your `assets/templates` folder, then
        include them here automatically (between TEMPLATES and TEMPLATES END).
        To customize this behavior to fit your needs, just edit `tasks/pipeline.js`.
        For example, here are a few things you could do:
            + Import templates from other directories
            + Use a different template engine (handlebars, jade, dust, etc.)
            + Internationalize your client-side templates using a server-side
              stringfile before they're served.
    -->

    <!--TEMPLATES-->
    
    <!--TEMPLATES END-->
```
Il n'y a rien entre les balises `TEMPLATES`, on n'utilise donc pas jst. On peut donc commenter la task jst je pense et enlever la dépendance dans le package...

Il faudrait également commenter les tasks qui utilisent `clientSideTemplates` et `clientSideTemplatesBuild` (`linkAssets.js`, `prod.js`, `linkAssetsBuildProd.js`, `linkAssetsBuild.js`).

Clm-Roig:
Ok, je m'en occupe !

Clm-Roig:
I renamed the jst.js file to jst.js.old because Grunt register all the .js files in the folder even if there are fully commented. Maybe there is a better solution, I didn't have the time to check it this morning, will see later maybe...

Clm-Roig:
I renamed the jst.js file to jst.js.old because Grunt register all the .js files in the folder even if there are fully commented. Maybe there is a better solution, I didn't have the time to check it this morning, will see later maybe...

PR #496, opened 2021-02-10
[SWAGGER API DOC] Update, clean API routes & update documentation

Clm-Roig:
- I updated all the Swagger yaml file to match the current API. I removed all the "500" error: it's obvious that any request can fail on a 500 server error.

- I also changed the order of the routes to match the routes order in *routes.js* file.

- All routes are now using the **api/v1/** path (instead of **api/**).

- Some routes are not used in the application currently and were written years ago (I think they don't work). I didn't add them to Swagger and they will be deleted in the future, in another PR.


bsoufflet (dismissed):

Clm-Roig:
J'ai oublié de MàJ les modèles de données, je pushe ça dans l'heure.

bsoufflet (approved):

Issue #500, opened 2021-02-13
[DOC VALIDATION] Hide some new columns


Clm-Roig:
De nouvelles données sont retournées par l'API à propos des documents. Dans la fonctionnalité de validation de ceux-ci, elles ne sont pas utiles. 

![image](https://user-images.githubusercontent.com/20704943/107854994-06619780-6e20-11eb-9976-16b0f6a7ca07.png)

Il faudrait donc cacher **@context**, **@id** et **@type** par défaut.

ClemRz:
@Clm-Roig could you please give a bit more context? Url of this view? Search criteria?
Also, I believe the same issue is happening in the "Manage Users" section, `/ui/admin/users`. Both the lists of administrators
 and moderators are showing `@` prefixed columns.

Could `@` prefix be a rule to omit it from rendering?

Clm-Roig:
Yes it's in /ui/admin/users. You need to be admin in order to see this page. 

By default, the table component displays all the attributes of the users. But we can specify which ones to hide with the exact attribute's name, not with a prefix or a regexp currently.

I will post the link of the concerned files here.

Clm-Roig:
See this file / line:
https://github.com/GrottoCenter/Grottocenter3/blob/7b9847516573c16c4ee06be1849454b1d1108906/assets/js/react/pages/Admin/UserList.jsx#L9

ClemRz:
Thank you @Clm-Roig 
The screenshot you attached shows "documents" not users.
Thank you for pointing the line of code, this is exactly where I was tinkering.

My question about the `@` prefix is to prevent further issues like this one. We could initialize the state with all the columns starting with `@` so we don't have to come back to this when another @ filed is added. You know what I mean?

Clm-Roig:
Oh yes sorry, the problem is the same for both users and documents. 
For documents, the URL is /ui/doc/validate I think. You can also search for a variable named **default hidden columns** or something like that.

Yes why not. I don't know if it's difficult or not to implement. The component is /common/Table and has been coded by Maxime.

ClemRz:
Thank you.
It's not difficult, I got it.

Clm-Roig:
Fixed by #507

PR #507, opened 2021-02-17
feat(cleanup): hides the columns prefixed with `@`

ClemRz:
For UserList and DocumentsTable only: if a column's ID starts with `@` it will be automatically added to the default hidden columns.
Closes issue #500


Clm-Roig (dismissed):

Clm-Roig (approved):
Ok for me! I added more columns to hide for the documents: there are more columns returned by the API recently but they are useless for this use case.

PR #529, opened 2021-04-05
[ACCOUNT API] Add change alert for news route api

Clm-Roig:
I also fixed the changeEmail route. Now, both controller methods have the same logic.


Clm-Roig:
[Spécifier si l'on veut recevoir news](https://trello.com/c/wTL4kYXD/190-sp%C3%A9cifier-si-lon-veut-recevoir-news)

bsoufflet (approved):

PR #530, opened 2021-04-05
[CAVE API] Add delete cave api route

Clm-Roig:


Clm-Roig:
[Supprimer une cavité](https://trello.com/c/S1IaZT7Q/173-supprimer-une-cavit%C3%A9)

bsoufflet (approved):

PR #598, opened 2021-06-16
fix(search): add special char to ignore

AxelCanton:
- Add french quote mark and backslash to the special characters we must ignore
- Remove the last dot of a search query, if there is any. If you add it, even if the field (title or description) possesses such dot at the end, it couldn't be found.

### DIFF
--- a/api/services/ElasticsearchService.js
+++ b/api/services/ElasticsearchService.js
@@ -282,10 +282,8 @@ const self = (module.exports = {
    */
    sanitizeQuery: (sourceString) => {
     // eslint-disable-next-line no-useless-escape
-    sourceString.replace(/[`~!@#$%^&*()_|+\-=?;:",<>«»\{\}\[\]\/\\]/gi, ' ')
-    if(sourceString[sourceString.length-1] === '.'){
-      sourceString = sourceString.slice(0, -1);
-    }
+    sourceString.replace(/[`~!@#$%^&*()_|+\-=?;:",<>«»\{\}\[\]\/\\]/gi, ' ');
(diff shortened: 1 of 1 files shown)


urien:
Super !

api/services/ElasticsearchService.js
```diff
@@ -280,7 +280,12 @@ const self = (module.exports = {
...
    * @param {*} sourceString
    */
-  sanitizeQuery: (sourceString) =>
+   sanitizeQuery: (sourceString) => {
     // eslint-disable-next-line no-useless-escape
-    sourceString.replace(/[`~!@#$%^&*()_|+\-=?;:",<>\{\}\[\]\/]/gi, ' '),
+    sourceString.replace(/[`~!@#$%^&*()_|+\-=?;:",<>«»\{\}\[\]\/\\]/gi, ' ')
+    if(sourceString[sourceString.length-1] === '.'){
+      sourceString = sourceString.slice(0, -1);
+    }
```
Clm-Roig, line 288: C'est bon pour moi, juste 2 petites améliorations de mise en forme :

```js
sourceString.replace(/[`~!@#$%^&*()_|+\-=?;:",<>«»\{\}\[\]\/\\]/gi, ' ');  // <---- le point virgule à la fin
sourceString = sourceString[sourceString.length-1] === '.' ? sourceString.slice(0, -1) : sourceString; // <---- if en une ligne

```

Clm-Roig (changes requested):

Clm-Roig (approved):

PR #621, opened 2021-10-04
[LANGUAGES API] Add Elasticsearch on languages

Clm-Roig:


Clm-Roig:
Oups ! I will PR this later, I need to tweak some things

PR #623, opened 2021-10-05
[DOC API] Return doc children

Clm-Roig:


Clm-Roig:
Oups... This PR needs another one, I will wait before opening it!

PR #645, opened 2021-11-01
[CHILD DOC] Improve document children performance

### LINKED ISSUE #644: [DOC] Getting all child documents is taking too much time
The largest Grottocenter collection has 1000 child documents. It doesn't load on the client because the server is taking way too much time to complete the request. Even with 100 child documents it can take around 10s. 

## Solution proposal

Send the documents without its children, ask them in another request. The children request can take a lot of time too, maybe it will be necessary to make it partial with some _**from**_ and _**size**_ parameters (ex: load the 0-20 children, then 20-40 etc.)

Clm-Roig:
Fix #644

Feature is not recursive anymore : just find children and grandchildren. This is enough for the main use case which is "When an user is viewing a Collection, he needs to view the issues and the articles (children and grandchildren)".

`GET /documents/:id` does not return children anymore. You need to call `GET /documents/:id/children` to get them.

### DIFF
--- a/api/controllers/DocumentController.js
+++ b/api/controllers/DocumentController.js
@@ -7,6 +7,7 @@
 
 const ramda = require('ramda');
 const getCountryISO3 = require('country-iso-2-to-3');
+const DocumentService = require('../services/DocumentService');
 
 // Tool methods
 // Set name of cave, entrance, massif, editor and library if present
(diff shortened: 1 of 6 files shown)


Clm-Roig:
C'est tout bon pour moi désormais :slightly_smiling_face:

bsoufflet (approved):

PR #816, opened 2022-04-21
813 massifs create/find

### LINKED ISSUE #813: [MASSIF] update Create/Find
- Update the create func to create massif with the geoJson data.
- Update the find func to get massif from PostGis geometry directly in geoJson

lageta:
fix #813


lageta:
This pull request is include in a new #817

Issue #1032, opened 2022-10-16
Interval format is not consistent


bsoufflet:
Interval format is not consistent between the GET command and the POST / PACTH commands for the comments object.

Indeed, currently the GET command provide the interval duration as JS object `{hours: 1, minutes: 45}` but the POST / PATCH route is waiting for an interval duration with the format `01:45:00`.

The fix is to always use the same format !

PR #1033, opened 2022-10-16
fix(#1032): make interval duration format consistent

### LINKED ISSUE #1037: [RIGGING] Update and Create routes should use the same object for obstacles  as the entrance find result
Replace the raw obstacles values (with "|!|" separator) with a obstacles object structures by lines as the one we currently have in the entrance object.
This will allow the front to not have to convert the obstacles into the DB format (with "|!|" separator) before sending the create/update request.

bsoufflet:
I try multiple solutions but I had to use this one which is not so elegant...

- I wanted to use ISO duration format for the API but it was difficult to parse in both direction the ISO format to postgre default format (hh:mm:ss)  even with the Postgre-interval node module...
- I also try to use the Postgre-interval object for both side of the API but Postgre-interval node module do not allow to convert a postgre default format to a postgreInterval object
-> So the easiest way was to use Moment Duration Format extension to be able to format the Postgre-interval object into the postgre default format (hh:mm:ss)

Also : 
- fix update route of comment to support the reset to null for not required properties like eTTrail, aestheticism, ...
- Add the reviewer to the comment response to be able to display it on the UI
- set the reviewer ID on the update route
- update the rights for comment update. Edit Own for regular users and Edit all for moderator and admin
- fix #1037


bsoufflet:
@Clm-Roig : Tu peux regarder ? Cela me permettra de deploy ensuite l'edit/create des commentaires :)

Clm-Roig (approved):

PR #1081, opened 2022-12-15
feat(cave): refactor cave find and update

vmarseguerra:
Refactor Cave find and update

Linked to https://github.com/GrottoCenter/grottocenter-front/pull/787


bsoufflet (dismissed):
Thanks for this refactoring @vmarseguerra !!
As it is sensitive it would still be better if @Clm-Roig could also review this PR before merging it ;)

api/models/TCave.js
```diff
@@ -58,13 +58,17 @@ module.exports = {
       allowNull: true,
     },
 
-    length: {
+    // Due to to an issue in sail ORM (waterline)
+    // We cannot have a property named 'length' or else
+    // the columnName resolution will not work anymore for this model
+    // https://github.com/balderdashy/sails/issues/7106#issuecomment-1341652168
+    caveLength: {
       type: 'number',
```
Clm-Roig, line 66: Si le problème vient bien de là, il faudrait supprimer le commentaire en haut du fichier TCave.js qui explique pourquoi on n'utilisait pas d'alias sur les autres attributs.

api/services/mapping/converters.js
```diff
@@ -17,44 +17,41 @@ const FileService = require('../FileService');
...
+    names: toList('names', source, c.toName),
+    descriptions: toList('descriptions', source, c.toDescription),
+    entrances: toList('entrances', source, c.toEntranceShort),
+    documents: toList('documents', source, c.toDocument),
+    massifs: toList('massifs', source, c.toMassifShort),
+  }),
 
-    return result;
-  },
   toCaveShort: (source) => ({
```
Clm-Roig, line 45: "toCaveShort" sounds a bit odd to me... Maybe I prefer "toSimpleCave", what do you think ?

api/services/mapping/converters.js
```diff
@@ -592,6 +589,11 @@ const c = {
     return result;
   },
 
+  toMassifShort: (source) => ({
```
Clm-Roig, line 592: Same as "toCaveShort" => "toSimpleMassif" ?

api/services/mapping/models/CaveModel.js
```diff
@@ -3,21 +3,4 @@ module.exports = {
...
-  histories: [],
-  id: undefined,
-  isDeleted: undefined,
-  isDiving: undefined,
-  length: undefined,
-  massifs: [],
-  name: undefined,
-  names: [],
-  reviewer: undefined,
-  temperature: undefined,
```
Clm-Roig, line 22: I understand we should do something about these models which are not very useful (because now it's pretty clear in the converters, what are the attributes used) but still : why deleting these lines ?

api/controllers/v1/cave/update.js
```diff
@@ -13,37 +12,50 @@ const { toCave } = require('../../../services/mapping/converters');
...
+        rightAction: RightService.RightActions.EDIT_ANY,
+      })
+      .intercept('rightNotFound', () =>
+        res.serverError(
+          'A server error occured when checking your right to update a cave.'
+        )
+      );
+    if (!hasRight) {
+      return res.forbidden('You are not authorized to update a cave.');
+    }
```
Clm-Roig, line 29: You don't need the `try` here: `intercept()` is syntaxic sugar for try/catch : https://sailsjs.com/documentation/reference/waterline-orm/queries/intercept

api/controllers/v1/cave/update.js
```diff
@@ -13,37 +12,50 @@ const { toCave } = require('../../../services/mapping/converters');
...
-  // Check if cave exists
-  const caveId = req.param('id');
-  const currentCave = await TCave.findOne(caveId);
-  if (!currentCave) {
-    return res.notFound({
-      message: `Cave with id ${caveId} not found.`,
-    });
-  }
+    const caveId = req.param('id');
+    const rawCave = await TCave.findOne(caveId);
```
Clm-Roig, line 32: The `try {` block should start here instead.

api/controllers/v1/cave/update.js
```diff
@@ -13,37 +12,50 @@ const { toCave } = require('../../../services/mapping/converters');
...
+      reviewer: req.token.id,
+      // dateReviewed will be updated automaticly by the SQL historisation trigger
+    };
+
+    if (newLatitude) updatedFields.latitude = newLatitude;
+    if (newLongitude) updatedFields.longitude = newLongitude;
+    if (newDepth) updatedFields.depth = newDepth;
+    if (newLength) updatedFields.caveLength = newLength;
+    if (newTemperature) updatedFields.temperature = newTemperature;
+    if (newIsDiving) updatedFields.isDiving = newIsDiving;
```
Clm-Roig, line 56: You can use the following pattern here : https://stackoverflow.com/a/40560953/16939610

```js
updatedFields = {
   ...(newLatitude && {latitude: newLatitude}),
   ...(newLongitude&& {longitude: newLongitude}
   // etc.
}
```

It's pretty much the same. I just don't like very much using if blocks (thinking in a "functional programming" way). As But as you want, I don't really mind it.

api/controllers/v1/cave/find.js
```diff
@@ -1,55 +1,64 @@
...
+  try {
+    const cave = await TCave.findOne(req.params.id)
+      .populate('descriptions')
+      .populate('documents')
+      .populate('entrances')
+      .populate('author')
+      .populate('reviewer')
+      .populate('names');
+
+    // TODO How to delete/restore entity ?
```
Clm-Roig, line 24: I don't understand this comment: why we should be able to delete/restore in the cave/find controller?

api/controllers/v1/cave/find-all.js
```diff
@@ -1,46 +1,37 @@
...
 module.exports = async (req, res) => {
-  const parameters = {};
+  const params = {
+    controllerMethod: 'CaveController.findAll',
+    searchedItem: 'all caves',
+    notFoundMessage: 'No caves found.',
+  };
+  try {
+    const parameters = {
+      // TODO Add parameters ?
```
Clm-Roig, line 14: ControllerService params are not very useful, I think: one day the `treat()` and `treatAndConvert()` methods will be totally reworked. They were supposed (I guess) to handle a lot of default cases (like errors or not found), but in practice, it's better to handle such cases in the controller. 

Anyway, add at least `searchedItem` and `controllerMethod` for debug purpose.

Clm-Roig (changes requested):

Clm-Roig:
The test for the `moveCave` feature doesn't pass here. I'm restarting the tests in case it solves the problem.

api/models/TCave.js
vmarseguerra: Modified

api/services/mapping/converters.js
vmarseguerra: Agree, to remain consistent I also renamed the other methods:
```
toCaveShort -> toSimpleCave
toCaverShort -> toSimpleCaver
toEntranceShort -> toSimpleEntrance
toMassifShort -> toSimpleMassif
```

api/services/mapping/models/CaveModel.js
vmarseguerra: Because, as you said theses properties are useless now and will be overwritten in the converter. To improve this maybe I could either delete the cave model and put the fixed properties directly in the converter or put them back for a future global models refactor ?

api/controllers/v1/cave/update.js
vmarseguerra: Modified

api/controllers/v1/cave/find.js
vmarseguerra: Like in the previous refactor on other entries it serves as a reminder that cave delete / restore is not currently implemented.

api/controllers/v1/cave/find-all.js
vmarseguerra: Agree, the Controller Service should be reworked
Here the variable parameters is used in the TCave.find() not for the ControllerService.

Clm-Roig: Oh sorry, I misread.

I think the front doesn't use this controller / route. Maybe we should add a default limit to avoid users querying too much the route ? Like 100 caves ?

Clm-Roig (approved):

api/controllers/v1/cave/find.js
Clm-Roig: Ok but why mentioning it here ? In the find code ? I don't get it 😅

api/services/mapping/models/CaveModel.js
Clm-Roig: Ok, let's remove the undefined fields.

About the JSON-LD values, I would keep them in the models to clearly see which properties are set by default without having to dive into the converters' code. And maybe rename the folder "defaultModels" instead of "models" ?

PR #1082, opened 2022-12-15
feat(log): add response status and timing for each request

bsoufflet:
also prepare version number v22.0.31 for next release

## 🤔 What

In addition to logging the HTTP request when it arrive on the API, we want to log the response status code of the API and the time spend to process the request.

## 📸 Previews

image

### DIFF
--- a/config/http.js
+++ b/config/http.js
@@ -42,6 +42,7 @@ module.exports.http = {
       'generalRateLimit',
       'userDeleteRateLimit',
       'moderatorDeleteRateLimit',
+      'responseTimeLogger',
       'requestLogger',
       'fileMiddleware',
       'bodyParser',
@@ -113,8 +114,22 @@ module.exports.http = {
 
     // Logs each request to the console
     requestLogger(req, res, next) {
-      sails.log.info('Requested ::', req.method, req.url);
+      sails.log.info('Req ::', req.method, req.url);
       return next();
+    },
+
+    // Logs each request response to the console (with status and time)
+    responseTimeLogger(req, res, next) {
+      res.on('finish', () => {
+        sails.log.info(
+          'Res ::',
+          req.method,
+          req.url,
+          res.statusCode,
+          res.get('X-Response-Time')
+        );
+      });
+      require('response-time')()(req, res, next);
     },
   },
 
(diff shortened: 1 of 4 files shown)


config/http.js
```diff
@@ -113,9 +114,23 @@ module.exports.http = {
...
+        sails.log.info(
+          'Res ::',
+          req.method,
+          req.url,
+          res.statusCode,
+          res.get('X-Response-Time')
+        );
+      });
+      require('response-time')()(req, res, next);
+    },
```
vmarseguerra, line 133: Each request will be logged twice, this will be very noisy in prod. May be only log requests once with the status code and timing ?

Also using an additional plugin for this functionality is may be overkill
What about something like this:
```suggestion
      const start = Date.now();
      res.once('finish', function reqFinish () {
          sails.log.info('Req ::', req.method, req.url, this.statusCode, `${Date.now() - start}ms`)
      })
      return next();
    },
```

vmarseguerra (changes requested):

config/http.js
bsoufflet: @vmarseguerra : It seems not good to me to only log the request once (when it is finished) for troubleshooting purpose because any API request that would make the API crash will just never be logged. And it seems that this is what we want to better understand... Which API request make the node app to crash...
In addition our current prod log could have more information that what we currently have ;)

bsoufflet: @vmarseguerra : Regarding response-time I am ok with you on using a manually build timing but a read multiple time that it was more reliable to use the response-time lib than doing it ourself. Not sure why... But it is the reason I used it...
But if you are confident with your proposal, I can go with this one.

Clm-Roig: @bsoufflet @vmarseguerra 
I agree with Benjamin, I prefer using a library for this kind of things. 

Moreover, this one is pretty small and understandable. I just took a look at its source code: it's basically what you are suggesting @vmarseguerra with more options and using `process.hrtime()` instead of `new Date()`.

Clm-Roig (approved):

config/http.js
vmarseguerra: What do you mean by API crash ? A full nodejs process crash ? Or only the crash of the request ?
From my testing it seems the error "Maximum call stack size exceeded" only crashes the request (error 500) and not the full process.

Do we know if the full node process crash append often ?
Anyway, good point, as a precaution we should leave the first request log

I already used this timing calculation in multiple other projects and never seen an issue so far. The main difference is before the timing computation was stopped when the headers were sent and now it is stopped when the last part of the body is sent

vmarseguerra: As you like

vmarseguerra (approved):

PR #1083, opened 2022-12-16
feat(entrance): handle deleted entrance in the api

### LINKED ISSUE #1075: [ENTRANCE] Handle deleted items properly
**Problem :**
Currently If I mark with is_deleted an entrance it will still be accessible the same way from the API.

In addition, Elastic Search will also continue to return it.

This issue is mostly to handle entities like Cave, Entrance, Massif, Document

For secondary entities like Description, History, Rigging, ... the solution for deleted items may be different

**Solution :** 
We need to return a minimalist entrance object when it is deleted. The goal of this object is to contain enough information to understand what was deleted, where I should go instead (redirect_to field) and potentially help a moderator later revert it if necessary.
But we should not provide information which may be potentially sensitive, as the entrance may have been deleted for the purpose of hiding this sensitive information...

> What about deleted documents?
> The astute reader may have noticed that if a document is deleted from MySQL, then that deletion will not be propagated to Elasticsearch. The following approaches may be considered to address this issue:
> MySQL records could include an "is_deleted" field, which indicates that they are no longer valid. This is known as a "soft delete". As with any other update to a record in MySQL, the "is_deleted" field will be propagated to Elasticsearch through Logstash. If this approach is implemented, then Elasticsearch and MySQL queries would need to be written so as to exclude records/documents where "is_deleted" is true. Eventually, background jobs can remove such documents from both MySQL and Elastic.
>  Another alternative is to ensure that any system that is responsible for deletion of records in MySQL must also subsequently execute a command to directly delete the corresponding documents from Elasticsearch.

Link: [How to keep Elasticsearch synchronized with a relational database using Logstash and JDBC](https://www.elastic.co/blog/how-to-keep-elasticsearch-synchronized-with-a-relational-database-using-logstash)

bsoufflet:
fix #1075

Filter the entrance object returned for deleted entrance to only return useful information to understand what was the deleted entrance and by filtering all possible sensitive information (which was potentially hiding by deleting this entrance)

Linked to : https://github.com/GrottoCenter/grottocenter-front/pull/788


api/controllers/v1/entrance/find.js
```diff
@@ -21,9 +21,8 @@ module.exports = async (req, res) => {
       .populate('names');
 
     const params = { searchedItem: `Entrance of id ${req.params.id}` };
-    // TODO How to delete/restore entity ?
-    if (!entrance || entrance.isDeleted)
-      return res.notFound(`${params.searchedItem} not found`);
+
+    if (!entrance) return res.notFound(`${params.searchedItem} not found`);
```
vmarseguerra, line 26: Is it needed to populate all child entities when the entrance is deleted ?
May be a early return could be added ?
```suggestion
if(entrance.isDeleted){
    return ControllerService.treatAndConvert(
      req,
      null,
      entrance,
      params,
      res,
      toEntrance
    );
}
```

vmarseguerra (changes requested):

api/services/mapping/converters.js
```diff
@@ -469,6 +470,26 @@ const c = {
...
+      deletedResult.dateInscription = result.dateInscription;
+      deletedResult.country = result.country;
+      deletedResult.county = result.county;
+      deletedResult.region = result.region;
+      deletedResult.city = result.city;
+      deletedResult.author = result.author;
+      deletedResult.reviewer = result.reviewer;
+      return deletedResult;
+    }
+
```
Clm-Roig, line 492: Ne vaudrait-il mieux pas créer un nouveau converter, `toDeletedEntrance` histoire de ne pas plus alourdir ce converter qui est déjà assez long ?

Clm-Roig (changes requested):

api/services/mapping/converters.js
bsoufflet: Done

api/controllers/v1/entrance/find.js
bsoufflet: Done !

bsoufflet:
@vmarseguerra ou @Clm-Roig : J'ai rebase devant les changements de @vmarseguerra et j'ai repris le swagger.
Je vous laisse valider si c'est ok ;)

Clm-Roig (approved):

PR #1085, opened 2022-12-16
Feature/polytech historisation

julien-vendran:
## 🤔 What

- Add a script to udpate h_location, add the title column
- Create a get by idLocation on table h_location 
- Create an endpoint to get all location by id from h_location 

## 🤷‍♂️ Why

To allow people to have access to a versioning of pages

## 🔍 How

Create a classical call to the database and retrieve it throught the rest api

## 🧪 Testing

No test were added for the moment. Just tested via endpoint call


config/policies.js
```diff
@@ -213,6 +213,9 @@ module.exports.policies = {
   // Swagger (API doc)
   'v1/swagger/get-yaml': true,
 
+  // Historisation
```
Clm-Roig, line 216: `History` instead of `Historisation` : use the resource name (see my other comment about english translation of french words "historique" and  "historisation")

config/routes.js
```diff
@@ -311,6 +311,9 @@ module.exports.routes = {
   'POST /api/v1/entrances/from-duplicate/:id':
     'v1/entrance-duplicate/create-from-duplicate',
 
+  // Historisation
+  'GET /api/v1/historisations/location/:id': 'v1/location/history',
```
Clm-Roig, line 315: `historisations` doesn't exist in english. Even `historization` has another meaning than the french word "historisation".

Moreover, are we speaking about a resource (un "historique") ou un process ("l'historisation") ? Generally, REST is about resources : so I would choose `history`.

About the REST route, I think it's more relevant to use `locations/:id/history`: we are getting the history of a particular location.

`historisations/location/:id`, in REST language, means :
- get all the historisations (let's say 'histories' instead, see above comment about french <-> english translation) => ok fine, I have all my histories 🟢 
- get the "location" of all the "histories" => doesn't make sense, there is no direct link between "all the histories" and "one location" 🔴 
- get the id of this "history location" => previous step doesn't have sense, so this one neither

api/controllers/v1/location/history.js
```diff
@@ -0,0 +1,30 @@
...
+    .populate('author')
+    .populate('entrance');
+
+    if (!locationsHPopulated) {
+      return res.notFound(`No history found for location ${req.params.id}`);
+    }
+
+    const params = {};
+    params.controllerMethod = 'GrottoController.findAll';
+    params.notFoundMessage = 'No location history found.';
```
Clm-Roig, line 18: You can drop the line `params.notFoundMessage` --> we don't use it any more because the case where the resource is not found is handled just before. 

Instead, just do : 
```js
const params = {
   controllerMethod: 'GrottoController.findAll'
}
```

Clm-Roig (changes requested):
Good PR ! 
I'm just bothering you with the correct naming of "un historique / historisation" because once we choose it, everything else will be named according to it : we need to be very careful about the word used 😉

config/routes.js
bsoufflet: @Clm-Roig @julien-vendran : This route is returning all the histories of a specific location ? Then if it is the case it should be `locations/:id/histories` no ?
And if you only want to get a specific history the route should be : 
`locations/:id/histories/:dateReviewed` (as dateReviewed is the second part of the primary key)

config/policies.js
bsoufflet: I agree vocabulary need to be clear from the start ;)

bsoufflet (changes requested):

config/routes.js
Clm-Roig: I used the word `history` as the english for `historique`: we don't say "les historiques de la localisation'" but "l'historique de la localisation". 

Is "histories" appropriate in english ? It is a plural form of "historique" legit in english ? Or does "history" should be only singular like "historique" in french ?

Clm-Roig: I agree for the dateReviewed, good point !

config/policies.js
Clm-Roig: I'm thinking about something else! We will have a problem when speaking about the history (subject of this PR) and a single history from the cave histories : 

- the first one is about the **data** history (cave.history => a collection of data modifications)
- the 2nd one is about the **cave** history (cave.histories which contains multiple history => a collection of "histoires" about the cave)

So do we keep "history" or do we need to use "dataHistory" instead for example ?

Clm-Roig:
We should create a guide about the PR naming. Please use something like this @julien-vendran :

`[DATA HISTORY] Add get route`

The fact you are from polytech is not needed when submitting a PR. Same for "feature" : `add get route` is more precise and we understand it's a feature.

Clm-Roig:
Un test simple sur la route serait également le bienvenu. Tu peux t'inspirer de `test\integration\4_routes\Massifs\find.test.js` par exemple

config/policies.js
julien-vendran: With @CecileDestIG, we were thinking about it. What do you think about the name `versionning` or something more related to the idea of version (Because we talk about different version of an entrance for instance) ? 
With that, we don't have problem with history and the distinction is clearer in our opinion.

config/routes.js
julien-vendran: Noted for the name of the route ! Same here, you need to validate, or not, the usage of `versioning` instead. 
About `locations/:id/histories/:dateReviewed`, in our case, we get all version of a location but we don't need details about this (We already have all information about each location). So, for the moment, we don't think that this route is usefull to us.

config/policies.js
Clm-Roig: I don't see any problems with `versioning`, it's ok for me 🙂 

I'm thinking to something else in the same "spirit" : `snapshot`. When you retrieve a past version of something, it's something frozen in the past, which will never change. So `snapshot` can be relevant too.

Juste to be precise (once again 😅): be careful between the name `version` and the process `versioning`. I think you will only need the first one for all this PR (i.e `version(s)`).

CecileDestIG: Snapshot seems to be clearer, we will use it.

urien:
Pour le nom des routes nous avons utilisé des noms au pluriel jusqu'à présent.
Est ce que versions conviendrait ?

config/policies.js
CecileDestIG: @urien we suggest to use `snapshot(s)` instead of `version(s)` as suggested @Clm-Roig, is it ok ?

urien: Oui

julien-vendran:
We made some changes according to your comments, let us know if you have any questions or changes requests

api/controllers/v1/location/get-snapshots.js
```diff
@@ -0,0 +1,54 @@
+const ControllerService = require('../../../services/ControllerService');
+const ErrorService = require('../../../services/ErrorService');
+const RightService = require('../../../services/RightService');
+
+module.exports = async (req, res) => {
```
julien-vendran, line 5: Hello @bsoufflet, here is the management of the sensitive entrance. 
Can you have a look on it and tell us if you have any comments ?

config/policies.js
bsoufflet: The "_h" SQL table will not be aligned with the new naming convention but I also prefer "version" and "snapshot" than history ;)

config/routes.js
bsoufflet: ok ! and yes use snapshots instead of histories in the route too when needed ;)

api/controllers/v1/location/get-snapshots.js
```diff
@@ -0,0 +1,54 @@
...
+        'A server error occured when checking the sensitivity of the entrance.'
+      );
+    }
+
+    const hasRight = isSensitive
+      ? await sails.helpers.checkRight
+          .with({
+            groups: req.token.groups,
+            rightEntity: RightService.RightEntities.ENTRANCE,
+            rightAction: RightService.RightActions.EDIT_ANY,
```
bsoufflet, line 28: Ce n'est pas ce right qu'il faut utiliser mais plutot cela il me semble : 
```
          rightEntity: RightService.RightEntities.ENTRANCE,
          rightAction: RightService.RightActions.VIEW_COMPLETE,
```

api/controllers/v1/location/get-snapshots.js
bsoufflet: Plutot que de faire ce check ici ne faudrait il pas mieux améliorer `remove-sensitive-entrances.js` ?
@Clm-Roig tu en penses quoi ?

bsoufflet (changes requested):

api/controllers/v1/location/get-snapshots.js
CecileDestIG: Ok we will change this

bsoufflet:
@julien-vendran @CecileDestIG Serait-il possible de régler les problèmes de prettier ?
image

api/models/HDescription.js
```diff
@@ -8,7 +8,7 @@
 module.exports = {
   tableName: 'h_description',
 
-  primaryKey: 'id',
+  primaryKey: 'dateReviewed',
```
bsoufflet, line 11: What is the reason of changing this ?
It seems that the primaryKey property is only used by waterline in "migration" -> alter mode but in our case we use the safe mode to prevent waterline to do any change automatically to our schema...
SailsJS doc discourage to change "id" with something else : 
image

And composite primary key (as it is the case for HDescription is still not supported by Waterline : https://github.com/balderdashy/sails/issues/6359

api/models/HDescription.js
```diff
@@ -38,7 +38,8 @@ module.exports = {
     },
 
     dateReviewed: {
-      type: 'ref',
+      type: 'string',
```
bsoufflet, line 41: why switching from ref to string if columnType: 'timestamp' ?

api/models/HHistory.js
```diff
@@ -0,0 +1,80 @@
...
+ * HHistory.js
+ *
+ * @description :: hHistory model
+ * @docs        :: http://sailsjs.org/#!documentation/models
+ */
+
+module.exports = {
+  tableName: 'h_history',
+
+  primaryKey: 'dateReviewed',
```
bsoufflet, line 11: same comment

api/models/HHistory.js
```diff
@@ -0,0 +1,80 @@
...
+
+  attributes: {
+    id: {
+      type: 'number',
+      autoIncrement: true,
```
bsoufflet, line 16: nop

bsoufflet: id is the history ID not an auto_increment

api/models/HHistory.js
```diff
@@ -0,0 +1,80 @@
...
+      columnName: 'id',
+      unique: true,
```
bsoufflet, line 18: nop, you can have multiple H_history with the same id. you can have multiple "snapshots" for the same history no ?

api/models/HHistory.js
```diff
@@ -0,0 +1,80 @@
...
+
+    dateInscription: {
+      type: 'ref',
+      allowNull: false,
+      columnName: 'date_inscription',
+      columnType: 'timestamp',
+    },
+
+    dateReviewed: {
+      type: 'string',
```
bsoufflet, line 41: why string ?

api/models/HLocation.js
```diff
@@ -0,0 +1,69 @@
...
+ * HLocation.js
+ *
+ * @description :: hLocation model
+ * @docs        :: http://sailsjs.org/#!documentation/models
+ */
+
+module.exports = {
+  tableName: 'h_location',
+
+  primaryKey: 'dateReviewed',
```
bsoufflet, line 11: same comment

api/models/HLocation.js
```diff
@@ -0,0 +1,69 @@
...
+
+    dateInscription: {
+      type: 'ref',
+      allowNull: false,
+      columnName: 'date_inscription',
+      columnType: 'timestamp',
+    },
+
+    dateReviewed: {
+      type: 'string',
```
bsoufflet, line 39: same comment

sql/1_2022_12_14_add_title_h_location.sql
```diff
@@ -0,0 +1,4 @@
+\c grottoce
```
bsoufflet, line 1: As this is an initial error in the model (and this table is still unused in production) could you please change the 0_tables.sql and 0_triggers.sql files instead (and delete these 2 files) ?

Can you just provide these 2 below queries (+ the missing query to fix the trigger) as Slack message to apply it manually on the prod database ?

bsoufflet (changes requested):

api/models/HDescription.js
CecileDestIG: If we put id as the primary key, as it is supposed to be unique, the .find of sails retrieves only one value from the table while there can be multiple. As it is not possible to make composite primary keys with sails, this is why we had put the other attribute of the primary key of h_**** which is dateReviewed and which is unique.

api/models/HDescription.js
CecileDestIG: We did that because we use this attribute as primary key and it must be string or number. We will check if there is other possibility.

api/models/HHistory.js
CecileDestIG: Same reason as HDescription.

api/models/HLocation.js
CecileDestIG: Same reason as HDescription.

api/models/HHistory.js
CecileDestIG: We took HDocument and HDescription as a base for HHistory and they have `autoIncrement : true` and `unique : true` but we will change this,

api/models/HDescription.js
CecileDestIG: We find a way to avoid using `primary_key:date_reviewed` by changing the model. 
`id` now represent `date_reviewed` because it is unique
`id_location` represent `id`

config/routes.js
```diff
@@ -311,6 +311,16 @@ module.exports.routes = {
   'POST /api/v1/entrances/from-duplicate/:id':
     'v1/entrance-duplicate/create-from-duplicate',
 
+  // Snapshots
+  'GET /api/v1/comments/:id/snapshots': 'v1/comment/get-snapshots',
```
bsoufflet, line 315: @julien-vendran @CecileDestIG : You need to update the Swagger file with all these new routes.
You can use https://editor.swagger.io/ to help you double check your changes on the swagger file.

bsoufflet (changes requested):

api/services/DescriptionService.js
```diff
@@ -48,11 +48,65 @@ module.exports = {
...
+      tDescription[0].t_id = tDescription[0].id;
+      tDescription[0].id = tDescription[0].dateReviewed;
+      hDescriptions.push(tDescription[0]);
+    }
+    return hDescriptions;
+  },
+
+  compareDescriptionDate(documentDate, newDescDate, oldDescDate) {
+    if (
+      documentDate.getFullYear() >= newDescDate.getFullYear() &&
```
bsoufflet, line 100: Je ne suis pas certain de comprendre mais il me semble que cela ne fonctionne pas non ?

Que se passe t'il si on a les dates suivantes : 
- 2022-12-30
- 2023-01-01

On a 2023 >= 2022 mais par contre 01 n'est pas >= à 12 ...

On bien je ne comprend pas ce que fait cette fonction...

api/services/EntranceService.js
```diff
@@ -111,6 +111,43 @@ module.exports = {
     return entranceData;
   },
 
+  getHEntrancesById: async (entranceId, isNetwork) => {
```
bsoufflet, line 114: isNetwork est peut etre à expliquer avec un commentaire ici non ?

api/services/EntranceService.js
```diff
@@ -111,6 +111,43 @@ module.exports = {
...
+        // eslint-disable-next-line no-param-reassign
+        entrance.name = name.name;
+        if (isNetwork === true) {
+          const caveName = await NameService.setName(
+            { id: entrance.id_cave },
+            'cave'
+          );
+          // eslint-disable-next-line no-param-reassign
+          entrance.caveName = caveName.name;
+          sails.log(caveName);
```
bsoufflet, line 145: log à garder ?

api/services/RiggingService.js
```diff
@@ -70,6 +70,30 @@ module.exports = {
...
+    return riggingsId;
+  },
+
+  getHRiggingById: async (riggingId) => {
+    const hRiggings = await HRigging.find({ t_id: riggingId })
+      .populate('reviewer')
+      .populate('author');
+    hRiggings.forEach((rigging) => {
+      // eslint-disable-next-line no-param-reassign
+      rigging.deserialize = module.exports.deserializeForAPI(rigging);
```
bsoufflet, line 93: Pourquoi ne pas utiliser le `converters.js` pour faire cela ?
Pourquoi ne pas utiliser la meme convention de nommage `obstacles` pour cela ?

bsoufflet (changes requested):

bsoufflet:
@julien-vendran @CecileDestIG : Serait il possible de rajouter quelques tests pour éviter à coveralls de décroitre de 0.9% ?

bsoufflet:
Il me semble y avoir un peu trop de commentaire pour pouvoir merge et deploy cette PR comme cela.
Je vous laisse revenir vers mois sur Slack dès que c'est pres pour etre merge. Je peux par exemple le faire lundi midi si vous etes rapide à corriger les commentaires de la PR ;)

api/services/DescriptionService.js
```diff
@@ -48,11 +48,65 @@ module.exports = {
...
+
   getDescription: async (descriptionId, { includeDeleted = false } = {}) =>
     TDescription.findOne({
       id: descriptionId,
       isDeleted: includeDeleted,
     })
       .populate('author')
       .populate('reviewer'),
+
+  getDescriptionsByDocument: async (documentId) => {
```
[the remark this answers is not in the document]
CecileDestIG: J'ai changé le nom pour `getHDescriptionsOfDocument`, cela vous convient-il ?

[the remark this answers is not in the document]
bsoufflet: Oui c'est mieux. Vous pouvez ajouter aussi un commentaire de fonction pour expliquer que vous retourner tous les snapshot d'un document en y incluant en plus la version actuelle.

api/services/RiggingService.js
CecileDestIG: Nous ne pouvons pas utiliser converters sinon on a une dépendance circulaire. 
Nous avons changé la fonction pour garder la même convention de nommage.

api/controllers/v1/rigging/get-snapshots.js
```diff
@@ -0,0 +1,26 @@
...
+    const riggingH = await RiggingService.getHRiggingById(req.params.id);
+
+    const params = {};
+    params.controllerMethod = 'GrottoController.getAllSnapshots';
+
+    if (Object.keys(riggingH).length === 0) {
+      return res.notFound(`Rigging ${req.params.id} has no snapshot.`);
+    }
+
+    return ControllerService.treat(
```
bsoufflet, line 16: @CecileDestIG : C'est ici que vous avec une dépendance circulaire si vous utilisez `treatAndConvert` au lieu de `treat` en utilisant un converter ?

CecileDestIG: Je ne peux pas appeler le `converters` dans `treatAndConvert` car j'envoie un tableau d'objets et que cette fonction est faite pour un seul objet il me semble. J'avais l'erreur de la dépendance circulaire en l'utilisant dans le `RiggingService`. Formatter le rigging dans le service permet de pouvoir réutiliser la fonction dans le futur pour la page d'historisation générale, c'est pour cela que j'ai conservé le formatage du rigging dans `getHRiggingById` de `getHRiggingById`.

bsoufflet: Je valide comme cela pour ne pas perdre 1j de plus mais meme si le pattern du treat et de converter n'est pas parfait il est en general mieux de respecter les conventions en place plutot que de faire un cas particulier...
Il est possible d'utiliser `treatAndConvert` avec une liste d'objets, il y a plusieurs exemples : 
```
  return ControllerService.treatAndConvert(
    req,
    null,
    adminsWithGroups,
    params,
    res,
    (data) => toListFromController('cavers', data, toCaver)
  );
```
Je vous conseille donc de regarder de nouveau le fonctionnement de `treatAndConvert` et du `converter` car il me semblerait plus propre de respecter la convention si cela est acceptable biensur ;)

api/services/DescriptionService.js
```diff
@@ -48,11 +48,62 @@ module.exports = {
...
+      // Transform TDescritpion object to keep same model as HDescription
+      tDescription[0].t_id = tDescription[0].id;
+      tDescription[0].id = tDescription[0].dateReviewed;
+      hDescriptions.push(tDescription[0]);
+    }
+    return hDescriptions;
+  },
+
+  compareDescriptionDate(documentDate, newDescDate, oldDescDate) {
+    newDescDate.setMinutes(newDescDate.getMinutes() - 2);
```
bsoufflet, line 101: ce "hack" de -2 min doit etre expliqué...

api/services/EntranceService.js
```diff
@@ -111,6 +111,43 @@ module.exports = {
     return entranceData;
   },
 
+  // If the entrance belong to a network we send different information
```
bsoufflet, line 114: pourquoi laisser la question ouverte ici...

// If the entrance do not belong to a network the associated cave is populated

api/services/RiggingService.js
```diff
@@ -70,6 +70,36 @@ module.exports = {
...
+    }
+    return riggingsId;
+  },
+
+  getHRiggingById: async (riggingId) => {
+    const hRiggings = await HRigging.find({ t_id: riggingId })
+      .populate('reviewer')
+      .populate('author');
+    hRiggings.forEach((rigging) => {
+      // eslint-disable-next-line no-param-reassign
```
bsoufflet, line 92: Plutot que de mettre le eslint-disable 4x il serait mieux d'utiliser la version "block" : 
```
/* eslint-disable no-return-assign, no-param-reassign */
( your code... )
/* eslint-enable no-return-assign, no-param-reassign */
```

bsoufflet (approved):

bsoufflet:
@julien-vendran @CecileDestIG : J'ai merge cette PR mais je vous invite à tout de meme prendre en compte mes commentaires dans votre prochaine PR de back ;)

PR #1108, opened 2023-01-16
Refactors caver endpoints

vmarseguerra:
Major changes:
- Refactor the name service to make at most only 2 requests to the DB
- Parallelize the caver/find endpoint
- Remove the unused, deprecated and dangerous caver find-all endpoint
- Only return the 10 most recent documents and exploredEntrances for a caver

The last point is open to discussion. It was mainly done to improve performance, but also because I don't think having a long list of documents and entrances really adds a lot of value to the Caver page in its current state (example: https://grottocenter.org/ui/persons/427)
What do you think ?


bsoufflet (approved):

bsoufflet:
I let you merge it when needed @vmarseguerra

urien:
Salut @vmarseguerra est ce que tu peux merger cette PR ?

PR #1155, opened 2023-03-02
improvement(login): caver email always as lowercase

### LINKED ISSUE #1154: [LOGIN] email address shouldn't be case sensitive for login
Currently, the email address of a caver which is used as credential for the login is case sensitive during the credential check.
It shouldn't be the case as email address are usually not case sensitive when used to send email... 

**proposed solution :**
convert to lower case both the email address provided during login and convert all email stored in database in lowercase.

https://ux.stackexchange.com/questions/16848/if-your-login-is-an-email-should-you-canonicalize-lower-case-it

**Describe the user story**
We received a message from Florian Wachter who cannot log in after changing his password. Here's what he does
i clicked forgot pwd
entered my email
got an email to reset
changed pwd and saved
then went bavk to home and tried to login
here i was told again, my mail or pwd are rejected

I did the procedure myself and everything worked fine. Are certain characters prohibited?

I can send Florian's email to the developer who wants to take care of this ticket

bsoufflet:
This is to :
- prevent duplicate email addresses in the t_caver table with different case
- make login easier event if the case of the email provided during login is not the same as the DB.

In addition to this code change, the following Query will need to be run in DB : 

```
update public.t_caver set mail = LOWER(mail);
```

Fix #1154

### DIFF
--- a/api/controllers/v1/account/change-email.js
+++ b/api/controllers/v1/account/change-email.js
@@ -31,7 +31,7 @@ module.exports = async (req, res) => {
     await TCaver.updateOne({
       id: req.token.id,
     }).set({
-      mail: req.param('email'),
+      mail: req.param('email').toLowerCase(),
     });
 
     return res.ok();
--- a/api/controllers/v1/account/forgot-password.js
+++ b/api/controllers/v1/account/forgot-password.js
@@ -10,9 +10,9 @@ module.exports = async (req, res) => {
   }
 
   // Get info about the user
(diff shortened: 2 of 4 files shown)


vmarseguerra (changes requested):
Looks good, thanks.
I think there is a `.toLowerCase()` change missing on sign-up when checking if the email already exists in the database ([sign-up.js:15](https://github.com/GrottoCenter/grottocenter-api/blob/develop/api/controllers/v1/auth/sign-up.js#L15))

vmarseguerra (approved):

PR #1189, opened 2023-05-19
Save and expose the latest contributions

vmarseguerra:
Save the latest contributions and add a new endpoint to retrieve them

https://github.com/GrottoCenter/grottocenter-front/issues/384

A new table is created to save the latest changes: `t_last_change`.
This new table is populated using DB triggers on the different entities.
Only the _create_ and _update_ changes are saved, the _delete_ and document validation changes are not saved.
The change for a document is only saved when it is validated

For (sub) entities like `location`, `rigging` or `history` the reference to the related main entities is also saved (entrance or document)
The Name of the entity (or related entity) is also retrieved when the change is made to conserve the name change history

The change list from the database is not returned directly, to be more concise and useful events are grouped together.
To keep the UI clean, at most the 10 recent groups are returned by the API

Each time the recent change list is retrieved, there is 5% chance to also clean changes that are older than a month

> In the production DB the new table should be created and the file `0_triggers.sql` should to be reapplied before merging.

Also fix issues discovered during testing:
- Set the reviewer during an entrance update
- Set the reviewer during an organization update
- Set the reviewer during a document update
- Refactor document multiple validation to only make one SQL update

Used by https://github.com/GrottoCenter/grottocenter-front/pull/906

Future works:
- Better handle bulk import of entities (with grouping or filtering)
- Allow to get the latest changes for only a massif or a country (https://github.com/GrottoCenter/grottocenter-front/issues/382)

### DIFF
--- a/api/controllers/v1/cave/update.js
+++ b/api/controllers/v1/cave/update.js
@@ -55,16 +55,17 @@ module.exports = async (req, res) => {
     if (newTemperature) updatedFields.temperature = newTemperature;
     if (newIsDiving) updatedFields.isDiving = newIsDiving;
 
-    await TCave.updateOne({ id: caveId }).set(updatedFields);
-
     // Handle name manually
     // Currently, use only one name per cave (even if the model can handle multiple names)
+    // Done before the TCave update so the last_change_cave DB trigger will fetch the last updated name
     await TName.updateOne({
       cave: caveId,
     }).set({
       name: req.param('name')?.text,
       language: req.param('name')?.language,
     });
+
+    await TCave.updateOne({ id: caveId }).set(updatedFields);
 
     const populatedCave = await TCave.findOne(caveId)
       .populate('author')
--- a/api/controllers/v1/change/get-recent.js
+++ b/api/controllers/v1/change/get-recent.js
@@ -1,0 +1,24 @@
+const ControllerService = require('../../../services/ControllerService');
+const RecentChangeService = require('../../../services/RecentChangeService');
+
+module.exports = async (req, res) => {
+  let offset = parseInt(req.param('offset'), 10);
+  if (Number.isNaN(offset)) offset = 0;
+  let limit = parseInt(req.param('limit'), 10);
+  if (Number.isNaN(limit)) limit = 10;
+  const start = offset >= 0 ? offset : offset - limit;
+  const end = offset >= 0 ? offset + limit : offset;
+
+  const changes = await RecentChangeService.getRecent();
+
+  return ControllerService.treat(
+    req,
+    null,
+    { changes: changes.slice(start, end) },
+    {
+      controllerMethod: 'ChangeController.getRecent',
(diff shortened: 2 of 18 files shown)


bsoufflet:
@vmarseguerra : @urien 
Not time yet to look at your code but here is some questions based on your description : 

> at most the 10 recent groups are returned by the API
Does it means that it is not possible at all the get more than the 10 latest changes from the API ? It is possible that a contributor will do a lot of changes in an area (100+) and that a moderator wants to review all of them. Could a "More" button to load 10 extra changes be used ?

> Only the create and update changes are saved, the delete and document validation changes are not saved.
Is it a technical limitation, or is it wanted to not record the deletion ? It seems important for moderators to easily know about deletions if they happened (even if the app do not allow yet to delete ;) )

> there is 5% chance to also clean changes that are older than a month
So it is like "builtin" TTL of 1 month for the last_change data ? t_last_change will be cleanup from old data by the API requests ?

urien:
Quand on réalise un import csv (entrées / documents) cela représente habituellement plusieurs centaines de lignes et souvent plusieurs milliers de documents sont créés / modifiés  en une journée

vmarseguerra:
@bsoufflet @urien:

Currently the API only returns the 10 most recent changes. This was an arbitrary choice to not overload the homepage (where changes are currently displayed).

But technically, all changes are pulled from the database and grouped together. The API could totaly be adapted to allow moderators to retrieve a longer list of changes with more details on a dedicated page. Same for deletion, DB triggers could be adapted to save them.

> So it is like "builtin" TTL of 1 month for the last_change data ?

Yes exactly

urien:
Salut @vmarseguerra, 
Effectivement s'il était possible d'avoir une liste plus longue a la demande et que dans liste on ai accès aux entités supprimées ce serait interessant

vmarseguerra:
@bsoufflet @urien 
J'ai rajouté le support pour les changements de type `delete` et `restore`
J'ai aussi ajouté les paramètres `offset` et `limit` sur l'API pour pouvoir récupérer une liste de changement plus longue

urien:
Merci @vmarseguerra.
@bsoufflet , est ce que tu souhaites jeter un oeil sur le code ou c'est ok pour toi si Vincent déploie ce qu'il a mis au point ?
A priori cela n'impacte en rien le reste de l'application

urien (dismissed):
Merci pour le boulot Vincent

bsoufflet (commented):
Tout me semble bon :)
Une seul remarque : il faudrait aussi mettre à jour le fichier swagger non ?
Tu appliqueras manuellement les changements SQL avant le deploy en prod ?

bsoufflet (dismissed):

vmarseguerra:
@bsoufflet 
J'ai mis à jour le swagger.

Oui juste avant de merge.

bsoufflet (approved):

PR #1428, opened 2025-12-16
feat(caves): adds exploring organizations in the cave response

ClemRz:
## 🤔 What

_Add exploring organizations support to cave entities_

- Added `getExploringOrganizations()` method to CaveService using ORM relationships instead of raw SQL
- Updated `getPopulatedCave()` to include exploring organizations in the populated cave data
- Modified `updateInSearch()` to properly exclude exploring organizations from search indexing
- Enhanced EntranceService to properly populate cave data with exploring organizations
- Added exploring organizations to both `toCave()` and `toSimpleCave()` converters for API responses
- Updated Swagger documentation to include the new `exploringOrganizations` field with proper schema
- Added ORM best practices documentation to PROJECT_SUMMARY.md

## 🤷‍♂️ Why

_Caves need to display information about which organizations (grottos) have explored them. This information was missing from the cave data structure and API responses, making it impossible for the frontend to show exploration history. Additionally, the codebase needed better ORM practices documentation._

## 🔍 How

_Implementing the change through ORM relationships and service layer updates:_

- **ORM Approach**: Used Waterline ORM's `.populate('exploringGrottos')` instead of raw SQL for better maintainability
- **Service Integration**: Integrated the new method into the existing `getPopulatedCave()` workflow using Promise.all for performance
- **Enhanced EntranceService**: Improved cave population in entrance context to include exploring organizations
- **API Response**: Updated both converters to include exploring organizations in cave responses
- **Search Optimization**: Excluded exploring organizations from search indexing to avoid bloating search documents
- **Documentation**: Added ORM best practices to PROJECT_SUMMARY.md
- **Code Cleanup**: Removed outdated TODO comment about exploring grottos since it's now implemented

## 🧪 Testing

_Comprehensive test coverage added for the new functionality:_

- Added unit tests for `getExploringOrganizations()` method (both with data and empty results)
- Updated existing `getPopulatedCave()` tests to verify exploring organizations are included
- Added converter tests for both `toSimpleCave()` scenarios (with and without exploring organizations)
- Updated EntranceService tests to verify cave population includes exploring organizations
- Added search update tests to ensure exploring organizations are excluded from search data

## 📸 Previews

* `GET http://127.0.0.1:1337/api/v1/entrances/5`
* `GET http://127.0.0.1:1337/api/v1/caves/5`

```json
{
  ...
  "cave": {
    ...
    "exploringOrganizations": [
      {
        "id": 2,
        "name": "CLUB 22179",
        "language": "fra",
        "isDeleted": false
      },
      {
        "id": 1,
        "name": "CLUB 773",
        "language": "fra",
        "isDeleted": false
      }
    ]
  }
  ...
}
```

## 📚 Documentation Updates

Added ORM best practices to PROJECT_SUMMARY.md:
- Prefer ORM over raw SQL queries
- Leverage model relationships with `.populate()`
- Performance optimization guidelines for data fetching

### DIFF
--- a/api/services/CaveService.js
+++ b/api/services/CaveService.js
@@ -150,6 +150,30 @@ module.exports = {
     }
   },
 
+  /**
+   * Get organizations that explored the cave
+   * @param {number} caveId
+   * @returns {Promise<Array>} Array of organizations with id and name
+   */
+  getExploringOrganizations: async (caveId) => {
+    const query = `
+      SELECT g.*
+      FROM t_grotto g
+             JOIN j_grotto_cave_explorer j ON g.id = j.id_grotto
+      WHERE j.id_cave = $1
+        AND g.is_deleted = false
+    `;
+    const result = await CommonService.query(query, [caveId]);
+    const grottos = result.rows;
+
+    if (!grottos || grottos.length === 0) {
+      return [];
+    }
+
+    await NameService.setNames(grottos, 'grotto');
+    return grottos;
+  },
+
   async getPopulatedCave(caveId, subEntitiesWhere = {}) {
     const cave = await TCave.findOne(caveId)
       .populate('author')
@@ -167,6 +191,9 @@ module.exports = {
       DocumentService.getDocuments(cave.documents?.map((d) => d.id) ?? []),
     ]);
 
+    cave.exploringOrganizations =
+      await module.exports.getExploringOrganizations(cave.id);
+
     const nameAsyncArr = [
       NameService.setNames(cave?.entrances, 'entrance'),
       NameService.setNames(cave?.massifs, 'massif'),
@@ -181,7 +208,6 @@ module.exports = {
     // - histories
     // - riggings
     // - comments
-    // - exploringGrottos
     // - partneringGrottos
 
     return cave;
--- a/api/services/EntranceService.js
+++ b/api/services/EntranceService.js
@@ -396,9 +396,14 @@ module.exports = {
     if (!entrance) return null;
 
     if (entrance.cave) {
-      [entrance.cave.massifs, entrance.cave.entrances] = await Promise.all([
+      [
+        entrance.cave.massifs,
+        entrance.cave.entrances,
+        entrance.cave.exploringOrganizations,
+      ] = await Promise.all([
         CaveService.getMassifs(entrance.cave.id),
         TEntrance.find({ cave: entrance.cave.id }),
+        CaveService.getExploringOrganizations(entrance.cave.id),
       ]);
       await Promise.all([
         NameService.setNames(entrance.cave.massifs, 'massif'),
--- a/api/services/mapping/converters.js
+++ b/api/services/mapping/converters.js
@@ -45,6 +45,11 @@ const c = {
     entrances: toList('entrances', source, c.toSimpleEntrance, { meta }),
     massifs: toList('massifs', source, c.toSimpleMassif),
     documents: toList('documents', source, c.toSimpleDocument),
+    exploringOrganizations: toList(
+      'exploringOrganizations',
+      source,
+      c.toSimpleOrganization
+    ),
   }),
 
   toSimpleCave: (source) => ({
@@ -56,6 +61,11 @@ const c = {
     temperature: source.temperature,
     isDiving: source.isDiving,
     entrances: source.entrances?.map((e) => e.id),
+    exploringOrganizations: toList(
+      'exploringOrganizations',
+      source,
+      c.toSimpleOrganization
+    ),
   }),
 
   toCaver: (source) => {
--- a/PROJECT_SUMMARY.md
+++ b/PROJECT_SUMMARY.md
@@ -247,6 +247,11 @@ npm test -- --grep "Auth"  # Specific test patterns
 - **Models**: Data validation, relationships, lifecycle callbacks
 - **Converters**: Transform database models to API responses
 
+### ORM & Database Best Practices
+- **Prefer ORM over Raw SQL**: Use Waterline ORM methods instead of `CommonService.query()` when possible
+- **Leverage Model Relationships**: Use `.populate()` with defined associations rather than manual joins
+- **Performance Optimization**: Only fetch required data - avoid full population when using converter functions like `toSimpleCave()`
+
 ### Naming Conventions
 ```javascript
 // Database tables
--- a/assets/swaggerV1.yaml
+++ b/assets/swaggerV1.yaml
@@ -4678,6 +4678,25 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/Entrance'
+        exploringOrganizations:
+          type: array
+          items:
+            type: object
+            properties:
+              id:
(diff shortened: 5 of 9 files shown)


api/services/mapping/converters.js
```diff
@@ -56,6 +56,7 @@ const c = {
     temperature: source.temperature,
     isDiving: source.isDiving,
     entrances: source.entrances?.map((e) => e.id),
+    exploringOrganizations: source.exploringOrganizations,
```
vmarseguerra: Not needed in `toSimpleCave()` should be added in `toCave`  with proper formating.

api/services/CaveService.js
```diff
@@ -150,6 +150,24 @@ module.exports = {
     }
   },
 
+  /**
+   * Get organizations that explored the cave
+   * @param {number} caveId
+   * @returns {Promise<Array>} Array of organizations with id and name
+   */
+  getExploringOrganizations: async (caveId) => {
```
vmarseguerra, line 158: Why not using the ORM for this query ?

api/services/EntranceService.js
```diff
@@ -396,14 +396,7 @@ module.exports = {
...
     if (entrance.cave) {
-      [entrance.cave.massifs, entrance.cave.entrances] = await Promise.all([
-        CaveService.getMassifs(entrance.cave.id),
-        TEntrance.find({ cave: entrance.cave.id }),
-      ]);
-      await Promise.all([
-        NameService.setNames(entrance.cave.massifs, 'massif'),
-        NameService.setNames([entrance.cave], 'cave'),
-      ]);
+      entrance.cave = await CaveService.getPopulatedCave(entrance.cave.id);
```
vmarseguerra: Cave here is a sub-entity of the entrance; fetching the fully populated cave is not required. In fact `toSimpleCave()` is used so I think even the `CaveService.getMassifs` can also be removed

vmarseguerra (changes requested):

api/services/CaveService.js
ClemRz: You're right

api/services/mapping/converters.js
ClemRz: I disagree. We want the exploring orgs as part of the `GET /api/v1/entrances/:id` response.

ClemRz: Now, I think it should **also** be added in `toCave` so the caves page can display them as well.

api/services/EntranceService.js
ClemRz: Good catch

api/services/CaveService.js
```diff
@@ -150,6 +150,19 @@ module.exports = {
...
+    const cave = await TCave.findOne(caveId).populate('exploringGrottos');
```
vmarseguerra: This is works, but it fetches the cave for nothing, maybe it is better to do it the other way around, by fetching all grottos that are associated with the cave ID?

api/services/CaveService.js
```diff
@@ -161,10 +174,16 @@ module.exports = {
 
     if (!cave) return null;
 
-    [cave.massifs, cave.descriptions, cave.documents] = await Promise.all([
+    [
+      cave.massifs,
+      cave.descriptions,
+      cave.documents,
+      cave.exploringOrganizations,
```
vmarseguerra: Instead of fetching `exploringOrganizations` here, adding a `.populate('entrances')` above with a `NameService.setNames()` should also work and it saves a request to the DB

api/services/CaveService.js
```diff
@@ -192,7 +210,7 @@ module.exports = {
   },
 
   async updateInSearch(populatedCave) {
-    const { names, ...c } = populatedCave;
+    const { names, exploringOrganizations, ...c } = populatedCave;
```
vmarseguerra: Not needed, as there is no spread in the `cave` object below

api/services/EntranceService.js
```diff
@@ -396,14 +396,22 @@ module.exports = {
...
     if (entrance.cave) {
-      [entrance.cave.massifs, entrance.cave.entrances] = await Promise.all([
-        CaveService.getMassifs(entrance.cave.id),
-        TEntrance.find({ cave: entrance.cave.id }),
-      ]);
-      await Promise.all([
-        NameService.setNames(entrance.cave.massifs, 'massif'),
-        NameService.setNames([entrance.cave], 'cave'),
-      ]);
+      const cave = await TCave.findOne(entrance.cave.id)
```
vmarseguerra: Why fetching the `cave` again ? 
It was already fetched with the `.populate('cave')` and `author` and `reviewer` won't be retained in `toSimpleCave()`

vmarseguerra (changes requested):

api/services/CaveService.js
ClemRz: This is a tricky one, Sails doesn't allow filtering by plural associations because of the lack of `id` on the junction model. The junction model has a composite key making it difficult.

ClemRz: So, it's either that inefficient Sails/ORM usage, or the native SQL query I had originally. Unless you have a better solution, in which case please post it here.

api/services/CaveService.js
ClemRz: Got it.

api/services/CaveService.js
ClemRz: Got it.

api/services/EntranceService.js
ClemRz: Because: shenanigans 😆

Let me fix that.

vmarseguerra (dismissed):

api/services/CaveService.js
ClemRz: What if, instead of expecting a cave ID in parameter, we expect a cave object? This way we wouldn't need to "fetch the cave for nothing"?

vmarseguerra: But the only time this function is used in the `EntranceService` there is only the `caveId`

The original proposition with the native SQL query was better in the end, I think. it does not fetch the `cave` and already has the join for the name

ClemRz: The cave's exploring organizations list is already available where `getExploringOrganizations` gets called, so we can save the cave fetching part. I made the change and added a test to cover it. It works. IDK if this is a pattern we want to adopt though (mixed type argument).

api/services/CaveService.js
```diff
@@ -150,14 +150,39 @@ module.exports = {
...
   async getPopulatedCave(caveId, subEntitiesWhere = {}) {
     const cave = await TCave.findOne(caveId)
       .populate('author')
       .populate('reviewer')
       .populate('names')
       .populate('descriptions')
       .populate('entrances')
-      .populate('documents');
+      .populate('documents')
+      .populate('exploringGrottos');
```
vmarseguerra: `.populate('exploringGrottos');`  is not needed when using `getExploringOrganizations(cave.id)` below

vmarseguerra (dismissed):

PR #1448, opened 2026-02-10
chore(deps): update dependencies to latest versions

ClemRz:
## 🤔 What

Update package dependencies to their latest versions:

- `@commitlint/config-conventional`: ^20.2.0 → ^20.4.1
- `ejs`: ^3.1.10 → ^4.0.1 (major version)
- `prettier`: ^3.7.4 → ^3.8.1
- `typesense`: ^2.1.0 → ^3.0.1 (major version)

Removed `@babel/runtime` (only needed as a transitive dependency).

Added `express` as dev dependency (required by rate limiter tests).

Fixed security vulnerabilities by adding overrides for `qs` (^6.14.1) and `diff` (^8.0.3).

Added `.npmrc` with `legacy-peer-deps=true` to handle peer dependency conflicts.

## 🤷‍♂️ Why

Keep dependencies up-to-date for security patches, bug fixes, and new features. Remove unused direct dependency (`@babel/runtime`). Fix high severity vulnerabilities in `qs` and `diff`.

## 🔍 How

- Ran `npx npm-check-updates -u` to update package.json
- Kept ESLint at v8 with existing `.eslintrc.js` config to maintain compatibility with `eslint-config-airbnb-base`
- Removed `@babel/runtime` as it's only needed as a transitive dependency of typesense
- Added `.npmrc` to configure npm to use legacy peer deps
- Added overrides for `qs` and `diff` to fix security vulnerabilities
- Added `express` as dev dependency since it's directly imported in rate limiter tests
- Ran `npm install` to update package-lock.json

## 🧪 Testing

- Build workflow will verify all tests pass
- Typesense v3 should be backward compatible with existing usage
- EJS v4 is a minor update with no breaking changes expected
- All security vulnerabilities resolved (0 vulnerabilities)

## 📸 Previews

N/A - dependency updates only

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+legacy-peer-deps=true
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,12 +11,11 @@
       "dependencies": {
         "@aws-sdk/client-ses": "^3.986.0",
         "@azure/storage-blob": "^12.30.0",
-        "@babel/runtime": "^7.28.4",
         "@natlibfi/marc-record": "^10.0.1",
         "@natlibfi/marc-record-serializers": "^11.0.1",
         "archiver": "^7.0.1",
         "argon2": "^0.44.0",
-        "ejs": "^3.1.10",
+        "ejs": "^4.0.1",
         "express-rate-limit": "^8.2.1",
         "jsonwebtoken": "^9.0.3",
         "mime-types": "^3.0.2",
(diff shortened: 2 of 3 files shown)


vmarseguerra (commented):
Why removing all eslint-disable comments? 

Their rules should still be enforced but it does not seem to be the case now

bsoufflet (approved):

PR #1460, opened 2026-02-12
fix(name): align h_name.name column to varchar(200) and add name length validation

ClemRz:
## 🤔 What

- Fix server crash (500) when updating an entrance/cave/massif name longer than 100 characters

- Add input validation for name length across all name create/update endpoints

- Add SQL migration to align `h_name.name` column with `t_name.name` (`varchar(100)` → `varchar(200)`)

## 🤷‍♂️ Why

Updating entrance 89682 with the name `"Grande Fontaine (?) [Trou Bleu] [Source de l'Ornel] [Exsurgence de l'Ornel] [Grande Fontaine de la Lonne]"` (101 characters) crashed the server with an unhandled `AdapterError: value too long for type character varying(100)`.

The root cause is a schema mismatch: `t_name.name` is `varchar(200)` but the history table `h_name.name` was `varchar(100)`. The PostgreSQL trigger `histo_update_name()` copies the old name into `h_name` before any update, and that insert fails for names exceeding 100 characters.

## 🔍 How

Two-part fix:

1. **SQL migration** (`sql/8_2026_02_12_fix_h_name_varchar.sql`): ALTERs `h_name.name` from `varchar(100)` to `varchar(200)` to match `t_name.name`. Also updated `sql/0_tables.sql` so new deployments get the correct schema from the start.

2. **Shared validation utility** (`api/utils/nameValidation.js`): Exports `validateNameLength` and `NAME_MAX_LENGTH` (200) so the limit is defined in one place.

3. **Input validation** in controllers that write to `TName`:

   - `api/controllers/v1/entrance/update.js`

   - `api/controllers/v1/entrance/create.js`

   - `api/controllers/v1/cave/update.js`

   - `api/controllers/v1/cave/create.js`

   - `api/controllers/v1/name/update.js`

   - `api/controllers/v1/massif/create.js`

   - `api/controllers/v1/organization/create.js`

   Each now calls `validateNameLength` before hitting the DB and returns a `400 Bad Request` with a clear message if the name exceeds 200 characters.

## 🧪 Testing

1. Reset local db: `npm run dev:clean`

2. Start the dev server and attempt to update an entrance with a name longer than 100 characters — it should now succeed.

3. Attempt to update with a name longer than 200 characters — should return a 400 error with a descriptive message.

4. Run existing tests: `npm run test`

## 📸 Previews

_N/A — backend-only change, no UI impact._

### DIFF
--- a/api/controllers/v1/massif/create.js
+++ b/api/controllers/v1/massif/create.js
@@ -3,6 +3,7 @@ const MassifService = require('../../../services/MassifService');
 const NotificationService = require('../../../services/NotificationService');
 const RecentChangeService = require('../../../services/RecentChangeService');
 const { toMassif } = require('../../../services/mapping/converters');
+const { validateNameLength } = require('../../../utils/nameValidation');
 
 // eslint-disable-next-line consistent-return
 module.exports = async (req, res) => {
@@ -25,6 +26,12 @@ module.exports = async (req, res) => {
   }
   if (missingParamaters.length > 0) {
     return res.badRequest(`${missingParamaters} parameter(s) must be provided`);
+  }
+
+  // Validate name length
+  const nameError = validateNameLength(req.body.name);
+  if (nameError) {
+    return res.badRequest(nameError);
   }
 
   // Launch creation request using transaction: it performs a rollback if an error occurs
--- a/api/controllers/v1/cave/create.js
+++ b/api/controllers/v1/cave/create.js
@@ -7,6 +7,7 @@
 const CaveService = require('../../../services/CaveService');
 const ControllerService = require('../../../services/ControllerService');
 const { toCave } = require('../../../services/mapping/converters');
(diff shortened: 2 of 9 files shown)


api/controllers/v1/massif/create.js
```diff
@@ -27,6 +28,13 @@ module.exports = async (req, res) => {
     return res.badRequest(`${missingParamaters} parameter(s) must be provided`);
   }
 
+  // Validate name length
+  const nameValue = req.param('name');
```
vmarseguerra: Use the same name parameter (`req.body.name`) than the one used in `TName.create` below

vmarseguerra (changes requested):
Missing TName validation for
- Organization create
- Cave update

vmarseguerra (changes requested):
Missing cave create TName validation

vmarseguerra (approved):

PR #1464, opened 2026-02-13
fix(search): handle nested objects in CSV export to prevent [object Object] output

ClemRz:
## 🤔 What

- Fix CSV export displaying `[object Object]` for Author, Country/Region, and other nested fields.
- Make `resolveField` robust against both parent keys (`"authors"`) and dot-notation keys (`"authors.nickname"`).
- Add `toPrimitive` helper to gracefully coerce nested objects/arrays into CSV-safe strings.
- Add test coverage for nested object handling in CSV export.

## 🤷‍♂️ Why

A user reported that CSV exports for document searches display `[object Object]` instead of actual values for "Author" and "Country/Region" fields.

Root cause: the front-end sends parent object keys (e.g. `"authors"`, `"iso3166"`) in the `columns` array, not the Typesense nested field paths (`"authors.nickname"`, `"iso3166.iso"`). Typesense returns these as nested objects, and `String()` coercion produces `[object Object]`.

A previous fix (commit `35424e4`) added a `resolveField` function for dot-notation traversal, but it didn't help because the front-end never sends dot-notation keys for these fields.

## 🔍 How

- Removed the `hasOwnProperty` early-return in `resolveField` for dot-notation keys — always traverse the nested structure to avoid returning raw objects if Typesense ever includes flat keys with object values.
- Added `toPrimitive(val)` as a safety net for simple (non-dot) keys:
  - Primitives pass through unchanged.
  - Arrays of primitives are joined with `"; "`.
  - Plain objects: picks the first string-valued property as display value.
  - Arrays of objects: recursively applies the above, then joins.
- This means `"authors"` → `"PANOS Vladimir; DUPONT Jean"` (extracts `nickname`, the first string property) even without the front-end fix.

A companion `FRONTEND_FIX.md` documents the front-end change needed: replace `"authors"` → `"authors.nickname"` and `"iso3166"` → `"iso3166.iso"` in the export columns payload.

## 🧪 Testing

Two new test cases in `AdvancedSearchExport.test.js`:

1. Parent keys as columns — verifies `toPrimitive` flattens `authors`, `iso3166`, and `editor` objects into readable strings.
2. Dot-notation keys — verifies traversal still works for `authors.nickname` and `editor.name`.

Manual verification against production:

```bash
curl -s -X POST 'https://api.grottocenter.org/api/v1/advanced-search/export' \
  -d '{"query":"","entity":"documents","filter":{"datePublication":"2025"},"matchAllFields":true,"columns":["id","title","authors","iso3166"],"columnsName":["ID","Title","Authors","Country / Region"]}'
```

Should output author nicknames and ISO codes, not `[object Object]`.

## 📸 Previews

### DIFF
--- a/api/controllers/v1/search/advanced-search-export.js
+++ b/api/controllers/v1/search/advanced-search-export.js
@@ -8,31 +8,59 @@ function escapeCSV(v) {
 }
 
 /**
+ * Convert a value to a CSV-safe primitive.
+ * - primitives pass through
+ * - arrays of primitives are joined with "; "
+ * - plain objects are stringified by picking the first string-valued field
+ * - arrays of objects: extract the first string field from each, then join
+ */
+function toPrimitive(val) {
+  if (val == null) return val;
+  if (typeof val !== 'object') return val;
+
+  if (Array.isArray(val)) {
+    const flat = val.map((item) => toPrimitive(item)).filter((v) => v != null);
+    return flat.join('; ');
+  }
+
+  // Plain object — pick the first string-valued property as a display value
+  const strVal = Object.values(val).find((v) => typeof v === 'string');
+  return strVal ?? JSON.stringify(val);
+}
+
+/**
  * Resolve a dot-notation key from a Typesense document.
  * Typesense with enable_nested_fields returns nested structures,
  * e.g. "authors.nickname" is stored as authors: [{ nickname: "X" }].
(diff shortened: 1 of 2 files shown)


ClemRz:
I'm gonna merge right away. @bsoufflet & @vmarseguerra feel free to comment even after the fact if there is anything concerning to you :)

PR #1465, opened 2026-02-13
fix(search): increase max_filter_by_candidates for prefix date filters

ClemRz:
## 🤔 What

- Fix document search by publication date returning incomplete results when using prefix filters (e.g. searching for year "2025" only returned a fraction of matching documents).
- Set `max_filter_by_candidates: 100` on all Typesense search requests that involve a prefix filter.

## 🤷‍♂️ Why

Typesense's `max_filter_by_candidates` parameter defaults to **4**, meaning it only considers the top 4 distinct prefix matches when evaluating a `filter_by` with a wildcard. For `datePublication:=2025*`, there are 13+ distinct matching values (`"2025"`, `"2025-01"`, `"2025-02"`, …, `"2025-12"`, plus full dates like `"2025-01-15"`). The default cap of 4 caused Typesense to silently ignore the rest, returning only a subset of documents.

This is why searching for "Le P'tit Usania" issues published in 2025 returned 3 instead of 12.

Ref: [Typesense docs — `max_filter_by_candidates`](https://typesense.org/docs/30.1/api/search.html)

## 🔍 How

- `buildFilter()` in `SearchService.js` now returns `{ filterBy, hasPrefixFilter }` instead of a plain string, tracking whether a prefix wildcard filter was generated.
- All three search functions (`collectionSearch`, `multiCollectionsSearch`, `fieldSearch`) conditionally add `max_filter_by_candidates: 100` to the Typesense query params when a prefix filter is present.
- No changes to the filter syntax itself — `datePublication:=2025*` was already correct.

### Files changed

- `api/services/SearchService.js` — `buildFilter` returns an object; all callers pass `max_filter_by_candidates` when needed.
- `test/integration/1_services/SearchService.test.js` — added 3 test cases for `max_filter_by_candidates` behavior.

## 🧪 Testing

```bash
npx mocha test/integration/1_services/SearchService.test.js --timeout 10000
```

To verify in production: search for documents with type "Issue" and datePublication "2025" — all 12 monthly "Le P'tit Usania" issues should now appear.

## 📸 Previews

### DIFF
--- a/api/services/SearchService.js
+++ b/api/services/SearchService.js
@@ -18,6 +18,7 @@ const allEntities = {
 const allEntitiesKeys = Object.keys(allEntities);
 
 function buildFilter(filter, isLogicalCompareAnd = true) {
+  let hasPrefixFilter = false;
   const out = Object.entries(filter)
     .filter(([, v]) => v)
     .flatMap(([k, v]) => {
@@ -26,6 +27,7 @@ function buildFilter(filter, isLogicalCompareAnd = true) {
       // e.g. "2025" matches "2025", "2025-01", "2025-01-15", etc.
       // Typesense supports prefix matching on string fields with := and *.
       if (k === 'datePublication' && typeof v === 'string') {
+        hasPrefixFilter = true;
         return [`${k}:=${v}*`];
       }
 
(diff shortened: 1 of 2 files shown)


ClemRz:
I'm gonna merge right away. @bsoufflet & @vmarseguerra feel free to comment even after the fact if there is anything concerning to you :)

PR #1466, opened 2026-02-14
feat(relevanceOrdering): auto-assign relevance on entity creation and add move-relevance endpoints

### LINKED ISSUE #574: [RELEVANCE] Handle relevance in API
In some entities of Grottocenter (Location, Description, Comment & Rigging), a **relevance** attribute is used to indicate the order in which the elements should be displayed. Currently, this is not handled by the API or the client. 

At the creation of one of these entities, we should set the relevance to **maxRelevance (of the table) + 1**.
Then, a route allowing the users to move the items up or down (i.e. swap the relevance between 2 entities) must be created.

ClemRz:
Closes #574

## 🤔 What

- Auto-assign relevance on creation of Location, Description, Comment, Rigging, and History entities (`MAX(relevance) + 1` within the parent scope, or `1` if empty)
- New `PATCH /api/v1/<entityType>/:id/move-relevance` endpoints for all five entity types, allowing users to swap an entity's position up (`-1`) or down (`+1`) with its adjacent neighbor
- SQL migration script to normalize existing relevance values to sequential positive integers (1, 2, 3, ...) per scope, fixing duplicates, negatives, and gaps in the current data
- Swagger documentation for all five new endpoints

## 🤷‍♂️ Why

The `relevance` attribute on Location, Description, Comment, Rigging, and History controls display ordering, but it was never managed by the API:
- On creation, relevance was left unset (there were `// TODO compute relevance` placeholders in the create controllers)
- There was no way for users to reorder entities
- Existing data had duplicate relevance values within the same scope, negative values, and gaps — making any future ordering logic unreliable

This change makes relevance a fully managed, consistent attribute so the front-end can implement drag-and-drop or up/down reordering.

## 🔍 How

- A shared `RelevanceService` (`api/services/RelevanceService.js`) handles all logic for the five entity types via an `ENTITY_CONFIG` map, avoiding code duplication:
  - `computeNextRelevance(entityType, parentFieldValues)` — queries `MAX(relevance)` within the scoped parent
  - `moveRelevance(entityType, entityId, direction)` — finds the adjacent neighbor and swaps their relevance values
  - `getParentScope(entityType, entity)` — determines the parent scope by inspecting which parent field is set on the entity
- The five existing create controllers (`location/create.js`, `description/create.js`, `comment/create.js`, `rigging/create.js`, `history/create.js`) now call `computeNextRelevance` before creating the entity
- Five new `move-relevance.js` controllers are thin wrappers that call `moveRelevance` and map service errors to HTTP responses (404, 400)
- Routes added in `config/routes.js`, protected by `tokenAuth` in `config/policies.js`
- Historisation entries (H-tables) are handled automatically by the existing PostgreSQL triggers — no application code needed
- `sql/6_04_2026_02_13_normalize-relevance.sql` uses `ROW_NUMBER() OVER (PARTITION BY <parent_fields> ORDER BY relevance, id)` CTEs to renumber each scope sequentially, and sets deleted entities' relevance to 0. The script is idempotent and wrapped in a transaction.

## 🧪 Testing

- Property-based tests (fast-check, 100 iterations each):
  - Property 1: Creation assigns `max + 1` (or 1 for empty scope)
  - Property 2: Move swaps exactly two entities and preserves all others
- Service-level unit tests: empty scope, deleted entities, 404/400 error cases
- Route-level integration tests: 401 without auth, 200 valid swap, 400 invalid direction, routes exist for all five types
- Run `npm run test` to execute all tests
- To test the migration: `docker cp sql/6_04_2026_02_13_normalize-relevance.sql grotto-postgres:/tmp/normalize-relevance.sql && docker exec grotto-postgres psql -U root -d grottoce -f /tmp/normalize-relevance.sql`
- To test the endpoints manually (requires a running dev server):
  1. Get a token: `curl -X POST http://localhost:1337/api/v1/login -H "Content-Type: application/json" -d '{"email":"admin@admin.com","password":"admin@admin.com"}'`
  2. Move a rigging down: `curl -X PATCH http://localhost:1337/api/v1/riggings/17/move-relevance -H "Content-Type: application/json" -H "Authorization: Bearer <token>" -d '{"direction":1}'`

## 📸 Previews

No UI changes — this is a backend-only feature. The endpoints return JSON:

```json
{
  "moved": { "id": 17, "relevance": 2, "title": "Accès par l'ancienne entrée", "..." },
  "swapped": { "id": 18, "relevance": 1, "title": "Des Meulières aux galeries...", "..." }
}
```

Error responses:
```json
{"message": "Rigging of id 17 cannot be moved further in that direction."}  // 400
{"message": "Location of id 99999 not found."}                              // 404
{"message": "Direction must be 1 or -1."}                                   // 400
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -123,5 +123,6 @@ coverage
 docker/esdata/*
 !docker/esdata/for-docker-compose
 .amazonq
+.kiro
 *.iml
 build-info.json
--- a/api/controllers/v1/comment/create.js
+++ b/api/controllers/v1/comment/create.js
@@ -1,6 +1,7 @@
 const ControllerService = require('../../../services/ControllerService');
 const NotificationService = require('../../../services/NotificationService');
 const CommentService = require('../../../services/CommentService');
+const RelevanceService = require('../../../services/RelevanceService');
 const ParametersValidatorService = require('../../../services/ParametersValidatorService');
 const { toSimpleComment } = require('../../../services/mapping/converters');
 
@@ -20,6 +21,10 @@ module.exports = async (req, res) => {
   );
   if (!linkedEntity) return null;
 
+  const relevance = await RelevanceService.computeNextRelevance('comment', {
+    [linkedEntity.type]: linkedEntity.id,
+  });
+
   const newComment = await TComment.create({
     author: req.token.id,
     body,
@@ -32,7 +37,7 @@ module.exports = async (req, res) => {
     dateInscription: new Date(),
     language,
     [linkedEntity.type]: linkedEntity.id,
-    // TODO compute relevance
+    relevance,
   }).fetch();
 
   const populatedComment = await CommentService.getComment(newComment.id);
--- a/api/controllers/v1/comment/move-relevance.js
+++ b/api/controllers/v1/comment/move-relevance.js
@@ -1,0 +1,18 @@
+const RelevanceService = require('../../../services/RelevanceService');
+
+module.exports = async (req, res) => {
+  const id = Number(req.params.id);
+  const direction = Number(req.body.direction);
+
+  try {
+    const result = await RelevanceService.moveRelevance(
+      'comment',
+      id,
+      direction
+    );
+    return res.ok(result);
+  } catch (err) {
+    if (err.status === 404) return res.notFound({ message: err.message });
+    return res.badRequest({ message: err.message });
+  }
+};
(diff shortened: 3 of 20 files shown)


urien:
Merci pour ce travail important. Est ce qu'il serait possible de mettre aussi history dans la liste ?

ClemRz:
@urien History support added ✅

bsoufflet (dismissed):

PR #1486, opened 2026-03-03
feat(geoloc): add in-memory coordinates snapshot with bbox filtering and cache headers

### LINKED ISSUE #1473: Performance: Pre-compute and cache all entrances for map heatmap endpoint
## Context

The map view (heatmap/cluster) currently loads lot of entrance points from the database on every request. This is the worst-case scenario: a world-wide view forces fetching the entire dataset every time, and it does not scale as the database grows.

**Current observed performance:**
- Server response time: ~1,470ms
- Download time: ~950ms
- **Total: ~2,420ms**
- And there's a SQL hard limit, so we don't have all the points anyway (hard limit of 100,000 results)

## Benchmark: Grottomap

Grottomap serves a similar dataset (`GET https://grottomap.org/entrances.json`, 236,137 points) with significantly better performance:
- Server response time: ~200ms
- Download time: ~540ms
- **Total: ~740ms**

Their approach appears (**probably**) to be a pre-computed file refreshed periodically (not a live database query), which explains the speed difference.

## Next steps - expected

- [ ] Investigate the best caching strategy on the backend
- [ ] Implement the strategy
- [ ] Evaluate client-side caching as a complementary optimization

ClemRz:
Closes #1473

## 🤔 What

Add an in-memory coordinates snapshot to the `GET /api/v1/geoloc/entrancesCoordinates` endpoint, replacing per-request PostgreSQL queries with sub-millisecond JavaScript filtering for non-massif requests.

- New `CoordinatesSnapshotService` that loads all public entrance coordinates into memory on bootstrap
- Non-massif requests filter the snapshot by bounding box in JS instead of hitting the DB
- Massif requests and cold-start fallback continue using the existing DB path
- `Cache-Control` header added to successful responses
- Snapshot is invalidated after bulk entrance CSV imports

## 🤷‍♂️ Why

The full world bounding box query takes over 1.5s in PostgreSQL. Since the front-end calls this endpoint on every map pan/zoom, this creates noticeable lag. An in-memory snapshot eliminates the DB round-trip entirely for the common case.

## 🔍 How

- `api/services/CoordinatesSnapshotService.js`: singleton service holding a `[[lng, lat], ...]` array. Exposes `load()`, `getCoordinates(swLat, swLng, neLat, neLng)`, `clear()`, `reset()`, `isLoaded()`, `getLastRefreshedAt()`, `getTTL()`.
- Bounding box filtering uses strict inequality (`>` / `<`) to match existing SQL behavior, with full-range short-circuits for lat and lng.
- Background stale-while-revalidate refresh: when TTL expires, the stale snapshot continues serving while a new one loads. Single-flight guard prevents concurrent refreshes.
- `config/custom.js`: `coordinatesSnapshotTTL` defaults to 86400s (24h).
- `config/bootstrap.js`: fire-and-forget `load()` call on startup.
- `api/controllers/v1/geoloc/find-entrances-coordinates.js`: routes non-massif requests through the snapshot, falls back to `GeoLocService.getEntrancesCoordinates()` when snapshot is null. Services are resolved lazily inside the handler to avoid Sails `include-all` cache issues.
- `api/controllers/v1/entrance/import-rows.js`: calls `clear()` after successful imports.

## 🧪 Testing

- Property-based tests (fast-check) for bounding box filter correctness, full-range lat skip, and full-range lng skip
- Unit tests for `CoordinatesSnapshotService` (load, error handling, TTL refresh, single-flight guard, clear, accessors)
- Integration tests for the controller (snapshot path, DB fallback, massif bypass, Cache-Control header presence/absence, max-age value)
- Integration tests for bootstrap wiring and import-rows `clear()` call

Run all tests:
```shell
npm run test
```

## 📸 Previews

### DIFF
--- a/api/controllers/v1/entrance/import-rows.js
+++ b/api/controllers/v1/entrance/import-rows.js
@@ -1,4 +1,5 @@
 const CaveService = require('../../../services/CaveService');
+const CoordinatesSnapshotService = require('../../../services/CoordinatesSnapshotService');
 const EntranceService = require('../../../services/EntranceService');
 const EntranceCSVImportService = require('../../../services/EntranceCSVImportService');
 const RightService = require('../../../services/RightService');
@@ -149,5 +150,11 @@ module.exports = async (req, res) => {
   requestResponse.total.successfulImportAsDuplicates =
     requestResponse.successfulImportAsDuplicates.length;
   requestResponse.total.failure = requestResponse.failureImport.length;
+
+  // Invalidate coordinates snapshot so newly imported entrances appear on the map
+  if (requestResponse.successfulImport.length > 0) {
+    CoordinatesSnapshotService.clear();
+  }
+
   return res.ok(requestResponse);
 };
--- a/api/controllers/v1/geoloc/find-entrances-coordinates.js
+++ b/api/controllers/v1/geoloc/find-entrances-coordinates.js
@@ -1,3 +1,4 @@
+const CoordinatesSnapshotService = require('../../../services/CoordinatesSnapshotService');
 const ErrorService = require('../../../services/ErrorService');
 const GeoLocService = require('../../../services/GeoLocService');
 
@@ -12,14 +13,49 @@ module.exports = async (req, res) => {
   if (errorResponse) return errorResponse;
 
   try {
-    const result = await GeoLocService.getEntrancesCoordinates(
-      southWestBound,
-      northEastBound,
-      100000,
-      massifId
-    );
+    let result;
+
+    if (massifId) {
+      // Massif requests bypass the snapshot (AC 1.4)
+      result = await GeoLocService.getEntrancesCoordinates(
+        southWestBound,
+        northEastBound,
+        100000,
+        massifId
+      );
+    } else {
+      // Try snapshot first (AC 1.2)
+      result = CoordinatesSnapshotService.getCoordinates(
+        parseFloat(southWestBound.lat),
+        parseFloat(southWestBound.lng),
+        parseFloat(northEastBound.lat),
+        parseFloat(northEastBound.lng)
+      );
+
+      if (result === null) {
+        // Snapshot not loaded yet — fallback (AC 1.3)
+        sails.log.info(
+          'CoordinatesSnapshot not loaded yet, falling back to DB query'
+        );
+        result = await GeoLocService.getEntrancesCoordinates(
+          southWestBound,
+          northEastBound,
+          100000
+        );
+      }
+    }
+
+    // Cache-Control header for all successful responses (AC 4.1)
+    const lastRefreshed = CoordinatesSnapshotService.getLastRefreshedAt();
+    const ttl = CoordinatesSnapshotService.getTTL();
+    const age = lastRefreshed
+      ? Math.floor((Date.now() - lastRefreshed.getTime()) / 1000)
+      : 0;
+    res.set('Cache-Control', `public, max-age=${Math.max(ttl - age, 0)}`);
+
     return res.json(result);
   } catch (e) {
+    // No Cache-Control on errors (AC 4.2)
     return ErrorService.getDefaultErrorHandler(res)(e);
   }
 };
--- a/api/services/CoordinatesSnapshotService.js
+++ b/api/services/CoordinatesSnapshotService.js
@@ -1,0 +1,132 @@
+/**
+ * CoordinatesSnapshotService
+ *
+ * Holds an in-memory snapshot of all public entrance coordinates.
+ * The snapshot is loaded from PostgreSQL on server bootstrap and
+ * refreshed in the background when the configured TTL expires.
+ *
+ * State is stored in module-level variables (not on the exported object)
+ * so that Sails' _.bindAll() cannot interfere with it.
+ */
+
+const CommonService = require('./CommonService');
+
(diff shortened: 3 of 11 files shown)


Paul-AUB (changes requested):
**Overall:** Well-structured feature with good test coverage (unit, property-based, and integration). The stale-while-revalidate pattern and single-flight guard are correctly implemented. There are a few genuine bugs and one misleading pattern worth fixing.

---

## Bug 1 — Dead `.catch()` in `bootstrap.js` (misleading)

**File:** `config/bootstrap.js:40-42`

```js
CoordinatesSnapshotService.load().catch((err) => {
  sails.log.error('Failed to load coordinates snapshot on bootstrap:', err);
});
```

`load()` wraps everything in an async IIFE with a `try/catch` that swallows all errors internally. The promise it returns **never rejects**. This `.catch()` is dead code and creates a false sense of safety — a future reader might assume errors are logged here when they're not.

**Fix:** Either remove the `.catch()` entirely, or make `load()` propagate errors by re-throwing after logging them.

---

## Bug 2 — Cache gets stuck after `clear()` + failed refresh

**File:** `api/services/CoordinatesSnapshotService.js:106-109`

```js
clear() {
  service.lastRefreshedAt = null;   // ← nulled here
  service.load().catch(() => {});   // ← if this fails silently...
},
```

And in `getCoordinates`, the stale-while-revalidate trigger requires `lastRefreshedAt` to be truthy:

```js
if (
  service.lastRefreshedAt &&   // ← this is null after a failed clear()
  Date.now() - service.lastRefreshedAt.getTime() > getTTL() * 1000 &&
  !service.loadPromise
) {
  service.load().catch(() => {});
}
```

**Scenario:** `clear()` is called → `lastRefreshedAt` is set to `null` → the background `load()` fails (DB transient error). Now:
- `coordinates` still has stale data (not nulled by `clear()`)
- `lastRefreshedAt` is permanently `null`
- The stale-while-revalidate guard will **never trigger** again (`service.lastRefreshedAt &&` is always false)
- No Cache-Control header is served for any subsequent response
- The cache is stuck until the server restarts or another explicit `clear()` is called

**Fix:** In the `catch` block of `load()`, if `lastRefreshedAt` is null but `coordinates` is not, restore it to a past epoch date so the TTL check can trigger a retry:

```js
} catch (err) {
  sails.log.error('CoordinatesSnapshotService.load() failed:', err);
  // Ensure stale-while-revalidate can retry on the next request
  if (service.lastRefreshedAt === null && service.coordinates !== null) {
    service.lastRefreshedAt = new Date(0); // epoch triggers immediate retry
  }
}
```

---

## Bug 3 — Inconsistency in full-range lat/lng shortcut vs. strict inequality

**File:** `api/services/CoordinatesSnapshotService.js:80-95`

The comment says strict `>` inequalities exclude bbox boundary coordinates. But the `fullLat` shortcut skips the latitude filter entirely:

```js
const fullLat = swLat <= -90 && neLat >= 90;

if (fullLat) {
  // Lat filter skipped — a coordinate at lat=-90 IS included here
  return service.coordinates.filter(
    (coord) => coord[0] > swLng && coord[0] < neLng
  );
}

// Without fullLat: coord[1] > swLat → lat=-90 IS excluded
```

A coordinate at exactly `lat = -90` would be:
- **Included** when bbox has `neLat >= 90` (e.g., full world view)
- **Excluded** when bbox has `neLat = 89.9` (e.g., slightly zoomed in)

In practice this is irrelevant (no cave entrance is at the poles), but the behavior contradicts the documented invariant. The property tests don't catch this because the bbox generator can hit the `fullLat` path and mask the inconsistency.

**Fix:** Add a comment acknowledging this intentional trade-off, or apply strict `>` consistently in the `fullLat`/`fullLng` branches too.

---

## Minor — `clear()` vs. single-flight guard interaction

**File:** `api/services/CoordinatesSnapshotService.js:106-109`

`clear()` calls `service.load()`, but `load()` has the single-flight guard:

```js
load() {
  if (service.loadPromise) return service.loadPromise; // returns immediately if in-flight
  ...
}
```

If `clear()` is called while a TTL-triggered background refresh is already in progress, it returns the existing promise and `lastRefreshedAt` is set to null — but the in-flight refresh will set it back to `new Date()` on completion. This means newly imported entrances will not appear until after the already-in-flight refresh completes rather than a fresh one. Worth a comment documenting this edge case.

---

## Observation — `Cache-Control: public` directive

**File:** `api/controllers/v1/geoloc/find-entrances-coordinates.js:41-45`

The `public` directive allows CDN/proxy caching. The SQL query already filters `is_sensitive = false`, so the data is safe to cache publicly. Worth a brief comment confirming this was intentional so future readers don't remove `public` thinking it is a security concern.

---

## Summary

| Severity | Issue |
|---|---|
| **Medium** | `clear()` + failed refresh leaves cache permanently stuck with no auto-retry |
| **Medium** | `.catch()` in `bootstrap.js` is dead code, creates a false safety signal |
| **Low** | Full-range lat/lng shortcut is inconsistent with the strict-inequality invariant |
| **Low** | Single-flight guard interaction with `clear()` leaves a race window worth documenting |

Paul-AUB:
Tested with the test database on the front branch, technicaly works (but I can't measure the perfs on the test database):
image

Paul-AUB (approved):
LGTM !

PR #1489, opened 2026-03-08
fix(convert): prevent server crash from unhandled exceptions in controllers

ClemRz:
## 🤔 What

- Fix a crash in the `/api/v1/convert` endpoint caused by CRS definitions missing the `+title=` segment
- Convert three legacy `.then()` controllers to `async/await` with `try/catch` to prevent unhandled exceptions from crashing the server
- Add process-level `uncaughtException` and `unhandledRejection` handlers as a last-resort safety net

## 🤷‍♂️ Why

The production server was rebooting every time `/api/v1/convert` was called (which happens on every page load). A CRS row in the database had a `definition` without `+title=`, causing `words[1].split()` to throw a `TypeError` on `undefined`. Because the controller used the old `.then()` pattern without `.catch()`, this became an unhandled promise rejection, which Node.js 24 treats as fatal — killing the entire process.

This also caused CORS errors in the browser: Azure's reverse proxy returned a bare 502 with no headers when the Node process died.

## 🔍 How

1. **Guard the title parsing** in `convert.js`: `words[1] ? words[1].split('+', 1)[0] : ''`
2. **Convert 3 controllers** from `.then(success, error)` to `async/await` + `try/catch`:
   - `api/controllers/v1/convert/convert.js`
   - `api/controllers/v1/comment/get-entrance-stats.js`
   - `api/controllers/v1/comment/get-entrance-time-infos.js`
3. **Add process-level handlers** in `app.js` for `uncaughtException` and `unhandledRejection` — these log the error and keep the process alive instead of crashing.

## 🧪 Testing

- Added a CRS fixture without `+title=` (`EPSG:NO_TITLE`) to reproduce the crash scenario
- Updated Convert tests to verify the guard works (title defaults to empty string)
- All 26 related tests pass: `PORT=1338 npx mocha --timeout 20000 --exit test/bootstrap.test.js "test/integration/4_routes/Convert.test.js" "test/integration/4_routes/Comments/get-entrance-stats.test.js" "test/integration/4_routes/Comments/get-entrance-time-infos.test.js" "test/integration/4_routes/Notifications/count-unread.test.js"`

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/app.js
+++ b/app.js
@@ -60,6 +60,17 @@ try {
   process.exit(1);
 } // -•
 
+// Last-resort safety net: log and survive instead of crashing.
+// Individual controllers should still handle their own errors —
+// these handlers exist only to prevent a full server reboot when
+// an edge case slips through.
+process.on('uncaughtException', (err) => {
+  console.error(new Date().toISOString(), 'Uncaught exception:', err);
+});
+process.on('unhandledRejection', (reason) => {
+  console.error(new Date().toISOString(), 'Unhandled rejection:', reason);
+});
+
 // Start server
 sails.lift(rc('sails'));
 sails.on('lifted', () => console.log(new Date().toISOString(), 'Sails lifted'));
--- a/api/controllers/v1/comment/get-entrance-stats.js
+++ b/api/controllers/v1/comment/get-entrance-stats.js
@@ -1,12 +1,10 @@
 const CommentService = require('../../../services/CommentService');
 
-module.exports = (req, res) => {
+module.exports = async (req, res) => {
   const entranceId = req.param('entranceId');
   if (!entranceId) {
     return res.badRequest('EntranceId param is missing');
   }
-  return CommentService.getStatsFromId(entranceId).then(
-    (result) => res.json(result),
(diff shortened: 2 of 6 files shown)


vmarseguerra (changes requested):
No need to wrap into try/catch (only convert the `then()` into an async function). Catching the error, logging it, and returning an HTTP 500 is handled automatically by Sails.
It allows a uniform error handling across all endpoints and is easier to parse and debug in the log

app.js
```diff
@@ -60,6 +60,17 @@ try {
   process.exit(1);
 } // -•
 
+// Last-resort safety net: log and survive instead of crashing.
+// Individual controllers should still handle their own errors —
+// these handlers exist only to prevent a full server reboot when
+// an edge case slips through.
+process.on('uncaughtException', (err) => {
```
vmarseguerra, line 67: Maybe it is better to let the service crash instead of catching the error. Azure will restart it anyway.
So we can clearly identify errors and fix them, this will only hide them in the logs

vmarseguerra (approved):

PR #1499, opened 2026-03-17
fix(search): sanitize empty and whitespace-only filter values in buildFilter

ClemRz:
## 🤔 What

Sanitize string filter values in `SearchService.buildFilter()` to prevent empty or whitespace-only values from being sent to Typesense.

## 🤷‍♂️ Why

The `POST /api/v1/advanced-search` endpoint was returning 500 errors when users submitted filter objects with empty or near-empty string values (e.g., `city: ". "`, `county: ""`). Typesense rejects empty filter values with a 400 (`Filter value cannot be empty`), which bubbled up as an unhandled 500 since `buildFilter` only checked for JS falsiness — missing whitespace-only and near-empty strings like `". "` that are truthy but meaningless.

## 🔍 How

In `buildFilter()` (`api/services/SearchService.js`):
- Replaced the simple `v => v` truthy check with explicit type-aware filtering that skips empty strings and whitespace-only strings via `v.trim().length > 0`
- Added trimming of string values before they're used in the filter expression, so `"  FR  "` becomes `FR` in the Typesense query
- Preserved existing behavior for `null`, `undefined`, `false`, and `0` (all still filtered out)

## 🧪 Testing

5 new unit tests added to `SearchService.test.js` under the `buildFilter()` describe block:
- Filters out empty strings
- Filters out whitespace-only strings
- Trims dot-space values like `". "` before filtering
- Trims leading/trailing whitespace from valid string values
- Produces no `filter_by` when all filter values are empty or whitespace

Run with: `PORT=1338 npx mocha --exit --timeout 30000 --grep "buildFilter" test/integration/1_services/SearchService.test.js`

### DIFF
--- a/api/services/SearchService.js
+++ b/api/services/SearchService.js
@@ -20,24 +20,30 @@ const allEntitiesKeys = Object.keys(allEntities);
 function buildFilter(filter, isLogicalCompareAnd = true) {
   let hasPrefixFilter = false;
   const out = Object.entries(filter)
-    .filter(([, v]) => v)
+    .filter(([, v]) => {
+      if (v === null || v === undefined) return false;
+      if (typeof v === 'string') return v.trim().length > 0;
+      return true;
+    })
     .flatMap(([k, v]) => {
       // For datePublication, use a prefix filter so that partial dates
       // (year, year-month, or full date) match all more-specific entries.
       // e.g. "2025" matches "2025", "2025-01", "2025-01-15", etc.
       // Typesense supports prefix matching on string fields with := and *.
-      if (k === 'datePublication' && typeof v === 'string') {
+      const trimmed = typeof v === 'string' ? v.trim() : v;
+
+      if (k === 'datePublication' && typeof trimmed === 'string') {
         hasPrefixFilter = true;
-        return [`${k}:=${v}*`];
+        return [`${k}:=${trimmed}*`];
       }
 
-      let vFmt = v;
+      let vFmt = trimmed;
       let operator = ':'; // Partial equal
-      if (Array.isArray(v))
-        vFmt = `[${v.join('..')}]`; // Range
-      else if (typeof v === 'boolean' || typeof v === 'number')
+      if (Array.isArray(trimmed))
+        vFmt = `[${trimmed.join('..')}]`; // Range
+      else if (typeof trimmed === 'boolean' || typeof trimmed === 'number')
         operator = ':='; // Exact equal
-      else vFmt = `\`${v}\``;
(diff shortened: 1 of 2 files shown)


Paul-AUB (commented):
# Code Review: fix(search): sanitize empty and whitespace-only filter values in buildFilter

Commit: `e40a394`

## Overview

The fix is minimal and well-targeted. The root cause is clearly identified (Typesense 400 surfacing as 500) and the solution is correct. Test coverage for the reported cases is solid. A few points deserve attention, including a latent bug that the refactor makes more visible.

---

## Issues

### 1. `false` and `0` are silently dropped as filter values — `SearchService.js:23`

```js
if (v === null || v === undefined || v === false || v === 0) return false;
```

The original `.filter(([, v]) => v)` already dropped these (both are falsy), so the behavior is **preserved**, not introduced. However, the new explicit form raises the question loudly: is dropping `false` intentional?

The `flatMap` below explicitly handles booleans and numbers with exact-equal operator:

```js
else if (typeof trimmed === 'boolean' || typeof trimmed === 'number')
  operator = ':='; // Exact equal
```

This means `true` and any non-zero number are valid filter values, but `false` and `0` are silently ignored. A real-world case: `{ isPublic: false }` (filter non-public entries) would produce no filter at all — likely a bug. If this is intentional by convention, it should be documented with a comment.

### 2. No test coverage for `false` and `0` filter values

No test asserts what happens with `{ isPublic: false }` or `{ altitude: 0 }`. Given the code explicitly names these cases, a test that documents the expected behavior (whether "drop it" or "pass it through") would prevent silent regressions.

### 3. Minor: test description is misleading — `SearchService.test.js:642`

```js
it('should trim string values like dot-space before filtering', ...)
```

The description says "like dot-space" but the actual intent is just "trim trailing whitespace from non-empty values". The wording is unclear — "dot-space" looks like it refers to a specific input artifact rather than describing the general behavior being tested.

---

## Positives

- The fix is surgical: only `buildFilter` is touched, no side effects elsewhere.
- Replacing the implicit `.filter(([, v]) => v)` with explicit conditions is a clear readability improvement.
- The `trimmed` variable is introduced once and reused consistently throughout the `flatMap`, avoiding repeated calls to `.trim()`.
- Five new tests cover all the meaningful cases: empty string, whitespace-only, mixed empty/non-empty, trimming, and all-empty producing no `filter_by`.
- Commit message clearly explains the chain of failure (empty value → Typesense 400 → unhandled 500).

ClemRz:
1, 2 and 3 fixed.

Paul-AUB (approved):

PR #1500, opened 2026-03-17
fix(db): add query performance fixes for recent changes, geoloc, notifications, and comments

ClemRz:
## 🤔 What

Fixes 6 query performance defects identified through PostgreSQL diagnostic analysis (sequential scan ratios, missing indexes, unbounded queries, row multiplication).

- Fix 1: Add 7-day window + LIMIT 500 to `RecentChangeService.getRecent()` (was full table scan)
- Fix 2: Add `is_main = true AND is_deleted = false` filters to GeoLocService name joins (was causing row multiplication)
- Fix 3: Add covering index `idx_t_notification_notified(id_notified, date_inscription DESC)` for notification queries
- Fix 4: Add synthetic PK (`id SERIAL`) to `t_last_change` to enable HOT updates
- Fix 5: Replace partial comment indexes with non-partial (planner couldn't use them for all query patterns)
- Fix 7: Tune autovacuum on `t_last_change` (trigger at 5% dead tuples instead of default 20%)

Fix 6 (materialized view refresh optimization) is deferred — it runs every 3 days at 4:40 AM via cron and needs separate analysis.

## 🤷‍♂️ Why

Production diagnostic queries revealed:
- `t_last_change`: 98.5% sequential scan ratio, 7.2% dead tuple ratio, no PK for HOT updates
- `t_notification`: 98.1% sequential scan ratio on user notification lookups
- GeoLocService map queries: row multiplication from unfiltered name joins (multiple `t_name` rows per entrance/cave)
- Comment indexes were partial (`WHERE id_entrance IS NOT NULL`) preventing the planner from using them for all query patterns

The `RecentChangeService.getRecent()` query was doing a full sequential scan of `t_last_change` with no date filter or row limit, despite only needing recent data for the dashboard.

## 🔍 How

**SQL Migration** (`sql/94_01_2026_03_17_query_performance_fixes.sql`):
- Idempotent DO block adds `id SERIAL PRIMARY KEY` to `t_last_change` if missing
- Creates covering index on `t_notification(id_notified, date_inscription DESC)`
- Drops partial comment indexes and recreates as non-partial
- Sets autovacuum tuning parameters on `t_last_change`

**Schema baseline** (`sql/0_tables.sql`):
- Updated `t_last_change` to include `id serial NOT NULL` PK and autovacuum WITH clause
- Changed comment indexes from partial to non-partial
- Added `idx_t_notification_notified` after notification table definition

**Service changes**:
- `RecentChangeService.getRecent()`: Added `WHERE date_change > current_timestamp - interval '7 days'` and `LIMIT 500`
- `GeoLocService`: Added `AND ne.is_main = true AND ne.is_deleted = false` to all `LEFT JOIN t_name` clauses in `PUBLIC_ENTRANCES_IN_BOUNDS`, `PUBLIC_ENTRANCES_IN_BOUNDS_AND_MASSIF`, and `NETWORKS_IN_BOUNDS`

### Automated tests

```bash
npm run dev:clean                          # Rebuild containers with new migration
docker logs grotto-postgres | grep ERROR   # Verify no init errors
PORT=1338 npm run test                     # All 1763 tests pass
npm run lint                               # Clean
```

New test files:
- `test/integration/0_models/QueryPerformanceFixes.test.js` — schema assertions for new indexes, non-partial comment indexes, `t_last_change` PK
- `test/integration/1_services/RecentChangeWindow.test.js` — verifies 7-day filter, LIMIT 500, ORDER BY in query
- `test/integration/1_services/GeoLocNameFilter.test.js` — verifies no row multiplication, `is_main` filter in SQL

### Manual spot tests

After deploying to a test environment with the migration applied:

1. **Recent changes** — `GET /api/v1/changes/recent`
   - Should return grouped changes from the last 7 days only
   - Response should be noticeably faster than before (was full table scan)
   - Verify the response still contains `date`, `authorId`, `author`, `mainEntityType`, `mainEntityId`, `mainAction` fields

2. **Map entrances** — `GET /api/v1/geoloc/entrances?sw_lat=43&sw_lng=1&ne_lat=46&ne_lng=4&limit=200`
   - Should return no duplicate entrance IDs in the response array
   - Each entrance should appear exactly once (no row multiplication from name joins)

3. **Map networks** — `GET /api/v1/geoloc/networks?sw_lat=43&sw_lng=1&ne_lat=46&ne_lng=4`
   - Should return no duplicate network IDs
   - Network names should still resolve correctly (COALESCE of cave name / entrance name)

4. **Map entrances with massif filter** — `GET /api/v1/geoloc/entrances?sw_lat=43&sw_lng=1&ne_lat=46&ne_lng=4&limit=200&massif=<valid_massif_id>`
   - Same as above, filtered to a specific massif — no duplicates

5. **Notifications** (requires auth) — Verify notification queries for a logged-in user still work and return results ordered by date

### Production deployment procedure

1. **Run the migration BEFORE deploying the new code** (the migration is additive and backward-compatible):
   ```sql
   -- Connect to the production PostgreSQL instance and run:
   \i sql/94_01_2026_03_17_query_performance_fixes.sql
   ```
   Or execute the statements manually:
   ```sql
   -- Fix 4: Add synthetic PK to t_last_change
   DO $
   BEGIN
     IF NOT EXISTS (
       SELECT 1 FROM information_schema.columns
       WHERE table_name = 't_last_change' AND column_name = 'id'
     ) THEN
       ALTER TABLE t_last_change ADD COLUMN id SERIAL PRIMARY KEY;
     END IF;
   END $;

   -- Fix 3: Notification covering index
   CREATE INDEX IF NOT EXISTS idx_t_notification_notified
     ON t_notification(id_notified, date_inscription DESC);

   -- Fix 5: Replace partial comment indexes with non-partial
   DROP INDEX IF EXISTS idx_t_comment_entrance;
   DROP INDEX IF EXISTS idx_t_comment_cave;
   CREATE INDEX IF NOT EXISTS idx_t_comment_entrance ON t_comment(id_entrance);
   CREATE INDEX IF NOT EXISTS idx_t_comment_cave ON t_comment(id_cave);

   -- Fix 7: Autovacuum tuning for t_last_change
   ALTER TABLE t_last_change SET (
     autovacuum_vacuum_scale_factor = 0.05,
     autovacuum_analyze_scale_factor = 0.02
   );
   ```

2. **Verify migration** — after running, confirm:
   ```sql
   -- PK exists on t_last_change
   SELECT column_name FROM information_schema.columns WHERE table_name = 't_last_change' AND column_name = 'id';
   -- New indexes exist
   SELECT indexname FROM pg_indexes WHERE indexname IN ('idx_t_notification_notified', 'idx_t_comment_entrance', 'idx_t_comment_cave');
   -- Comment indexes are non-partial (no WHERE in indexdef)
   SELECT indexname, indexdef FROM pg_indexes WHERE indexname IN ('idx_t_comment_entrance', 'idx_t_comment_cave');
   ```

3. **Deploy the code** — merge to `develop`, GitHub Actions deploys to Azure App Service automatically

4. **Post-deploy verification** — run the manual spot tests above against production, and monitor:
   ```sql
   -- Check sequential scan ratios improved
   SELECT relname, seq_scan, idx_scan,
          CASE WHEN (seq_scan + idx_scan) > 0
               THEN round(100.0 * seq_scan / (seq_scan + idx_scan), 1)
               ELSE 0 END AS seq_pct
   FROM pg_stat_user_tables
   WHERE relname IN ('t_last_change', 't_notification', 't_comment')
   ORDER BY relname;

   -- Check dead tuple ratio on t_last_change (should drop from ~7% toward <2%)
   SELECT relname, n_dead_tup, n_live_tup,
          CASE WHEN n_live_tup > 0
               THEN round(100.0 * n_dead_tup / n_live_tup, 1)
               ELSE 0 END AS dead_pct
   FROM pg_stat_user_tables
   WHERE relname = 't_last_change';
   ```

### Rollback

All changes are backward-compatible. If needed:
- The new indexes and PK column can be dropped without affecting existing code
- The `RecentChangeService` 7-day window and LIMIT can be reverted by removing the WHERE/LIMIT clauses
- The GeoLocService join filters can be reverted by removing the `AND ne.is_main = true AND ne.is_deleted = false` conditions

## 📸 Previews

N/A — backend-only changes, no UI impact.

### DIFF
--- a/api/services/GeoLocService.js
+++ b/api/services/GeoLocService.js
@@ -4,8 +4,8 @@ const PUBLIC_ENTRANCES_IN_BOUNDS = `
   c.size_coef as size_coef, e.id_cave as idCave, nc.name as nameCave, c.depth as depthCave,
   c.length as lengthCave
   FROM t_entrance as e
-  LEFT JOIN t_name as ne ON ne.id_entrance = e.id
-  LEFT JOIN t_name as nc ON nc.id_cave = e.id_cave
+  LEFT JOIN t_name as ne ON ne.id_entrance = e.id AND ne.is_main = true AND ne.is_deleted = false
+  LEFT JOIN t_name as nc ON nc.id_cave = e.id_cave AND nc.is_main = true AND nc.is_deleted = false
   LEFT JOIN t_cave as c ON c.Id = e.id_cave
   WHERE ST_Within(e.point_geom, ST_MakeEnvelope($1, $2, $3, $4, 4326))
   AND e.is_sensitive = false
@@ -20,8 +20,8 @@ const PUBLIC_ENTRANCES_IN_BOUNDS_AND_MASSIF = `
   c.size_coef as size_coef, e.id_cave as idCave, nc.name as nameCave, c.depth as depthCave,
   c.length as lengthCave
   FROM t_entrance as e
-  LEFT JOIN t_name as ne ON ne.id_entrance = e.id
-  LEFT JOIN t_name as nc ON nc.id_cave = e.id_cave
+  LEFT JOIN t_name as ne ON ne.id_entrance = e.id AND ne.is_main = true AND ne.is_deleted = false
+  LEFT JOIN t_name as nc ON nc.id_cave = e.id_cave AND nc.is_main = true AND nc.is_deleted = false
   LEFT JOIN t_cave as c ON c.Id = e.id_cave
   JOIN t_massif AS m ON m.id = $6
   WHERE ST_Within(e.point_geom, ST_MakeEnvelope($1, $2, $3, $4, 4326))
@@ -55,8 +55,8 @@ const NETWORKS_IN_BOUNDS = `
   SELECT c.id as id, COALESCE(nc.name, ne.name) as name, avg(en.longitude) as longitude, avg(en.latitude) as latitude
   FROM t_entrance as en
   INNER JOIN t_cave c ON c.id = en.id_cave
-  LEFT JOIN t_name AS nc ON nc.id_cave = c.id
-  LEFT JOIN t_name as ne ON ne.id_entrance = en.id
+  LEFT JOIN t_name AS nc ON nc.id_cave = c.id AND nc.is_main = true AND nc.is_deleted = false
+  LEFT JOIN t_name as ne ON ne.id_entrance = en.id AND ne.is_main = true AND ne.is_deleted = false
   WHERE ST_Within(en.point_geom, ST_MakeEnvelope($1, $2, $3, $4, 4326))
   AND en.is_sensitive = false
   AND en.is_deleted = false
--- a/api/services/RecentChangeService.js
+++ b/api/services/RecentChangeService.js
@@ -2,6 +2,12 @@ const CommonService = require('./CommonService');
 
 const GROUP_MAX_TIME_DIFF_S = 60 * 60 * 6; // 6 Hours
 
+// Dashboard only needs a short lookback; 7 days covers the typical review cycle
+// while keeping the query fast on the unpartitioned t_last_change table.
+const RECENT_CHANGES_WINDOW_DAYS = 7;
+// Hard cap to avoid oversized payloads — the dashboard is a summary, not a full audit log.
+const RECENT_CHANGES_LIMIT = 500;
+
 async function removeOlderChanges() {
   const query = `DELETE FROM t_last_change WHERE date_change < current_timestamp - interval '1 month';`;
   await CommonService.query(query);
@@ -108,7 +114,9 @@ async function getRecent() {
   const query = `
   SELECT tbl.*, author.nickname FROM t_last_change tbl
   LEFT JOIN t_caver author ON tbl.id_author = author.id
+  WHERE tbl.date_change > current_timestamp - interval '${RECENT_CHANGES_WINDOW_DAYS} days'
   ORDER BY date_change DESC
+  LIMIT ${RECENT_CHANGES_LIMIT}
   `;
   const rep = await CommonService.query(query);
 
--- a/sql/0_tables.sql
+++ b/sql/0_tables.sql
@@ -721,8 +721,8 @@ CREATE TABLE t_comment (
 	CONSTRAINT t_comment_t_entrance2_fk FOREIGN KEY (id_exit) REFERENCES t_entrance(id),
 	CONSTRAINT t_comment_t_language0_fk FOREIGN KEY (id_language) REFERENCES t_language(id)
 );
-CREATE INDEX idx_t_comment_entrance ON t_comment(id_entrance) WHERE id_entrance IS NOT NULL;
-CREATE INDEX idx_t_comment_cave ON t_comment(id_cave) WHERE id_cave IS NOT NULL;
+CREATE INDEX idx_t_comment_entrance ON t_comment(id_entrance);
+CREATE INDEX idx_t_comment_cave ON t_comment(id_cave);
 
 -- t_document definition
 -- Drop table
@@ -1286,6 +1286,7 @@ CREATE TABLE t_notification (
   CONSTRAINT t_notification_t_massif_fk FOREIGN KEY (id_massif) REFERENCES t_massif(id),
   CONSTRAINT t_notification_t_rigging_fk FOREIGN KEY (id_rigging) REFERENCES t_rigging(id)
 );
+CREATE INDEX idx_t_notification_notified ON t_notification(id_notified, date_inscription DESC);
 
 -- DROP TABLE j_document_iso3166_2;
 CREATE TABLE j_document_iso3166_2 (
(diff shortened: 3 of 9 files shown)


Paul-AUB (dismissed):
# Code Review: fix(db): add query performance fixes for recent changes, geoloc, notifications, and comments

Commit: `32d05e1`


## Overview

Four distinct performance fixes in one commit: JOIN filtering on `t_name` to prevent row multiplication in GeoLoc queries, a time window + LIMIT on `getRecent()`, a composite index on notifications, and index/PK fixes on comments and `t_last_change`. The approach is correct in each case, and the migration is idempotent. A few issues deserve attention before merging.

---

## Issues


**NB : These are LLM comments which seems only nits. I approved in advance the PR as they might not all even need a fix**


### 1. GeoLocNameFilter test passes trivially on empty results — `GeoLocNameFilter.test.js:17`

```js
const ids = entrances.map((e) => e.id);
const uniqueIds = [...new Set(ids)];
should(ids.length).equal(uniqueIds.length, '...');
```

If no fixture entrances fall inside the hardcoded bounding box (`lat 40–52, lng -5–15`), `entrances` is an empty array and `ids.length === uniqueIds.length` is `0 === 0` — the assertion passes without testing anything. The test needs a guard:

```js
should(entrances.length).be.greaterThan(0, 'No fixture entrances in bounding box — test is vacuous');
```

The same applies to `getNetworksMap()`. Without this, a broken JOIN filter would go undetected if the bounding box happens to return no rows.

### 2. `LIMIT 500` is a silent data truncation — `RecentChangeService.js:111`

```js
LIMIT 500
```

If more than 500 changes occur within the 7-day window (plausible during active periods), `getRecent()` silently returns a truncated result. The controller at `get-recent.js` passes this straight to the client with no indication that the list is incomplete. At minimum this should be a named constant with a comment; ideally the response should communicate pagination or a cap:

```js
const RECENT_CHANGES_LIMIT = 500; // hard cap — client shows at most this many entries
```

### 3. 7-day window is hardcoded without rationale — `RecentChangeService.js:110`

```js
WHERE tbl.date_change > current_timestamp - interval '7 days'
```

Why 7 days and not 14 or 30? No comment explains the business rule. If the window ever needs to change, the value is buried in a SQL string. A named constant at the top of the file, or at least an inline comment, would make the intent clear and the value easier to find.

### 4. Commit message says "7 defects" — diff shows 4

The message references "Fix 7 query performance defects" and the migration comments are labeled Fix 3, 4, 5, 7. Fixes 1, 2, and 6 are absent from the diff with no explanation. Either they were applied elsewhere (untracked), deferred, or the numbering comes from a diagnostic document not included in the PR. This makes it impossible to verify completeness during review. A short note in the migration file or PR description explaining the missing fixes would resolve the ambiguity.

### 5. Autovacuum tuning is missing from `customSQL.js` test bootstrap — `customSQL.js:103-127`

The migration file applies autovacuum tuning to `t_last_change`:

```sql
ALTER TABLE t_last_change SET (
  autovacuum_vacuum_scale_factor = 0.05,
  autovacuum_analyze_scale_factor = 0.02
);
```

But `QUERY_PERFORMANCE_FIXES_MIGRATION` in `customSQL.js` omits this step. This is probably intentional (autovacuum settings are a server concern, not something you assert in tests), but the omission is undocumented. A comment explaining why would prevent future contributors from thinking the migration is incomplete.

### 6. Replacing partial comment indexes increases index size — `sql/0_tables.sql:723`

The change from:
```sql
CREATE INDEX idx_t_comment_entrance ON t_comment(id_entrance) WHERE id_entrance IS NOT NULL;
```
to:
```sql
CREATE INDEX idx_t_comment_entrance ON t_comment(id_entrance);
```

The rationale (partial index prevents planner from using it when the query lacks an explicit `IS NOT NULL` predicate) is correct and documented in the migration. However, the full index now indexes NULL rows, which adds storage overhead for rows where `id_entrance IS NULL` (comments attached to a cave, not an entrance). If NULL rows are a significant fraction of the table, the index will grow noticeably. Worth confirming the NULL ratio before deploying on a large dataset.

---

## Positives

- The `is_main = true AND is_deleted = false` filter on `t_name` JOINs in `GeoLocService.js` correctly eliminates row multiplication at the SQL level rather than in application code.
- The migration is fully idempotent (`IF NOT EXISTS`, `DO $$ IF NOT EXISTS $$`, `DROP INDEX IF EXISTS`) — safe to run on a partially migrated database.
- `0_tables.sql` is kept in sync with the migration, so fresh installs and upgrades converge to the same schema.
- The `QueryPerformanceFixes.test.js` schema assertions (index existence, non-partial check via `indexdef`, PK presence) are a good pattern for verifying DDL migrations.
- The spy-based tests in `GeoLocNameFilter.test.js` and `RecentChangeWindow.test.js` directly inspect the SQL string, giving confidence that the fix is in the query and not accidentally filtered in memory.

ClemRz:
Issue 1: Added non-vacuous guard to GeoLocNameFilter.test.js and fixed the bounding box from France (lat 40-52, lng -5 to 15) to match actual fixture coordinates (lat 55-65, lng 70-80). For networks, the guard is conditional since fixtures don't have multi-entrance caves.
Issue 2: Extracted LIMIT 500 into a named constant RECENT_CHANGES_LIMIT with a comment.
Issue 3: Extracted the 7-day window into RECENT_CHANGES_WINDOW_DAYS with a comment explaining the rationale.
Issue 4: No change needed.
Issue 5: No change needed.
Issue 6: The partial→full index tradeoff is already documented in the migration. No code change needed.

Paul-AUB (approved):

PR #1501, opened 2026-03-17
feat(massif): add massif geometry endpoints for map rendering

### LINKED ISSUE #1497: Add massif geometry endpoints for map display
# Add massif geometry endpoints for map display

## Summary

To display massifs on the interactive map (heatmap at low zoom, polygon layer with popups at high zoom), two new API endpoints are needed under the existing `/api/v1/geoloc/` namespace.

---

## Context

The frontend map currently supports:
- **Hexbin heatmap** (zoom < 13): one-time worldwide fetch of `[lng, lat]` coordinate arrays → D3-hexbin aggregates by cell
- **Bounded markers** (zoom ≥ 13): fetched on every `moveend` event using viewport bounds

Massifs have a `geogPolygon` (PostGIS MultiPolygon) that could be displayed as:
1. Centroid points feeding the heatmap at low zoom
2. Full polygon geometries with popups at higher zoom

---

### 1. Massif centroids — heatmap feed

```
GET /api/v1/geoloc/massifsCoordinates
```

**Strategy:** single worldwide fetch on page load, same as `entrancesCoordinates` and `networksCoordinates`.

**Query params:** none required.

**Response:**
```json
[[5.52, 44.91], [3.12, 43.20], ...]
```

Array of `[longitude, latitude]` — one point per massif, computed with `ST_Centroid`.

---

### 2. Massif polygons — bounded fetch with popup data

```
GET /api/v1/geoloc/massifs
```

**Strategy:** called by frontend on every `moveend` event when zoom ≥ threshold (suggested: zoom ≥ 8), with viewport bounding box.

**Query params:**

| Param | Type | Description |
|-------|------|-------------|
| `sw_lat` | float | South-West latitude |
| `sw_lng` | float | South-West longitude |
| `ne_lat` | float | North-East latitude |
| `ne_lng` | float | North-East longitude |
| `zoom` | int | Current zoom level |

**Backend filter:** return massifs whose geometry intersects the bounding box (`ST_Intersects`).

**Response:**
```json
[
  {
    "id": 42,
    "name": "Vercors",
    "geogPolygon": {
      "type": "MultiPolygon",
      "coordinates": [[[[5.4, 44.8], [5.6, 44.8], [5.6, 45.0], [5.4, 45.0], [5.4, 44.8]]]]
    },
    "entranceCount": 312,
    "networkCount": 28
  }
]
```

**Response fields:**

| Field | Type | Description |
|-------|------|-------------|
| `id` | int | Massif ID |
| `name` | string | Display name (default language) |
| `geogPolygon` | GeoJSON object | Parsed MultiPolygon (not a string) |
| `entranceCount` | int | Number of entrances in the massif |
| `networkCount` | int | Number of cave networks in the massif |

**Usage:** The frontend popup will display: name (linked to `/massif/{id}`), entrance count, network count.

---

## Summary Table

| Endpoint | Zoom trigger | Response format | Fetch strategy |
|----------|-------------|-----------------|----------------|
| `GET /geoloc/massifsCoordinates` | < threshold | `[[lng, lat], ...]` | Once on load (worldwide) |
| `GET /geoloc/massifs` | ≥ threshold | `[{id, name, geogPolygon, entranceCount, networkCount}]` | On every `moveend` |

---

## Backend Implementation Notes

- Pagination not required if the number of massifs per viewport stays reasonable (< a few hundred), which should be the case at the zoom level the polygons are displayed
- No authentication required (public read, same as other geoloc endpoints), but potentially the sensitive massifs should be excluded from the list

---

## Acceptance Criteria

- [ ] `GET /api/v1/geoloc/massifsCoordinates` returns `[[lng, lat], ...]` for all massifs with a defined polygon
- [ ] `GET /api/v1/geoloc/massifs` returns polygon + popup fields for massifs intersecting the given bbox
- [ ] `geogPolygon` is returned as a parsed GeoJSON object (not a raw string)
- [ ] Both endpoints respond within acceptable latency for the map use case

ClemRz:
## 🤔 What

Add two new public geolocation endpoints for massif map data:

- `GET /api/v1/geoloc/massifsCoordinates` — massif centroid coordinates for the hexbin heatmap at low zoom levels, backed by an in-memory snapshot cache
- `GET /api/v1/geoloc/massifs` — massif polygon geometries with popup metadata (name, entrance count, network count) for higher zoom levels

Closes #1497

## 🤷‍♂️ Why

The front-end map needs massif geometry data to render massif layers. At low zoom, centroids feed the hexbin heatmap. At higher zoom, full polygons with metadata enable clickable massif overlays with popup information.

### Service layer
- Added `getMassifsCoordinates()` and `getMassifsMap()` to `GeoLocService` using raw SQL with PostGIS spatial functions (`ST_Centroid`, `ST_Within`, `ST_Intersects`, `ST_AsGeoJSON`, `ST_Contains`)
- Created `MassifCoordinatesSnapshotService` mirroring `CoordinatesSnapshotService` — in-memory cache with stale-while-revalidate and single-flight guard
- The polygon query uses a single SQL statement with correlated subqueries for entrance/network counts and a `LEFT JOIN t_name` for the main name

### Controllers
- `find-massifs-coordinates.js`: snapshot-first with DB fallback, `Cache-Control` header
- `find-massifs.js`: direct DB query via `GeoLocService.getMassifsMap()`
- Both use `GeoLocService.checkAndGetCoordinatesParams()` for bbox validation
- Public access (`true` policy), no authentication required

### OpenAPI
- Added both endpoint definitions to `assets/swaggerV1.yaml` with full request/response schemas

## 🧪 Testing

```bash
PORT=1338 npm run test
```

New tests added (10 integration + 9 property tests):

**Route integration tests** (`test/integration/4_routes/Geoloc.test.js`):
- massifsCoordinates: 400 on missing params, 200 with valid bbox, empty array for no-match, no-auth access, Cache-Control header
- massifs: 400 on missing params, 200 with correct shape, geogPolygon is parsed object, empty array for no-match, no-auth access

**Property tests** (fast-check, 100 runs each):
- Property 1: Centroid computation matches direct ST_Centroid query
- Property 5: Response shape and GeoJSON parsing correctness
- Property 6: Only non-deleted massifs with non-null polygons returned
- Property 7: Every returned polygon intersects the requested bbox
- Property 8: Invalid bbox params rejected with 400
- Property 9: Name, entranceCount, networkCount match reference queries

**Service property tests** (`test/integration/1_services/`):
- Property 2: Centroid exclusion filtering
- Property 3: Snapshot bbox filter correctness
- Property 4: Snapshot load and stale-while-revalidate

## 📸 Previews

N/A — backend-only changes.

### DIFF
--- a/api/controllers/v1/entrance/import-rows.js
+++ b/api/controllers/v1/entrance/import-rows.js
@@ -153,7 +153,7 @@ module.exports = async (req, res) => {
 
   // Invalidate coordinates snapshot so newly imported entrances appear on the map
   if (requestResponse.successfulImport.length > 0) {
-    CoordinatesSnapshotService.clear();
+    CoordinatesSnapshotService.invalidate();
   }
 
   return res.ok(requestResponse);
--- a/api/controllers/v1/geoloc/find-entrances-coordinates.js
+++ b/api/controllers/v1/geoloc/find-entrances-coordinates.js
@@ -46,12 +46,16 @@ module.exports = async (req, res) => {
     }
 
     // Cache-Control header for all successful responses (AC 4.1)
+    // When lastRefreshed is null (snapshot not yet loaded), emit max-age=0
+    // to prevent CDNs from caching a DB-fallback response for the full TTL.
     const lastRefreshed = CoordinatesSnapshotService.getLastRefreshedAt();
     const ttl = CoordinatesSnapshotService.getTTL();
-    const age = lastRefreshed
-      ? Math.floor((Date.now() - lastRefreshed.getTime()) / 1000)
-      : 0;
-    res.set('Cache-Control', `public, max-age=${Math.max(ttl - age, 0)}`);
+    if (lastRefreshed) {
+      const age = Math.floor((Date.now() - lastRefreshed.getTime()) / 1000);
+      res.set('Cache-Control', `public, max-age=${Math.max(ttl - age, 0)}`);
+    } else {
+      res.set('Cache-Control', 'public, max-age=0');
+    }
 
     return res.json(result);
   } catch (e) {
--- a/api/controllers/v1/geoloc/find-massifs-coordinates.js
+++ b/api/controllers/v1/geoloc/find-massifs-coordinates.js
@@ -1,0 +1,44 @@
+const MassifCoordinatesSnapshotService = require('../../../services/MassifCoordinatesSnapshotService');
+const ErrorService = require('../../../services/ErrorService');
+const GeoLocService = require('../../../services/GeoLocService');
+
+module.exports = async (req, res) => {
+  const { southWestBound, northEastBound, errorMessage } =
+    GeoLocService.checkAndGetCoordinatesParams(req);
+
+  if (errorMessage !== '') return res.badRequest(errorMessage);
+
+  try {
+    // Try snapshot first
+    let result = MassifCoordinatesSnapshotService.getCoordinates(
+      parseFloat(southWestBound.lat),
+      parseFloat(southWestBound.lng),
+      parseFloat(northEastBound.lat),
+      parseFloat(northEastBound.lng)
+    );
+
+    if (result === null) {
+      // Snapshot not loaded yet — fallback to DB query
+      sails.log.info(
+        'MassifCoordinatesSnapshot not loaded yet, falling back to DB query'
+      );
+      result = await GeoLocService.getMassifsCoordinates(
+        southWestBound,
+        northEastBound
+      );
+    }
+
+    // Cache-Control header for successful responses
+    const lastRefreshed =
+      MassifCoordinatesSnapshotService.getLastRefreshedAt();
+    const ttl = MassifCoordinatesSnapshotService.getTTL();
+    const age = lastRefreshed
+      ? Math.floor((Date.now() - lastRefreshed.getTime()) / 1000)
+      : 0;
+    res.set('Cache-Control', `public, max-age=${Math.max(ttl - age, 0)}`);
+
+    return res.json(result);
+  } catch (e) {
+    return ErrorService.getDefaultErrorHandler(res)(e);
+  }
+};
--- a/api/controllers/v1/geoloc/find-massifs-coordinates.js
+++ b/api/controllers/v1/geoloc/find-massifs-coordinates.js
@@ -28,14 +28,17 @@ module.exports = async (req, res) => {
       );
     }
 
-    // Cache-Control header for successful responses
-    const lastRefreshed =
-      MassifCoordinatesSnapshotService.getLastRefreshedAt();
+    // Cache-Control header for successful responses.
+    // When lastRefreshed is null (snapshot not yet loaded), emit max-age=0
+    // to prevent CDNs from caching a DB-fallback response for the full TTL.
+    const lastRefreshed = MassifCoordinatesSnapshotService.getLastRefreshedAt();
     const ttl = MassifCoordinatesSnapshotService.getTTL();
-    const age = lastRefreshed
-      ? Math.floor((Date.now() - lastRefreshed.getTime()) / 1000)
-      : 0;
-    res.set('Cache-Control', `public, max-age=${Math.max(ttl - age, 0)}`);
+    if (lastRefreshed) {
+      const age = Math.floor((Date.now() - lastRefreshed.getTime()) / 1000);
+      res.set('Cache-Control', `public, max-age=${Math.max(ttl - age, 0)}`);
+    } else {
+      res.set('Cache-Control', 'public, max-age=0');
+    }
 
     return res.json(result);
   } catch (e) {
--- a/api/controllers/v1/geoloc/find-massifs.js
+++ b/api/controllers/v1/geoloc/find-massifs.js
@@ -1,0 +1,19 @@
+const ErrorService = require('../../../services/ErrorService');
+const GeoLocService = require('../../../services/GeoLocService');
+
+module.exports = async (req, res) => {
+  const { southWestBound, northEastBound, errorMessage } =
+    GeoLocService.checkAndGetCoordinatesParams(req);
+
+  if (errorMessage !== '') return res.badRequest(errorMessage);
+
+  try {
+    const result = await GeoLocService.getMassifsMap(
+      southWestBound,
+      northEastBound
+    );
+    return res.json(result);
+  } catch (e) {
+    return ErrorService.getDefaultErrorHandler(res)(e);
+  }
+};
--- a/api/services/CoordinatesSnapshotService.js
+++ b/api/services/CoordinatesSnapshotService.js
@@ -7,6 +7,11 @@
  *
  * State is stored in module-level variables (not on the exported object)
  * so that Sails' _.bindAll() cannot interfere with it.
+ *
+ * NOTE: If load() fails at bootstrap, there is no automatic retry.
+ * The DB fallback in the controller will handle requests correctly,
+ * but without the snapshot's performance benefit. A server restart
+ * is required to restore the in-memory snapshot.
  */
 
 const CommonService = require('./CommonService');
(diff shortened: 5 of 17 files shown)


Paul-AUB (changes requested):
# Code Review — `feat/massif-geometry-endpoints`

## Bugs / Correctness Issues

### 1. Cache-Control is `max-age=TTL` on DB fallback responses (boot window)

**File:** [api/controllers/v1/geoloc/find-massifs-coordinates.js](api/controllers/v1/geoloc/find-massifs-coordinates.js#L32)

When the snapshot has not yet been loaded (server just started), `getLastRefreshedAt()`
returns `null`. The controller treats `null` as age=0:

```js
const age = lastRefreshed
  ? Math.floor((Date.now() - lastRefreshed.getTime()) / 1000)
  : 0;                                  // ← 0 when snapshot not ready
res.set('Cache-Control', `public, max-age=${Math.max(ttl - age, 0)}`);
// → Cache-Control: public, max-age=86400
```

A client that hits the API during the boot window (typically a few seconds) receives a
live DB result with a 24-hour cache directive. If that client is a CDN, it will serve the
snapshot-free result for the next 24 hours.

**Suggestion:** When `lastRefreshed` is null (snapshot not yet loaded), emit `max-age=0`
or omit the `Cache-Control` header entirely.

### 2. `clear()` causes the same stale Cache-Control for fresh requests

**File:** [api/services/MassifCoordinatesSnapshotService.js](api/services/MassifCoordinatesSnapshotService.js#L118)

```js
clear() {
  lastRefreshedAt = null;   // ← reset to null, but coordinates kept
  load().catch(() => {});
},
```

`clear()` sets `lastRefreshedAt = null` but does not clear `coordinates`. The old
snapshot is still served while `load()` runs in the background (stale-while-revalidate).

The problem: the controller reads `getLastRefreshedAt()` to compute the Cache-Control
header. With `lastRefreshedAt = null`, it falls into the `age = 0` path and emits
`max-age=TTL` (e.g. `max-age=86400`) for data that could be up to one full TTL old.

Two independent fixes are needed:

- Either keep `lastRefreshedAt` at its previous value (not null) when doing a
  stale-while-revalidate, so the age continues to increase correctly.
- Or do not set the Cache-Control header based on `lastRefreshedAt` when coordinates
  were returned from a stale/fallback path.

---

## Design Concerns

### 3. `clear()` name and log message are misleading

**File:** [api/services/MassifCoordinatesSnapshotService.js](api/services/MassifCoordinatesSnapshotService.js#L118)

The log says `"MassifCoordinatesSnapshot cleared"` and the method is called `clear()`,
but the snapshot is not cleared — `isLoaded()` returns `true` and `getCoordinates()`
continues serving old data. The intent is a background revalidation, not a reset.

This creates confusion for any caller that calls `clear()` expecting the service to stop
serving stale data. Compare with `reset()` which actually clears state.

**Suggestion:** Rename to `invalidate()` or `scheduleRefresh()` and update the log to say
something like `"TTL invalidated, triggering background refresh"`.

### 4. Correlated subqueries in `MASSIFS_IN_BOUNDS` for `entranceCount` / `networkCount`

**File:** [api/services/GeoLocService.js](api/services/GeoLocService.js#L93)

```sql
(
  SELECT COUNT(e.id)::integer
  FROM t_entrance AS e
  WHERE e.is_deleted = false
    AND ST_Contains(m.geog_polygon::geometry, e.point_geom)
) AS "entranceCount",
```

Each matched massif triggers a separate `ST_Contains` scan over `t_entrance`. For a
bounding box covering a large region (e.g. all of France), this is O(massifs ×
entrances). Whether this is acceptable depends on the expected cardinality — worth
benchmarking on production data before shipping.

A lateral join or a pre-aggregated materialized view would avoid the repeated scans.

---

## Minor Issues

### 5. `fullLat` / `fullLng` branches use inclusive inequalities, general case uses exclusive

**File:** [api/services/MassifCoordinatesSnapshotService.js](api/services/MassifCoordinatesSnapshotService.js#L92)

```js
if (fullLat) {
  return coordinates.filter(
    (coord) => coord[0] >= swLng && coord[0] <= neLng  // inclusive
  );
}
return coordinates.filter(
  (coord) =>
    coord[0] > swLng && coord[0] < neLng &&            // exclusive
    coord[1] > swLat && coord[1] < neLat
);
```

The comment explaining strict inequalities only covers the general branch. The
`fullLat`/`fullLng` special cases silently use inclusive bounds without explanation. In
practice, coordinates sitting exactly on ±180° longitude or ±90° latitude don't exist in
cave data, so the divergence has no real-world impact — but a comment justifying the
asymmetry would prevent future confusion.

### 6. Empty test file committed

**File:** [test/integration/1_services/MassifCoordinatesSnapshotService.property.test.js](test/integration/1_services/MassifCoordinatesSnapshotService.property.test.js)

The file is 0 bytes. A test runner will collect it, report 0 tests, and give silent
false confidence. Either add a `describe` skeleton with `it.todo(...)` entries, or
delete the file until the tests are ready.

### 7. No automatic recovery from a failed initial snapshot load

**File:** [config/bootstrap.js](config/bootstrap.js)

If `load()` fails at bootstrap, `coordinates` stays `null` and there is no retry —
every request falls back to the DB for the lifetime of the process. This is safe but
means an availability event during server startup permanently degrades the endpoint
performance until the next restart.

Low severity since the DB fallback is correct and the boot failure is logged. Worth
documenting in the service's JSDoc so operators know a restart is needed to restore
the snapshot.

---

## Positive Points

- **`ORDER BY id ASC` added to all paginated dbSync queries** — Important correctness
  fix. Without a stable sort, `OFFSET`/`LIMIT` pagination can return duplicate or
  missing rows when the planner changes the execution plan between pages.

- **`MassifCoordinatesSnapshotService` stale-while-revalidate pattern** — The design
  (module-level state, `loadPromise` deduplication, background refresh on TTL expiry)
  is well-implemented and clearly commented. The note about Sails `_.bindAll()` is a
  good explanation for a non-obvious choice.

- **`resync-search.js` script** — Clean and focused. Minimal Sails boot (no views,
  sockets, or http hooks), `--export` flag, graceful `process.exit`. Well-documented in
  the README.

- **Log level standardisation** — `sails.log()` → `sails.log.info()` and `[dbSync]`
  prefix throughout makes production log filtering much easier.

- **`MassifCoordinatesSnapshotService.test.js`** — Solid coverage: load deduplication,
  stale-while-revalidate trigger, all bounding-box branches, TTL expiry, error path.

- **`massif.js` GROUP BY and duplicate `m.id` fixes** — Both are real correctness bugs
  fixed incidentally; worth calling out in the PR description.

---

## Summary Table

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | Medium | `find-massifs-coordinates.js` | Cache-Control `max-age=TTL` during boot window (DB fallback) |
| 2 | Medium | `MassifCoordinatesSnapshotService.js` | `clear()` → `lastRefreshedAt=null` causes same stale Cache-Control |
| 3 | Low | `MassifCoordinatesSnapshotService.js` | `clear()` name/log imply empty state, but snapshot still served |
| 4 | Low | `GeoLocService.js` | Correlated subqueries for `entranceCount`/`networkCount` — O(n) per massif |
| 5 | Low | `MassifCoordinatesSnapshotService.js` | `fullLat`/`fullLng` use inclusive vs exclusive bounds, undocumented |
| 6 | Low | `MassifCoordinatesSnapshotService.property.test.js` | Empty test file |
| 7 | Info | `config/bootstrap.js` | No retry after failed initial load; restart required to restore snapshot |

ClemRz:

**1 Cache-Control `max-age=TTL` during boot window** ✅ Fixed
Both `find-massifs-coordinates.js` and `find-entrances-coordinates.js` now emit `max-age=0` when `lastRefreshed` is null, preventing CDNs from caching DB-fallback responses.

**2 `clear()` → `lastRefreshedAt=null` causes stale Cache-Control** ✅ Fixed
`invalidate()` (renamed from `clear()`) no longer resets `lastRefreshedAt`, so the controller computes a correct increasing age while the background refresh runs.

**3 `clear()` name/log misleading** ✅ Fixed
Renamed to `invalidate()` with accurate log message `"TTL invalidated, triggering background refresh"`. Updated all callers and tests.

**4 Correlated subqueries for `entranceCount`/`networkCount`** ⏳ Deferred
Agreed this is worth benchmarking on production data. Will profile separately and optimize if needed.

**5 `fullLat`/`fullLng` inclusive vs exclusive bounds undocumented** ✅ Fixed
Added comment explaining why full-range branches use inclusive bounds (no adjacent tile beyond geographic limits).

**6 Empty test file** ✅ Fixed
Deleted the 0-byte `MassifCoordinatesSnapshotService.property.test.js`.

**7 No retry after failed initial load** ✅ Fixed
Added JSDoc note to both snapshot services documenting that a restart is needed to restore the snapshot after a bootstrap failure.

Note: fixes for 1–3 and 5 were also applied to `CoordinatesSnapshotService` / `find-entrances-coordinates.js` which had the same patterns. Also fixed a flaky Property 8 test where `fc.double` produced subnormal values that broke URL serialization — switched to `fc.integer` for out-of-range arbitraries.

Paul-AUB (approved):

PR #1504, opened 2026-03-19
fix(document): set multer defParamCharset to utf8 and repair corrupted filenames

### LINKED ISSUE #1503: File names with accented characters are garbled in API responses
# Bug: File names with accented characters are garbled in API responses

## Summary

File names containing non-ASCII characters (accented Latin characters, etc.) are returned garbled by the API. For example, a file named `Soupirail d'entrée de l'Antre de Vénus.jpg` is received by the frontend as `Soupirail d'entr©e de l'Antre de V©nus.jpg`.

## Root cause

This is a classic Latin-1 / UTF-8 mojibake. The file names are stored as UTF-8 in the database, but at some point in the response pipeline (serialisation, ORM, or HTTP layer) the raw UTF-8 bytes are treated as Latin-1 / ISO-8859-1 code points. Each multi-byte UTF-8 sequence is therefore split into individual Latin-1 characters:

| Original | UTF-8 bytes | Received as |
|---|---|---|
| `é` (U+00E9) | `0xC3 0xA9` | `Ã©` (U+00C3 U+00A9) |
| `è` (U+00E8) | `0xC3 0xA8` | `Ã¨` |
| `ü` (U+00FC) | `0xC3 0xBC` | `Ã¼` |

## Expected behaviour

The API should return file names as properly encoded UTF-8 strings. The JSON response body is already UTF-8 (as declared by `Content-Type: application/json; charset=utf-8`), so the file name strings should be correct Unicode — no re-encoding needed.

## Affected endpoints

Any endpoint that returns file metadata with a `fileName` field, in particular:

- `GET /api/v1/documents/:id` — `data.files[].fileName`
- Any other endpoint returning file attachment metadata

## How to reproduce

1. Upload a document with a file whose name contains accented characters (e.g. `Entrée.pdf`).
2. Fetch the document via `GET /api/v1/documents/:id`.
3. Inspect `data.files[0].fileName` in the JSON response — it will be garbled.

## Current workaround

The frontend applies a client-side re-decode (`TextDecoder('utf-8', { fatal: true })`) to recover the original string. This is a best-effort workaround with a known edge case: a filename whose bytes are simultaneously valid Latin-1 and valid UTF-8 would be wrongly decoded.

This workaround should be removed once the backend is fixed.

## Suggested fix

Ensure the ORM / serialiser layer returns `fileName` as a Unicode string without re-interpreting its bytes. Depending on the stack, this may mean:

- Setting the database connection charset to `utf8mb4` (MySQL/MariaDB)
- Ensuring the ORM reads text columns as UTF-8
- Or explicitly decoding the value with `str.encode('latin-1').decode('utf-8')` (Python) before serialising to JSON

ClemRz:
## 🤔 What

- Fix non-ASCII filename encoding (mojibake) in file uploads
- Repair existing corrupted filenames in the database
- Closes #1503

## 🤷‍♂️ Why

Multer's default `defParamCharset` is `'latin1'`, which causes UTF-8 filenames containing accented characters (é, ü, ö, ñ, etc.) to be misinterpreted as Latin-1 before storage. For example, `Entrée.pdf` gets stored as `EntrÃ©e.pdf`.

## 🔍 How

1. **Multer config** (`config/http.js`): Added `defParamCharset: 'utf8'` to the multer options so incoming filenames are decoded as UTF-8 instead of Latin-1.

2. **SQL migration** (`sql/93_02_fix_filename_encoding.sql`): A PL/pgSQL block iterates over `t_file` rows containing non-ASCII characters and applies `convert_from(convert_to(filename, 'LATIN1'), 'UTF8')` to reverse the mojibake. Rows where conversion fails (e.g., CJK characters already stored correctly) are skipped gracefully. The migration is idempotent.

3. **Property-based tests** (`test/integration/1_services/FileService.encoding.property.test.js`):
   - Bug condition exploration tests (11 concrete cases) demonstrating the mojibake problem
   - Preservation tests (5 fast-check PBT tests) verifying ASCII-only filenames and record field integrity are unaffected

## 🧪 Testing

All 1774 existing tests pass, plus 16 new tests (11 bug condition + 5 preservation).

```bash
PORT=1338 npm run test
```

## 🚀 Deployment Procedure

1. **Deploy the code** — the multer `defParamCharset: 'utf8'` change takes effect immediately for new uploads
2. **Run the SQL migration** against the production database:
   ```bash
   psql -U <user> -h <host> -d grottoce -f sql/93_02_fix_filename_encoding.sql
   ```
   This repairs existing corrupted filenames. It is idempotent and safe to run multiple times.
3. **Frontend workaround** — the `TextDecoder` re-decode workaround in grottocenter-front can be removed after this backend fix is deployed

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/config/http.js
+++ b/config/http.js
@@ -122,7 +122,16 @@ module.exports.http = {
     fileMiddleware: (function () {
       const inMemoryStorage = multer.memoryStorage();
       // File size is 100 Mo (Mb)
-      const upload = multer({ storage: inMemoryStorage, fileSize: 100000000 });
+      //
+      // defParamCharset: 'utf8' fixes mojibake for non-ASCII filenames.
+      // It only affects the legacy `filename=` parameter in Content-Disposition
+      // headers (RFC 2047). Modern browsers using RFC 5987 `filename*=UTF-8''…`
+      // are already decoded correctly by busboy regardless of this setting.
+      const upload = multer({
+        storage: inMemoryStorage,
+        fileSize: 100000000,
+        defParamCharset: 'utf8',
+      });
       return upload.fields([{ name: 'files' }]);
     })(),
 
--- a/sql/93_02_fix_filename_encoding.sql
+++ b/sql/93_02_fix_filename_encoding.sql
@@ -1,0 +1,50 @@
+\c grottoce;
+
+-- Fix Latin-1/UTF-8 mojibake in t_file.filename
+--
+-- Multer's default defParamCharset was 'latin1', causing UTF-8 filenames
+-- to be misinterpreted as Latin-1 before storage. This migration re-encodes
+-- the corrupted bytes back to proper UTF-8.
+--
+-- How it works:
+--   convert_to(filename, 'LATIN1') — treats each character as a Latin-1 code
+--     point and emits the corresponding byte (reversing the original mis-decode)
+--   convert_from(..., 'UTF8') — re-interprets those bytes as UTF-8
+--
+-- Safety:
+--   - ASCII-only filenames are unchanged (ASCII is identical in both encodings)
+--   - The WHERE clause makes this idempotent (no-op on already-fixed rows)
+--   - Rows with characters outside Latin-1 range (e.g. CJK already stored
+--     correctly) will fail convert_to(..., 'LATIN1'), so we skip them
+--     gracefully using a PL/pgSQL block with EXCEPTION handling.
+
+BEGIN;
+
+DO $$
+DECLARE
+  rec RECORD;
+  fixed_name TEXT;
+BEGIN
+  FOR rec IN
+    SELECT id, filename
+    FROM t_file
+    WHERE filename ~ '[^\x00-\x7F]'
+  LOOP
+    BEGIN
+      fixed_name := convert_from(convert_to(rec.filename, 'LATIN1'), 'UTF8');
+      -- 200-char guard matches t_file.filename varchar(200) / Waterline maxLength: 200.
+      -- In practice, UTF-8 re-encoding produces shorter strings (multi-byte mojibake
+      -- chars collapse back to single code points), so this is a safety net, not a filter.
+      IF fixed_name IS DISTINCT FROM rec.filename
+         AND length(fixed_name) <= 200
+      THEN
+        UPDATE t_file SET filename = fixed_name WHERE id = rec.id;
+      END IF;
+    EXCEPTION WHEN OTHERS THEN
+      -- Skip rows where conversion fails (characters outside Latin-1 range)
+      RAISE NOTICE 'Skipping t_file id=% filename=%: %', rec.id, rec.filename, SQLERRM;
+    END;
+  END LOOP;
+END $$;
+
+COMMIT;
--- a/test/integration/1_services/FileService.encoding.property.test.js
+++ b/test/integration/1_services/FileService.encoding.property.test.js
@@ -1,0 +1,362 @@
+/* eslint-disable func-names */
+const should = require('should');
+const fc = require('fast-check');
+const FileService = require('../../../api/services/FileService');
+
+/**
+ * Bug Condition Exploration — Non-ASCII Filename Mojibake
+ *
+ * Demonstrates that multer's default `defParamCharset: 'latin1'` causes
+ * UTF-8 filenames to be misinterpreted as Latin-1, producing mojibake.
+ *
+ * The bug lives in the multer middleware layer (config/http.js), not in
+ * FileService itself. FileService is a passthrough — it stores whatever
+ * `file.originalname` it receives. When multer decodes UTF-8 bytes as
+ * Latin-1, the garbled string flows through FileService into the database.
+ *
+ * These tests use concrete counterexamples to surface the encoding problem.
+ *
+ * Validates: Requirements 1.1, 1.2, 2.1, 2.2
+ */
+describe('FileService — Property 1: Bug Condition — Non-ASCII Filename Mojibake', () => {
+  // -------------------------------------------------------------------------
+  // Test 1: Pure encoding demonstration (no DB)
+  // Shows that UTF-8 bytes decoded as Latin-1 produce mojibake.
+  // -------------------------------------------------------------------------
+  describe('UTF-8 bytes decoded as Latin-1 produce mojibake', () => {
+    const cases = [
+      { original: 'Entrée.pdf', expectedMojibake: 'EntrÃ©e.pdf' },
+      { original: 'Höhle.pdf', expectedMojibake: 'HÃ¶hle.pdf' },
+      { original: 'über.pdf', expectedMojibake: 'Ã¼ber.pdf' },
+      { original: 'café.pdf', expectedMojibake: 'cafÃ©.pdf' },
+    ];
+
+    cases.forEach(({ original, expectedMojibake }) => {
+      it(`should garble "${original}" into "${expectedMojibake}" when decoded as Latin-1`, () => {
+        // Simulate what multer does: take UTF-8 string, get its bytes,
+        // then decode those bytes as Latin-1
+        const utf8Bytes = Buffer.from(original, 'utf8');
+        const mojibake = utf8Bytes.toString('latin1');
+
+        // The mojibake should NOT equal the original
+        should(mojibake).not.equal(
+          original,
+          `Latin-1 decoding of UTF-8 bytes should produce mojibake, but got the original back`
+        );
+
+        // The mojibake should match the expected garbled string
+        should(mojibake).equal(
+          expectedMojibake,
+          `Expected mojibake "${expectedMojibake}" but got "${mojibake}"`
+        );
+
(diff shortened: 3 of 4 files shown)


Paul-AUB (changes requested):
# Code Review — `fix/filename-encoding-1503`

## Overview

1 commit, 3 files changed. The branch fixes a UTF-8 mojibake bug in multer's filename
decoding, adds a one-shot SQL migration to repair existing corrupted rows in `t_file`,
and introduces property-based integration tests documenting the bug.

---

## Bugs / Correctness Issues

### 1. `defParamCharset` has no effect on `filename*` (RFC 5987) parameters

**File:** [config/http.js](config/http.js#L125)

`defParamCharset` controls the fallback decoder for unquoted legacy `filename=`
parameters in `Content-Disposition`. Browsers that follow RFC 5987 send
`filename*=UTF-8''...` (percent-encoded), which busboy always decodes correctly
regardless of `defParamCharset`.

The fix is correct for old/legacy clients that send a bare `filename=` without encoding,
but it provides no safety net for RFC 5987 paths (already fine) or for clients that send
raw binary bytes in the legacy field (rare but possible). Document this scope explicitly
so future maintainers understand what the option covers.

### 2. Migration not wrapped in a transaction

**File:** [sql/93_02_fix_filename_encoding.sql](sql/93_02_fix_filename_encoding.sql)

The `DO $$` block iterates and `UPDATE`s rows one by one. If the script is interrupted
mid-run (crash, `pg_cancel_backend`, timeout), some rows will have been fixed and others
not. The script is idempotent so re-running is safe, but silence around partial
application can cause confusion.

**Suggestion:** Wrap the `DO` block in an explicit `BEGIN` / `COMMIT`:

```sql
BEGIN;
DO $$ … END $$;
COMMIT;
```

### 3. `length(fixed_name) <= 200` limit may silently skip valid filenames

**File:** [sql/93_02_fix_filename_encoding.sql](sql/93_02_fix_filename_encoding.sql#L38)

The migration skips any corrected filename longer than 200 characters. UTF-8 → Latin-1
decoding expands multi-byte characters (e.g. `é` becomes two Latin-1 chars), so a
filename stored near the column limit could decode to a correct UTF-8 string that is
_shorter_ than the corrupted one — meaning the 200-char guard on `fixed_name` is safe in
the common direction, but it is worth verifying against the actual column DDL
(`t_file.filename` `varchar(200)`?) to ensure no valid filenames near the limit are
silently left uncorrected.

---

## Design Concerns

### 4. Tests document the bug but do not verify the fix

**File:** [test/integration/1_services/FileService.encoding.property.test.js](test/integration/1_services/FileService.encoding.property.test.js)

The test file is split into two groups:

- **Property 1** — injects a pre-computed mojibake string directly into
  `FileService.document.create()`, bypassing multer entirely. These tests pass on
  _both_ fixed and unfixed code because they never exercise the middleware.
- **Property 2** — verifies ASCII-only filename baseline behaviour, which also does not
  change with the fix.

Neither group sends an actual multipart HTTP request through the middleware stack. As a
result, the only assertion that `defParamCharset: 'utf8'` actually corrects the
filename decoding is absent from the test suite. A route-level integration test
(e.g. using `supertest` to POST a multipart form with a non-ASCII filename and asserting
the stored `fileName` matches the original) would close this gap.

### 5. `FileService.document.create(file, 1, true)` assumes document id=1 exists

**File:** [test/integration/1_services/FileService.encoding.property.test.js](test/integration/1_services/FileService.encoding.property.test.js#L81)

The hardcoded `idDocument = 1` will trigger a foreign-key constraint error if no document
with that id is present in the test database, causing misleading test failures unrelated
to encoding. Use a fixture document created in a `before` hook or pick an id known to
exist in the test seed data.

---

## Minor Issues

### 6. Azure blob URL hardcoded in property test assertion

**File:** [test/integration/1_services/FileService.encoding.property.test.js](test/integration/1_services/FileService.encoding.property.test.js#L310)

```js
should(url).containEql('grottocenter.blob.core.windows.net/documents/');
```

This ties the test to a specific deployment environment. If the Azure URL is in
`sails.config`, derive it from there instead. At minimum, the assertion should test the
_structure_ of the URL (contains the path segment `documents/` and a non-empty blob
name), not the hardcoded hostname.

### 7. Property 2 runs 500 DB round-trips

**File:** [test/integration/1_services/FileService.encoding.property.test.js](test/integration/1_services/FileService.encoding.property.test.js#L196)

Five `fc.asyncProperty` tests each run `numRuns: 100`, creating and destroying 500
`t_file` rows. For properties that do not require random variation (e.g. `isValidated`
always defaults to `true` regardless of input), a single concrete assertion achieves the
same confidence with far less overhead. Consider lowering `numRuns` to 10–20 for these
structural properties, or replacing them with plain `it()` tests.

### 8. Database name `grottoce` hardcoded in migration

**File:** [sql/93_02_fix_filename_encoding.sql](sql/93_02_fix_filename_encoding.sql#L1)

```sql
\c grottoce;
```

The `\c` command is a psql meta-command that works only when the script is run via
`psql` directly. If a migration runner connects to the database before executing the
script (as most CI pipelines do), the `\c` line is either ignored or causes an error.
If the database name differs between environments (staging, review apps), the script
would connect to the wrong database silently. Remove the `\c` line and rely on the
connection string provided by the migration runner.

---

## Positive Points

- **Root cause correctly identified and fixed** — `defParamCharset: 'utf8'` is passed
  through multer v2 → busboy v1 → `parseDisposition`, which is the exact call site that
  decodes the `filename=` parameter in `Content-Disposition`. Confirmed by reading
  `node_modules/multer/lib/make-middleware.js` and `node_modules/busboy/lib/types/multipart.js`.

- **SQL migration logic is correct** — `convert_from(convert_to(filename, 'LATIN1'),
  'UTF8')` is the exact inverse of the original mis-decoding. ASCII-only filenames are
  unaffected (ASCII is a subset of both encodings). The per-row `EXCEPTION` block
  gracefully skips rows with characters outside the Latin-1 range (e.g. CJK already
  stored correctly).

- **Idempotency** — The `WHERE filename ~ '[^\x00-\x7F]'` clause combined with
  `IS DISTINCT FROM` ensures re-running the migration is a no-op on already-fixed rows.

- **Property 1 tests are good bug documentation** — The encoding demonstration tests
  (Test 1) clearly show the mojibake mechanics and will serve as a regression reference.
  The `try/finally` cleanup pattern is correct throughout.

---

## Summary Table

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | Low | `config/http.js` | `defParamCharset` scope limited to legacy `filename=`; document it |
| 2 | Low | `93_02_fix_filename_encoding.sql` | No explicit transaction — partial updates on interruption |
| 3 | Low | `93_02_fix_filename_encoding.sql` | 200-char guard should be verified against actual column DDL |
| 4 | Medium | `FileService.encoding.property.test.js` | No E2E test verifying the multer fix end-to-end |
| 5 | Medium | `FileService.encoding.property.test.js` | `idDocument=1` hardcoded — fragile FK dependency |
| 6 | Low | `FileService.encoding.property.test.js` | Azure hostname hardcoded in URL assertion |
| 7 | Info | `FileService.encoding.property.test.js` | 500 DB writes for structural properties; lower `numRuns` |
| 8 | Low | `93_02_fix_filename_encoding.sql` | `\c grottoce` hardcodes DB name; incompatible with migration runners |

ClemRz (commented):

**Fixed (items 1, 2, 3, 4, 5, 7):**

- **1** — Added a comment in `config/http.js` documenting that `defParamCharset` only affects the legacy `filename=` parameter (RFC 2047), and that RFC 5987 `filename*=UTF-8''…` paths are already handled correctly by busboy.
- **2** — Wrapped the `DO $$` block in an explicit `BEGIN` / `COMMIT` transaction.
- **3** — Added a comment clarifying the 200-char guard matches `t_file.filename varchar(200)` (confirmed via `TFile.fileName.maxLength: 200` in the Waterline model). In practice UTF-8 re-encoding produces *shorter* strings, so this is a safety net.
- **4** — Added a route-level E2E test (`test/integration/4_routes/Documents/create-with-encoding.test.js`) that sends an actual multipart HTTP request with a UTF-8 filename (`Entrée.pdf`) through the full Sails middleware pipeline and asserts the stored `fileName` matches the original. This closes the gap.
- **5** — Added comments referencing the fixture (`test/fixtures/tdocument.json`, id=1 "Spelunca [COLLECTION]") wherever `idDocument = 1` is used.
- **7** — Lowered `numRuns` from 100 to 20 for the three structural properties (`isValidated`, `fileFormat`, `getUrl`) that don't benefit from high random variation. Kept 100 for the two properties that exercise filename content (`fileName` storage and `path` generation).

**Not changed (items 6, 8):**

- **6** — The Azure hostname is hardcoded as a constant in `FileService.js` itself (not configurable). The existing `FileService.test.js` uses the same assertion pattern. Changing this would be inconsistent with the rest of the codebase.
- **8** — `\c grottoce;` is a project convention.

Paul-AUB (approved):

PR #1505, opened 2026-03-20
feat(auth): add iat-based token blacklist for immediate role revocation

### LINKED ISSUE #378: [AUTH TOKEN] Automatically refresh it
Currently, the auth token is valid only for 24 hours. We should revalidate it everytime the user perform an action for example (simple proposition, not sure if it's the correct pattern). 
Maybe we should use 2 tokens: authToken and refreshToken, see this article => [https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them/](https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them/)

At least, the user must be aware that his token is not valid anymore: there is no information displayed in the interface about this currently.

ClemRz:
## 🤔 What

Adds an `iat`-based token blacklist that invalidates JWTs immediately when a user's roles change.

Closes #378

## 🤷‍♂️ Why

The API uses stateless JWTs with a 90-day TTL. When an administrator changes a user's roles (add/remove/set groups), the change only takes effect after the current token expires — up to 90 days later. There is no token revocation mechanism.

## 🔍 How

- New `t_token_blacklist` table with `id_caver` (PK, FK to `t_caver`) and `revoked_before` (timestamptz)
- `BlacklistService` with an in-memory `Map` cache for O(1) lookups:
  - `revoke(caverId)` — UPSERT to DB + update cache
  - `isRevoked(caverId, iat)` — check `iat < Math.ceil(revoked_before)`
  - `loadCache()` — populate cache from DB on bootstrap (blocking)
- `parseAuthToken` middleware in `config/http.js` now:
  - Rejects tokens missing the `iat` claim with 401
  - Rejects tokens where `iat < revoked_before` with 401
  - Uses `sails.services.blacklistservice` (Sails global instance) to avoid dual-instance cache issues
- Controllers `put-on-group`, `remove-from-group`, `set-groups` call `BlacklistService.revoke()` after modifying groups
- Cache loaded in `config/bootstrap.js` before the server accepts requests
- Fully backward-compatible: empty blacklist table on deploy means no existing tokens are affected

## 🚀 Deployment

Run the SQL migration against the production database **before** deploying the new code:

```sql
CREATE TABLE IF NOT EXISTS t_token_blacklist (
    id_caver integer NOT NULL,
    revoked_before timestamp with time zone NOT NULL,
    CONSTRAINT t_token_blacklist_pkey PRIMARY KEY (id_caver),
    CONSTRAINT t_token_blacklist_id_caver_fkey FOREIGN KEY (id_caver)
        REFERENCES t_caver(id) ON DELETE CASCADE
);
```

The migration is idempotent (`IF NOT EXISTS`), so it's safe to run multiple times.

Once the table exists, deploy the application normally. On startup, `BlacklistService.loadCache()` will read the (empty) table into memory. No existing tokens are affected — the blacklist starts empty, so all current sessions continue working.

No rollback concerns: if the code is reverted, the table simply goes unused. It can be dropped later if needed.

## 🧪 Testing

- 5 property-based tests (fast-check) validating correctness properties of `BlacklistService`
- Unit tests for `BlacklistService` edge cases
- Integration tests for the middleware (revoked token → 401, fresh token → 200, missing iat → 401)
- Route integration tests verifying revocation triggers on all 3 group endpoints
- Requires `npm run dev:clean` to rebuild Docker containers (new SQL migration)

```shell
npm run dev:clean
npm run dev:up
PORT=1338 npm run test
```

## 📸 Previews

N/A — backend only, no UI changes.


Paul-AUB (changes requested):
# Code Review: feat(auth): add iat-based token blacklist for immediate role revocation

Commit: `7feb543`

## Overview

The feature is well-designed: IAT-based blacklist pattern with in-memory cache + DB persistence, clean and efficient. Test coverage (unit, property-based, integration) is excellent. A few points deserve attention.

---

## Issues

### 1. Subtle race condition in `revoke()` — `BlacklistService.js:118-127`

```js
await CommonService.query(`... VALUES ($1, NOW()) ...`, [caverId]);
const now = new Date();
cache.set(caverId, now);
```

The cache is populated with `new Date()` *after* the SQL call resolves. The DB timestamp (`NOW()`) is therefore slightly earlier than the cache timestamp. If a token has an `iat` falling within that millisecond gap:
- `isRevoked()` returns `false` (accepted per the cache)
- But after the next `loadCache()`, it would be rejected

In practice this is negligible (JWT `iat` values are whole seconds), but it is a real inconsistency.
**Simple fix:** retrieve the timestamp from the SQL response instead of calling `new Date()`.

### 2. Overly permissive CORS — and out of scope — `config/http.js:200-214`

```js
res.set('Access-Control-Allow-Origin', '*');
res.set('Access-Control-Allow-Headers', 'content-type, authorization');
```

- `Access-Control-Allow-Origin: *` combined with `authorization` in `Allow-Headers`: browsers will not send credentials with a wildcard, but any origin can make non-credentialed requests. This is likely a regression against the existing CORS configuration.
- **This change is out of scope for this commit** (it has nothing to do with the blacklist). It should live in a separate commit, and the existing CORS config (`config/security.js` or equivalent) should be updated rather than short-circuited by a custom middleware.

### 3. Property 5 stub is probably a no-op — `blacklist-service.property.test.js:595`

```js
const queryStub = sinon
  .stub(sails, 'sendNativeQuery')
  .rejects(new Error(errorMsg));
```

The test assumes `CommonService.query` calls `sails.sendNativeQuery` internally. If `CommonService.query` wraps something else (e.g. a `pg` pool directly), the stub does nothing and `BlacklistService.revoke()` will not throw — the test would pass silently without testing what it claims to test.
**Action:** verify the internal implementation of `CommonService.query` and adjust the stub target accordingly.

### 4. No blacklist cleanup — unbounded growth

The `t_token_blacklist` table grows indefinitely. Once tokens expire (e.g. after 24 h), their blacklist entries are no longer useful. A periodic cleanup job (or a TTL derived from token expiry) would prevent the table and cache from growing without bound. Not blocking now, but worth planning for.

### 5. Inconsistent service access in route tests — `put-on-group.test.js:734` vs `745`

```js
// afterEach — via sails.services:
sails.services.blacklistservice.getCache().delete(testCaverId);

// inside the test — via direct require:
BlacklistService.getCache().delete(testCaverId);
```

Two different ways of accessing the service. If Sails and `require()` return different instances (which can happen with Sails' module loading), the cleanup and assertions would not target the same cache object.
**Fix:** use `sails.services.blacklistservice` consistently everywhere in test files.

---

## Positives

- IAT-based approach is correct and avoids storing full token strings.
- `loadCache()` is blocking at bootstrap (intentional and documented) — the right call to guarantee consistency before accepting requests.
- `ON CONFLICT (id_caver) DO UPDATE`: clean UPSERT, no risk of duplicate inserts.
- `ON DELETE CASCADE` in the SQL migration: consistent with caver deletion.
- The atomicity property test (Property 5) is a good idea, even if the stub target is suspect.
- Middleware order (`corsHeaders` before `parseAuthToken`) is correct for OPTIONS preflight handling.

ClemRz:
1. Fixed
2. Intentional
3. No fix needed. `CommonService.query` directly delegates to `sails.sendNativeQuery`, so the stub is correct
4. Skipped
5. Fixed

Paul-AUB (approved):

PR #1509, opened 2026-03-30
feat(entrance): add data quality score to entrance detail endpoint

### LINKED ISSUE #1506: Get entrance quality score in entrance API
# Show data quality score on individual entrance pages

## Context

GrottoCenter already computes a data quality score (0–100) for each entrance. This score is currently displayed in the entrance lists accessible from massif and country pages, where each entrance is shown with a circular progress indicator colored according to the score (red < 40, orange 40–70, green > 70).

However, this score is **not visible on the individual entrance detail page**, which is the primary page most users land on when consulting or editing a cave entry. This creates a blind spot: contributors have no immediate feedback on the quality of the data they are viewing or about to update, and casual users cannot assess how reliable the information is.

Displaying the score on the detail page would:
- Help contributors identify what is missing and prioritize their edits
- Give all users a quick signal of data reliability without navigating to a list page
- Create visual consistency between list and detail views (the circular indicator is already a recognized pattern in the app)

---

### Option A — Include `data_quality` in `GET /api/v1/entrances/:id` (recommended)

Add the quality score directly to the existing entrance response payload.

**Pros:**
- Single request, no waterfall
- Consistent with how other entrance metadata (`altitude`, `precision`, etc.) is fetched
- Zero additional frontend data-fetching logic needed — the score is already in the Redux store once the entrance loads

**Cons:**
- Depends on the cost of the compute/get in db

---

### Option B — New endpoint `GET /api/v1/entrances/:id/quality`

Create a dedicated endpoint that returns only the quality score for a given entrance.

**Pros:**
- No change to the existing entrance response contract
- Score can be loaded lazily with its own loading skeleton

**Cons:**
- Extra HTTP request on every entrance page load
- Requires a new backend endpoint + new Redux action + new reducer on the frontend
- More complexity for what is ultimately a single numeric value

ClemRz:
## 🤔 What

Add the data quality score and per-category breakdown to the entrance detail endpoint (`GET /api/v1/entrances/:id`). Normalize all quality scores to 0–100 across both detail and list endpoints.

- Closes #1506
- Refactor `computeEntranceDataQuality.js`: centralize category list, scoring tier constants, fix `moment().subtract()` mutation bug, normalize scores to 0–100
- Add `getEntranceQualityById` to `DataQualityComputeService` for single-entrance lookup
- Extend `toEntrance` converter with `dataQuality` field (total + 7 category breakdown + dateOfUpdate)
- Add error logging to all `DataQualityComputeService` catch blocks
- Update Swagger spec with `DataQuality` schema
- Add 4 property-based tests and 2 integration tests

## 🤷‍♂️ Why

The quality score was only visible in list views (massif/country pages) but not on the individual entrance detail page. Contributors had no immediate feedback on data quality when viewing or editing an entrance.

The frontend already assumes a 0–100 scale, but the backend was returning raw 0–98 values. This PR aligns both.

## 🔍 How

**Score computation refactor** (`api/utils/computeEntranceDataQuality.js`):
- Extracted `QUALITY_CATEGORIES` array and all scoring tier/threshold constants (no more magic numbers)
- Fixed a subtle bug where `moment().subtract()` was mutating the same moment object across comparisons — replaced with `moment().diff()` 
- `getQualityData()` now iterates over `QUALITY_CATEGORIES` and returns `Math.round(raw / 98 * 100)`
- New `getQualityBreakdown()` returns per-category normalized scores (0–100 each)
- Changed module export from single function to named exports

**Service layer** (`api/services/DataQualityComputeService.js`):
- New `getEntranceQualityById(entranceId)` using raw SQL (not Waterline — materialized views lack a real primary key, so `findOne` doesn't work reliably)
- Added `sails.log.error(e)` to all 7 catch blocks (6 existing + 1 new)

**Controller** (`api/controllers/v1/entrance/find.js`):
- Quality data fetched in parallel with `getPopulatedEntrance` via `Promise.all` — no additional latency
- Attached as `entrance.qualityData` before passing to converter

**Converter** (`api/services/mapping/converters.js`):
- `toEntrance` now outputs `dataQuality: { total, categories, dateOfUpdate }` or `null`
- List endpoints automatically return normalized 0–100 scores since they call the same `getQualityData()`

## 🧪 Testing

```bash
PORT=1338 npm run test
```

4 property-based tests verify score bounds, total consistency, and list/detail parity. 2 integration tests verify the endpoint returns quality data (or null when missing).

## 📸 Previews

Response for `GET /api/v1/entrances/:id`:
```json
{
  "dataQuality": {
    "total": 71,
    "categories": {
      "general": 100,
      "location": 71,
      "description": 50,
      "document": 71,
      "rigging": 50,
      "history": 71,
      "comment": 100
    },
    "dateOfUpdate": "2026-03-30T10:20:00.000Z"
  }
}
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -57,6 +57,7 @@ module.exports = {
     TRigging: 'readonly',
     TSubject: 'readonly',
     TType: 'readonly',
+    VDataQualityComputeEntrance: 'readonly',
     sails: 'readonly',
   },
   rules: {
--- a/api/controllers/v1/entrance/find.js
+++ b/api/controllers/v1/entrance/find.js
@@ -1,4 +1,5 @@
 const ControllerService = require('../../../services/ControllerService');
+const DataQualityComputeService = require('../../../services/DataQualityComputeService');
 const EntranceService = require('../../../services/EntranceService');
 const RightService = require('../../../services/RightService');
 const {
@@ -18,16 +19,22 @@ module.exports = async (req, res) => {
   if (!hasRight) where.isDeleted = false;
 
   const params = { searchedItem: `Entrance of id ${entranceId}` };
-  const entrance = await EntranceService.getPopulatedEntrance(
-    entranceId,
-    where
-  );
+
+  // Fetch entrance and quality data in parallel.
+  // The quality query may return null for deleted entrances (the materialized
+  // view filters is_deleted = false) or for newly created entrances not yet
+  // in the view. In 404 cases the quality result is simply discarded.
+  const [entrance, qualityRow] = await Promise.all([
+    EntranceService.getPopulatedEntrance(entranceId, where),
+    DataQualityComputeService.getEntranceQualityById(entranceId),
+  ]);
 
   if (!entrance) return res.notFound(`${params.searchedItem} not found`);
+
   return ControllerService.treatAndConvert(
     req,
     null,
-    entrance,
+    { ...entrance, qualityData: qualityRow },
     params,
     res,
     entrance.isDeleted && !hasRight ? toDeletedEntity : toEntrance
--- a/api/services/DataQualityComputeService.js
+++ b/api/services/DataQualityComputeService.js
@@ -74,6 +74,7 @@ module.exports = {
       const queryResult = await CommonService.query(query, params);
       return queryResult.rows;
     } catch (e) {
+      sails.log.error(e);
       return null;
     }
   },
@@ -91,6 +92,7 @@ module.exports = {
       );
       return parseInt(queryResult.rows[0].count, 10);
     } catch (e) {
+      sails.log.error(e);
       return 0;
     }
   },
@@ -123,6 +125,7 @@ module.exports = {
       const queryResult = await CommonService.query(query, params);
       return queryResult.rows;
     } catch (e) {
+      sails.log.error(e);
       return null;
     }
   },
@@ -140,6 +143,7 @@ module.exports = {
       );
       return parseInt(queryResult.rows[0].count, 10);
     } catch (e) {
+      sails.log.error(e);
       return 0;
     }
   },
@@ -172,6 +176,7 @@ module.exports = {
       const queryResult = await CommonService.query(query, params);
       return queryResult.rows;
     } catch (e) {
+      sails.log.error(e);
       return null;
     }
   },
@@ -189,7 +194,26 @@ module.exports = {
       );
       return parseInt(queryResult.rows[0].count, 10);
     } catch (e) {
+      sails.log.error(e);
       return 0;
+    }
+  },
+
+  /**
+   *
+   * @param {number} entranceId
+   * @returns {Object|null} the materialized view row, or null if not found
+   */
+  getEntranceQualityById: async (entranceId) => {
+    try {
+      const queryResult = await CommonService.query(
+        'SELECT * FROM v_data_quality_compute_entrance WHERE id_entrance = $1 ORDER BY id_massif NULLS LAST LIMIT 1',
+        [entranceId]
+      );
+      return queryResult.rows[0] || null;
+    } catch (e) {
+      sails.log.error(e);
+      return null;
     }
   },
 };
--- a/api/services/mapping/converters.js
+++ b/api/services/mapping/converters.js
@@ -17,7 +17,10 @@ const {
 } = require('./utils');
 const FileService = require('../FileService');
 const RiggingService = require('../RiggingService');
-const { getQualityData } = require('../../utils/computeEntranceDataQuality');
+const {
+  getQualityData,
+  getQualityBreakdown,
+} = require('../../utils/computeEntranceDataQuality');
 
 const c = {
   toCave: (source, meta) => ({
@@ -367,6 +370,17 @@ const c = {
     result.riggings = toList('riggings', source, c.toSimpleRigging, {
       filterDeleted: false,
     });
+
+    if (source.qualityData) {
+      result.dataQuality = {
+        total: getQualityData(source.qualityData),
+        categories: getQualityBreakdown(source.qualityData),
+        lastComputedAt: source.qualityData.date_of_update,
+      };
+    } else {
+      result.dataQuality = null;
+    }
+
     return result;
   },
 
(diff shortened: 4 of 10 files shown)


Paul-AUB (changes requested):
## CRITICAL

### #1 — Non-deterministic `LIMIT 1` on multi-row materialized view

**File:** [`api/services/DataQualityComputeService.js:210`](api/services/DataQualityComputeService.js#L210)

```js
'SELECT * FROM v_data_quality_compute_entrance WHERE id_entrance = $1 LIMIT 1'
```

The materialized view has a unique index on `(id_massif, id_entrance)`, meaning a single entrance can produce **multiple rows** when it belongs to more than one massif. Without an `ORDER BY`, PostgreSQL returns an arbitrary row — and which one may change across view refreshes.

In practice the quality score columns are identical across all massif rows for a given entrance, so the impact is limited today. But the query is semantically fragile and should be made deterministic.

**Fix:**
```sql
SELECT * FROM v_data_quality_compute_entrance
WHERE id_entrance = $1
ORDER BY id_massif NULLS LAST
LIMIT 1
```

---

## MEDIUM

---

### #3 — Unconditional DB query on every entrance fetch

**File:** [`api/controllers/v1/entrance/find.js:22`](api/controllers/v1/entrance/find.js#L22)

```js
const [entrance, qualityRow] = await Promise.all([
  EntranceService.getPopulatedEntrance(entranceId, where),
  DataQualityComputeService.getEntranceQualityById(entranceId),
]);
```

The quality query fires unconditionally on every `GET /api/v1/entrances/:id`, including:
- **404 cases** — the entrance doesn't exist, `qualityRow` is fetched and immediately discarded.
- **Deleted entrances** — the materialized view filters `WHERE is_deleted = false`, so `qualityRow` will always be `null` for moderator requests on deleted entrances, yet the query still runs.

This adds a fixed DB latency cost to every entrance request. Consider guarding the quality query (e.g., fetch it only after confirming the entrance exists, or gate it with a query parameter like `?includeQuality=true`).

---

### #6 — Invalid `nullable` + `$ref` combination in OpenAPI 3.0

**File:** [`assets/swaggerV1.yaml:5438`](assets/swaggerV1.yaml#L5438)

```yaml
dataQuality:
  nullable: true
  $ref: '#/components/schemas/DataQuality'
```

In OpenAPI 3.0, `$ref` replaces all sibling keywords — `nullable: true` is silently ignored by validators and code generators. This is a known 3.0 limitation addressed in 3.1.

**Fix for OpenAPI 3.0:**
```yaml
dataQuality:
  allOf:
    - $ref: '#/components/schemas/DataQuality'
  nullable: true
```

---

## LOW

### #4 — ORM model object mutation for data plumbing

**File:** [`api/controllers/v1/entrance/find.js:29`](api/controllers/v1/entrance/find.js#L29)

```js
entrance.qualityData = qualityRow;
```

`entrance` is the raw Waterline model object. Mutating it to carry extra data into the converter is inconsistent with how the rest of the codebase handles model objects, and is fragile if `getPopulatedEntrance` is ever cached or reused.

**Suggested alternative:** pass `qualityRow` as a separate argument to the converter, or wrap it: `{ ...entrance, qualityData: qualityRow }`.

---

### #5 — Wasted query for deleted entrances (moderator path)

**File:** [`api/controllers/v1/entrance/find.js:22`](api/controllers/v1/entrance/find.js#L22)

When a moderator requests a deleted entrance, `getEntranceQualityById` always returns `null` because the materialized view filters out deleted entrances. The query still runs and its result is discarded. A comment explaining this would at least clarify the intent.

---

### #7 — No `required` fields in the `DataQuality` OpenAPI schema

**File:** [`assets/swaggerV1.yaml:5442`](assets/swaggerV1.yaml#L5442)

The `DataQuality` schema and its `categories` sub-object declare properties but mark none as `required`. Clients cannot rely on `total`, `categories`, or `dateOfUpdate` being present in a non-null `DataQuality` object according to the spec, even though they always are in practice.

**Suggested addition:**
```yaml
DataQuality:
  required: [total, categories, dateOfUpdate]
  properties:
    categories:
      required: [general, location, description, document, rigging, history, comment]
      ...
```

---

### #8 — Misleading `dateOfUpdate` semantics

**File:** [`api/services/mapping/converters.js:378`](api/services/mapping/converters.js#L378)

```js
dateOfUpdate: source.qualityData.date_of_update,
```

In the materialized view, `date_of_update` is set to `NOW()` at view **refresh time** (cron job), not at the time the entrance data was last modified. The OpenAPI spec exposes this as `format: date-time` with no description, which is ambiguous — consumers will likely assume it means "last data update."

**Suggested fix:** add a description in the spec:
```yaml
dateOfUpdate:
  type: string
  format: date-time
  description: Timestamp at which the quality scores were last recomputed (materialized view refresh time).
```

Consider renaming the field to `lastComputedAt` to make the semantics explicit.

---

### #9 — Property test calls the full `toEntrance` converter

**File:** [`test/integration/2_utils/computeEntranceDataQuality.property.test.js`](test/integration/2_utils/computeEntranceDataQuality.property.test.js)

Calling `toEntrance` in a unit-style property test (`2_utils/`) is fragile: it relies on a minimal stub that may silently fail on sub-properties if `toEntrance` is extended in the future. It would be cleaner to test `getQualityData` and `getQualityBreakdown` directly in `2_utils/` tests, and leave `toEntrance` integration to `4_routes/` tests.

ClemRz:
@Paul-AUB Thanks for the thorough review! All items addressed:

- **1 (CRITICAL)** — Added `ORDER BY id_massif NULLS LAST` to make the query deterministic.
- **3 / 5 (MEDIUM/LOW)** — Added a comment explaining the parallel fetch tradeoff (quality query may return null for deleted/404 cases but avoids sequential latency on the happy path).
- **4 (LOW)** — Replaced `entrance.qualityData = qualityRow` mutation with a spread: `{ ...entrance, qualityData: qualityRow }`.
- **6 (MEDIUM)** — Fixed the `nullable` + `$ref` issue with the `allOf` wrapper for OpenAPI 3.0 compatibility.
- **7 (LOW)** — Added `required` fields to both `DataQuality` and its `categories` sub-object.
- **8 (LOW)** — Renamed `dateOfUpdate` → `lastComputedAt` and added a description clarifying it's the materialized view refresh time, not the last data modification.
- **9 (LOW)** — Moved the converter test out of `2_utils/`. Property 1 now tests `getQualityData` and `getQualityBreakdown` directly. The `toEntrance` integration is covered by the `4_routes/` tests.

Paul-AUB (approved):
LGTM !

PR #1510, opened 2026-03-31
feat(entrance): add nine boolean characteristic flags

### LINKED ISSUE #1507: Support new boolean characteristics on entrances
# API spec: new boolean characteristics on entrances

## Context

The frontend now displays a set of boolean flags on the Entrance detail page, under the **Characteristics** section (`Properties.jsx`). These fields are rendered on the read side but the API does not yet return them, and the write endpoints do not yet accept them.

---

## New fields

All boolean fields default to `false`. `null` and absent values are treated as `false` on the frontend (the chip is hidden).

| Field | Type | Description |
|---|---|---|
| `hasBat` | boolean | The cave is a known bat habitat |
| `dangerFlooding` | boolean | Flooding risk present |
| `dangerCO2` | boolean | CO2 accumulation risk |
| `dangerRockfall` | boolean | Rockfall risk |
| `dangerPollution` | boolean | Pollution risk (water contamination etc.) |
| `needCleanGear` | boolean | Clean/decontaminated gear required (biosecurity) |
| `needStayOnTrail` | boolean | Visitors must stay on the marked trail |
| `hasRules` | boolean | Specific entry rules apply (permit, reservation…) |
| `isTouristic` | boolean | Developed/touristic cave open to the public |

---

## API changes required

**Important note : I don't know if the `/cave/*` should also be updated, same for snapshot. Please consider**

### 1. `GET /api/v1/entrances/{id}`

Add all nine fields to the `Entrance` response schema (`components/schemas/Entrance`).

Example response addition:

```yaml
hasBat:
  type: boolean
  default: false
dangerFlooding:
  type: boolean
  default: false
dangerCO2:
  type: boolean
  default: false
dangerRockfall:
  type: boolean
  default: false
dangerPollution:
  type: boolean
  default: false
needCleanGear:
  type: boolean
  default: false
needStayOnTrail:
  type: boolean
  default: false
hasRules:
  type: boolean
  default: false
isTouristic:
  type: boolean
  default: false
```

### 2. `POST /api/v1/entrances`

Add the nine fields as optional query parameters (following the same pattern as existing params like `isSensitive`, `isPublic`):

```yaml
- name: hasBat
  in: query
  description: The cave is a known bat habitat
  schema:
    type: boolean
    default: false
- name: dangerFlooding
  in: query
  description: Flooding risk present
  schema:
    type: boolean
    default: false
- name: dangerCO2
  in: query
  description: CO2 accumulation risk
  schema:
    type: boolean
    default: false
- name: dangerRockfall
  in: query
  description: Rockfall risk
  schema:
    type: boolean
    default: false
- name: dangerPollution
  in: query
  description: Pollution risk
  schema:
    type: boolean
    default: false
- name: needCleanGear
  in: query
  description: Clean/decontaminated gear required (biosecurity)
  schema:
    type: boolean
    default: false
- name: needStayOnTrail
  in: query
  description: Visitors must stay on the marked trail
  schema:
    type: boolean
    default: false
- name: hasRules
  in: query
  description: Specific entry rules apply (permit, reservation…)
  schema:
    type: boolean
    default: false
- name: isTouristic
  in: query
  description: Developed/touristic cave open to the public
  schema:
    type: boolean
    default: false
```

### 3. `PUT /api/v1/entrances/{id}`

The description already states _"See the POST /entrances route explanation about the parameters you can use to update an entrance"_ — no additional path-level change needed beyond updating the POST params above.

ClemRz:
## 🤔 What

Add nine boolean characteristic flags to the entrance entity: `hasBat`, `dangerFlooding`, `dangerCo2`, `dangerRockfall`, `dangerPollution`, `needCleanGear`, `needStayOnTrail`, `hasRules`, `isTouristic`.

- Closes #1507
- All nine fields live on `t_entrance` (and `h_entrance` for snapshots), default to `false`, no special permission restrictions
- Only `isTouristic` and `dangerPollution` are indexed in Typesense search (per maintainer decision — other hazard/rules fields are intentionally not searchable to avoid facilitating degradation)
- CSV import supports all nine fields
- JSON export (dbSync) includes all nine fields
- Swagger spec updated

## 🤷‍♂️ Why

The frontend already displays boolean characteristic chips on the Entrance detail page (Properties.jsx), but the API doesn't return or accept them yet. This closes the gap.

## 🔍 How

- **SQL migration** (`sql/4_01_2026_07_15_entrance_boolean_characteristics.sql`): adds nine `bool NOT NULL DEFAULT false` columns to both `t_entrance` and `h_entrance` in a single transaction with idempotent guards
- **Waterline models**: `TEntrance.js` and `HEntrance.js` declare the nine attributes
- **Service layer**: extracted a `coerceBool(req, field)` helper in `EntranceService.js` to handle string-to-bool coercion for query params — refactored `isSensitive` to use it too
- **Response shape**: `EntranceModel.js` + `toEntrance()` converter updated
- **Search**: `dbSync/entities/entrance.js` selects all nine columns (for JSON export) but only `isTouristic` and `dangerPollution` are in the Typesense schema. The seven non-indexed fields are stripped in both `updateInSearch` and `importFormater` so they don't leak into search results
- **CSV import**: `EntranceCSVImportService.js` maps nine optional `karstlink:*` columns with a shared `parseBool` helper
- **Swagger**: entrance schema and POST parameters updated

### Automated tests
```bash
PORT=1338 npm run test
```

8 property-based tests (fast-check) covering:
1. Model attribute consistency (TEntrance + HEntrance)
2. Converter preserves all boolean characteristics
3. String-to-bool coercion round-trip
4. Search document includes only indexed fields (excludes the seven non-indexed ones)
5. CSV import parses boolean fields

Integration tests for create, update, and find endpoints verifying the nine fields.

### Manual CSV import test file

Tab-separated CSV for testing via the frontend import UI. Row 1 has all nine booleans with mixed casing, row 2 has none (should default to `false`):

```
id	rdf:type	dct:rights/cc:attributionName	dct:rights/karstlink:licenseType	rdfs:label	rdfs:label/dc:language	gn:countryCode	w3geo:latitude	w3geo:longitude	karstlink:hasBat	karstlink:dangerFlooding	karstlink:dangerCo2	karstlink:dangerRockfall	karstlink:dangerPollution	karstlink:needCleanGear	karstlink:needStayOnTrail	karstlink:hasRules	karstlink:isTouristic
99001	karstlink:UndergroundCavity	TestImport	CC-BY-SA	Test Bat Cave Entrance	eng	FR	45.123	6.456	true	false	True	FALSE	true	true	false	True	true
99002	karstlink:UndergroundCavity	TestImport	CC-BY-SA	Test Clean Entrance	eng	FR	46.789	7.012									
```

## 📸 Previews

N/A — backend only, no UI changes in this repo.


Paul-AUB (changes requested):
## Critical

### `coerceBool` can return non-boolean values

**File:** [api/services/EntranceService.js](api/services/EntranceService.js)

```js
function coerceBool(req, field) {
  let value = req.param(field);
  if (value !== undefined) {
    value = typeof value === 'string' ? value === 'true' : value; // ← returns value as-is if not a string
  }
  return value;
}
```

When the input is not a string and not `undefined` (e.g., a numeric `1`/`0`, or `null`), the function returns it unchanged. The model declares `type: 'boolean'` for all these fields, so returning a non-boolean breaks the type contract. PostgreSQL may silently coerce, but it creates subtle bugs in downstream logic that expects actual booleans.

Note: `isSensitive` previously had the same inline logic and has the same issue after refactoring.

**Fix:**
```js
function coerceBool(req, field) {
  const value = req.param(field);
  if (value === undefined || value === null) return value;
  return typeof value === 'string' ? value === 'true' : Boolean(value);
}
```

---

## Major

### Strip list duplicated in two places with inconsistent style

The list of boolean fields excluded from the TypeSense search document is duplicated:

- **[api/dbSync/entities/entrance.js](api/dbSync/entities/entrance.js)** uses anonymous placeholders `_1, _2, ..._7`
- **[api/services/EntranceService.js](api/services/EntranceService.js)** uses named placeholders `_hasBat, _dangerFlooding, ...`

Both lists must stay in sync. If a new boolean field is added to one and forgotten in the other, it will either leak into search documents or get incorrectly stripped. The anonymous naming in `entrance.js` also makes the intent harder to read.

**Fix:** Extract the list to a shared constant, e.g., in `config/constants/entrance.js`:
```js
const NON_INDEXED_BOOLEAN_FIELDS = [
  'hasBat', 'dangerFlooding', 'dangerCo2', 'dangerRockfall',
  'needCleanGear', 'needStayOnTrail', 'hasRules',
];
```
Then use it in both files to build the destructuring/omit logic.

### `parseBool` crashes on non-string input

**File:** [api/services/EntranceCSVImportService.js](api/services/EntranceCSVImportService.js)

```js
const parseBool = (v) => v.toLowerCase() === 'true';
```

This throws a `TypeError` if `v` is not a string (e.g., already a boolean or `undefined`). Since `doubleCheck` calls this with `defaultValue: false`, a missing CSV field returns `false` correctly — but if a CSV row contains a non-string value for a boolean column, this will throw at import time.

**Fix:**
```js
const parseBool = (v) => String(v).toLowerCase() === 'true';
```

---

## Minor

### Property tests don't cover edge cases for coercion

**File:** [test/property/EntranceBooleanCoercion.property.test.js](test/property/EntranceBooleanCoercion.property.test.js)

The tests cover `'true'`, `'false'`, and `boolean` types, but not:
- Numeric strings: `'1'`, `'0'`
- Numeric values: `1`, `0`
- `null`
- Case variants: `'TRUE'`, `'False'`

Since `coerceBool` currently fails on some of these, adding them would also catch the critical bug above.

---

## Positive observations

- **Strong test coverage:** Property-based tests with `fast-check` across 5 dimensions (characteristics, coercion, converters, CSV import, search). This is well above average for this codebase.
- **No breaking changes:** All boolean fields default to `false`, so existing entrances are unaffected.
- **Naming consistency:** camelCase field names follow existing conventions throughout models, converters, and API responses.
- **Search optimization:** Only `isTouristic` and `dangerPollution` are indexed in TypeSense; the rest are correctly stripped from the search document.
- **Migration is clean:** The SQL migration adds columns with `DEFAULT FALSE NOT NULL`, which is safe for a live database.

---

## Summary

| Severity | # | Description |
|----------|---|-------------|
| Critical | 1 | `coerceBool` returns non-boolean for numeric/null inputs |
| Major    | 2 | Strip list duplicated + inconsistent style; `parseBool` throws on non-string |
| Minor    | 1 | Missing coercion edge cases in tests|

ClemRz:
All four items addressed in `df93615`:

| Severity | Issue | Fix |
|----------|-------|-----|
| Critical | `coerceBool` returns non-boolean for numeric/null | Now uses `Boolean(value)` for non-string inputs, passes through `undefined`/`null` |
| Major | Strip list duplicated in two places | Extracted `NON_INDEXED_BOOLEAN_FIELDS` to `config/constants/entrance.js`, used in both `EntranceService.updateInSearch` and `dbSync importFormater` |
| Major | `parseBool` crashes on non-string | Changed to `String(v).toLowerCase() === 'true'` |
| Minor | Missing coercion edge cases in tests | Added 3 new property tests: `null` input, numeric values (`0`, `1`, `-1`, `42`), case variants (`'TRUE'`, `'False'`, `'tRuE'`) |

Paul-AUB (approved):
LGTM

PR #1511, opened 2026-03-31
fix(search): validate sort parameter in advanced-search

ClemRz:
## 🤔 What

- Add sort parameter validation in the advanced search endpoint (`POST /api/v1/advanced-search`)
- Invalid sort fields now return 400 Bad Request instead of crashing with 500

## 🤷‍♂️ Why

Production logs showed a 500 Internal Server Error when a user sorted entrances by `depth:desc`. The entrances Typesense collection has `cave.depth` (nested) but no top-level `depth` field. The sort parameter was passed directly to Typesense without validation, causing an `ObjectNotFound` error that bubbled up as a 500.

## 🔍 How

- Added a `validateSort(entity, sort)` helper in `SearchService.js` that checks:
  1. Sort format is `field:asc` or `field:desc`
  2. The field exists in the target entity's Typesense schema
  3. The field is marked as sortable (`sort: true`)
- Called `validateSort` in `collectionSearch()` before building Typesense query params
- Added try/catch in the `advanced-search.js` controller to catch `E_SORT_VALIDATION` errors and return `res.badRequest()` with a descriptive message
- Non-validation errors (e.g., Typesense connectivity) still propagate as 500

## 🧪 Testing

- Added property-based tests in `test/integration/1_services/SearchService.property.test.js`:
  - **Property 1 (Bug Condition)**: Generates random invalid sort inputs (non-existent fields, non-sortable fields, bad format/direction) and verifies validation errors are thrown
  - **Property 2a (Preservation)**: Verifies all valid entity + sortable field + direction combos produce correct `sort_by` params
  - **Property 2b (Preservation)**: Verifies no-sort requests omit `sort_by`
  - **Property 2c (Preservation)**: Verifies invalid entity names return `null`

```bash
PORT=1338 npm run test -- --grep "SearchService - Property"
```

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,31 @@ Grottocenter API implements a Role-Based Access Control (RBAC) system with 5 use
 
 For detailed information about roles, permissions, and implementation, see [here](PERMISSION_SYSTEM.md).
 
+### Sortable fields
+
+The advanced search endpoints (`POST /api/v1/advanced-search` and `POST /api/v1/advanced-search/export`) accept a `sort` parameter in the format `field:asc` or `field:desc`. Only fields that are sortable in the target entity's Typesense schema are accepted; invalid fields return a 400 error.
+
+The source of truth for sortable fields is the `search.schema.fields` array in each entity module under `api/dbSync/entities/`. A field is sortable when:
+
+- it has `sort: true` in its schema definition, or
+- its type is numeric or boolean (`int32`, `int64`, `float`, `bool`, and their array variants), which Typesense makes sortable by default
+
+To list all sortable fields for a given entity:
+
+```bash
+node -e "
+const SORTABLE_BY_DEFAULT = ['int32','int64','float','bool','int32[]','int64[]','float[]','bool[]'];
+const entities = ['organization','person','massif','cave','entrance','document'];
+entities.forEach(e => {
+  const mod = require('./api/dbSync/entities/' + e);
+  const sortable = mod.search.schema.fields
+    .filter(f => f.sort === true || SORTABLE_BY_DEFAULT.includes(f.type))
+    .map(f => f.name);
+  console.log(mod.search.schema.name + ': ' + sortable.join(', '));
+});
+"
+```
+
 ### Git
 
 #### Workflow
--- a/api/controllers/v1/search/advanced-search-export.js
+++ b/api/controllers/v1/search/advanced-search-export.js
@@ -97,12 +97,24 @@ module.exports = async (req, res) => {
   let hasSentHeader = false;
   let page = 1;
   for (;;) {
-    // eslint-disable-next-line no-await-in-loop
-    const results = await SearchService.collectionSearch({
-      ...params,
-      page,
-      size: BATCH_SIZE,
-    }).catch((err) => err);
+    let results;
+    try {
+      // eslint-disable-next-line no-await-in-loop
+      results = await SearchService.collectionSearch({
+        ...params,
+        page,
+        size: BATCH_SIZE,
+      });
+    } catch (err) {
+      if (err.code === 'E_SORT_VALIDATION') {
+        if (!hasSentHeader) {
+          res.badRequest({ error: err.message });
+          return;
+        }
+        break;
+      }
+      results = err;
+    }
 
     if (!results || !results.hits) {
       sails.log.error('Export to CSV error', params, page, results);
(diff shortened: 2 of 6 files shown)


Paul-AUB (changes requested):
# Code Review: fix(search): validate sort parameter in advanced-search

Commits: `e95bc86` (fix) · `392927d` (docs)

## Overview

The fix is correct and well-structured: a dedicated `validateSort` function with a clear error code, a minimal try/catch in the controller, and a solid property-based test suite. However, the validation is applied to only one of the two controllers that call `collectionSearch`, leaving the export endpoint exposed to the same failure mode.

---

## Issues

### 1. `advanced-search-export.js` is not protected — same bug still reachable

`collectionSearch` is called with a user-supplied `sort` in two controllers:

- `advanced-search.js` → now catches `E_SORT_VALIDATION` → returns 400 ✓
- `advanced-search-export.js:101` → uses `.catch((err) => err)` → an `E_SORT_VALIDATION` error is caught, treated as a failed result, logs it as an internal error, and returns a 500 ✗

The fix leaves the export endpoint with the same original failure mode (invalid sort → 500). The catch block in that controller needs the same `E_SORT_VALIDATION` handling.

### 2. Empty string sort skips validation — `SearchService.js:163`

```js
if (sort) validateSort(entity, sort);
```

`sort = ''` is falsy and skips validation, then `sort_by` is omitted from the Typesense query (because of `sort && { sort_by: ... }` at line 178). This happens to be safe, but only by accident. A client sending `sort=''` silently receives results without sorting rather than a 400. This is a latent inconsistency — the controller should strip or reject empty strings before passing to the service.

### 3. Ambiguous error message for empty field name — `SearchService.js:119`

`sort = ':asc'` splits to `['', 'asc']` — format is technically valid (two parts, valid direction), so it passes the first check and reaches field lookup. The error thrown is:

```
Field '' does not exist in 'entrances' schema.
```

This message is confusing; the root problem is the missing field name, not a non-existent one. The format check at the top should also reject an empty field segment.

### 4. Hardcoded sortable field list in Swagger will drift — `swaggerV1.yaml:3745-3753`

The enumerated list of sortable fields per entity is manually written into the YAML description. It will silently go out of date whenever a schema changes. This is a documentation maintenance risk. A comment linking to the source-of-truth schema files would at least make the drift visible during code review.

### 5. `allEntities` is duplicated in the property test — `SearchService.property.test.js:14-27`

The test file manually re-imports all entity modules and rebuilds `allEntities` and `allEntitiesKeys`. If a new entity is added to `SearchService.js` but forgotten here, the new entity is not covered by any property test and no error is raised. Exporting `allEntities` from the service (or a shared module) would keep the two in sync automatically.

### 6. `nonExistentFieldArb` convergence is not guaranteed — `SearchService.property.test.js:62`

```js
fc.stringMatching(/^[a-z]{2,15}$/)
  .filter((field) => !existingFields.includes(field))
```

`fast-check` will retry filtered-out values, but there is no bound on how many exist in `existingFields`. In practice the existing schemas have few short lowercase field names, so this converges fine. A structurally simpler alternative is to prefix generated strings (e.g. `zzz_${field}`) to guarantee they cannot match any real field name.

---

## Positives

- `validateSort` is a clean, pure function with no side effects — easy to test in isolation.
- The `E_SORT_VALIDATION` error code pattern (rather than a custom class) is lightweight and consistent with how Sails/Waterline surfaces errors.
- The guard order in `collectionSearch` (invalid entity → early return null, then validate sort) is correct: `allEntities[entity]` is always defined when `validateSort` runs.
- The three-property structure in the test file (bug condition / valid sort preserved / no-sort preserved) is well-organised and clearly motivated.
- The Swagger `400` response code addition and the cross-reference from the GET endpoint description to the POST one are both good documentation choices.

ClemRz:
1. **Export endpoint unprotected** — `advanced-search-export.js` now catches `E_SORT_VALIDATION` and returns 400
2. **Empty string sort skips validation** — `sort.trim()` check added in `collectionSearch`
3. **Empty field name ambiguous error** — `:asc` now rejected at format check (`!parts[0]`)
4. **Hardcoded sortable fields in Swagger** — Removed enumerated list, replaced with pointer to new "Sortable fields" README section
5. **`allEntities` duplicated in test** — Exported from `SearchService`, reused in property test file
6. **`nonExistentFieldArb` convergence** — Uses `zzz_` prefix instead of `.filter()` on generated strings

Not addressed: no integration tests for the export endpoint's `E_SORT_VALIDATION` handling

Paul-AUB (approved):

PR #1525, opened 2026-04-03
feat(caver): add admin ban/unban endpoints with security hardening

ClemRz:
## 🤔 What

Add admin-only ban/unban endpoints, harden all unauthenticated flows against information leakage, expose ban status to admins, and provide a banned cavers listing.

- `POST /api/v1/cavers/:caverId/ban` — sets `banned=true`, revokes all active tokens via BlacklistService
- `POST /api/v1/cavers/:caverId/unban` — sets `banned=false`
- `GET /api/v1/cavers/:id` — now includes `isBanned` field when the caller is an administrator
- `GET /api/v1/cavers/banned` — admin-only endpoint returning all banned cavers (`{ banned: [{ id, nickname, name, surname }] }`)
- Activate the existing `banned`, `date_last_connection`, and `connection_counter` columns on `t_caver`
- Harden login, forgot-password, change-password, and sign-up endpoints
- Introduce `toListCaver` converter (`{ id, nickname, name, surname }`) used by `get-banned`, `get-groups`, `get-admins`, and `get-moderators` — keeps `toSimpleCaver` (`{ id, nickname }`) unchanged for author/reviewer references

## 🤷‍♂️ Why

Administrators currently revoke caver tokens manually via direct database inserts into `t_token_blacklist`. This feature provides a proper REST API for banning cavers, with persistent ban tracking and comprehensive security hardening so that ban status is never leaked to callers. Admins can now see ban status on individual caver profiles and list all banned accounts from a single endpoint.

## 🔍 How

**Ban/Unban endpoints** (`api/controllers/v1/caver/ban.js`, `unban.js`):
- Admin-only (checked via `RightService.hasGroup`), self-ban prevention, idempotent
- Ban sets `banned=true` and calls `BlacklistService.revoke()` to invalidate all pre-ban tokens

**Login hardening** (`api/controllers/v1/auth/login.js`):
- Banned cavers get a generic `401 { status: 'Mismatch', message: 'Invalid email or password.' }` — identical to wrong credentials
- `date_last_connection` and `connection_counter` updated on every successful credential verification (including banned cavers)
- `sails.log.warn` emitted for banned login attempts

**Forgot-password hardening** (`api/controllers/v1/account/forgot-password.js`):
- Banned cavers get a silent `200 OK` with no token generated and no email sent
- `sails.log.warn` emitted

**Change-password hardening** (`api/controllers/v1/account/change-password.js`):
- Banned cavers using a pre-ban reset token get `403 'The password reset token has expired.'` — same as a genuinely expired token
- `sails.log.warn` emitted

**Sign-up enumeration prevention** (`api/controllers/v1/auth/sign-up.js`):
- Duplicate email/nickname conflicts now return a generic `409 'Email or nickname is already used.'` instead of specifying which field
- `sails.log.warn` with the specific detail emitted server-side

**Ban status in caver profile** (`api/controllers/v1/caver/find.js`, `api/services/mapping/converters.js`):
- `GET /api/v1/cavers/:id` conditionally includes `isBanned` (boolean) when the caller is an admin
- Non-admin and unauthenticated callers never see this field
- `CaverService.getCaver` now returns the `banned` field

**List banned cavers** (`api/controllers/v1/caver/get-banned.js`):
- `GET /api/v1/cavers/banned` returns `{ banned: [{ id, nickname, name, surname }] }` — all banned cavers in one response
- Follows the same pattern as `get-groups.js` (no pagination)
- Admin-only access via `RightService.hasGroup`

**New `toListCaver` converter** (`api/services/mapping/converters.js`):
- Returns `{ id, nickname, name, surname }` — richer than `toSimpleCaver` but scoped to listing endpoints
- Used by `get-banned`, `get-groups`, `get-admins`, `get-moderators`
- `toSimpleCaver` (`{ id, nickname }`) remains unchanged for author/reviewer nested references

**Swagger** (`assets/swaggerV1.yaml`): Updated for all modified and new endpoints.

## 🧪 Testing

```bash
PORT=1338 npm run test
```

Integration tests added:
- `test/integration/4_routes/Cavers/ban.test.js`
- `test/integration/4_routes/Cavers/unban.test.js`
- `test/integration/4_routes/Cavers/login-banned.test.js`
- `test/integration/4_routes/Cavers/find-banned.test.js`
- `test/integration/4_routes/Cavers/get-banned.test.js`
- `test/integration/4_routes/Account/forgot-password-banned.test.js`
- `test/integration/4_routes/Account/change-password-banned.test.js`
- `test/integration/4_routes/Auth/sign-up-enumeration.test.js`

Property-based tests (fast-check, 14 properties):
- `test/integration/4_routes/Cavers/ban-unban.property.test.js`

## 📸 Previews

N/A — backend-only changes.

### DIFF
--- a/api/controllers/v1/account/change-password.js
+++ b/api/controllers/v1/account/change-password.js
@@ -46,6 +46,14 @@ module.exports = async (req, res) => {
     });
   }
 
+  // Block banned cavers — return same response as expired token to hide ban status
+  if (userFound.banned === true) {
+    sails.log.warn(
+      `Banned caver ${userFound.id} attempted password change via reset token`
+    );
+    return res.forbidden('The password reset token has expired.');
+  }
+
   // Check token
   const verifyTokenCallback = async (err) => {
     if (err) {
--- a/api/controllers/v1/account/forgot-password.js
+++ b/api/controllers/v1/account/forgot-password.js
@@ -19,6 +19,13 @@ module.exports = async (req, res) => {
     });
   }
 
+  // If the caver is banned, return 200 OK without generating a token or sending an email
+  // to prevent information leakage about ban status
+  if (userFound.banned === true) {
+    sails.log.warn(`Banned caver ${userFound.id} attempted password reset`);
+    return res.ok();
+  }
+
   // Generate reset password token
   const token = TokenService.issue(
     {
--- a/api/controllers/v1/auth/login.js
+++ b/api/controllers/v1/auth/login.js
@@ -26,6 +26,24 @@ module.exports = async (req, res) => {
       message: 'Invalid email or password.',
     });
   }
+
+  // Update login metadata (fire-and-forget)
+  CommonService.query(
+    `UPDATE t_caver SET date_last_connection = NOW(), connection_counter = connection_counter + 1 WHERE id = $1`,
+    [result.user.id]
+  ).catch((e) =>
+    sails.log.error('Failed to update login metadata:', e.message)
+  );
+
+  // Reject banned cavers — return generic 401 to hide ban status
+  if (result.user.banned) {
+    sails.log.warn(`Banned caver ${result.user.id} attempted login`);
+    return res.unauthorized({
+      status: 'Mismatch',
+      message: 'Invalid email or password.',
+    });
+  }
+
   const token = TokenService.issue(
     {
       id: result.user.id,
--- a/api/controllers/v1/auth/sign-up.js
+++ b/api/controllers/v1/auth/sign-up.js
@@ -12,7 +12,8 @@ module.exports = async (req, res) => {
   email = email.toLowerCase();
   const caverEmail = await TCaver.findOne({ mail: email });
   if (caverEmail) {
-    return res.conflict(`The email ${email} is already used.`);
+    sails.log.warn(`Sign-up attempt with existing email: ${email}`);
+    return res.conflict('Email or nickname is already used.');
   }
 
   const password = req.param('password');
@@ -31,7 +32,8 @@ module.exports = async (req, res) => {
   }
   const caverNickname = await TCaver.findOne({ nickname });
   if (caverNickname) {
-    return res.conflict(`The nickname ${nickname} is already used.`);
+    sails.log.warn(`Sign-up attempt with existing nickname: ${nickname}`);
+    return res.conflict('Email or nickname is already used.');
   }
 
   try {
--- a/api/controllers/v1/caver/ban.js
+++ b/api/controllers/v1/caver/ban.js
@@ -1,0 +1,30 @@
+const RightService = require('../../../services/RightService');
+
+module.exports = async (req, res) => {
+  const hasRight = RightService.hasGroup(
+    req.token.groups,
+    RightService.G.ADMINISTRATOR
+  );
+  if (!hasRight) {
+    return res.forbidden('You are not authorized to ban a caver.');
+  }
+
+  const { caverId } = req.params;
+
+  if (req.token.id === parseInt(caverId, 10)) {
+    return res.forbidden('You cannot ban yourself.');
+  }
+
+  const caver = await TCaver.findOne({ id: caverId });
+  if (!caver) {
+    return res.notFound(`Caver with id ${caverId} not found.`);
+  }
+
+  if (!caver.banned) {
+    await TCaver.updateOne({ id: caverId }).set({ banned: true });
+  }
+
+  await BlacklistService.revoke(parseInt(caverId, 10));
+
+  return res.ok({ banned: true });
+};
--- a/api/controllers/v1/caver/find.js
+++ b/api/controllers/v1/caver/find.js
@@ -1,6 +1,7 @@
 const CaverService = require('../../../services/CaverService');
 const ControllerService = require('../../../services/ControllerService');
 const DocumentService = require('../../../services/DocumentService');
+const RightService = require('../../../services/RightService');
 const { toCaver } = require('../../../services/mapping/converters');
 
 module.exports = async (req, res) => {
@@ -16,12 +17,16 @@ module.exports = async (req, res) => {
     caverFound.documents.map((d) => d.id)
   );
 
+  const isAdmin =
+    req.token &&
+    RightService.hasGroup(req.token.groups, RightService.G.ADMINISTRATOR);
+
   return ControllerService.treatAndConvert(
     req,
     null,
     caverFound,
     params,
     res,
-    toCaver
+    (source) => toCaver(source, { isAdmin })
   );
 };
--- a/api/controllers/v1/caver/get-admins.js
+++ b/api/controllers/v1/caver/get-admins.js
@@ -1,5 +1,5 @@
 const ControllerService = require('../../../services/ControllerService');
-const { toSimpleCaver } = require('../../../services/mapping/converters');
+const { toListCaver } = require('../../../services/mapping/converters');
 const { toListFromController } = require('../../../services/mapping/utils');
 
 module.exports = async (req, res) => {
@@ -15,6 +15,6 @@ module.exports = async (req, res) => {
     adminGroup.cavers,
     { controllerMethod: 'CaverController.getAdmins' },
     res,
-    (data) => toListFromController('cavers', data, toSimpleCaver)
+    (data) => toListFromController('cavers', data, toListCaver)
   );
 };
--- a/api/controllers/v1/caver/get-banned.js
+++ b/api/controllers/v1/caver/get-banned.js
@@ -1,0 +1,16 @@
+const { toListCaver } = require('../../../services/mapping/converters');
+const RightService = require('../../../services/RightService');
+
+module.exports = async (req, res) => {
+  const hasRight = RightService.hasGroup(
+    req.token.groups,
+    RightService.G.ADMINISTRATOR
+  );
+  if (!hasRight) {
+    return res.forbidden('You are not authorized to list banned cavers.');
+  }
+
+  const bannedCavers = await TCaver.find({ banned: true });
+
+  return res.ok({ banned: bannedCavers.map(toListCaver) });
+};
--- a/api/controllers/v1/caver/get-groups.js
+++ b/api/controllers/v1/caver/get-groups.js
@@ -1,4 +1,4 @@
-const { toSimpleCaver } = require('../../../services/mapping/converters');
+const { toListCaver } = require('../../../services/mapping/converters');
 const { toList } = require('../../../services/mapping/utils');
 
 module.exports = async (req, res) => {
@@ -9,8 +9,8 @@ module.exports = async (req, res) => {
   ]);
 
   return res.ok({
-    administrators: toList('cavers', administrators, toSimpleCaver),
-    moderators: toList('cavers', moderators, toSimpleCaver),
-    leaders: toList('cavers', leaders, toSimpleCaver),
+    administrators: toList('cavers', administrators, toListCaver),
+    moderators: toList('cavers', moderators, toListCaver),
+    leaders: toList('cavers', leaders, toListCaver),
   });
 };
--- a/api/controllers/v1/caver/get-moderators.js
+++ b/api/controllers/v1/caver/get-moderators.js
@@ -1,5 +1,5 @@
 const ControllerService = require('../../../services/ControllerService');
-const { toSimpleCaver } = require('../../../services/mapping/converters');
+const { toListCaver } = require('../../../services/mapping/converters');
 const { toListFromController } = require('../../../services/mapping/utils');
 
 module.exports = async (req, res) => {
@@ -15,6 +15,6 @@ module.exports = async (req, res) => {
     moderatorGroup.cavers,
     { controllerMethod: 'CaverController.getModerators' },
     res,
-    (data) => toListFromController('cavers', data, toSimpleCaver)
+    (data) => toListFromController('cavers', data, toListCaver)
   );
 };
--- a/api/controllers/v1/caver/unban.js
+++ b/api/controllers/v1/caver/unban.js
@@ -1,0 +1,26 @@
+const RightService = require('../../../services/RightService');
+
+module.exports = async (req, res) => {
+  const hasRight = RightService.hasGroup(
+    req.token.groups,
+    RightService.G.ADMINISTRATOR
+  );
+  if (!hasRight) {
+    return res.forbidden('You are not authorized to unban a caver.');
+  }
+
+  const { caverId } = req.params;
+
+  const caver = await TCaver.findOne({ id: caverId });
+  if (!caver) {
+    return res.notFound(`Caver with id ${caverId} not found.`);
+  }
+
+  if (!caver.banned) {
+    return res.ok({ banned: false });
+  }
+
+  await TCaver.updateOne({ id: caverId }).set({ banned: false });
+
+  return res.ok({ banned: false });
+};
--- a/api/models/TCaver.js
+++ b/api/models/TCaver.js
@@ -48,14 +48,12 @@ module.exports = {
       maxLength: 64,
     },
 
-    // Unsued
     banned: {
       type: 'boolean',
       columnName: 'banned',
       defaultsTo: false,
     },
 
-    // Unsued
     connectionCounter: {
       type: 'number',
       allowNull: false,
@@ -116,7 +114,6 @@ module.exports = {
       columnType: 'timestamp',
     },
 
-    // Unsued
     dateLastConnection: {
       type: 'ref',
       columnName: 'date_last_connection',
--- a/api/services/CaverService.js
+++ b/api/services/CaverService.js
@@ -109,6 +109,7 @@ module.exports = {
     return {
       id: caver.id,
       type: caver.type,
+      banned: caver.banned,
       name: caver.name,
       surname: caver.surname,
       nickname: caver.nickname,
--- a/api/services/mapping/converters.js
+++ b/api/services/mapping/converters.js
@@ -71,7 +71,7 @@ const c = {
     ),
   }),
 
-  toCaver: (source) => {
+  toCaver: (source, meta) => {
     const result = {
       ...CaverModel,
       id: source.id,
@@ -110,12 +110,23 @@ const c = {
     listParser('documents', c.toSimpleDocument);
     listParser('grottos', c.toSimpleOrganization, 'organizations');
 
+    if (meta?.isAdmin) {
+      result.isBanned = Boolean(source.banned);
+    }
+
     return result;
   },
 
   toSimpleCaver: (source) => ({
     id: source.id,
     nickname: source.nickname,
+  }),
+
+  toListCaver: (source) => ({
+    id: source.id,
+    nickname: source.nickname,
+    name: source.name,
+    surname: source.surname,
   }),
 
   toSimpleComment: (source) => {
--- a/assets/swaggerV1.yaml
+++ b/assets/swaggerV1.yaml
@@ -959,6 +959,102 @@ paths:
                     items:
                       $ref: '#/components/schemas/Document'
 
+  '/cavers/{caverId}/ban':
+    post:
+      tags:
+        - cavers
+      summary: Ban a caver
+      description: Bans a caver, revoking all active tokens and preventing future logins. Requires Administrator role.
+      security:
+        - bearerAuth: []
+      parameters:
+        - name: caverId
+          in: path
+          description: Caver id to ban
+          required: true
+          schema:
+            type: integer
+      responses:
+        '200':
+          description: Caver banned successfully
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  banned:
+                    type: boolean
+                    example: true
+        '401':
+          description: Unauthorized — missing or invalid bearer token.
+        '403':
+          description: Forbidden — caller is not an administrator, or attempted to ban themselves.
+        '404':
+          description: Caver not found.
+
+  '/cavers/{caverId}/unban':
+    post:
+      tags:
+        - cavers
+      summary: Unban a caver
+      description: Unbans a previously banned caver, allowing them to log in again. Requires Administrator role.
+      security:
+        - bearerAuth: []
+      parameters:
+        - name: caverId
+          in: path
+          description: Caver id to unban
+          required: true
+          schema:
+            type: integer
+      responses:
+        '200':
+          description: Caver unbanned successfully
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  banned:
+                    type: boolean
+                    example: false
+        '401':
+          description: Unauthorized — missing or invalid bearer token.
+        '403':
+          description: Forbidden — caller is not an administrator.
+        '404':
+          description: Caver not found.
+
+  '/cavers/banned':
+    get:
+      tags:
+        - cavers
+      summary: List banned cavers
+      description: Returns all currently banned cavers. Requires Administrator role.
+      security:
+        - bearerAuth: []
+      responses:
+        '200':
+          description: Successful operation
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  banned:
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        id:
+                          type: integer
+                        nickname:
+                          type: string
+        '401':
+          description: Unauthorized — missing or invalid bearer token.
+        '403':
+          description: Forbidden — caller is not an administrator.
+
   '/descriptions':
     post:
       tags:
@@ -3023,7 +3119,7 @@ paths:
         '400':
           description: You must provide a password and a reset password token in the body of your request. The password must be at least 8 characters long.
         '403':
-          description: The password reset token has expired or is invalid.
+          description: The password reset token has expired, is invalid, or the account is banned. The response is intentionally identical for all cases to prevent information leakage.
         '404':
           description: User in the password reset token not found.
 
@@ -3053,6 +3149,9 @@ paths:
         - authentication
       description: >-
         Launch the procedure of password change. This will send an email to the email adress you provided if it exists as a Grottocenter user's email. This email contains a password reset link expiring after 24 hours using the Grottocenter application.
+
+
+        The endpoint always returns 200 OK regardless of whether the account is banned or not, to prevent information leakage about ban status. If the account is banned, no reset email is sent.
 
 
         __IMPORTANT NOTICE__
@@ -3103,7 +3202,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/Auth-success'
         '401':
-          description: Invalid email or password or the password must be reset first.
+          description: Invalid email or password, password must be reset, or account is banned. The response is intentionally identical for all failure cases to prevent information leakage.
           content:
             application/json:
               schema:
@@ -3151,7 +3250,7 @@ paths:
         '400':
           description: Bad request (password too short, missing value(s)...).
         '409':
-          description: Email or nickname already used.
+          description: Email or nickname is already used. The response intentionally does not specify which field caused the conflict to prevent account enumeration.
 
   '/entrances':
     post:
@@ -5091,6 +5190,9 @@ components:
             $ref: '#/components/schemas/Group'
         id:
           type: integer
+        isBanned:
+          type: boolean
+          description: Whether the caver is banned. Only present when the caller is an authenticated administrator.
         language:
           type: string
         name:
--- a/config/policies.js
+++ b/config/policies.js
@@ -50,6 +50,7 @@ module.exports.policies = {
   'v1/organization/remove-explored-cave': 'tokenAuth',
 
   // Caver
+  'v1/caver/get-banned': 'tokenAuth',
   'v1/caver/count': true,
   'v1/caver/users-count': true,
   'v1/caver/find': ['validateId'],
@@ -64,6 +65,8 @@ module.exports.policies = {
   'v1/caver/get-users': 'tokenAuth',
   'v1/caver/put-on-group': 'tokenAuth',
   'v1/caver/remove-explored-cave': 'tokenAuth',
+  'v1/caver/ban': 'tokenAuth',
+  'v1/caver/unban': 'tokenAuth',
   'v1/caver/remove-from-group': 'tokenAuth',
   'v1/caver/set-groups': 'tokenAuth',
   'v1/caver/add-to-organization': 'tokenAuth',
--- a/config/routes.js
+++ b/config/routes.js
@@ -73,6 +73,7 @@ module.exports.routes = {
   // Caver
   'DELETE /api/v1/cavers/:caverId/groups/:groupId':
     'v1/caver/remove-from-group',
+  'GET /api/v1/cavers/banned': 'v1/caver/get-banned',
   'GET /api/v1/cavers/:id': 'v1/caver/find',
   'GET /api/v1/cavers/:caverId/documents': 'v1/document/find-by-caver-id',
   'GET /api/v1/cavers/:caverId/subscriptions': 'v1/caver/get-subscriptions',
@@ -94,6 +95,8 @@ module.exports.routes = {
     'v1/caver/add-to-organization',
   'DELETE /api/v1/cavers/:caverId/organizations/:organizationId':
     'v1/caver/remove-from-organization',
+  'POST /api/v1/cavers/:caverId/ban': 'v1/caver/ban',
+  'POST /api/v1/cavers/:caverId/unban': 'v1/caver/unban',
   'DELETE /api/v1/cavers/:id': 'v1/caver/delete',
 
   // Entrance
--- a/test/integration/4_routes/Account/change-password-banned.test.js
+++ b/test/integration/4_routes/Account/change-password-banned.test.js
@@ -1,0 +1,89 @@
+const supertest = require('supertest');
+const should = require('should');
+const TokenService = require('../../../../api/services/TokenService');
+
+const NEW_PASSWORD = 'newpassword123';
+const targetCaverId = 3; // user1
+
+describe('Account features', () => {
+  describe('Change password - Banned caver', () => {
+    let originalPasswordHash;
+
+    beforeEach(async () => {
+      const caver = await TCaver.findOne({ id: targetCaverId });
+      originalPasswordHash = caver.password;
+    });
+
+    afterEach(async () => {
+      // Restore banned flag and original password
+      await TCaver.updateOne({ id: targetCaverId }).set({
+        banned: false,
+        password: originalPasswordHash,
+      });
+    });
+
+    it('should return 403 with expired token message when a banned caver changes password via reset token', async () => {
+      const userFound = await TCaver.findOne({ id: targetCaverId });
+      const resetToken = TokenService.issue(
+        { userId: userFound.id },
+        sails.config.custom.passwordResetTokenTTL,
+        'Reset password',
+        TokenService.getResetPasswordTokenSalt(userFound)
+      );
+
+      await TCaver.updateOne({ id: targetCaverId }).set({ banned: true });
+
+      const res = await supertest(sails.hooks.http.app)
+        .patch('/api/v1/account/password')
+        .send({ password: NEW_PASSWORD, token: resetToken })
+        .set('Content-type', 'application/json')
+        .set('Accept', 'application/json')
+        .expect(403);
+
+      should(res.body).have.property(
+        'message',
+        'The password reset token has expired.'
+      );
+    });
+
+    it('should NOT change the password for a banned caver', async () => {
+      const userFound = await TCaver.findOne({ id: targetCaverId });
+      const resetToken = TokenService.issue(
+        { userId: userFound.id },
+        sails.config.custom.passwordResetTokenTTL,
+        'Reset password',
+        TokenService.getResetPasswordTokenSalt(userFound)
+      );
+
+      await TCaver.updateOne({ id: targetCaverId }).set({ banned: true });
+
+      await supertest(sails.hooks.http.app)
+        .patch('/api/v1/account/password')
+        .send({ password: NEW_PASSWORD, token: resetToken })
+        .set('Content-type', 'application/json')
+        .set('Accept', 'application/json')
+        .expect(403);
+
+      // Verify password hash is unchanged
+      const caverAfter = await TCaver.findOne({ id: targetCaverId });
+      should(caverAfter.password).equal(originalPasswordHash);
+    });
+
+    it('should allow a non-banned caver to change password via valid reset token (returns 204)', async () => {
+      const userFound = await TCaver.findOne({ id: targetCaverId });
+      const resetToken = TokenService.issue(
+        { userId: userFound.id },
+        sails.config.custom.passwordResetTokenTTL,
+        'Reset password',
+        TokenService.getResetPasswordTokenSalt(userFound)
+      );
+
+      await supertest(sails.hooks.http.app)
+        .patch('/api/v1/account/password')
+        .send({ password: NEW_PASSWORD, token: resetToken })
+        .set('Content-type', 'application/json')
+        .set('Accept', 'application/json')
+        .expect(204);
+    });
+  });
+});
--- a/test/integration/4_routes/Account/forgot-password-banned.test.js
+++ b/test/integration/4_routes/Account/forgot-password-banned.test.js
@@ -1,0 +1,92 @@
+const supertest = require('supertest');
+const should = require('should');
+const sinon = require('sinon');
+
+const targetCaverId = 3; // user1
+const targetEmail = 'user1@user1.com';
+
+describe('Account features', () => {
+  describe('Forgot password - Banned caver', () => {
+    beforeEach(() => {
+      // Stub the sendEmail helper to track calls without actually sending
+      // sails.helpers.sendEmail.with({...}).intercept(...) is the call pattern
+      // We replace the entire helper with a stub that returns a chainable object
+      sinon.stub(sails.helpers, 'sendEmail').value({
+        with: sinon.stub().returns({
+          intercept: sinon.stub().resolves(),
+        }),
+      });
+    });
+
+    afterEach(async () => {
+      // Restore banned flag
+      await TCaver.updateOne({ id: targetCaverId }).set({ banned: false });
+      // Restore the original sendEmail helper
+      sinon.restore();
+    });
+
+    it('should return 204 for a banned caver forgot-password request', async () => {
+      await TCaver.updateOne({ id: targetCaverId }).set({ banned: true });
+
+      await supertest(sails.hooks.http.app)
+        .post('/api/v1/forgotPassword')
+        .send({ email: targetEmail })
+        .set('Content-type', 'application/json')
+        .set('Accept', 'application/json')
+        .expect(204);
+    });
+
+    it('should not send email for a banned caver', async () => {
+      await TCaver.updateOne({ id: targetCaverId }).set({ banned: true });
+
+      await supertest(sails.hooks.http.app)
+        .post('/api/v1/forgotPassword')
+        .send({ email: targetEmail })
+        .set('Content-type', 'application/json')
+        .set('Accept', 'application/json')
+        .expect(204);
+
+      // sendEmail.with() should NOT have been called for banned caver
+      should(sails.helpers.sendEmail.with.called).be.false();
+    });
+
+    it('should return 204 for a non-banned caver and send email', async () => {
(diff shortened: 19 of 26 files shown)


ClemRz:
# Requirements Document

## Introduction

Grottocenter administrators currently revoke caver tokens manually via direct database inserts into `t_token_blacklist`. This feature introduces a proper REST endpoint that allows administrators to ban a caver, revoking all their active tokens and preventing future authenticated access. The existing `banned` column on `t_caver` (currently unused) will be activated to persistently track ban status, and the login flow will be updated to reject banned cavers. Additionally, the forgot-password flow is updated to silently skip sending reset emails to banned cavers, preventing information leakage about ban status. The change-password flow is hardened to block banned cavers from resetting their password via a previously obtained reset token. The sign-up flow is also hardened to prevent account enumeration by returning generic conflict messages. Finally, the caver profile endpoint is updated to conditionally expose the `banned` field to administrators only, giving them visibility into account status without leaking ban information to regular users. An additional endpoint allows administrators to list all currently banned cavers, providing a centralized view of banned accounts.

## Glossary

- **API**: The Grottocenter Sails.js REST API
- **Administrator**: A caver with the Administrator role, responsible for user management
- **Caver**: A registered user in the Grottocenter system, stored in the `t_caver` table
- **Ban_Endpoint**: The `POST /api/v1/cavers/:caverId/ban` route that bans a caver
- **Unban_Endpoint**: The `POST /api/v1/cavers/:caverId/unban` route that unbans a caver
- **BlacklistService**: The existing service that manages per-user token revocation timestamps
- **Forgot_Password_Endpoint**: The existing `POST /api/v1/forgotPassword` route that initiates a password reset by sending a reset email to the caver
- **Forgot_Password_Controller**: The controller at `api/controllers/v1/account/forgot-password.js` that handles forgot-password requests
- **Login_Endpoint**: The existing `POST /api/v1/login` route that authenticates a caver
- **Token**: A JWT bearer token issued upon login, containing the caver's id, groups, and nickname
- **Authentication_Service**: The service (`AuthService`) responsible for verifying caver credentials against stored password hashes
- **Change_Password_Endpoint**: The existing `PUT /api/v1/account/password` route that changes a caver's password (via authenticated session or reset token)
- **Change_Password_Controller**: The controller at `api/controllers/v1/account/change-password.js` that handles password change requests
- **SignUp_Endpoint**: The existing `POST /api/v1/auth/sign-up` route that registers a new caver
- **SignUp_Controller**: The controller at `api/controllers/v1/auth/sign-up.js` that handles sign-up requests
- **GetCaver_Endpoint**: The existing `GET /api/v1/cavers/:id` route that returns a single caver's public profile
- **GetCaver_Controller**: The controller at `api/controllers/v1/caver/find.js` that handles get-caver requests
- **GetBannedCavers_Endpoint**: The `GET /api/v1/cavers/banned` route that returns all currently banned cavers
- **GetBannedCavers_Controller**: The controller at `api/controllers/v1/caver/get-banned.js` that handles list-banned-cavers requests

## Requirements

### Requirement 1: Ban a Caver

**User Story:** As an administrator, I want to ban a caver via an API endpoint, so that the caver's access is immediately revoked and future logins are prevented.

#### Acceptance Criteria

1. WHEN a valid ban request is received for an existing caver, THE Ban_Endpoint SHALL set the `banned` flag to `true` on the target caver record and revoke all active tokens for that caver via BlacklistService.
2. WHEN a ban request is received, THE Ban_Endpoint SHALL verify that the requesting caver has the Administrator role before processing the request.
3. IF a ban request is received from a caver without the Administrator role, THEN THE Ban_Endpoint SHALL return a 403 Forbidden response.
4. IF a ban request references a caver that does not exist, THEN THE Ban_Endpoint SHALL return a 404 Not Found response.
5. IF a ban request targets a caver that is already banned, THEN THE Ban_Endpoint SHALL return a 200 OK response without modifying the caver record.
6. WHEN a ban request is received without a valid bearer token, THE API SHALL return a 401 Unauthorized response.
7. IF an administrator attempts to ban their own account, THEN THE Ban_Endpoint SHALL return a 403 Forbidden response.

### Requirement 2: Unban a Caver

**User Story:** As an administrator, I want to unban a previously banned caver, so that the caver can log in and use the platform again.

#### Acceptance Criteria

1. WHEN a valid unban request is received for an existing banned caver, THE Unban_Endpoint SHALL set the `banned` flag to `false` on the target caver record.
2. WHEN an unban request is received, THE Unban_Endpoint SHALL verify that the requesting caver has the Administrator role before processing the request.
3. IF an unban request is received from a caver without the Administrator role, THEN THE Unban_Endpoint SHALL return a 403 Forbidden response.
4. IF an unban request references a caver that does not exist, THEN THE Unban_Endpoint SHALL return a 404 Not Found response.
5. IF an unban request targets a caver that is not currently banned, THEN THE Unban_Endpoint SHALL return a 200 OK response without modifying the caver record.
6. WHEN an unban request is received without a valid bearer token, THE API SHALL return a 401 Unauthorized response.

### Requirement 3: Prevent Banned Caver Login

**User Story:** As an administrator, I want banned cavers to be unable to log in, so that the ban is enforced beyond token revocation.

#### Acceptance Criteria

1. WHEN a banned caver submits valid credentials to the Login_Endpoint, THE Login_Endpoint SHALL return a 401 Unauthorized response with the same generic message used for invalid credentials, so that the caller cannot distinguish a banned account from incorrect credentials.
2. WHEN a banned caver submits valid credentials, THE Login_Endpoint SHALL not issue a new Token.
3. WHEN a banned caver submits valid credentials, THE Login_Endpoint SHALL log a warning message indicating the banned caver attempted to log in.

### Requirement 4: Update Last Connection Date on Successful Authentication

**User Story:** As a system operator, I want the `date_last_connection` timestamp to be recorded when a caver successfully authenticates, so that administrators can see when each caver last logged in.

#### Acceptance Criteria

1. WHEN the Authentication_Service verifies a caver's email and password successfully, THE Login_Endpoint SHALL update the `date_last_connection` column on the corresponding `t_caver` record to the current timestamp.
2. WHEN a banned caver provides valid credentials, THE Login_Endpoint SHALL update the `date_last_connection` column before returning the 401 Unauthorized response.
3. IF the `date_last_connection` update fails, THEN THE Login_Endpoint SHALL log the error and continue processing the login request without interrupting the authentication flow.

### Requirement 5: Increment Connection Counter on Successful Authentication

**User Story:** As a system operator, I want the `connection_counter` to be incremented when a caver successfully authenticates, so that administrators can track how many times each caver has logged in.

#### Acceptance Criteria

1. WHEN the Authentication_Service verifies a caver's email and password successfully, THE Login_Endpoint SHALL increment the `connection_counter` column on the corresponding `t_caver` record by 1.
2. WHEN a banned caver provides valid credentials, THE Login_Endpoint SHALL increment the `connection_counter` column before returning the 401 Unauthorized response.
3. IF the `connection_counter` update fails, THEN THE Login_Endpoint SHALL log the error and continue processing the login request without interrupting the authentication flow.

### Requirement 6: Suppress Password Reset Email for Banned Cavers

**User Story:** As an administrator, I want the forgot-password flow to silently skip sending reset emails to banned cavers, so that ban status is not leaked to the caller.

#### Acceptance Criteria

1. WHEN a forgot-password request is received for a caver whose `banned` flag is `true`, THE Forgot_Password_Controller SHALL return a 200 OK response without generating a reset password token and without sending a reset email.
2. WHEN a forgot-password request is received for a caver whose `banned` flag is `false`, THE Forgot_Password_Controller SHALL generate a reset password token and send the reset email as normal.
3. THE Forgot_Password_Endpoint SHALL return the same 200 OK response for both banned and non-banned cavers, so that the caller cannot distinguish between the two cases.
4. WHEN a forgot-password request is received for a caver whose `banned` flag is `true`, THE Forgot_Password_Controller SHALL log a warning message indicating the banned caver attempted a password reset.

### Requirement 7: Block Password Change for Banned Cavers

**User Story:** As an administrator, I want banned cavers to be unable to change their password via a previously obtained reset token, so that the ban cannot be circumvented through the forgot-password flow.

#### Acceptance Criteria

1. WHEN a password change request via reset token is received for a caver whose `banned` flag is `true`, THE Change_Password_Controller SHALL reject the request with the same response as an expired token, without changing the password.
2. WHEN a password change request via reset token is received for a banned caver, THE Change_Password_Controller SHALL log a warning message indicating the banned caver attempted a password change.
3. THE Change_Password_Endpoint SHALL return the same error response for banned cavers as for expired tokens, so that the caller cannot distinguish between the two cases.

### Requirement 8: Prevent Account Enumeration on Sign-Up

**User Story:** As a system operator, I want the sign-up endpoint to return generic conflict messages, so that callers cannot enumerate which emails or nicknames are already registered.

#### Acceptance Criteria

1. WHEN a sign-up request is received with an email that is already used, THE SignUp_Controller SHALL return a generic 409 Conflict response without specifying whether the email or nickname caused the conflict.
2. WHEN a sign-up request is received with a nickname that is already used, THE SignUp_Controller SHALL return the same generic 409 Conflict response.
3. WHEN a sign-up request is rejected due to a duplicate email or nickname, THE SignUp_Controller SHALL log a warning message with the specific detail indicating which field caused the conflict.


### Requirement 9: Expose Ban Status in Caver Profile for Administrators

**User Story:** As an administrator, I want to see whether a caver is banned when viewing their profile, so that I can quickly assess a caver's account status without querying the database directly.

#### Acceptance Criteria

1. WHEN an administrator sends a GET request to the GetCaver_Endpoint, THE GetCaver_Controller SHALL include the `isBanned` field in the response body for the requested caver.
2. WHEN a non-administrator authenticated caver sends a GET request to the GetCaver_Endpoint, THE GetCaver_Controller SHALL not include the `isBanned` field in the response body.
3. WHEN an unauthenticated caller sends a GET request to the GetCaver_Endpoint, THE GetCaver_Controller SHALL not include the `isBanned` field in the response body.

### Requirement 10: List Banned Cavers

**User Story:** As an administrator, I want the ability to list the banned cavers, so that I can review and manage all currently banned accounts from a single view.

#### Acceptance Criteria

1. WHEN an administrator sends a GET request to the GetBannedCavers_Endpoint, THE GetBannedCavers_Controller SHALL return a list of all cavers whose `banned` flag is `true` in a single response with the shape `{ banned: [{ id, nickname }, ...] }`.
2. WHEN a list-banned-cavers request is received, THE GetBannedCavers_Endpoint SHALL verify that the requesting caver has the Administrator role before processing the request.
3. IF a list-banned-cavers request is received from a caver without the Administrator role, THEN THE GetBannedCavers_Endpoint SHALL return a 403 Forbidden response.
4. WHEN a list-banned-cavers request is received without a valid bearer token, THE API SHALL return a 401 Unauthorized response.
5. WHEN no cavers are currently banned, THE GetBannedCavers_Controller SHALL return a 200 OK response with an empty array: `{ banned: [] }`.
6. THE GetBannedCavers_Controller SHALL include only the `id` and `nickname` fields for each banned caver in the response, matching the toSimpleCaver shape.

ClemRz:
# Design Document: Admin Ban Caver

## Overview

This feature activates the existing `banned` column on `t_caver` by adding two admin-only REST endpoints (`POST /api/v1/cavers/:caverId/ban` and `POST /api/v1/cavers/:caverId/unban`) and modifying the login flow to reject banned cavers. On ban, the caver's `banned` flag is set to `true` and all active tokens are revoked via `BlacklistService.revoke()`. On unban, the flag is reset to `false`. The login controller is extended to check the `banned` flag after successful credential verification, update `date_last_connection` and increment `connection_counter`, and return a 401 with the same generic "Invalid email or password" message if the caver is banned — hiding ban status from the caller.

Additionally, the forgot-password controller is modified to silently suppress token generation and email sending for banned cavers, returning the same 200 OK response regardless of ban status to prevent information leakage.

The change-password controller is hardened to block banned cavers from resetting their password via a previously obtained reset token. When the reset-token path detects a banned caver, it returns the same `res.forbidden('The password reset token has expired.')` response as an expired token — preventing the caller from distinguishing a ban from a genuine expiry.

The sign-up controller is hardened to prevent account enumeration by replacing specific conflict messages (which revealed whether the email or nickname was taken) with a single generic message: `'Email or nickname is already used.'`. Specific details are logged server-side via `sails.log.warn`.

The get-caver controller (`find.js`) is modified to conditionally include the `isBanned` field in the response when the requesting user is an administrator. The field is derived from the caver's `banned` column. Non-admin and unauthenticated callers never see this field, preventing ban status leakage to regular users.

A new admin-only endpoint `GET /api/v1/cavers/banned` returns all currently banned cavers in a single response, giving administrators a centralized view of banned accounts. The endpoint follows the same simple pattern as `get-groups.js`: query the database, map through `toSimpleCaver`, and return the result. Each entry includes `id` and `nickname`.

No new database tables or columns are required — the `banned`, `date_last_connection`, and `connection_counter` columns already exist on `t_caver` but are currently unused.

## Architecture

The feature follows the existing Sails.js controller → service → model pattern:

```mermaid
sequenceDiagram
    participant Admin as Admin Client
    participant Route as config/routes.js
    participant Policy as tokenAuth policy
    participant Ctrl as ban.js / unban.js
    participant Right as RightService
    participant Model as TCaver (Waterline)
    participant BL as BlacklistService

    Admin->>Route: POST /api/v1/cavers/:caverId/ban
    Route->>Policy: tokenAuth
    Policy->>Ctrl: req (with req.token)
    Ctrl->>Right: hasGroup(req.token.groups, ADMINISTRATOR)
    Right-->>Ctrl: true/false
    Ctrl->>Model: findOne({ id: caverId })
    Ctrl->>Model: updateOne({ id: caverId }).set({ banned: true })
    Ctrl->>BL: revoke(caverId)
    Ctrl-->>Admin: 200 OK
```

```mermaid
sequenceDiagram
    participant User as Caver Client
    participant Login as login.js
    participant Auth as AuthService
    participant Model as TCaver (Waterline)
    participant Token as TokenService

    User->>Login: POST /api/v1/login { email, password }
    Login->>Auth: authenticate(email, password)
    Auth-->>Login: { status: SUCCESS, user }
    Login->>Model: updateOne({ id }).set({ dateLastConnection, connectionCounter })
    alt user.banned === true
        Login->>Login: sails.log.warn("Banned caver attempted login")
        Login-->>User: 401 Unauthorized "Invalid email or password"
    else
        Login->>Token: issue(payload)
        Login-->>User: 200 { token }
    end
```

```mermaid
sequenceDiagram
    participant Client as Client
    participant FP as forgot-password.js
    participant Model as TCaver (Waterline)
    participant Token as TokenService
    participant Email as sails.helpers.sendEmail

    Client->>FP: POST /api/v1/forgotPassword { email }
    FP->>Model: findOne({ mail }).populate('language')
    alt caver not found
        FP-->>Client: 404 Not Found
    else caver.banned === true
        FP->>FP: sails.log.warn("Banned caver attempted password reset")
        FP-->>Client: 200 OK (no token, no email)
    else caver.banned === false
        FP->>Token: issue(payload, ttl, salt)
        FP->>Email: sendEmail(resetLink)
        FP-->>Client: 200 OK
    end
```

```mermaid
sequenceDiagram
    participant Client as Client
    participant CP as change-password.js
    participant JWT as jwt.decode
    participant Model as TCaver (Waterline)
    participant Token as TokenService
    participant Auth as AuthService

    Client->>CP: PUT /api/v1/account/password { token, password }
    CP->>JWT: decode(token)
    JWT-->>CP: { userId }
    CP->>Model: findOne(userId)
    alt caver not found
        CP-->>Client: 404 Not Found
    else caver.banned === true
        CP->>CP: sails.log.warn("Banned caver attempted password change via reset token")
        CP-->>Client: 403 Forbidden "The password reset token has expired."
    else caver.banned === false
        CP->>Token: verify(token, callback, salt)
        Token-->>CP: callback(err)
        alt token valid
            CP->>Auth: createHashedPassword(password)
            CP->>Model: updateOne({ id }).set({ password })
            CP-->>Client: 200 OK
        else token expired/invalid
            CP-->>Client: 403 Forbidden (error message)
        end
    end
```

```mermaid
sequenceDiagram
    participant Client as Client
    participant GC as find.js (GetCaver)
    participant CS as CaverService
    participant DS as DocumentService
    participant Right as RightService
    participant Ctrl as ControllerService

    Client->>GC: GET /api/v1/cavers/:id
    GC->>CS: getCaver(caverId)
    CS-->>GC: caverFound (includes banned field)
    GC->>DS: getDocuments(caverFound.documents)
    DS-->>GC: documents
    alt req.token exists AND req.token.groups includes Administrator
        GC->>Right: hasGroup(req.token.groups, ADMINISTRATOR)
        Right-->>GC: true
        GC->>GC: Attach isBanned = caverFound.banned to response
    else non-admin or unauthenticated
        GC->>GC: Do not attach isBanned
    end
    GC->>Ctrl: treatAndConvert(req, null, caverFound, params, res, toCaver)
    Ctrl-->>Client: 200 OK (with or without isBanned)
```

```mermaid
sequenceDiagram
    participant Admin as Admin Client
    participant Route as config/routes.js
    participant Policy as tokenAuth policy
    participant Ctrl as get-banned.js
    participant Right as RightService
    participant Model as TCaver (Waterline)

    Admin->>Route: GET /api/v1/cavers/banned
    Route->>Policy: tokenAuth
    Policy->>Ctrl: req (with req.token)
    Ctrl->>Right: hasGroup(req.token.groups, ADMINISTRATOR)
    Right-->>Ctrl: true/false
    alt not admin
        Ctrl-->>Admin: 403 Forbidden
    else admin
        Ctrl->>Model: find({ banned: true })
        Model-->>Ctrl: bannedCavers
        Ctrl->>Ctrl: bannedCavers.map(toSimpleCaver)
        Ctrl-->>Admin: 200 OK { banned: [{ id, nickname }, ...] }
    end
```

### Design Decisions

1. **Two separate endpoints (ban/unban) instead of a PATCH on `banned`**: Matches the existing pattern for state-changing actions in this codebase (e.g., `subscribe`/`unsubscribe`, `put-on-group`/`remove-from-group`). POST actions are explicit and auditable.

2. **Token revocation on ban only, not on unban**: Banning must immediately invalidate sessions. Unbanning has no tokens to revoke since the caver couldn't log in while banned.

3. **Self-ban prevention**: An admin cannot ban themselves. This prevents accidental lockout of the only admin account.

4. **Idempotent ban/unban**: Banning an already-banned caver or unbanning a non-banned caver returns 200 without side effects. This simplifies client logic and avoids race conditions.

5. **`date_last_connection` and `connection_counter` update on all successful credential verifications, including banned cavers**: Per requirements 4.2 and 5.2, these fields track authentication attempts, not successful logins. This gives admins visibility into banned cavers still trying to access the system.

6. **Fire-and-forget for metadata updates**: Per requirements 4.3 and 5.3, failures in updating `date_last_connection` or `connection_counter` are logged but do not interrupt the login flow.

7. **Silent suppression on forgot-password for banned cavers**: Per requirement 6, the forgot-password controller returns 200 OK for banned cavers without generating a token or sending an email. The response is identical to the non-banned case, preventing callers from inferring ban status. Both the login and forgot-password flows follow the same principle: never reveal ban status to the caller. A `sails.log.warn` is emitted in both cases so administrators can monitor banned caver activity in server logs.

8. **Banned caver login returns generic 401, not 403**: Per requirement 3.1, a banned caver's login attempt returns `res.unauthorized({ status: 'Mismatch', message: 'Invalid email or password.' })` — the exact same response shape and status code used for invalid credentials. Returning a distinct 403 or "Banned" status would let an attacker enumerate which accounts are banned. A `sails.log.warn` is emitted before returning so admins have visibility without leaking information to the client.

9. **Banned caver password change via reset token returns expired-token response**: Per requirement 7, when a banned caver attempts to change their password using a reset token, the controller returns `res.forbidden('The password reset token has expired.')` — the exact same response as a genuinely expired token. The check is placed after `TCaver.findOne(decodedToken.userId)` and before `TokenService.verify()`, so the token is never validated and the password is never updated. This follows the same information-hiding principle as the login and forgot-password flows: never reveal ban status to the caller.

10. **Generic conflict message on sign-up to prevent account enumeration**: Per requirement 8, the sign-up controller replaces specific conflict messages (`The email X is already used.` / `The nickname X is already used.`) with a single generic message: `'Email or nickname is already used.'`. The specific detail is logged via `sails.log.warn` before returning, giving admins visibility without leaking information to the caller. This prevents attackers from probing the sign-up endpoint to discover which emails or nicknames are registered.

11. **Conditional `isBanned` field in GetCaver response**: Per requirement 9, the `find.js` controller checks `req.token` and `RightService.hasGroup` to determine if the caller is an administrator. If so, the `isBanned` boolean (derived from `caverFound.banned`) is attached to the response object before it passes through the `toCaver` converter. Non-admin and unauthenticated callers never see this field. The check is performed in the controller rather than the converter because the converter has no access to request context (`req.token`). The field is named `isBanned` (not `banned`) to match the API's camelCase convention and distinguish it from the internal column name.

12. **Dedicated `/cavers/banned` endpoint with simple response (no pagination)**: Per requirement 10, a dedicated `GET /api/v1/cavers/banned` endpoint is used rather than adding a `?banned=true` query parameter to an existing caver listing endpoint. Rationale: (a) the existing paginated caver list endpoints (`get-users`, `get-authors`, `get-contributors`) each query the `v_caver_roles` view with role-specific filters — there is no general-purpose "list all cavers" endpoint to extend; (b) a dedicated endpoint makes the admin-only access control explicit in `config/policies.js` without conditional logic; (c) the response shape is intentionally minimal (`id`, `nickname` via `toSimpleCaver`) matching the same pattern used by `get-groups.js`. Pagination is not needed because the number of banned cavers is expected to be a small set — this is an admin management tool, not a general-purpose listing. The simple `{ banned: [...] }` response keeps the implementation lean and consistent with `get-groups.js`.

## Components and Interfaces

### New Files

| File | Purpose |
|------|---------|
| `api/controllers/v1/caver/ban.js` | Ban endpoint controller |
| `api/controllers/v1/caver/unban.js` | Unban endpoint controller |
| `api/controllers/v1/caver/get-banned.js` | List banned cavers endpoint controller |

### Modified Files

| File | Change |
|------|--------|
| `config/routes.js` | Add two new route entries for ban/unban, one new route entry for get-banned |
| `config/policies.js` | Add policy entries for ban, unban, and get-banned actions |
| `api/controllers/v1/auth/login.js` | Add banned check, metadata updates |
| `api/controllers/v1/account/forgot-password.js` | Add early return for banned cavers before token generation and email sending |
| `api/controllers/v1/account/change-password.js` | Add banned check before token verification in the reset-token path |
| `api/controllers/v1/auth/sign-up.js` | Replace specific conflict messages with generic message |
| `api/controllers/v1/caver/find.js` | Conditionally set `isBanned` on caverFound for admin callers |
| `api/services/mapping/converters.js` | Pass through `isBanned` in `toCaver` when present on source |

### Controller: `ban.js`

```javascript
// POST /api/v1/cavers/:caverId/ban
module.exports = async (req, res) => {
  // 1. Check admin permission via RightService.hasGroup
  // 2. Validate caverId param
  // 3. Prevent self-ban (req.token.id === caverId)
  // 4. Find caver or return 404
  // 5. If already banned, return 200 (idempotent)
  // 6. Update banned = true
  // 7. Revoke tokens via BlacklistService.revoke(caverId)
  // 8. Return 200
};
```

### Controller: `unban.js`

```javascript
// POST /api/v1/cavers/:caverId/unban
module.exports = async (req, res) => {
  // 1. Check admin permission via RightService.hasGroup
  // 2. Validate caverId param
  // 3. Find caver or return 404
  // 4. If not banned, return 200 (idempotent)
  // 5. Update banned = false
  // 6. Return 200
};
```

### Modified Controller: `login.js`

```javascript
// After AuthService.authenticate returns SUCCESS:
// 1. Update date_last_connection and increment connection_counter (fire-and-forget with error logging)
// 2. Check user.banned — if true:
//    a. sails.log.warn(`Banned caver ${result.user.id} attempted login`)
//    b. return res.unauthorized({ status: 'Mismatch', message: 'Invalid email or password.' })
//       (same response as invalid credentials — hides ban status)
// 3. Otherwise issue token and return 200
```

### Modified Controller: `forgot-password.js`

```javascript
// POST /api/v1/forgotPassword
// After finding the caver via TCaver.findOne:
// 1. If userFound.banned === true:
//    a. sails.log.warn(`Banned caver ${userFound.id} attempted password reset`)
//    b. return res.ok() immediately (skip token generation and email sending)
// 2. Otherwise proceed with existing logic (generate token, send email, return res.ok())
```

### Modified Controller: `change-password.js`

```javascript
// PUT /api/v1/account/password
// In the reset-token path (no req.token), after TCaver.findOne(decodedToken.userId):
// 1. If userFound.banned === true:
//    a. sails.log.warn(`Banned caver ${userFound.id} attempted password change via reset token`)
//    b. return res.forbidden('The password reset token has expired.')
//       (same response as an expired token — hides ban status)
//    c. Do NOT proceed to TokenService.verify() or update the password
// 2. Otherwise proceed with existing logic (verify token, update password)
```

### Modified Controller: `sign-up.js`

```javascript
// POST /api/v1/auth/sign-up
// Replace specific conflict messages with generic ones:
// 1. After TCaver.findOne({ mail: email }):
//    - If caverEmail exists:
//      a. sails.log.warn(`Sign-up attempt with existing email: ${email}`)
//      b. return res.conflict('Email or nickname is already used.')
// 2. After TCaver.findOne({ nickname }):
//    - If caverNickname exists:
//      a. sails.log.warn(`Sign-up attempt with existing nickname: ${nickname}`)
//      b. return res.conflict('Email or nickname is already used.')
```

### Modified Controller: `find.js`

```javascript
// GET /api/v1/cavers/:id
// After fetching caverFound and populating documents, before calling treatAndConvert:
// 1. Check if req.token exists AND RightService.hasGroup(req.token.groups, RightService.G.ADMINISTRATOR)
// 2. If admin: set caverFound.isBanned = Boolean(caverFound.banned)
// 3. If not admin or unauthenticated: do nothing (isBanned is not set on caverFound)
// 4. Pass caverFound through ControllerService.treatAndConvert with toCaver converter as before
```

### Modified Converter: `toCaver` in `converters.js`

```javascript
// In the toCaver function, after building the result object:
// If source.isBanned is not undefined, include it in the result:
//   if (source.isBanned !== undefined) {
//     result.isBanned = source.isBanned;
//   }
// This ensures isBanned only appears when explicitly set by the controller (admin requests).
```

### Controller: `get-banned.js`

```javascript
// GET /api/v1/cavers/banned
const { toSimpleCaver } = require('../../../services/mapping/converters');

module.exports = async (req, res) => {
  // 1. Check admin permission via RightService.hasGroup
  // 2. If not admin, return 403
  // 3. Find all banned cavers: TCaver.find({ banned: true })
  // 4. Map through toSimpleCaver converter
  // 5. Return res.ok({ banned: bannedCavers.map(toSimpleCaver) })
};
```

### Route Definitions

```javascript
// config/routes.js — added to the Caver section
'POST /api/v1/cavers/:caverId/ban': 'v1/caver/ban',
'POST /api/v1/cavers/:caverId/unban': 'v1/caver/unban',
'GET /api/v1/cavers/banned': 'v1/caver/get-banned',
```

### Policy Definitions

```javascript
// config/policies.js — added to the Caver section
'v1/caver/ban': 'tokenAuth',
'v1/caver/unban': 'tokenAuth',
'v1/caver/get-banned': 'tokenAuth',
```

## Data Models

No schema changes are required. The following existing columns on `t_caver` are activated:

| Column | Type | Current State | New Usage |
|--------|------|---------------|-----------|
| `banned` | `boolean` (default `false`) | Marked "Unused" | Set to `true`/`false` by ban/unban endpoints; checked during login |
| `date_last_connection` | `timestamp` | Marked "Unused" | Updated to `NOW()` on every successful credential verification |
| `connection_counter` | `integer` (default `0`) | Marked "Unused" | Incremented by 1 on every successful credential verification |

### Response Shapes

**Ban/Unban success (200)**:
```json
{ "banned": true }
```
or
```json
{ "banned": false }
```

**Ban/Unban — already in target state (200)**:
```json
{ "banned": true }
```

**Banned caver login attempt (401)** — uses the same response as invalid credentials to hide ban status:
```json
{
  "status": "Mismatch",
  "message": "Invalid email or password."
}
```

**List banned cavers (200)**:
```json
{
  "banned": [
    {
      "id": 42,
      "nickname": "spelunker99"
    }
  ]
}
```

**List banned cavers — empty (200)**:
```json
{
  "banned": []
}
```


## Correctness Properties

*A property is a characteristic or behavior that should hold true across all valid executions of a system — essentially, a formal statement about what the system should do. Properties serve as the bridge between human-readable specifications and machine-verifiable correctness guarantees.*

### Property 1: Ban/unban round trip

*For any* existing non-banned caver, banning and then unbanning that caver should result in `banned` being `false`, restoring the caver to their original ban state.

**Validates: Requirements 1.1, 2.1**

### Property 2: Ban revokes all active tokens

*For any* existing non-banned caver, after a successful ban request, `BlacklistService.isRevoked(caverId, iatBeforeBan)` should return `true` for any token issued before the ban.

**Validates: Requirements 1.1**

### Property 3: Non-admin ban is rejected

*For any* authenticated caver whose groups do not include Administrator, a ban request should return a 403 Forbidden response and the target caver's `banned` flag should remain unchanged.

**Validates: Requirements 1.2, 1.3**

### Property 4: Non-admin unban is rejected

*For any* authenticated caver whose groups do not include Administrator, an unban request should return a 403 Forbidden response and the target caver's `banned` flag should remain unchanged.

**Validates: Requirements 2.2, 2.3**

### Property 5: Ban is idempotent

*For any* already-banned caver, calling the ban endpoint should return 200 and the caver's `banned` flag should remain `true` with no additional token revocation side effects.

**Validates: Requirements 1.5**

### Property 6: Unban is idempotent

*For any* non-banned caver, calling the unban endpoint should return 200 and the caver's `banned` flag should remain `false`.

**Validates: Requirements 2.5**

### Property 7: Banned caver login is rejected without token issuance

*For any* caver with `banned = true` who provides valid credentials, the login endpoint should return a 401 Unauthorized response with `{ status: "Mismatch", message: "Invalid email or password." }` — identical to the invalid-credentials response — and the response body should not contain a `token` field.

**Validates: Requirements 3.1, 3.2, 3.3**

### Property 8: Successful credential verification updates date_last_connection

*For any* caver (banned or not) who provides valid credentials, after calling the login endpoint the `date_last_connection` column should be updated to a timestamp within a reasonable window of the current time.

**Validates: Requirements 4.1, 4.2**

### Property 9: Successful credential verification increments connection_counter

*For any* caver (banned or not) who provides valid credentials and whose `connection_counter` is `N` before login, after calling the login endpoint the `connection_counter` should be `N + 1`.

**Validates: Requirements 5.1, 5.2**

### Property 10: Forgot-password silently suppresses reset for banned cavers

*For any* caver, the forgot-password endpoint should return a 200 OK response. If the caver's `banned` flag is `true`, no reset password token should be generated and no email should be sent. If the caver's `banned` flag is `false`, a reset password token should be generated and an email should be sent. The HTTP status code and response body should be identical in both cases.

Note: Both the banned-login and banned-forgot-password paths emit a `sails.log.warn` before returning. This logging behavior is verified in integration tests (not as a property) since it is a side effect observable only via log output, not via HTTP response.

**Validates: Requirements 6.1, 6.2, 6.3, 6.4**

### Property 11: Banned caver password change via reset token is rejected

*For any* caver with `banned = true` who has a valid reset token, calling the change-password endpoint with that token should return a 403 Forbidden response with the message `'The password reset token has expired.'` — identical to the expired-token response — and the caver's password should remain unchanged.

**Validates: Requirements 7.1, 7.3**

### Property 12: Sign-up conflict returns generic message

*For any* sign-up request where the email or nickname is already registered, the sign-up endpoint should return a 409 Conflict response with the message `'Email or nickname is already used.'`. The response should not contain the specific email address or nickname that caused the conflict.

**Validates: Requirements 8.1, 8.2**

### Property 13: Admin-only ban status exposure

*For any* caver and any caller, the `isBanned` field should appear in the `GET /api/v1/cavers/:id` response if and only if the caller is an administrator. When present, `isBanned` should equal the caver's `banned` column value. For non-admin authenticated callers and unauthenticated callers, the response should not contain the `isBanned` field.

**Validates: Requirements 9.1, 9.2, 9.3**

### Property 14: Admin-only banned cavers listing returns only banned cavers

*For any* set of cavers with varying banned states, when an administrator calls `GET /api/v1/cavers/banned`, the response should contain only cavers whose `banned` flag is `true`, each entry should include `id` and `nickname` fields (matching the `toSimpleCaver` shape), and the response should have the shape `{ banned: [...] }`. For non-admin authenticated callers, the endpoint should return a 403 Forbidden response.

**Validates: Requirements 10.1, 10.2, 10.3, 10.6**

## Error Handling

| Scenario | Response | Notes |
|----------|----------|-------|
| No bearer token on ban/unban | 401 Unauthorized | Handled by existing `tokenAuth` policy |
| Non-admin calls ban/unban | 403 Forbidden | Checked via `RightService.hasGroup` |
| Admin bans themselves | 403 Forbidden | `req.token.id === caverId` guard |
| Target caver not found | 404 Not Found | `TCaver.findOne` returns `undefined` |
| Already banned / already unbanned | 200 OK | Idempotent — no state change, returns current state |
| Banned caver login | 401 Unauthorized | `{ status: "Mismatch", message: "Invalid email or password." }` — same as invalid credentials; `sails.log.warn` emitted server-side |
| `date_last_connection` update fails | Log error, continue | Non-blocking — login flow proceeds normally |
| `connection_counter` update fails | Log error, continue | Non-blocking — login flow proceeds normally |
| Database error during ban/unban | 500 Server Error | Unhandled Waterline errors bubble up to Sails default error handler |
| Forgot-password for banned caver | 200 OK | Silent suppression — no token generated, no email sent, `sails.log.warn` emitted, response identical to non-banned case |
| Forgot-password email service error | 500 Server Error | Existing `sendSESEmailError` intercept in forgot-password controller (non-banned path only) |
| Banned caver password change via reset token | 403 Forbidden | `'The password reset token has expired.'` — same as expired token; `sails.log.warn` emitted server-side; token not verified, password not updated |
| Sign-up with duplicate email | 409 Conflict | `'Email or nickname is already used.'` — generic message; `sails.log.warn` with specific detail emitted server-side |
| Sign-up with duplicate nickname | 409 Conflict | `'Email or nickname is already used.'` — same generic message; `sails.log.warn` with specific detail emitted server-side |
| GetCaver by admin | 200 OK | Response includes `isBanned` field with the caver's ban status |
| GetCaver by non-admin or unauthenticated | 200 OK | Response does not include `isBanned` field |
| No bearer token on get-banned | 401 Unauthorized | Handled by existing `tokenAuth` policy |
| Non-admin calls get-banned | 403 Forbidden | Checked via `RightService.hasGroup` |
| No banned cavers found | 200 OK | Returns `{ banned: [] }` |
| Database error during get-banned | 500 Server Error | Unhandled Waterline errors bubble up to Sails default error handler |

## Testing Strategy

### Dual Testing Approach

This feature uses both unit/integration tests and property-based tests for comprehensive coverage.

**Integration tests** (Mocha + Supertest + should.js) cover:
- Specific HTTP status codes for each endpoint scenario (401, 403, 404, 200)
- Self-ban prevention (edge case)
- Non-existent caver (edge case)
- Unauthenticated requests (edge case)
- `date_last_connection`/`connection_counter` update failure resilience (error conditions, using Sinon stubs)
- Forgot-password silent suppression for banned cavers (200 OK, no email sent)
- Forgot-password normal flow for non-banned cavers (200 OK, email sent)
- Banned caver password change via reset token returns 403 with expired-token message
- Non-banned caver password change via reset token proceeds normally
- `sails.log.warn` emitted for banned caver password change attempt (Sinon stub verification)
- Sign-up with duplicate email returns generic 409 conflict message
- Sign-up with duplicate nickname returns generic 409 conflict message
- `sails.log.warn` emitted with specific detail on sign-up conflict (Sinon stub verification)
- GetCaver as admin returns `isBanned` field matching the caver's `banned` column
- GetCaver as non-admin authenticated caver does not return `isBanned` field
- GetCaver as unauthenticated caller does not return `isBanned` field
- GetBannedCavers returns 401 when no bearer token
- GetBannedCavers returns 403 when non-admin calls endpoint
- GetBannedCavers returns 200 with empty banned array when no cavers are banned
- GetBannedCavers returns 200 with correct banned cavers, each having id and nickname

**Property-based tests** (fast-check) cover:
- Universal properties that hold across all valid inputs (Properties 1–14)
- Each property test runs a minimum of 100 iterations
- Each property test references its design document property with a tag comment

### Property-Based Testing Configuration

- Library: `fast-check` (already available in the project's test ecosystem)
- Minimum iterations: 100 per property (`{ numRuns: 100 }`)
- Test file suffix: `.property.test.js`
- Test location: `test/integration/4_routes/caver/`
- Tag format: `Feature: admin-ban-caver, Property {number}: {property_text}`

Each correctness property (1–14) is implemented by a single `fc.assert(fc.property(...))` or `fc.assert(fc.asyncProperty(...))` call. For database-interacting properties, use `fc.asyncProperty` with generous timeouts and clean up created records in a `finally` block.

### Test File Organization

```
test/integration/4_routes/caver/
├── ban.test.js                    — integration tests for ban endpoint
├── unban.test.js                  — integration tests for unban endpoint
├── ban-unban.property.test.js     — property-based tests for Properties 1–14
├── find-banned.test.js            — integration tests for GetCaver isBanned exposure
├── get-banned.test.js             — integration tests for GetBannedCavers endpoint
test/integration/4_routes/auth/
├── login.test.js                  — integration tests for login banned check + metadata updates
└── sign-up.test.js                — integration tests for sign-up enumeration prevention
test/integration/4_routes/account/
├── forgot-password-banned.test.js — integration tests for forgot-password banned suppression
└── change-password-banned.test.js — integration tests for change-password banned check
```

Paul-AUB (changes requested):
# Code Review — commit `056805e8` `feat(caver): add admin ban/unban endpoints with security hardening`

## Context

This commit introduces ban/unban management for cavers, with a deliberately hardened security posture: generic error messages to prevent information leakage, immediate token revocation on ban, login metadata tracking, and account enumeration prevention on sign-up.

---

## Issues

### 1. Dead guard for `caverId` — [`api/controllers/v1/caver/ban.js:12-14`](api/controllers/v1/caver/ban.js#L12)

```js
const { caverId } = req.params;
if (!caverId) {
  return res.badRequest('Missing required parameter: caverId.');
}
```

The route is `POST /api/v1/cavers/:caverId/ban`. Sails.js only dispatches to this handler when `:caverId` is matched in the URL, so `req.params.caverId` is always defined at this point. The guard is unreachable dead code and should be removed.

Same applies to `unban.js:12-14`.

---

### 2. Idempotent early-return skips token revocation — [`api/controllers/v1/caver/ban.js:25-27`](api/controllers/v1/caver/ban.js#L25)

```js
if (caver.banned) {
  return res.ok({ banned: true });
}
```

When the caver is already flagged as `banned` in the database (e.g., set manually or via a direct DB migration), calling `POST /ban` returns early **without** calling `BlacklistService.revoke`. This means any active sessions for that caver are not invalidated by the API call.

Recommendation: move the idempotency return **after** the revoke call, or always call `BlacklistService.revoke` unconditionally before the early return.

---

### 3. "Fire-and-forget" that actually blocks — [`api/controllers/v1/auth/login.js:37-44`](api/controllers/v1/auth/login.js#L37)

```js
// Update login metadata (fire-and-forget)
try {
  await CommonService.query(
    `UPDATE t_caver SET date_last_connection = NOW(), connection_counter = connection_counter + 1 WHERE id = $1`,
    [result.user.id]
  );
} catch (e) { ... }
```

The comment says "fire-and-forget" but `await` is used, meaning every login blocks until the metadata UPDATE completes. True fire-and-forget would drop the `await`:

```js
CommonService.query(...).catch((e) => sails.log.error(...));
```

As written, a slow DB write adds latency to every successful login attempt.

---

### 4. Unbounded query in `get-banned` — [`api/controllers/v1/caver/get-banned.js:13`](api/controllers/v1/caver/get-banned.js#L13)

```js
const bannedCavers = await TCaver.find({ banned: true });
```

`TCaver.find()` with no `limit` fetches all matching rows in a single query. On a large dataset this can cause memory pressure and slow responses. Consider adding pagination (`skip` / `limit`) consistent with other list endpoints in the codebase.

---

### 5. ORM object mutation before conversion — [`api/controllers/v1/caver/find.js:20-22`](api/controllers/v1/caver/find.js#L20)

```js
caverFound.isBanned = Boolean(caverFound.banned);
```

This mutates the raw Waterline ORM instance before passing it to `toCaver`. The converter then checks `source.isBanned !== undefined` to decide whether to include the field. This creates an implicit coupling between the controller and the converter via a sentinel value (`undefined` as absence). A cleaner approach would be to pass extra context as a second argument to `toCaver` rather than patching the source object.

---

### 6. `toListCaver` silently expands existing endpoint shapes — [`api/services/mapping/converters.js:122-128`](api/services/mapping/converters.js#L122)

The three endpoints `get-admins`, `get-groups`, and `get-moderators` previously used `toSimpleCaver` (which returned only `{ id, nickname }`). They now use `toListCaver` (which adds `name` and `surname`). While the change is likely intentional (and benign), it silently extends the public API contract of those three endpoints without a mention in the commit message or Swagger. Any client relying on the exact shape of those responses could be affected.

---

### 7. Missing `security` declaration for `/cavers/{caverId}/ban` in Swagger — [`assets/swaggerV1.yaml`](assets/swaggerV1.yaml)

The `POST /cavers/{caverId}/unban` block includes:
```yaml
security:
  - bearerAuth: []
```
The `POST /cavers/{caverId}/ban` block does not. Authentication is still enforced at runtime through the `tokenAuth` policy, but the omission means the Swagger UI will not show the padlock or prompt for a token when calling the ban endpoint interactively.

---

## Positive notes

- The security model is coherent: every user-facing error for banned cavers returns the same shape as the equivalent non-ban error, with no status code divergence. The "identical response" property test (`Property 10`, `Property 12`) verifies this explicitly.
- Immediate token revocation via `BlacklistService.revoke` on ban is the right call — no window where a banned caver can continue using an in-flight session.
- Route ordering in `routes.js` is correct: `GET /api/v1/cavers/banned` is declared before `GET /api/v1/cavers/:id`, ensuring the static segment is not consumed as a numeric id.
- 14 property-based tests with `fast-check` cover the entire feature surface, including round-trips, idempotency, role matrix, connection metadata, and sign-up enumeration. This is a high-quality test suite.
- The `// Unsued` comments in `TCaver.js` (pre-existing typo) are removed as the fields are activated — clean housekeeping.
- Self-ban guard in `ban.js` prevents an administrator from locking themselves out.

---

## Summary

| Severity | File | Finding |
|---|---|---|
| Medium | `ban.js:12`, `unban.js:12` | Dead guard — `caverId` is always present via route matching |
| Medium | `ban.js:25-27` | Idempotent early-return skips `BlacklistService.revoke` |
| Low | `login.js:37-44` | "Fire-and-forget" blocks on `await` — adds latency to every login |
| Low | `get-banned.js:13` | No pagination on potentially unbounded banned-caver list |
| Nit | `find.js:20-22` | ORM object mutation as implicit converter input channel |
| Nit | `converters.js:122-128` | `toListCaver` extends existing endpoint shapes without mention |
| Nit | `swaggerV1.yaml` | Missing `security: bearerAuth` for the ban endpoint |

ClemRz (commented):
Thanks for the thorough review @Paul-AUB — here's what I addressed:

### 1. Dead guard for `caverId` ✅ Fixed
Removed the unreachable `if (!caverId)` guard from both `ban.js` and `unban.js`. Sails route matching guarantees the param is present.

### 2. Idempotent early-return skips token revocation ✅ Fixed
`BlacklistService.revoke()` is now called unconditionally — the DB update is skipped if already banned, but revocation always runs. This handles the case where `banned` was set via direct DB access without going through the API.

### 3. "Fire-and-forget" that actually blocks ✅ Fixed
Dropped the `await` and switched to `.catch()` for true fire-and-forget. Tests now include a small settle delay (`waitForMetadata`) to account for the async write.

### 4. Unbounded query in `get-banned` — Deferred
This follows the same pattern as `get-groups`, `get-admins`, and `get-moderators` (no pagination). The number of banned cavers is expected to stay small, and this is admin-only. Happy to add pagination in a follow-up if needed.

### 5. ORM object mutation before conversion ✅ Fixed
`toCaver` now accepts a `meta` argument (like other converters). The controller passes `{ isAdmin: true }` and the converter reads `meta.isAdmin` to conditionally include `isBanned` from `source.banned`. No more patching the ORM instance.

### 6. `toListCaver` extends existing endpoint shapes — Acknowledged
This is intentional (mentioned in the PR body). The swagger for `get-admins` and `get-moderators` already references the full `Caver` schema, which is a superset of what `toListCaver` returns — so no contract breakage.

### 7. Missing `security` for ban in Swagger — Already present
The `security: - bearerAuth: []` block is already on the `POST /cavers/{caverId}/ban` path (line 969 of `swaggerV1.yaml`). Might have been looking at a stale diff?

Paul-AUB (approved):

PR #1526, opened 2026-04-03
fix(search): forward Typesense 4xx errors as HTTP 400

ClemRz:
## 🤔 What

- Forward Typesense 4xx client errors as HTTP 400 instead of 500 on all search endpoints
- Remove the local `validateSort()` pre-validation in favor of letting Typesense reject invalid sort parameters directly
- Affects: `POST /api/v1/search`, `POST /api/v1/advanced-search`, `POST /api/v1/advanced-search/export`, `POST /api/v1/field-search`

## 🤷‍♂️ Why

When a client sends a malformed search request (e.g., a misspelled filter field like `isTouristicz`), Typesense correctly rejects it with HTTP 400 `RequestMalformed`. However, the API was swallowing that error and returning a generic HTTP 500 "An internal server error occurred." to the client — making it impossible to understand what went wrong.

This applied to all Typesense client errors: invalid filter fields, bad query syntax, non-existent collections, unsortable fields, etc.

## 🔍 How

- Created a centralized `TypesenseErrorService.handleTypesenseError(res, error)` helper that duck-types on `error.httpStatus` — if it's in the 400–499 range, it calls `res.badRequest()` with the Typesense error message and returns truthy; otherwise returns `false`
- Added try/catch with `handleTypesenseError` in all four search controllers
- Removed `validateSort()` from `SearchService.js` — invalid sort parameters now flow through to Typesense and come back as 400s via the same centralized handler, simplifying the code without degrading the user experience
- The `hasSentHeader` guard in `advanced-search-export.js` ensures we only return a JSON 400 if CSV headers haven't been written yet (streaming safety)

## 🧪 Testing

Property-based tests with `fast-check` covering:
- **Bug condition**: random `httpStatus` in [400, 499] with random error messages on all 4 endpoints → assert HTTP 400 with error message
- **Preservation**: valid searches → 200, Typesense 5xx → 500, input validation (missing query/field/entity) → 400
- **Sort delegation**: invalid sort inputs are passed through to Typesense (no local interception)

### Manual testing with curl

These requests would have returned 500 before this fix and now return 400 with a descriptive error message:

**Invalid filter field name:**
```bash
curl -s -X POST http://localhost:1337/api/v1/advanced-search \
  -d '{"query":"","entity":"entrances","filter":{"isTouristicz":true}}'
```

**Invalid sort field:**
```bash
curl -s -X POST http://localhost:1337/api/v1/advanced-search \
  -d '{"query":"test","entity":"entrances","sort":"nonExistentField:asc"}'
```

**Invalid filter syntax in export:**
```bash
curl -s -X POST http://localhost:1337/api/v1/advanced-search/export \
  -d '{"query":"","entity":"entrances","filter":{"isTouristicz":true},"columns":["id"],"columnsName":["ID"]}'
```

**Invalid field in field-search:**
```bash
curl -s -X POST http://localhost:1337/api/v1/field-search \
  -d '{"entity":"caves","field":"nonExistentField","query":"test"}'
```

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/api/controllers/v1/search/advanced-search-export.js
+++ b/api/controllers/v1/search/advanced-search-export.js
@@ -1,4 +1,7 @@
 const SearchService = require('../../../services/SearchService');
+const {
+  handleTypesenseError,
+} = require('../../../services/TypesenseErrorService');
 
 function escapeCSV(v) {
   // Escape double quotes
@@ -106,13 +109,8 @@ module.exports = async (req, res) => {
         size: BATCH_SIZE,
       });
     } catch (err) {
-      if (err.code === 'E_SORT_VALIDATION') {
-        if (!hasSentHeader) {
-          res.badRequest({ error: err.message });
-          return;
-        }
-        break;
-      }
+      if (!hasSentHeader && handleTypesenseError(res, err)) return;
+      if (hasSentHeader) break;
       results = err;
     }
 
--- a/api/controllers/v1/search/advanced-search.js
+++ b/api/controllers/v1/search/advanced-search.js
@@ -1,5 +1,8 @@
 const ControllerService = require('../../../services/ControllerService');
 const SearchService = require('../../../services/SearchService');
+const {
+  handleTypesenseError,
+} = require('../../../services/TypesenseErrorService');
 const { toSearchResult } = require('../../../services/mapping/converters');
 
 module.exports = async (req, res) => {
@@ -18,9 +21,7 @@ module.exports = async (req, res) => {
       size: req.param('size') ?? 10,
     });
   } catch (error) {
-    if (error.code === 'E_SORT_VALIDATION') {
-      return res.badRequest({ error: error.message });
-    }
+    if (handleTypesenseError(res, error)) return undefined;
     throw error;
   }
 
--- a/api/controllers/v1/search/field-search.js
+++ b/api/controllers/v1/search/field-search.js
@@ -1,5 +1,8 @@
 const ControllerService = require('../../../services/ControllerService');
 const SearchService = require('../../../services/SearchService');
+const {
+  handleTypesenseError,
+} = require('../../../services/TypesenseErrorService');
 
 module.exports = async (req, res) => {
   const field = req.param('field');
@@ -16,14 +19,20 @@ module.exports = async (req, res) => {
   let matchAllFields = req.param('matchAllFields') ?? true;
   if (!matchAllFields || matchAllFields === 'false') matchAllFields = false;
 
-  const r = await SearchService.fieldSearch({
-    entity,
-    field,
-    query: req.param('query'),
-    size: req.param('size') ?? 10,
-    filter: req.param('filter') ?? {},
-    isLogicalCompareAnd: !!matchAllFields,
-  });
+  let r;
+  try {
+    r = await SearchService.fieldSearch({
+      entity,
+      field,
+      query: req.param('query'),
+      size: req.param('size') ?? 10,
+      filter: req.param('filter') ?? {},
+      isLogicalCompareAnd: !!matchAllFields,
+    });
+  } catch (error) {
+    if (handleTypesenseError(res, error)) return;
+    throw error;
+  }
 
   const out = {
     totalDistinct: r.found,
--- a/api/controllers/v1/search/quick-search.js
+++ b/api/controllers/v1/search/quick-search.js
@@ -1,5 +1,8 @@
 const ControllerService = require('../../../services/ControllerService');
 const SearchService = require('../../../services/SearchService');
+const {
+  handleTypesenseError,
+} = require('../../../services/TypesenseErrorService');
 const { toSearchResult } = require('../../../services/mapping/converters');
 
 module.exports = async (req, res) => {
@@ -10,11 +13,17 @@ module.exports = async (req, res) => {
   }
 
   // Search on multiple entities but with no pagging and sort
-  const results = await SearchService.multiCollectionsSearch({
-    query,
-    entities: req.param('entities'),
-    filter: req.param('filter') ?? {},
-  });
+  let results;
+  try {
+    results = await SearchService.multiCollectionsSearch({
+      query,
+      entities: req.param('entities'),
+      filter: req.param('filter') ?? {},
+    });
+  } catch (error) {
+    if (handleTypesenseError(res, error)) return;
+    throw error;
+  }
 
   ControllerService.treatAndConvert(
     req,
--- a/api/services/SearchService.js
+++ b/api/services/SearchService.js
@@ -110,51 +110,6 @@ async function multiCollectionsSearch({
   });
 }
 
-function validateSort(entity, sort) {
-  const parts = sort.split(':');
-  if (parts.length !== 2 || !parts[0] || !['asc', 'desc'].includes(parts[1])) {
-    const err = new Error(
-      `Invalid sort format '${sort}'. Expected 'field:asc' or 'field:desc'.`
-    );
-    err.code = 'E_SORT_VALIDATION';
-    throw err;
-  }
-
-  const [fieldName] = parts;
-  const { fields } = allEntities[entity].schema;
-  const fieldDef = fields.find((f) => f.name === fieldName);
-
-  if (!fieldDef) {
-    const err = new Error(
-      `Field '${fieldName}' does not exist in '${entity}' schema.`
-    );
-    err.code = 'E_SORT_VALIDATION';
-    throw err;
-  }
-
-  // Typesense allows sorting on numeric and boolean fields by default.
-  // Only string fields require explicit `sort: true` in the schema.
-  const numericAndBoolTypes = [
-    'int32',
-    'int64',
-    'float',
-    'bool',
-    'int32[]',
-    'int64[]',
-    'float[]',
-    'bool[]',
-  ];
-  const isSortableByDefault = numericAndBoolTypes.includes(fieldDef.type);
-
-  if (!isSortableByDefault && fieldDef.sort !== true) {
-    const err = new Error(
-      `Field '${fieldName}' is not sortable in '${entity}' schema.`
-    );
-    err.code = 'E_SORT_VALIDATION';
-    throw err;
-  }
-}
-
 async function collectionSearch({
   query,
   entity = [],
@@ -166,7 +121,6 @@ async function collectionSearch({
   fields,
 } = {}) {
   if (!allEntitiesKeys.includes(entity)) return null;
-  if (sort && sort.trim()) validateSort(entity, sort);
   const q = query || '*';
   const { filterBy, hasPrefixFilter } = buildFilter(
     filter,
--- a/api/services/TypesenseErrorService.js
+++ b/api/services/TypesenseErrorService.js
@@ -1,0 +1,26 @@
+/**
+ * Centralized Typesense error handling.
+ * Duck-types on error.httpStatus to classify 4xx client errors.
+ */
+
+/**
+ * Handle a Typesense client error by returning HTTP 400 if the error
+ * has an httpStatus in the 4xx range.
+ *
+ * @param {Object} res - Sails response object
+ * @param {Error} error - The caught error
+ * @returns {boolean} true if the error was handled (4xx), false otherwise
+ */
+function handleTypesenseError(res, error) {
+  const status = error?.httpStatus;
+  const isClientError =
+    typeof status === 'number' && status >= 400 && status < 500;
+  const isAuthError = status === 401 || status === 403;
+  if (isClientError && !isAuthError) {
+    res.badRequest({ error: error.message });
+    return true;
+  }
+  return false;
+}
+
+module.exports = { handleTypesenseError };
--- a/test/integration/1_services/SearchService.property.test.js
+++ b/test/integration/1_services/SearchService.property.test.js
@@ -1,3 +1,4 @@
+/* eslint-disable func-names */
 const should = require('should');
 const sinon = require('sinon');
 const fc = require('fast-check');
@@ -137,37 +138,32 @@ const bugConditionArb = fc.oneof(
 );
 
 /**
- * Property 1: Bug Condition — Invalid Sort Field Causes Unhandled Typesense Error
+ * Property 1: Invalid Sort — Typesense Rejects Invalid Sort Inputs
  *
  * Validates: Requirements 1.1, 1.2, 1.3
  *
- * For any collectionSearch() input where isBugCondition(input) is true,
- * the function should throw a validation error with a descriptive message
- * and typesense.search should NOT be called.
- *
- * EXPECTED: This test FAILS on unfixed code because the current implementation
- * passes invalid sort directly to Typesense without validation.
+ * For any collectionSearch() input with an invalid sort parameter,
+ * Typesense will reject the request. The sort is passed through to
+ * Typesense without local validation.
  */
-describe('SearchService - Property 1: Bug Condition', () => {
-  let searchStub;
-
-  beforeEach(() => {
-    searchStub = sinon.stub(typesense, 'search').resolves({ hits: [] });
-  });
-
+describe('SearchService - Property 1: Invalid Sort', () => {
   afterEach(() => {
     sinon.restore();
   });
 
-  it('Property 1: invalid sort inputs should throw a validation error and not call Typesense', function () {
+  it('Property 1: invalid sort inputs are passed to Typesense (no local validation)', function () {
     this.timeout(30000);
     return fc.assert(
       fc.asyncProperty(bugConditionArb, async (input) => {
-        // Reset stub call count for each generated case
-        searchStub.resetHistory();
+        sinon.restore();
+        // Stub typesense.search to reject with a 400 (simulating Typesense rejection)
+        const searchStub = sinon
+          .stub(typesense, 'search')
+          .rejects(
+            Object.assign(new Error('Bad sort field'), { httpStatus: 400 })
+          );
 
         let threw = false;
-        let errorMessage = '';
         try {
           await SearchService.collectionSearch({
             query: 'test',
@@ -176,23 +172,16 @@ describe('SearchService - Property 1: Bug Condition', () => {
           });
         } catch (err) {
           threw = true;
-          errorMessage = err.message || '';
         }
 
-        // Assert: should have thrown a validation error
+        // The call should reach Typesense (no local interception)
+        should(searchStub.called).be.true(
+          `Expected typesense.search to be called for entity="${input.entity}" sort="${input.sort}" (${input.category})`
+        );
+
+        // And the Typesense error should propagate
         should(threw).be.true(
-          `Expected collectionSearch to throw for entity="${input.entity}" sort="${input.sort}" (${input.category}), but it did not throw`
-        );
-
-        // Assert: error message should be descriptive
-        should(errorMessage.length).be.greaterThan(
-          0,
-          `Expected a descriptive error message for entity="${input.entity}" sort="${input.sort}" (${input.category})`
-        );
-
-        // Assert: typesense.search should NOT have been called
-        should(searchStub.called).be.false(
-          `Expected typesense.search to NOT be called for entity="${input.entity}" sort="${input.sort}" (${input.category}), but it was called`
+          `Expected collectionSearch to throw for entity="${input.entity}" sort="${input.sort}" (${input.category})`
         );
       }),
       { numRuns: 100 }
--- a/test/integration/4_routes/Search/typesense-error-handling.property.test.js
+++ b/test/integration/4_routes/Search/typesense-error-handling.property.test.js
@@ -1,0 +1,206 @@
+/* eslint-disable func-names, global-require */
+const supertest = require('supertest');
+const sinon = require('sinon');
+const should = require('should');
+const fc = require('fast-check');
+
+/**
+ * Bug Condition Exploration — Property 1: Typesense 4xx errors return HTTP 400
+ *
+ * Validates: Requirements 1.1, 1.2, 1.3, 1.4
+ *
+ * For any Typesense error with httpStatus in [400, 499] (excluding auth errors
+ * 401/403 which are server config issues), the API should return HTTP 400 with
+ * the error message.
+ */
+describe('Bug Condition — Typesense 4xx errors on search endpoints', () => {
+  let SearchService;
+
+  before(() => {
+    SearchService = require('../../../../api/services/SearchService');
+  });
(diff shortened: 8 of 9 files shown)


ClemRz:

## Introduction

All search endpoints (`POST /api/v1/search`, `POST /api/v1/advanced-search`, `POST /api/v1/advanced-search/export`, `POST /api/v1/field-search`) return HTTP 500 "An internal server error occurred." when Typesense rejects a request due to a client error (HTTP 4xx). The Typesense JS client throws typed errors (`RequestMalformed`, `ObjectNotFound`, `ObjectUnprocessable`, `RequestUnauthorized`, etc.) that carry an `httpStatus` property, but the controllers either have no try/catch at all or only catch `E_SORT_VALIDATION`, causing Sails to convert any Typesense client error into a generic 500 response. Any Typesense error where the root cause is a bad client request should be forwarded as a 400 to the caller, not masked as a 500.

## Bug Analysis

### Current Behavior (Defect)

1.1 WHEN a client sends a search request that causes Typesense to return a 4xx error (e.g., invalid filter field name, malformed filter expression, bad query_by field, non-existent collection) to the `POST /api/v1/advanced-search` endpoint THEN the system returns HTTP 500 "An internal server error occurred." because the catch block only handles `E_SORT_VALIDATION` and re-throws everything else

1.2 WHEN a client sends a search request that causes Typesense to return a 4xx error to the `POST /api/v1/search` (quick-search) endpoint THEN the system returns HTTP 500 because the controller has no try/catch around the Typesense call

1.3 WHEN a client sends a search request that causes Typesense to return a 4xx error to the `POST /api/v1/field-search` endpoint THEN the system returns HTTP 500 because the controller has no try/catch around the Typesense call

1.4 WHEN a client sends a search request that causes Typesense to return a 4xx error to the `POST /api/v1/advanced-search/export` endpoint THEN the system returns HTTP 500 because the catch block only handles `E_SORT_VALIDATION` and assigns the error to `results` for other errors, which then fails the `!results.hits` check and returns a generic server error

2.1 WHEN a client sends a search request that causes Typesense to return a 4xx error to the `POST /api/v1/advanced-search` endpoint THEN the system SHALL return HTTP 400 with the Typesense error message (e.g., "Could not find a filter field named `isTouristicz` in the schema.")

2.2 WHEN a client sends a search request that causes Typesense to return a 4xx error to the `POST /api/v1/search` (quick-search) endpoint THEN the system SHALL return HTTP 400 with the Typesense error message

2.3 WHEN a client sends a search request that causes Typesense to return a 4xx error to the `POST /api/v1/field-search` endpoint THEN the system SHALL return HTTP 400 with the Typesense error message

2.4 WHEN a client sends a search request that causes Typesense to return a 4xx error to the `POST /api/v1/advanced-search/export` endpoint THEN the system SHALL return HTTP 400 with the Typesense error message (before any CSV headers have been sent)

3.1 WHEN a client sends a valid search request with correct filter fields and query parameters THEN the system SHALL CONTINUE TO return HTTP 200 with the search results on all search endpoints

3.2 WHEN a client sends a search request with an invalid sort field THEN the system SHALL CONTINUE TO return HTTP 400 via the existing `E_SORT_VALIDATION` handling on `advanced-search` and `advanced-search/export`

3.3 WHEN Typesense itself experiences a genuine server error (HTTP 5xx) THEN the system SHALL CONTINUE TO return HTTP 500 to the client on all search endpoints

3.4 WHEN a client sends a search request for a non-existent entity (not in `allEntitiesKeys`) THEN the system SHALL CONTINUE TO return the current response (null result handling) on all search endpoints

3.5 WHEN a client sends a quick-search request without a query parameter THEN the system SHALL CONTINUE TO return HTTP 400 "You must provide a query"

3.6 WHEN a client sends a field-search request without a field or entity parameter THEN the system SHALL CONTINUE TO return HTTP 400 with the existing validation message

ClemRz:
# Typesense Error Handling Bugfix Design

## Overview

All four search endpoints (`quick-search`, `advanced-search`, `advanced-search/export`, `field-search`) return HTTP 500 when Typesense rejects a request with a 4xx client error. The Typesense JS client throws typed `TypesenseError` subclasses (`RequestMalformed`, `ObjectNotFound`, `ObjectUnprocessable`, etc.) that carry an `httpStatus` property and a descriptive `message`. The fix introduces a centralized helper that inspects `error.httpStatus` and forwards 4xx errors as HTTP 400 to the caller, while letting 5xx errors continue to produce HTTP 500. Each controller will use this helper in its catch block.

## Glossary

- **Bug_Condition (C)**: A Typesense client error with `httpStatus` in the 400–499 range is thrown during a search call
- **Property (P)**: The API returns HTTP 400 with the Typesense error message to the caller
- **Preservation**: Existing behaviors (valid searches returning 200, `E_SORT_VALIDATION` returning 400, Typesense 5xx producing 500, input validation returning 400) remain unchanged
- **TypesenseError**: Base error class from `typesense` JS client; has `message`, `httpStatus`, and `httpBody` properties
- **RequestMalformed**: TypesenseError subclass thrown for HTTP 400 (e.g., invalid filter field)
- **ObjectNotFound**: TypesenseError subclass thrown for HTTP 404 (e.g., non-existent collection alias)
- **ObjectUnprocessable**: TypesenseError subclass thrown for HTTP 422 (e.g., invalid query_by field)
- **E_SORT_VALIDATION**: Application-level error code thrown by `SearchService.validateSort()` for invalid sort parameters

## Bug Details

### Bug Condition

The bug manifests when a client sends a search request with parameters that Typesense rejects (invalid filter field, malformed filter expression, bad `query_by` field, non-existent collection). The Typesense JS client throws a `TypesenseError` subclass with an `httpStatus` in the 400–499 range, but the controllers either have no try/catch (`quick-search`, `field-search`) or only catch `E_SORT_VALIDATION` (`advanced-search`, `advanced-search/export`), causing Sails to convert the unhandled error into a generic HTTP 500.

**Formal Specification:**
```
FUNCTION isBugCondition(error)
  INPUT: error of type Error (thrown during a search call)
  OUTPUT: boolean

  RETURN error IS instanceof TypesenseError
         AND error.httpStatus >= 400
         AND error.httpStatus < 500
END FUNCTION
```

### Examples

- `POST /api/v1/advanced-search` with `filter: { isTouristicz: true }` (misspelled field) → Typesense throws `RequestMalformed` (400) with message "Could not find a filter field named `isTouristicz` in the schema." → API currently returns 500, should return 400 with that message
- `POST /api/v1/search` with `entities: ['nonexistent']` where the alias doesn't exist → Typesense throws `ObjectNotFound` (404) → API currently returns 500, should return 400
- `POST /api/v1/field-search` with `field: 'nonexistent_field'` → Typesense throws `ObjectUnprocessable` (422) with message about invalid `query_by` → API currently returns 500, should return 400
- `POST /api/v1/advanced-search/export` with `filter: { depth:=[abc] }` (malformed range) → Typesense throws `RequestMalformed` (400) → API currently returns 500, should return 400

## Expected Behavior

### Preservation Requirements

**Unchanged Behaviors:**
- Valid search requests continue to return HTTP 200 with results on all endpoints
- `E_SORT_VALIDATION` errors in `advanced-search` and `advanced-search/export` continue to return HTTP 400 via the existing code path
- Typesense 5xx server errors continue to produce HTTP 500 on all endpoints
- Requests for non-existent entities (not in `allEntitiesKeys`) continue to return null/empty results
- `quick-search` without a `query` parameter continues to return HTTP 400 "You must provide a query"
- `field-search` without `field` or `entity` continues to return HTTP 400 with existing validation messages
- The `advanced-search/export` CSV streaming behavior (headers, BOM, batching) remains unchanged

**Scope:**
All inputs that do NOT cause a Typesense 4xx error should be completely unaffected by this fix. This includes:
- Valid search queries that return results
- Requests filtered out before reaching Typesense (input validation, unknown entity)
- Typesense 5xx errors (server-side failures)
- `E_SORT_VALIDATION` errors (caught before the Typesense call)

## Hypothesized Root Cause

Based on code analysis, the root causes are confirmed (not hypothesized):

1. **`quick-search.js` — No try/catch**: The `multiCollectionsSearch()` call has no error handling at all. Any Typesense error propagates as an unhandled rejection, which Sails converts to HTTP 500.

2. **`field-search.js` — No try/catch**: The `fieldSearch()` call has no error handling. Same issue as quick-search.

3. **`advanced-search.js` — Incomplete catch**: The catch block only handles `E_SORT_VALIDATION` and re-throws everything else with `throw error`, which Sails converts to HTTP 500.

4. **`advanced-search-export.js` — Swallowed error**: The catch block handles `E_SORT_VALIDATION`, but for other errors it assigns the error object to `results`. The subsequent `!results.hits` check then triggers `res.serverError()`, masking the original Typesense error message.

## Correctness Properties

Property 1: Bug Condition — Typesense 4xx errors return HTTP 400

_For any_ search request to any of the four endpoints where Typesense throws a `TypesenseError` with `httpStatus` in [400, 499], the API SHALL return HTTP 400 with a response body containing the Typesense error message.

**Validates: Requirements 2.1, 2.2, 2.3, 2.4**

Property 2: Preservation — Non-4xx errors and valid requests unchanged

_For any_ search request where Typesense does NOT throw a 4xx error (valid requests returning results, Typesense 5xx errors, `E_SORT_VALIDATION` errors, input validation errors), the API SHALL produce the same HTTP status and response body as the original unfixed code.

**Validates: Requirements 3.1, 3.2, 3.3, 3.4, 3.5, 3.6**

## Fix Implementation

### Changes Required

**New file**: `api/services/TypesenseErrorService.js`

A centralized helper that all four controllers will use:

**Function**: `handleTypesenseError(res, error)`

**Logic**:
1. Check if `error` has a numeric `httpStatus` property in the 400–499 range
2. If yes: call `res.badRequest()` with the Typesense error message and return `true`
3. If no: return `false` (caller should re-throw or handle otherwise)

This avoids importing the `TypesenseError` class directly — duck-typing on `httpStatus` is sufficient and more resilient to library version changes.

**File**: `api/controllers/v1/search/advanced-search.js`

**Change**: In the existing catch block, add a check for Typesense 4xx errors before the `throw error` fallback:
```
catch (error) {
  if (error.code === 'E_SORT_VALIDATION') return res.badRequest(...)
  if (handleTypesenseError(res, error)) return;
  throw error;
}
```

**File**: `api/controllers/v1/search/quick-search.js`

**Change**: Wrap the `multiCollectionsSearch()` call in a try/catch that uses the helper:
```
try {
  results = await SearchService.multiCollectionsSearch(...)
} catch (error) {
  if (handleTypesenseError(res, error)) return;
  throw error;
}
```

**File**: `api/controllers/v1/search/field-search.js`

**Change**: Wrap the `fieldSearch()` call in a try/catch that uses the helper:
```
try {
  r = await SearchService.fieldSearch(...)
} catch (error) {
  if (handleTypesenseError(res, error)) return;
  throw error;
}
```

**File**: `api/controllers/v1/search/advanced-search-export.js`

**Change**: In the existing catch block, add the Typesense 4xx check before the `results = err` fallback. Only return 400 if CSV headers have not been sent yet:
```
catch (err) {
  if (err.code === 'E_SORT_VALIDATION') { ... existing handling ... }
  if (!hasSentHeader && handleTypesenseError(res, err)) return;
  results = err;
}
```

## Testing Strategy

### Validation Approach

**Goal**: Surface counterexamples that demonstrate the bug BEFORE implementing the fix. Confirm the root cause analysis by observing that Typesense 4xx errors produce HTTP 500.

**Test Plan**: Stub `typesense.search()` and `typesense.multiSearch()` to throw `TypesenseError` instances with `httpStatus: 400` and a descriptive message. Call each endpoint via supertest and assert the response status.

**Test Cases**:
1. **advanced-search with invalid filter**: Stub throws `RequestMalformed(400)` → expect 500 on unfixed code
2. **quick-search with bad entity alias**: Stub throws `ObjectNotFound(404)` → expect 500 on unfixed code
3. **field-search with invalid query_by**: Stub throws `ObjectUnprocessable(422)` → expect 500 on unfixed code
4. **advanced-search/export with malformed filter**: Stub throws `RequestMalformed(400)` → expect 500 on unfixed code

**Expected Counterexamples**:
- All four endpoints return HTTP 500 instead of 400 when Typesense throws a 4xx error
- The Typesense error message is lost, replaced by "An internal server error occurred."

### Fix Checking

**Goal**: Verify that for all inputs where the bug condition holds, the fixed function produces the expected behavior.

**Pseudocode:**
```
FOR ALL error WHERE isBugCondition(error) DO
  result := callEndpoint(requestThatTriggersError)
  ASSERT result.status = 400
  ASSERT result.body CONTAINS error.message
END FOR
```

### Preservation Checking

**Goal**: Verify that for all inputs where the bug condition does NOT hold, the fixed function produces the same result as the original function.

**Pseudocode:**
```
FOR ALL request WHERE NOT isBugCondition(errorThrownBy(request)) DO
  ASSERT fixedEndpoint(request).status = originalEndpoint(request).status
  ASSERT fixedEndpoint(request).body = originalEndpoint(request).body
END FOR
```

**Testing Approach**: Property-based testing is recommended for preservation checking because:
- It generates many combinations of valid/invalid inputs to verify behavior is unchanged
- It catches edge cases around the boundary between 4xx and 5xx status codes
- It provides strong guarantees that `E_SORT_VALIDATION` and input validation paths are unaffected

**Test Plan**: Observe behavior on UNFIXED code first for valid searches, sort validation errors, and 5xx errors, then write property-based tests capturing that behavior.

**Test Cases**:
1. **Valid search preservation**: Verify valid requests continue to return 200 with results
2. **E_SORT_VALIDATION preservation**: Verify invalid sort fields continue to return 400 via existing path
3. **5xx error preservation**: Verify Typesense `ServerError(500)` continues to produce HTTP 500
4. **Input validation preservation**: Verify missing query/field/entity continues to return 400

### Unit Tests

- Test `TypesenseErrorService.handleTypesenseError()` with various error shapes: TypesenseError with httpStatus 400, 404, 422, 499; ServerError with httpStatus 500, 503; plain Error without httpStatus; error with non-numeric httpStatus
- Test that the helper returns `true` for 4xx and `false` for everything else
- Test that the helper calls `res.badRequest()` with the error message for 4xx

### Property-Based Tests

- Generate random `httpStatus` values in [400, 599] and verify the helper correctly classifies 4xx vs 5xx
- Generate random error objects (with/without httpStatus, with various message strings) and verify the helper's behavior is consistent
- Generate random search parameters and verify that the endpoint behavior matches expectations based on whether the stubbed Typesense call throws a 4xx, 5xx, or succeeds

### Integration Tests

- Test each of the four endpoints with a stubbed Typesense 4xx error → expect HTTP 400 with message
- Test each endpoint with a stubbed Typesense 5xx error → expect HTTP 500
- Test each endpoint with a successful Typesense response → expect HTTP 200
- Test `advanced-search` and `advanced-search/export` with `E_SORT_VALIDATION` → expect HTTP 400 (preservation)
- Test `quick-search` without query → expect HTTP 400 "You must provide a query" (preservation)
- Test `field-search` without field/entity → expect HTTP 400 with validation message (preservation)

Paul-AUB (changes requested):
# Code Review — `b18354ce` fix(search): forward Typesense 4xx errors as HTTP 400

**Author:** Clément Ronzon
**Date:** 2026-04-03
**Files changed:** 9 (+472 / -99)

---

## Summary

This commit replaces the custom `validateSort()` pre-validation logic in `SearchService` with a generic Typesense 4xx forwarding mechanism (`TypesenseErrorService`). All four search controllers are updated to catch Typesense errors and return HTTP 400 to the client when the error status is in the 4xx range.

---

## What's Good

- **Centralization (DRY):** Extracting error handling into `TypesenseErrorService.handleTypesenseError()` eliminates duplicated `if (error.code === 'E_SORT_VALIDATION')` checks across controllers.
- **Broader coverage:** The new approach catches any 4xx from Typesense, not just sort validation errors — so filter syntax errors, unknown fields, etc. are now properly forwarded as 400 instead of crashing as 500.
- **Property-based testing:** Using `fast-check` to generate arbitrary 4xx status codes and messages is a solid choice for this kind of boundary contract.
- **Preservation tests:** `typesense-preservation.property.test.js` explicitly verifies that 5xx errors still propagate as 500 and that valid inputs still return 200 — good regression safety net.

---

## Issues and Concerns

### 1. Potential regression in `advanced-search-export.js` — mid-stream 4xx not `break`-ing

**Severity: High**

The old code had an explicit `break` for validation errors when the response header had already been sent:

```js
// Before
if (err.code === 'E_SORT_VALIDATION') {
  if (!hasSentHeader) {
    res.badRequest({ error: err.message });
    return;
  }
  break; // <-- stops the streaming loop cleanly
}
```

The new code removes the `break`:

```js
// After
if (!hasSentHeader && handleTypesenseError(res, err)) return;
results = err;
```

When `hasSentHeader` is `true` and a 4xx Typesense error occurs mid-stream, the condition short-circuits and the code falls through to `results = err`. Depending on how the rest of the export loop uses `results`, this could lead to silent corruption of the streamed CSV, the loop continuing with an `Error` object as `results`, or an unhandled exception downstream.

**Fix:** add the `break` back for the mid-stream case:

```js
} catch (err) {
  if (!hasSentHeader && handleTypesenseError(res, err)) return;
  if (hasSentHeader) break; // terminate stream cleanly on any error
  results = err;
}
```

---

### 2. `handleTypesenseError` returns `res.badRequest(...)` instead of `true`

**Severity: Low**

```js
function handleTypesenseError(res, error) {
  const status = error?.httpStatus;
  if (typeof status === 'number' && status >= 400 && status < 500) {
    return res.badRequest({ error: error.message }); // returns a Sails response object
  }
  return false;
}
```

The callers rely on the truthiness of `res.badRequest()`. This works in Sails.js because the response object is truthy, but it's an implicit contract on a framework internals detail. If the response object ever returned `null` or `undefined` in some Sails version or error path, the guard would silently fail.

**Fix:** explicitly return `true` after sending the response:

```js
if (typeof status === 'number' && status >= 400 && status < 500) {
  res.badRequest({ error: error.message });
  return true;
}
```

---

### 3. Auth-related 4xx errors (401, 403) forwarded to clients as HTTP 400

**Severity: Medium**

Typesense returns `401 Unauthorized` and `403 Forbidden` when the API key is missing or invalid. These are server configuration issues — not the client's fault. The current handler blindly maps all `[400, 499]` errors to `res.badRequest()`, so a misconfigured Typesense API key would result in a `400` response body with the raw Typesense auth error message, leaking internal infrastructure details to end users.

**Fix:** either narrow the range, or explicitly exclude 401/403:

```js
const isClientError = typeof status === 'number' && status >= 400 && status < 500;
const isAuthError = status === 401 || status === 403;
if (isClientError && !isAuthError) {
  res.badRequest({ error: error.message });
  return true;
}
```

---

### 4. Loss of pre-validation: network round-trip for every invalid sort

**Severity: Low / Design trade-off**

The old `validateSort()` rejected bad sort parameters before hitting Typesense, at the cost of maintaining schema-level field knowledge in the application layer. The new approach always makes the network call.

For the application's load profile this is likely acceptable, but it should be a conscious choice: every invalid sort input now costs a Typesense round-trip. If sort validation is hot (e.g., called from an autocomplete loop), this could matter.

---

### 5. `require()` inside `it()` blocks — stale module cache risk

**Severity: Low**

Both new test files use `require()` inside `it()` or `asyncProperty` callbacks:

```js
it('...', function () {
  const SearchService = require('../../../../api/services/SearchService');
  sinon.stub(SearchService, 'collectionSearch').rejects(error);
  ...
});
```

Node's module cache means all calls resolve to the same object, so the stubs are applied correctly — but this pattern is easy to misread and can cause hard-to-debug issues if the module is ever reset between tests (e.g., via `proxyquire` or test isolation tools). Moving `require` to the top of each `describe` block is idiomatic and safer.

---

### 6. Double `sinon.restore()` in property test

**Severity: Very Low**

In `SearchService.property.test.js`, `sinon.restore()` is called both inside the `asyncProperty` callback and in `afterEach`. This is benign (Sinon tolerates double-restore), but it makes the test harder to reason about and masks whether the restore is happening at the right time.

---

### 7. Missing test: mid-stream 4xx in export endpoint

**Severity: Medium**

The new `typesense-error-handling.property.test.js` tests the export endpoint for 4xx errors, but only in the case where no headers have been sent yet (first batch). There is no test covering the case where headers are already sent (subsequent batch) and a 4xx error occurs — which is exactly the regression identified in issue #1 above.

---

## Summary Table

| # | Location | Severity | Topic |
|---|----------|----------|-------|
| 1 | `advanced-search-export.js` | **High** | Mid-stream `break` removed — potential loop/corruption regression |
| 2 | `TypesenseErrorService.js` | Low | Implicit truthiness on `res.badRequest()` return value |
| 3 | `TypesenseErrorService.js` | **Medium** | 401/403 Typesense errors forwarded to clients as 400 |
| 4 | `SearchService.js` | Low | Network round-trip added for every invalid sort (design trade-off) |
| 5 | Both new test files | Low | `require()` inside `it()` — stale cache risk |
| 6 | `SearchService.property.test.js` | Very Low | Double `sinon.restore()` |
| 7 | Tests | **Medium** | No test for mid-stream error in export |

ClemRz:

**#1 (High) — Mid-stream `break` removed in export:** Fixed. Added `if (hasSentHeader) break;` after the `handleTypesenseError` check so the streaming loop terminates cleanly when a 4xx error occurs after CSV headers have already been sent.

**#2 (Low) — Implicit truthiness on `res.badRequest()` return:** Fixed. `handleTypesenseError` now explicitly returns `true` after calling `res.badRequest()` instead of relying on the Sails response object's truthiness.

**#3 (Medium) — Auth-related 4xx errors (401, 403) forwarded as 400:** Fixed. `handleTypesenseError` now excludes 401 and 403 from the 4xx-to-400 mapping — these are server configuration issues and should continue to surface as 500. Added a preservation test for this.

**#4 (Low) — Loss of pre-validation:** Acknowledged as a conscious design trade-off, no change.

**#5 (Low) — `require()` inside `it()` blocks:** Moved the `require()` into a `before()` hook at the `describe` level. This runs after Sails has lifted (so the module cache is populated correctly) while keeping the import out of individual test callbacks. The `global-require` eslint disable is retained since the deferred require is intentional.

**#6 (Very Low) — Double `sinon.restore()`:** The inner `sinon.restore()` inside `asyncProperty` callbacks is necessary (it runs between property iterations to reset stubs), while the `afterEach` acts as a safety net for the overall test block. Kept both but added a comment clarifying the intent.

**#7 (Medium) — Missing test for mid-stream 4xx in export:** Added a new test case that stubs `collectionSearch` to succeed on the first call (sending CSV headers) and reject with a 4xx on the second call, then verifies the stream terminates cleanly with status 200 and valid CSV content.

Paul-AUB (approved):

PR #1527, opened 2026-04-03
fix(entrance,cave): add coerceToInt to round decimal values for integer DB columns

ClemRz:
## 🤔 What

- Add a `coerceToInt` utility that rounds decimal numeric values to the nearest integer before they reach PostgreSQL integer columns (`int2`/`int4`)
- Apply `coerceToInt` across all entry points that write to integer-typed DB columns:
  - `EntranceService.getConvertedDataFromClientRequest()` — `altitude`, `precision`, `yearDiscovery`
  - `CaveService.getConvertedDataFromClient()` — `depth`, `caveLength`
  - `cave/update.js` controller — `depth`, `length`
  - `EntranceCSVImportService.getConvertedEntranceFromCsv()` — `altitude`, `precision`
  - `EntranceCSVImportService.getConvertedCaveFromCsv()` — `depth`, `length`

## 🤷‍♂️ Why

When creating an entrance with a decimal altitude (e.g., `1132.6`), PostgreSQL rejects the value with `invalid input syntax for type smallint: "1132.6"`, causing a 500 Server Error. The Waterline ORM declares these fields as `type: 'number'` which accepts any JavaScript number, but the underlying DB columns are `int2`/`int4` and reject non-integer values.

## 🔍 How

A shared utility `api/utils/coerceToInt.js` handles the coercion:
- `null`/`undefined` → pass through unchanged
- Finite numbers → `Math.round(value)`
- Everything else (NaN, Infinity, strings) → pass through unchanged, letting Waterline's own validation handle rejection

Applied at the data conversion layer (closest to the input boundary) in all 5 entry points, keeping the change minimal and localized.

## 🧪 Testing

- Property-based tests with fast-check validating the bug condition fix and preservation of existing behavior
- Unit tests for `coerceToInt` covering integers, decimals, null, undefined, NaN, Infinity, strings
- Unit tests for `EntranceService.getConvertedDataFromClientRequest` and `CaveService.getConvertedDataFromClient` with decimal inputs
- Full test suite passes (1878 passing, 6 pre-existing failures unrelated to this change)

### Manual testing

Create an entrance with a decimal altitude (previously returned 500):
```bash
curl -X POST http://localhost:1337/api/v1/entrances \
  -H "Authorization: Bearer <TOKEN>" \
  -d '{
    "name": {"language": "fra", "text": "Test entrance"},
    "cave": 75070,
    "isSensitive": false,
    "longitude": "6.498327",
    "latitude": "46.926410",
    "altitude": 1132.6,
    "yearDiscovery": 1973
  }'
```

Update a cave with decimal depth and length:
```bash
curl -X PUT http://localhost:1337/api/v1/caves/5 \
  -H "Authorization: Bearer <TOKEN>" \
  -d '{
    "name": {"language": "fra", "text": "Test cave"},
    "depth": 250.4,
    "length": 1500.7
  }'
```

Verify integer values still work as before:
```bash
curl -X POST http://localhost:1337/api/v1/entrances \
  -H "Authorization: Bearer <TOKEN>" \
  -d '{
    "name": {"language": "fra", "text": "Test entrance"},
    "cave": 75070,
    "isSensitive": false,
    "longitude": "6.498327",
    "latitude": "46.926410",
    "altitude": 1132,
    "yearDiscovery": 1973
  }'
```

## 📸 Previews

N/A — backend-only change, no UI impact.

### DIFF
--- a/api/controllers/v1/cave/update.js
+++ b/api/controllers/v1/cave/update.js
@@ -1,4 +1,5 @@
 const ControllerService = require('../../../services/ControllerService');
+const coerceToInt = require('../../../utils/coerceToInt');
 const CaveService = require('../../../services/CaveService');
 const NotificationService = require('../../../services/NotificationService');
 const { toCave } = require('../../../services/mapping/converters');
@@ -27,8 +28,8 @@ module.exports = async (req, res) => {
 
   if (newLatitude != null) updatedFields.latitude = newLatitude;
   if (newLongitude != null) updatedFields.longitude = newLongitude;
-  if (newDepth != null) updatedFields.depth = newDepth;
-  if (newLength != null) updatedFields.caveLength = newLength;
+  if (newDepth != null) updatedFields.depth = coerceToInt(newDepth);
+  if (newLength != null) updatedFields.caveLength = coerceToInt(newLength);
   if (newTemperature != null) updatedFields.temperature = newTemperature;
   if (newIsDiving != null) updatedFields.isDiving = newIsDiving;
 
--- a/api/services/CaveService.js
+++ b/api/services/CaveService.js
@@ -1,4 +1,5 @@
 const CommonService = require('./CommonService');
+const coerceToInt = require('../utils/coerceToInt');
 const DocumentService = require('./DocumentService');
 const DescriptionService = require('./DescriptionService');
 const NameService = require('./NameService');
@@ -95,12 +96,12 @@ module.exports = {
   // Extract everything from the request body except id and dateInscription
   getConvertedDataFromClient: (req) => ({
     // The TCave.create() function doesn't work with TCave field alias. See TCave.js Model
-    depth: req.param('depth'),
+    depth: coerceToInt(req.param('depth')),
     documents: req.param('documents'),
     isDiving: req.param('isDiving'),
     latitude: req.param('latitude'),
     longitude: req.param('longitude'),
-    caveLength: req.param('length'),
+    caveLength: coerceToInt(req.param('length')),
     massif: req.param('massif'),
     temperature: req.param('temperature'),
   }),
--- a/api/services/EntranceCSVImportService.js
+++ b/api/services/EntranceCSVImportService.js
@@ -1,4 +1,5 @@
 const { getDateFromKarstlink } = require('../../config/constants/karstlink');
+const coerceToInt = require('../utils/coerceToInt');
 
 const doubleCheck = sails.helpers.csvhelpers.doubleCheck.with;
 const { getCreator } = require('../utils/csvHelper');
@@ -65,10 +66,12 @@ module.exports = {
       longitude: doubleCheckWithData({
         key: 'w3geo:longitude',
       }),
-      length: doubleCheckWithData({
-        key: 'karstlink:length',
-      }),
-      depth,
+      length: coerceToInt(
+        doubleCheckWithData({
+          key: 'karstlink:length',
+        })
+      ),
+      depth: coerceToInt(depth),
       dateInscription: doubleCheckWithData({
         key: 'dct:rights/dct:created',
         defaultValue: new Date(),
@@ -188,12 +191,16 @@ module.exports = {
       country: doubleCheckWithData({
         key: 'gn:countryCode',
       }),
-      precision: doubleCheckWithData({
-        key: 'dwc:coordinatePrecision',
-      }),
-      altitude: doubleCheckWithData({
-        key: 'w3geo:altitude',
-      }),
+      precision: coerceToInt(
+        doubleCheckWithData({
+          key: 'dwc:coordinatePrecision',
+        })
+      ),
+      altitude: coerceToInt(
+        doubleCheckWithData({
+          key: 'w3geo:altitude',
+        })
+      ),
       latitude: cave.latitude,
       longitude: cave.longitude,
       cave: cave.id,
--- a/api/services/EntranceService.js
+++ b/api/services/EntranceService.js
@@ -22,6 +22,7 @@ const DescriptionService = require('./DescriptionService');
 const HistoryService = require('./HistoryService');
 const LocationService = require('./LocationService');
 const RightService = require('./RightService');
+const coerceToInt = require('../utils/coerceToInt');
 
 function coerceBool(req, field) {
   const value = req.param(field);
@@ -49,6 +50,9 @@ module.exports = {
     const { id, ...reqBodyWithoutId } = req.body;
     return {
       ...reqBodyWithoutId,
+      altitude: coerceToInt(reqBodyWithoutId.altitude),
+      precision: coerceToInt(reqBodyWithoutId.precision),
+      yearDiscovery: coerceToInt(reqBodyWithoutId.yearDiscovery),
       geology: req.body.geology ?? 'Q35758',
       isSensitive: coerceBool(req, 'isSensitive'),
       hasBat: coerceBool(req, 'hasBat'),
--- a/api/utils/coerceToInt.js
+++ b/api/utils/coerceToInt.js
@@ -1,0 +1,17 @@
+/**
+ * Coerce a value to an integer via Math.round().
+ * Handles both numbers and numeric strings (e.g. from CSV parsing).
+ * Passes through null, undefined, non-finite, and non-numeric values unchanged
+ * (letting Waterline handle validation for invalid types).
+ *
+ * @param {*} value
+ * @returns {*} rounded integer for finite numbers/numeric strings, original value otherwise
+ */
+const coerceToInt = (value) => {
+  if (value === null || value === undefined) return value;
+  const n = typeof value === 'string' ? parseFloat(value) : value;
+  if (typeof n === 'number' && Number.isFinite(n)) return Math.round(n);
+  return value;
+};
+
+module.exports = coerceToInt;
--- a/test/integration/2_utils/coerceToInt-preservation.property.test.js
+++ b/test/integration/2_utils/coerceToInt-preservation.property.test.js
@@ -1,0 +1,48 @@
+const should = require('should');
+const fc = require('fast-check');
+
+/**
+ * Property 2: Preservation — Integer and Nullish Values Pass Through Unchanged
+ *
+ * Verifies that coerceToInt acts as an identity function for values that are
+ * already integers, null, undefined, or non-finite (NaN, Infinity).
+ */
+
+const coerceToInt = require('../../../api/utils/coerceToInt');
+
+describe('coerceToInt - Property: Preservation', () => {
+  describe('Property 2a: Integer values pass through unchanged', () => {
+    it('should return the same value for any integer', function () {
+      this.timeout(10000);
+      fc.assert(
+        fc.property(fc.integer(), (n) => {
+          should(coerceToInt(n)).equal(n);
+        }),
+        { numRuns: 100 }
+      );
+    });
+  });
+
+  describe('Property 2b: Nullish values pass through unchanged', () => {
+    it('should return null for null and undefined for undefined', function () {
+      this.timeout(10000);
+      fc.assert(
+        fc.property(
+          fc.oneof(fc.constant(null), fc.constant(undefined)),
+          (v) => {
+            should(coerceToInt(v)).equal(v);
+          }
+        ),
+        { numRuns: 100 }
+      );
+    });
+  });
+
+  describe('Property 2c: Non-finite values pass through unchanged', () => {
+    it('should return NaN for NaN and Infinity for Infinity', () => {
+      should(Number.isNaN(coerceToInt(NaN))).be.true();
+      should(coerceToInt(Infinity)).equal(Infinity);
+      should(coerceToInt(-Infinity)).equal(-Infinity);
+    });
+  });
+});
--- a/test/integration/2_utils/coerceToInt.property.test.js
+++ b/test/integration/2_utils/coerceToInt.property.test.js
@@ -1,0 +1,60 @@
+const should = require('should');
+const fc = require('fast-check');
+
+/**
+ * Property 1: Rounding Behavior — Decimal Values Are Rounded to Nearest Integer
+ *
+ * For all finite double values with a fractional part, coerceToInt(v) should
+ * equal Math.round(v) and the result should be an integer.
+ */
+
+const coerceToInt = require('../../../api/utils/coerceToInt');
+
+describe('coerceToInt - Property: Rounding Behavior', () => {
+  describe('Property 1: Decimal values are rounded to nearest integer', () => {
+    it('should round all finite decimal values to the nearest integer', function () {
+      this.timeout(10000);
+      fc.assert(
+        fc.property(
+          fc
+            .double({ noNaN: true, noDefaultInfinity: true })
+            .filter((v) => v !== Math.round(v)),
+          (v) => {
+            const result = coerceToInt(v);
+            should(result).equal(
+              Math.round(v),
+              `coerceToInt(${v}) should be ${Math.round(v)} but got ${result}`
+            );
+            should(Number.isInteger(result)).be.true(
+              `coerceToInt(${v}) should be an integer but got ${result}`
+            );
+          }
+        ),
+        { numRuns: 100 }
+      );
+    });
+
+    it('should round numeric strings with decimals to the nearest integer', function () {
+      this.timeout(10000);
+      fc.assert(
+        fc.property(
+          fc
+            .double({ noNaN: true, noDefaultInfinity: true })
+            .filter((v) => v !== Math.round(v))
+            .map((v) => String(v)),
+          (s) => {
+            const result = coerceToInt(s);
+            should(result).equal(
+              Math.round(parseFloat(s)),
+              `coerceToInt('${s}') should be ${Math.round(parseFloat(s))} but got ${result}`
+            );
+            should(Number.isInteger(result)).be.true(
+              `coerceToInt('${s}') should be an integer but got ${result}`
+            );
+          }
+        ),
+        { numRuns: 100 }
+      );
+    });
+  });
+});
--- a/test/integration/2_utils/coerceToInt.test.js
+++ b/test/integration/2_utils/coerceToInt.test.js
@@ -1,0 +1,122 @@
+const should = require('should');
+const coerceToInt = require('../../../api/utils/coerceToInt');
+const EntranceService = require('../../../api/services/EntranceService');
+const CaveService = require('../../../api/services/CaveService');
+
+describe('coerceToInt - Unit Tests', () => {
+  describe('coerceToInt()', () => {
+    it('should pass through integers unchanged', () => {
+      should(coerceToInt(0)).equal(0);
+      should(coerceToInt(1132)).equal(1132);
+      should(coerceToInt(-500)).equal(-500);
+    });
+
+    it('should round decimals to nearest integer', () => {
+      should(coerceToInt(1132.6)).equal(1133);
+      should(coerceToInt(1132.4)).equal(1132);
+      should(coerceToInt(1132.5)).equal(1133);
+      should(coerceToInt(-1132.6)).equal(-1133);
+      should(coerceToInt(0.1)).equal(0);
+      should(coerceToInt(0.9)).equal(1);
+    });
+
+    it('should pass through null unchanged', () => {
+      should(coerceToInt(null)).equal(null);
+    });
+
+    it('should pass through undefined unchanged', () => {
+      should(coerceToInt(undefined)).equal(undefined);
+    });
+
+    it('should pass through NaN unchanged', () => {
+      should(Number.isNaN(coerceToInt(NaN))).be.true();
+    });
+
+    it('should pass through Infinity unchanged', () => {
+      should(coerceToInt(Infinity)).equal(Infinity);
+      should(coerceToInt(-Infinity)).equal(-Infinity);
+    });
+
+    it('should pass through strings unchanged', () => {
+      should(coerceToInt('abc')).equal('abc');
+      should(coerceToInt('')).equal('');
+    });
+
+    it('should round numeric strings to nearest integer', () => {
+      should(coerceToInt('1132.6')).equal(1133);
+      should(coerceToInt('1132.4')).equal(1132);
+      should(coerceToInt('1132')).equal(1132);
+      should(coerceToInt('-500.3')).equal(-500);
+    });
+  });
+
+  describe('EntranceService.getConvertedDataFromClientRequest()', () => {
+    const makeReq = (body) => ({
+      body,
+      param: (key) => body[key],
+      token: { id: 1 },
+    });
+
+    it('should round decimal altitude to nearest integer', () => {
+      const req = makeReq({ altitude: 1132.6 });
+      const result = EntranceService.getConvertedDataFromClientRequest(req);
+      should(result.altitude).equal(1133);
+    });
+
+    it('should round decimal precision to nearest integer', () => {
+      const req = makeReq({ precision: 5.7 });
+      const result = EntranceService.getConvertedDataFromClientRequest(req);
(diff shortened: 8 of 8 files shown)


ClemRz:

## Introduction

When a user sends a decimal value (e.g., `1132.6`) for a numeric field that maps to a PostgreSQL integer column (`int2`/`int4`), the Waterline ORM passes the value through without coercion. PostgreSQL rejects it with `invalid input syntax for type smallint` (error code 22P02), causing a 500 Server Error. This affects all numeric fields across multiple models and entry points (API controllers and CSV import) where the Sails model declares `type: 'number'` but the underlying column is an integer type.

## Bug Analysis

### Current Behavior (Defect)

1.1 WHEN a decimal value (e.g., `1132.6`) is provided for the `altitude` field via POST/PUT `/api/v1/entrances` THEN the system returns a 500 Server Error with `invalid input syntax for type smallint: "1132.6"`

1.2 WHEN a decimal value is provided for the `precision` field via POST/PUT `/api/v1/entrances` THEN the system returns a 500 Server Error with `invalid input syntax for type smallint`

1.3 WHEN a decimal value is provided for the `yearDiscovery` field via POST/PUT `/api/v1/entrances` THEN the system returns a 500 Server Error with `invalid input syntax for type smallint`

1.4 WHEN a decimal value is provided for the `depth` field via POST/PUT `/api/v1/caves` THEN the system returns a 500 Server Error with `invalid input syntax for type integer`

1.5 WHEN a decimal value is provided for the `length` field via POST/PUT `/api/v1/caves` THEN the system returns a 500 Server Error with `invalid input syntax for type integer`

1.6 WHEN a decimal value is provided for the `min_depth` or `max_depth` fields via POST/PUT `/api/v1/caves` THEN the system returns a 500 Server Error with `invalid input syntax for type integer`

1.7 WHEN a CSV import row contains a decimal value for `altitude` or `precision` (entrance fields) THEN the system fails the import row with a database error

1.8 WHEN a CSV import row contains a decimal value for `depth` or `length` (cave fields) THEN the system fails the import row with a database error

2.1 WHEN a decimal value is provided for the `altitude` field via POST/PUT `/api/v1/entrances` THEN the system SHALL round the value to the nearest integer before persisting and return a successful response

2.2 WHEN a decimal value is provided for the `precision` field via POST/PUT `/api/v1/entrances` THEN the system SHALL round the value to the nearest integer before persisting and return a successful response

2.3 WHEN a decimal value is provided for the `yearDiscovery` field via POST/PUT `/api/v1/entrances` THEN the system SHALL round the value to the nearest integer before persisting and return a successful response

2.4 WHEN a decimal value is provided for the `depth` field via POST/PUT `/api/v1/caves` THEN the system SHALL round the value to the nearest integer before persisting and return a successful response

2.5 WHEN a decimal value is provided for the `length` field via POST/PUT `/api/v1/caves` THEN the system SHALL round the value to the nearest integer before persisting and return a successful response

2.6 WHEN a decimal value is provided for the `min_depth` or `max_depth` fields via POST/PUT `/api/v1/caves` THEN the system SHALL round the value to the nearest integer before persisting and return a successful response

2.7 WHEN a CSV import row contains a decimal value for `altitude` or `precision` (entrance fields) THEN the system SHALL round the value to the nearest integer before persisting and import the row successfully

2.8 WHEN a CSV import row contains a decimal value for `depth` or `length` (cave fields) THEN the system SHALL round the value to the nearest integer before persisting and import the row successfully

3.1 WHEN an integer value (e.g., `1132`) is provided for any integer-typed field via the API THEN the system SHALL CONTINUE TO persist the value as-is and return a successful response

3.2 WHEN a `null` or absent value is provided for an optional integer-typed field via the API THEN the system SHALL CONTINUE TO persist `null` and return a successful response

3.3 WHEN a non-numeric value (e.g., `"abc"`) is provided for an integer-typed field THEN the system SHALL CONTINUE TO reject the request with a validation error (existing Waterline type validation)

3.4 WHEN decimal values are provided for fields that map to non-integer DB columns (e.g., `temperature` which is `float8`, or `latitude`/`longitude` which are `numeric(24,20)`) THEN the system SHALL CONTINUE TO persist the exact decimal value without rounding

3.5 WHEN integer values are provided for cave and entrance fields via CSV import THEN the system SHALL CONTINUE TO import the rows successfully with the exact values

ClemRz:
# Numeric Input Validation Bugfix Design

## Overview

Decimal values (e.g., `1132.6`) sent to PostgreSQL integer columns (`int2`/`int4`) via the Waterline ORM cause `invalid input syntax for type smallint/integer` errors, resulting in 500 Server Errors. The fix introduces a shared `coerceToInt` utility that rounds numeric values to the nearest integer while preserving `null`/`undefined`, applied at each data conversion entry point before values reach the ORM.

## Glossary

- **Bug_Condition (C)**: A numeric value with a fractional part is provided for a field that maps to a PostgreSQL `int2` or `int4` column
- **Property (P)**: The value is rounded to the nearest integer via `Math.round()` before being passed to the ORM, so the database accepts it
- **Preservation**: Integer values, `null`/`undefined` values, and non-integer-column fields (e.g., `temperature`, `latitude`, `longitude`) must remain unchanged by the fix
- **coerceToInt**: A new utility function in `api/utils/coerceToInt.js` that rounds a numeric value to the nearest integer, passing through `null`/`undefined` unchanged
- **getConvertedDataFromClientRequest** (EntranceService): Extracts entrance data from the HTTP request body — affected fields: `altitude`, `precision`, `yearDiscovery`
- **getConvertedDataFromClient** (CaveService): Extracts cave data from the HTTP request — affected fields: `depth`, `caveLength`
- **cave/update.js controller**: Manually extracts cave fields — affected fields: `depth`, `caveLength` (as `newDepth`, `newLength`)
- **getConvertedEntranceFromCsv** (EntranceCSVImportService): Extracts entrance data from CSV row — affected fields: `altitude`, `precision`
- **getConvertedCaveFromCsv** (EntranceCSVImportService): Extracts cave data from CSV row — affected fields: `depth`, `length`

## Bug Details

### Bug Condition

The bug manifests when a client sends a decimal numeric value for any field that maps to a PostgreSQL `int2` or `int4` column. The Waterline ORM (`type: 'number'`) accepts any JavaScript number without integer validation, passing the raw decimal to PostgreSQL, which rejects it.

**Formal Specification:**
```
FUNCTION isBugCondition(input)
  INPUT: input of type { fieldName: string, value: any }
  OUTPUT: boolean

  INTEGER_FIELDS := {
    'altitude', 'precision', 'yearDiscovery',   -- TEntrance (int2)
    'depth', 'caveLength', 'min_depth', 'max_depth'  -- TCave (int4)
  }

  RETURN input.fieldName IN INTEGER_FIELDS
         AND input.value IS NOT NULL
         AND input.value IS NOT undefined
         AND typeof(input.value) == 'number'
         AND input.value !== Math.round(input.value)
END FUNCTION
```

### Examples

- `POST /api/v1/entrances` with `{ altitude: 1132.6 }` → 500 error `invalid input syntax for type smallint: "1132.6"` (expected: persists `1133`)
- `POST /api/v1/caves` with `{ depth: 250.4 }` → 500 error `invalid input syntax for type integer: "250.4"` (expected: persists `250`)
- `PUT /api/v1/caves/:id` with `{ length: 1500.7 }` → 500 error `invalid input syntax for type integer: "1500.7"` (expected: persists `1501`)
- CSV import with `w3geo:altitude` = `"845.3"` → import row fails (expected: persists `845`)
- `POST /api/v1/entrances` with `{ altitude: 1132 }` → works fine (no fractional part, not a bug condition)
- `POST /api/v1/caves` with `{ temperature: 12.5 }` → works fine (`temperature` is `float8`, not an integer column)

## Expected Behavior

### Preservation Requirements

**Unchanged Behaviors:**
- Integer values for integer-column fields must persist as-is (e.g., `altitude: 1132` → `1132`)
- `null` or `undefined` values for optional integer-column fields must persist as `null`
- Non-numeric values (e.g., `"abc"`) must continue to be rejected by existing Waterline type validation
- Decimal values for non-integer columns (`temperature` as `float8`, `latitude`/`longitude` as `numeric(24,20)`) must persist with full decimal precision
- All boolean coercion logic in `EntranceService.getConvertedDataFromClientRequest` must remain unchanged
- All non-integer fields in every converter function must pass through unchanged

**Scope:**
All inputs that do NOT involve decimal values for integer-column fields should be completely unaffected by this fix. This includes:
- Integer values for integer fields
- Any value for non-integer fields (temperature, latitude, longitude, booleans, strings)
- Null/undefined values for optional fields

## Hypothesized Root Cause

Based on the bug description, the root cause is straightforward:

1. **Missing integer coercion in data conversion layer**: The `getConvertedDataFromClientRequest` (EntranceService), `getConvertedDataFromClient` (CaveService), cave update controller, and CSV import converters pass raw numeric values directly to the ORM without rounding. Waterline's `type: 'number'` accepts any JavaScript number (including decimals), but PostgreSQL `int2`/`int4` columns reject non-integer values.

2. **No validation at the Waterline model level**: The Sails model definitions (`TCave.js`, `TEntrance.js`) declare `type: 'number'` without any custom validation to enforce integer-only values. Waterline does not auto-coerce to integers.

3. **Cave update controller bypasses service converter**: The `cave/update.js` controller manually extracts `depth` and `length` via `req.param()` without going through `CaveService.getConvertedDataFromClient`, so even fixing the service converter alone would miss this path.

The most likely root cause is #1 — the fix should be applied at the data conversion layer (closest to the input boundary) rather than at the model level, to keep the change minimal and localized.

## Correctness Properties

Property 1: Bug Condition - Decimal values are rounded to nearest integer

_For any_ numeric value `v` with a fractional part (i.e., `v !== Math.round(v)`), `coerceToInt(v)` SHALL return `Math.round(v)`, which is a safe integer that PostgreSQL `int2`/`int4` columns will accept.

**Validates: Requirements 2.1, 2.2, 2.3, 2.4, 2.5, 2.6, 2.7, 2.8**

Property 2: Preservation - Integer and nullish values pass through unchanged

_For any_ value `v` that is either an integer (`v === Math.round(v)`), `null`, or `undefined`, `coerceToInt(v)` SHALL return `v` unchanged, preserving existing behavior for all non-buggy inputs.

**Validates: Requirements 3.1, 3.2, 3.5**

Property 3: Preservation - Non-integer-column fields are not affected

_For any_ input to the converter functions, fields that map to non-integer database columns (`temperature`, `latitude`, `longitude`, boolean fields, string fields) SHALL pass through without any rounding or modification, preserving their exact values.

**Validates: Requirements 3.3, 3.4**

## Fix Implementation

### Changes Required

Assuming our root cause analysis is correct:

**File**: `api/utils/coerceToInt.js` (new file)

**Function**: `coerceToInt(value)`

**Specific Changes**:
1. **Create shared utility**: Implement `coerceToInt(value)` that returns `null`/`undefined` unchanged, returns `Math.round(value)` for finite numbers, and returns `value` as-is for non-numeric types (letting Waterline's own validation handle rejection)

**File**: `api/services/EntranceService.js`

**Function**: `getConvertedDataFromClientRequest(req)`

**Specific Changes**:
2. **Coerce entrance integer fields**: Apply `coerceToInt` to `altitude`, `precision`, and `yearDiscovery` in the returned object. These fields come from `req.body` via the spread operator, so wrap them after the spread.

**File**: `api/services/CaveService.js`

**Function**: `getConvertedDataFromClient(req)`

**Specific Changes**:
3. **Coerce cave integer fields**: Apply `coerceToInt` to `depth` and `caveLength` in the returned object.

**File**: `api/controllers/v1/cave/update.js`

**Specific Changes**:
4. **Coerce cave update fields**: Apply `coerceToInt` to `newDepth` and `newLength` before assigning to `updatedFields`.

**File**: `api/services/EntranceCSVImportService.js`

**Function**: `getConvertedEntranceFromCsv(rawData, idAuthor, cave)`

**Specific Changes**:
5. **Coerce CSV entrance integer fields**: Apply `coerceToInt` to `altitude` and `precision` in the returned object.

**File**: `api/services/EntranceCSVImportService.js`

**Function**: `getConvertedCaveFromCsv(rawData, idAuthor)`

**Specific Changes**:
6. **Coerce CSV cave integer fields**: Apply `coerceToInt` to `depth` and `length` in the returned object. Note that `depth` is computed from `extendBelowEntrance + extendAboveEntrance` with `parseInt`, but the individual `parseInt` calls may still produce a sum that is already an integer — however, the direct `karstlink:verticalExtend` path does not go through `parseInt`, so it needs coercion.

## Testing Strategy

### Validation Approach

**Goal**: Surface counterexamples that demonstrate the bug BEFORE implementing the fix. Confirm or refute the root cause analysis. If we refute, we will need to re-hypothesize.

**Test Plan**: Write tests that pass decimal values through the converter functions and assert the output. On unfixed code, the decimal values will pass through unmodified, confirming the root cause.

**Test Cases**:
1. **Entrance decimal passthrough**: Call `getConvertedDataFromClientRequest` with `altitude: 1132.6` and verify the output contains `1132.6` (unfixed) — confirms no rounding exists (will fail after fix, which is expected)
2. **Cave decimal passthrough**: Call `getConvertedDataFromClient` with `depth: 250.4` and verify the output contains `250.4` (unfixed) — confirms no rounding exists
3. **CSV entrance decimal passthrough**: Call `getConvertedEntranceFromCsv` with decimal altitude and verify passthrough
4. **CSV cave decimal passthrough**: Call `getConvertedCaveFromCsv` with decimal depth and verify passthrough

**Expected Counterexamples**:
- Decimal values pass through all converter functions unmodified
- Root cause confirmed: no integer coercion exists at the data conversion layer

### Fix Checking

**Goal**: Verify that for all inputs where the bug condition holds, the fixed function produces the expected behavior.

**Pseudocode:**
```
FOR ALL input WHERE isBugCondition(input) DO
  result := coerceToInt(input.value)
  ASSERT result == Math.round(input.value)
  ASSERT Number.isInteger(result)
END FOR
```

### Preservation Checking

**Goal**: Verify that for all inputs where the bug condition does NOT hold, the fixed function produces the same result as the original function.

**Pseudocode:**
```
FOR ALL input WHERE NOT isBugCondition(input) DO
  ASSERT coerceToInt(input.value) === input.value
END FOR
```

**Testing Approach**: Property-based testing with fast-check is recommended for both fix and preservation checking because:
- It generates many numeric values automatically (integers, decimals, edge cases like `NaN`, `Infinity`, `0`, `-0`)
- It catches edge cases that manual unit tests might miss (e.g., very large numbers, negative decimals)
- It provides strong guarantees that `coerceToInt` behaves correctly across the full numeric domain

**Test Plan**: Write property-based tests for `coerceToInt` directly, then integration-level tests for each converter function.

**Test Cases**:
1. **Integer preservation**: Generate random integers, verify `coerceToInt(n) === n`
2. **Null/undefined preservation**: Verify `coerceToInt(null) === null` and `coerceToInt(undefined) === undefined`
3. **Non-integer field preservation**: Verify converter functions do not round `temperature`, `latitude`, `longitude`

### Unit Tests

- Test `coerceToInt` with integers, decimals, null, undefined, NaN, Infinity, string numbers
- Test each converter function with decimal values for integer fields and verify rounding
- Test each converter function with integer values and verify passthrough

### Property-Based Tests

- Generate random `fc.double()` values and verify `coerceToInt` returns `Math.round(value)` for finite numbers
- Generate random `fc.integer()` values and verify `coerceToInt` returns the value unchanged (preservation)
- Generate random `fc.oneof(fc.constant(null), fc.constant(undefined))` and verify passthrough
- Generate random entrance/cave data objects with mixed decimal/integer fields and verify only integer-column fields are rounded

### Integration Tests

- Test `POST /api/v1/entrances` with decimal `altitude` and verify successful response
- Test `PUT /api/v1/caves/:id` with decimal `depth` and verify successful response
- Test that `temperature` decimal values are preserved exactly in cave create/update

Paul-AUB (changes requested):
# Code Review 

**Files changed:** 8 (+252 / -14)

---

## Summary

This commit introduces a `coerceToInt` utility that rounds floating-point values to the nearest integer via `Math.round()`, and applies it at the data-conversion layer for integer DB columns: `depth`, `caveLength` (cave), and `altitude`, `precision`, `yearDiscovery` (entrance). Three test files cover the utility directly, with property-based tests via `fast-check`.

---

## What's Good

- **Minimal, well-placed utility:** `coerceToInt` is small, focused, and correctly handles `null`, `undefined`, and non-finite numbers by passing them through — letting Waterline handle those cases downstream.
- **Applied at the right layer:** coercion happens in the service-level data conversion functions, not in the controller or model, which is consistent with the existing architecture.
- **Property-based tests:** using `fast-check` to verify rounding and preservation behavior is a solid approach for a numeric utility.
- **Preservation contract is explicit:** testing that integers, nulls, and non-finites are untouched is as important as testing the rounding itself, and both are covered.
- **`temperature` correctly excluded:** `CaveService` leaves `temperature` as a float, which is correct given the DB schema.

---

## Issues and Concerns

### 1. `coerceToInt` is a no-op for string inputs — CSV import fix is silently ineffective

**Severity: High**

`coerceToInt` only acts when `typeof value === 'number'`. The `doubleCheck` CSV helper (`api/helpers/CsvHelpers/double-check.js`) returns values as-is from the raw data object — which for CSV parsing means **strings** (e.g., `'1132.6'`).

```js
// double-check.js — returns data[key] directly, no type cast
return exits.success(data[key] && data[key] !== '' ? func(data[key]) : defaultValue);
```

So in `EntranceCSVImportService.js`:

```js
length: coerceToInt(
  doubleCheckWithData({ key: 'karstlink:length' }) // returns '1132.6' (string)
)
// coerceToInt('1132.6') → '1132.6' (no-op, typeof is 'string')
```

The bug that prompted this commit — decimal values being stored in integer columns — is **not fixed for CSV imports**. The coercion only works for the API endpoints where `req.param()` returns a JavaScript number.

Note: the `depth` field is a partial exception — when computed from `parseInt()` sums it is already an integer, but when it comes from `karstlink:verticalExtend` directly, it is a string and is also unaffected.

**Fix:** parse the string to a number before rounding, or make `coerceToInt` handle numeric strings:

```js
const coerceToInt = (value) => {
  if (value === null || value === undefined) return value;
  const n = typeof value === 'string' ? parseFloat(value) : value;
  if (typeof n === 'number' && Number.isFinite(n)) return Math.round(n);
  return value;
};
```

---

### 2. Stale TDD comments in test files are misleading

**Severity: Low**

`coerceToInt.property.test.js` contains:

```js
/**
 * This test imports `coerceToInt` from `api/utils/coerceToInt.js`, which does
 * not exist yet. The import failure itself confirms the bug condition…
 */
```

And `coerceToInt-preservation.property.test.js` says:

```js
/**
 * Before the fix exists, we stub coerceToInt as an identity function to encode
 * the preservation baseline…
 */
```

Both comments describe a pre-fix TDD state that no longer applies. The file exists, the import succeeds, and there is no stubbing anywhere in the file. A reader unfamiliar with the TDD workflow will be confused by the reference to a missing file and a stub that is never defined.

**Fix:** remove or replace the TDD-context paragraphs with a plain description of what the test verifies.

---

### 3. Test describe/file names say "Bug Condition" — misleading after the fix

**Severity: Low**

`coerceToInt.property.test.js`:

```js
describe('coerceToInt - Property: Bug Condition', () => {
  describe('Property 1: Decimal values are rounded to nearest integer', () => {
```

Once the fix is merged, "Bug Condition" describes correct, expected behavior. Future contributors may wonder whether this test is documenting a known defect or asserting normal operation.

**Fix:** rename to `'Property: Rounding Behavior'` or similar.

---

### 4. No test covering `EntranceCSVImportService` with `coerceToInt`

**Severity: Medium**

`coerceToInt.test.js` tests `EntranceService` and `CaveService`, but `EntranceCSVImportService` — which has the most complex integration and the silent no-op issue described in #1 — has no test coverage at all in this commit.

A test like this would have caught the string-input issue:

```js
it('should round decimal length from CSV string to nearest integer', () => {
  const rawData = { 'karstlink:length': '1132.6' };
  const result = EntranceCSVImportService.getConvertedCaveFromCsv(rawData, 1);
  should(result.length).equal(1133); // would currently fail: returns '1132.6'
});
```

---

### 5. `Math.round` vs `Math.trunc` for physical measurements

**Severity: Low / Design trade-off**

`Math.round(250.5)` → `251`, `Math.round(250.4)` → `250`. For speleological measurements like depth, length, and altitude, it is not obvious that rounding to nearest is preferable to truncation — surveyors typically record conservative (floor) values. This is a product/domain decision, but it should be a conscious one.

---

### 6. `yearDiscovery` silently coerced instead of validated

**Severity: Low**

Rounding `yearDiscovery: 1973.8` to `1974` is permissive. A year of discovery is an exact historical fact; a fractional year input is almost certainly a client error, not a precision issue. This field is arguably better served by a validation rejection (400 Bad Request) than silent rounding.

---

### 7. `coerceToInt.test.js` loads Sails-dependent services in a `2_utils` test file

**Severity: Very Low**

The unit test file imports `EntranceService` and `CaveService`, which depend on Sails context (e.g., `sails.helpers` is called at module load time in `EntranceCSVImportService`). This works as long as the integration suite lifts Sails before `2_utils` tests run, but it creates an implicit ordering dependency. A test named `coerceToInt.test.js` in a `utils` folder is unexpected to require a live Sails instance.

---

## Summary Table

| # | Location | Severity | Topic |
|---|----------|----------|-------|
| 1 | `EntranceCSVImportService.js` | **High** | `coerceToInt` is a no-op for string CSV values — decimal bug not actually fixed for imports |
| 2 | Both property test files | Low | Stale TDD comments reference a missing file/stub that no longer apply |
| 3 | `coerceToInt.property.test.js` | Low | Describe block named "Bug Condition" after the fix is misleading |
| 4 | Tests | **Medium** | No test for `EntranceCSVImportService` with `coerceToInt` — the silent regression in #1 would go undetected |
| 5 | `coerceToInt.js` | Low | `Math.round` vs `Math.trunc` for physical measurements — design trade-off worth documenting |
| 6 | `EntranceService.js` | Low | `yearDiscovery` should arguably be validated, not silently rounded |
| 7 | `coerceToInt.test.js` | Very Low | `2_utils` test file has implicit dependency on a live Sails instance |

ClemRz (commented):
Thanks for the thorough review @Paul-AUB — addressed the actionable items, here's the breakdown:

### #1 — `coerceToInt` no-op for string CSV values ✅ Fixed

Good catch. `coerceToInt` now parses numeric strings via `parseFloat()` before rounding:

```js
const n = typeof value === 'string' ? parseFloat(value) : value;
if (typeof n === 'number' && Number.isFinite(n)) return Math.round(n);
```

Non-numeric strings (e.g. `'abc'`, `''`) still pass through unchanged since `parseFloat` returns `NaN` for those.

### #2 — Stale TDD comments ✅ Fixed

Removed the pre-fix TDD-context paragraphs and replaced with plain descriptions of what each test verifies.

### #3 — "Bug Condition" naming ✅ Fixed

Renamed to `'coerceToInt - Property: Rounding Behavior'`.

### #4 — No test for CSV import path ✅ Fixed

Added a property-based test that generates numeric strings with decimals and verifies `coerceToInt` rounds them correctly. Also added unit tests for numeric string inputs. This covers the CSV import code path where values arrive as strings.

### #5 — `Math.round` vs `Math.trunc` — Acknowledged

Fair point on the domain semantics. `Math.round` was a deliberate choice — it's the least surprising default for general-purpose coercion. If the product team prefers truncation for physical measurements, that's a one-line change. Leaving as-is for now.

### #6 — `yearDiscovery` validation — Acknowledged

Agreed that a fractional year is likely a client error. However, adding validation/rejection here would expand the scope of this fix beyond the original bug (500 on decimal input → silent coercion). Happy to address this in a follow-up if desired.

### #7 — Test file Sails dependency — Acknowledged

The numbered prefix convention (`0_models`, `1_services`, `2_utils`) already guarantees execution order, so the implicit dependency is safe in practice. Not changing this to avoid unnecessary file moves.

---

All 1886 tests passing, no regressions. Force-pushed the amended commit.

Paul-AUB (approved):

PR #1528, opened 2026-04-04
fix(search): return language field from Typesense hit documents in advanced-search

ClemRz:
## 🤔 What

Fix the `language` field being `null`/`undefined`/missing in advanced-search results for all entity types (entrances, caves, massifs, organizations, documents).

## 🤷‍♂️ Why

The advanced-search endpoint (`POST /api/v1/advanced-search`) returns `language: null` for every entity type, even though:
- The Typesense index stores the language field correctly (proven by field-search faceting)
- Individual entity endpoints (e.g., `GET /api/v1/entrances/:id`) return it correctly
- Filtering by language works (matching results come back, but with `language: null`)

This means the language column in search result tables renders empty when users toggle it visible.

## 🔍 How

Three root causes, three fixes:

1. **`getMainLanguage()` in `api/services/mapping/utils.js`** — only extracted language from a `names` Array (database shape). Typesense hit documents have a flat `language` string. Added a fallback to `source.language ?? null` when `source.names` is not an Array, mirroring how `getMainName()` already handles this.

2. **`toOrganization` in `api/services/mapping/converters.js`** — derived language inline via `Array.isArray(source.names) ? ... : undefined`. Changed the fallback from `undefined` to `source.language`. Also added `language: source.language` to `toDocument`.

3. **Document Typesense schema in `api/dbSync/entities/document.js`** — was missing the `language` field entirely. Added it to the schema (faceted, sortable, optional), added it to `query_by`, and updated `importFormater` to extract the first language from `j_document_language` as the primary language.

## 🧪 Testing

7 property-based tests added in `test/integration/1_services/SearchLanguageField.property.test.js`:

- 4 bug condition tests: verify `getMainLanguage`, `toSimpleCave`, `toOrganization`, and `toDocument` return the language from flat Typesense sources
- 3 preservation tests: verify the database path (names Array) continues working identically for `getMainLanguage`, `toOrganization`, and `toDocument`

Run: `PORT=1338 npm run test -- --grep "SearchLanguageField"`

After deploying, the Typesense document collection needs to be re-synced (`npm run dev:clean` locally) for the new `language` field to be populated in existing document records.

## 📸 Previews

Before (all entity types):
```json
"language": null
```

After:
```json
"language": "fra"
```

### DIFF
--- a/api/dbSync/entities/document.js
+++ b/api/dbSync/entities/document.js
@@ -202,6 +202,7 @@ function importFormater(d) {
   d.dateInscription = new Date(d.dateInscription).getTime();
   if (d.dateReviewed) d.dateReviewed = new Date(d.dateReviewed).getTime();
   if (d.dateValidation) d.dateValidation = new Date(d.dateValidation).getTime();
+  d.language = d.languages?.[0];
   return d;
 }
 /* eslint-enable no-param-reassign */
@@ -293,6 +294,13 @@ module.exports = {
         { name: 'cave.name', type: 'string', optional: true, sort: true },
         { name: 'entrances.name', type: 'string[]', optional: true },
         { name: 'massifs.name', type: 'string[]', optional: true },
+        {
+          name: 'language',
+          type: 'string',
+          facet: true,
+          optional: true,
+          sort: true,
+        },
       ],
       default_sorting_field: 'dateInscription',
     },
--- a/api/services/mapping/converters.js
+++ b/api/services/mapping/converters.js
@@ -245,6 +245,12 @@ const c = {
       option: source.option?.name,
       mainLanguage: source.mainLanguage?.id ?? source.languages?.[0]?.id,
       languages: source.languages?.map((e) => e.id),
+      // language: from Typesense flat field, or derived from DB mainLanguage/languages
+      language:
+        source.language ??
+        source.mainLanguage?.id ??
+        source.languages?.[0]?.id ??
+        null,
 
       iso3166: source.iso3166 ?? [
         ...(source.countries?.map((e) => ({ iso: e.id, name: e.nativeName })) ??
@@ -606,7 +612,7 @@ const c = {
       : undefined,
     language: Array.isArray(source.names)
       ? source.names.find((name) => name.isMain)?.language
-      : undefined,
+      : source.language,
     latitude: parseFloat(source.latitude),
     longitude: parseFloat(source.longitude),
     country: source.country?.id ?? source.country,
--- a/api/services/mapping/utils.js
+++ b/api/services/mapping/utils.js
@@ -56,7 +56,7 @@ module.exports = {
     return mainName;
   },
   getMainLanguage: (source) => {
-    if (!(source.names instanceof Array)) return null;
+    if (!(source.names instanceof Array)) return source.language ?? null;
     return source.names.find((name) => name.isMain)?.language ?? null;
   },
 
--- a/test/integration/1_services/SearchLanguageField.property.test.js
+++ b/test/integration/1_services/SearchLanguageField.property.test.js
@@ -1,0 +1,213 @@
+/* eslint-disable func-names */
+const should = require('should');
+const fc = require('fast-check');
+const { getMainLanguage } = require('../../../api/services/mapping/utils');
+const {
+  toSimpleCave,
+  toOrganization,
+  toDocument,
+  toSimpleOrganization,
+} = require('../../../api/services/mapping/converters');
+
+// --- Shared arbitraries ---
+
+// ISO 639-3-like codes: 3 lowercase letters
+const lowerAlpha = fc.constantFrom(...'abcdefghijklmnopqrstuvwxyz');
+const iso639Arb = fc
+  .tuple(lowerAlpha, lowerAlpha, lowerAlpha)
+  .map(([a, b, c]) => a + b + c);
+
+// Unix timestamp in milliseconds (realistic range: 2020–2030)
+const timestampArb = fc.integer({ min: 1577836800000, max: 1893456000000 });
+
+/**
+ * Property 1: Bug Condition — Search Language Field Returns Null
+ * for Typesense Hit Documents
+ *
+ * When the source object has a flat `language` string (Typesense shape)
+ * and no `names` Array (database shape), the converters should return
+ * the language string — not null/undefined/missing.
+ *
+ * Validates: Requirements 1.1, 1.2, 1.3, 1.4, 1.5, 2.1, 2.2, 2.3, 2.4, 2.5
+ */
+describe('SearchLanguageField - Property 1: Bug Condition', () => {
+  it('1a: getMainLanguage returns the language from a flat Typesense source', function () {
+    this.timeout(30000);
+    fc.assert(
+      fc.property(iso639Arb, (lang) => {
+        const result = getMainLanguage({ language: lang });
+        should(result).equal(lang);
+      }),
+      { numRuns: 100 }
+    );
+  });
+
+  it('1b: toSimpleCave returns the language from a flat Typesense source', function () {
+    this.timeout(30000);
+    fc.assert(
+      fc.property(iso639Arb, (lang) => {
+        const result = toSimpleCave({ id: '1', name: 'Cave', language: lang });
+        should(result.language).equal(lang);
+      }),
+      { numRuns: 100 }
+    );
+  });
+
+  it('1c: toOrganization returns the language from a flat Typesense source', function () {
+    this.timeout(30000);
+    fc.assert(
+      fc.property(iso639Arb, timestampArb, (lang, ts) => {
+        const result = toOrganization({
+          id: '1',
+          name: 'Org',
+          language: lang,
+          dateInscription: ts,
+        });
+        should(result.language).equal(lang);
+      }),
+      { numRuns: 100 }
+    );
+  });
+
+  it('1d: toDocument returns the language from a flat Typesense source', function () {
+    this.timeout(30000);
+    fc.assert(
+      fc.property(iso639Arb, timestampArb, (lang, ts) => {
+        const result = toDocument({
+          id: '1',
+          title: 'Doc',
+          language: lang,
+          dateInscription: ts,
+        });
+        should(result.language).equal(lang);
+      }),
+      { numRuns: 100 }
+    );
+  });
+
+  it('1e: toSimpleOrganization returns the language from a flat Typesense source', function () {
+    this.timeout(30000);
+    fc.assert(
+      fc.property(iso639Arb, (lang) => {
+        const result = toSimpleOrganization({
+          id: '1',
+          name: 'Org',
+          language: lang,
+        });
+        should(result.language).equal(lang);
+      }),
+      { numRuns: 100 }
+    );
+  });
+});
+
+/**
+ * Property 2: Preservation — Database Sources Continue Deriving
+ * Language from Names Array
+ *
+ * When the source object has a `names` Array (database shape), the
+ * converters must continue to derive language from the main name entry.
+ * These tests capture baseline behavior on UNFIXED code so we can
+ * verify the fix does not regress the database path.
+ *
+ * Validates: Requirements 3.1, 3.2, 3.3, 3.4, 3.5
+ */
+describe('SearchLanguageField - Property 2: Preservation', () => {
+  // Arbitrary: random names arrays with varying structures
+  const namesArb = fc.array(
+    fc.record({
+      name: fc.string(),
+      isMain: fc.boolean(),
+      language: iso639Arb,
+      id: fc.nat(),
+    })
+  );
+
+  /**
+   * 2a: getMainLanguage preservation
+   *
+   * Constrains: For all source objects where source.names IS an Array,
+   * getMainLanguage returns the language of the first entry with
+   * isMain === true, or null if no such entry exists.
+   *
+   * Covers: empty arrays, arrays with no main name, arrays with one
+   * or more main names (first match wins via Array.find).
+   */
+  it('2a: getMainLanguage derives language from names Array', function () {
+    this.timeout(30000);
+    fc.assert(
+      fc.property(namesArb, (names) => {
+        const source = { names };
+        const result = getMainLanguage(source);
+        const expected = names.find((n) => n.isMain)?.language ?? null;
+        should(result).equal(expected);
+      }),
+      { numRuns: 100 }
+    );
+  });
+
+  /**
+   * 2b: toOrganization preservation
+   *
+   * Constrains: For all source objects where source.names IS an Array,
+   * toOrganization derives language and nameId from the main name entry
+   * in the Array, matching the inline Array.isArray derivation.
+   *
+   * Covers: organizations with database-shaped names arrays.
+   */
+  it('2b: toOrganization derives language and nameId from names Array', function () {
+    this.timeout(30000);
+    fc.assert(
+      fc.property(namesArb, timestampArb, (names, ts) => {
+        const source = {
+          id: 1,
+          names,
+          dateInscription: ts,
+        };
+        const result = toOrganization(source);
+        const mainEntry = names.find((n) => n.isMain);
+        const expectedLanguage = mainEntry?.language;
+        const expectedNameId = mainEntry?.id;
+        should(result.language).equal(expectedLanguage);
+        should(result.nameId).equal(expectedNameId);
+      }),
+      { numRuns: 100 }
+    );
+  });
+
+  /**
+   * 2c: toDocument preservation
+   *
+   * Constrains: For all document source objects with database shape
+   * (having a mainLanguage object with an id, and a languages array
+   * of objects with ids), the output mainLanguage and languages fields
+   * are unchanged — mainLanguage maps to source.mainLanguage.id and
+   * languages maps to source.languages.map(e => e.id).
+   *
+   * Covers: documents loaded from the database via Waterline ORM.
+   */
+  it('2c: toDocument preserves mainLanguage and languages from DB shape', function () {
+    this.timeout(30000);
+    fc.assert(
+      fc.property(
+        iso639Arb,
+        fc.array(iso639Arb, { minLength: 1, maxLength: 5 }),
+        timestampArb,
+        (mainLang, langIds, ts) => {
+          const source = {
+            id: 1,
+            dateInscription: ts,
+            mainLanguage: { id: mainLang },
+            languages: langIds.map((lid) => ({ id: lid })),
+          };
+          const result = toDocument(source);
+          should(result.mainLanguage).equal(mainLang);
+          should(result.languages).deepEqual(langIds);
+          // language is now always populated, derived from mainLanguage for DB sources
+          should(result.language).equal(mainLang);
+        }
+      ),
+      { numRuns: 100 }
+    );
+  });
+});


ClemRz:

## Introduction

The advanced-search API endpoint (`POST /api/v1/advanced-search`) returns `language: null` for every entity type (entrances, caves, massifs, documents, organizations), even though the Typesense index stores the language field correctly and individual entity endpoints return it. The root cause is twofold:

1. The `getMainLanguage()` utility only extracts language from a `names` Array (populated from the database). Typesense hit documents have a flat `language` string field, so `getMainLanguage()` returns `null` for all search results.
2. The `toOrganization` converter derives language from `source.names` array directly, which also fails for Typesense hits.
3. The `toDocument` converter never maps a `language` field at all, and the document Typesense schema is missing the `language` field definition entirely.

This affects all users who rely on the advanced-search endpoint to see language information in search result tables.

## Bug Analysis

### Current Behavior (Defect)

1.1 WHEN the advanced-search endpoint returns entrance results THEN the system returns `language: null` because `getMainLanguage()` finds no `names` Array on the flat Typesense hit document, even though the hit document contains a `language` string field

1.2 WHEN the advanced-search endpoint returns cave results THEN the system returns `language: null` because `toSimpleCave` calls `getMainLanguage()` which requires a `names` Array not present on Typesense hit documents

1.3 WHEN the advanced-search endpoint returns massif results THEN the system returns `language: null` because `toMassif` calls `getMainLanguage()` which requires a `names` Array not present on Typesense hit documents

1.4 WHEN the advanced-search endpoint returns organization results THEN the system returns `language: undefined` because `toOrganization` checks `Array.isArray(source.names)` which is false for Typesense hit documents that have a flat `language` string field

1.5 WHEN the advanced-search endpoint returns document results THEN the system returns no `language` field because `toDocument` does not map any language field from the Typesense hit document, and the document Typesense schema (`api/dbSync/entities/document.js`) does not include a `language` field definition

2.1 WHEN the advanced-search endpoint returns entrance results THEN the system SHALL return the `language` value from the Typesense hit document (e.g., `"fra"`, `"eng"`)

2.2 WHEN the advanced-search endpoint returns cave results THEN the system SHALL return the `language` value from the Typesense hit document

2.3 WHEN the advanced-search endpoint returns massif results THEN the system SHALL return the `language` value from the Typesense hit document

2.4 WHEN the advanced-search endpoint returns organization results THEN the system SHALL return the `language` value from the Typesense hit document

2.5 WHEN the advanced-search endpoint returns document results THEN the system SHALL return the `language` value from the Typesense hit document, after adding the `language` field to the document Typesense schema and sync query

3.1 WHEN individual entity endpoints return entrance data (e.g., `GET /api/v1/entrances/:id`) THEN the system SHALL CONTINUE TO return the correct `language` value derived from the `names` Array via `getMainLanguage()`

3.2 WHEN individual entity endpoints return cave data (e.g., `GET /api/v1/caves/:id`) THEN the system SHALL CONTINUE TO return the correct `language` value derived from the `names` Array via `getMainLanguage()`

3.3 WHEN individual entity endpoints return massif data (e.g., `GET /api/v1/massifs/:id`) THEN the system SHALL CONTINUE TO return the correct `language` value derived from the `names` Array via `getMainLanguage()`

3.4 WHEN individual entity endpoints return organization data (e.g., `GET /api/v1/organizations/:id`) THEN the system SHALL CONTINUE TO return the correct `language` value derived from the `names` Array

3.5 WHEN individual entity endpoints return document data (e.g., `GET /api/v1/documents/:id`) THEN the system SHALL CONTINUE TO return the correct `mainLanguage` and `languages` values from the database

3.6 WHEN the field-search endpoint is used to facet or filter by language THEN the system SHALL CONTINUE TO return correct language facet data for all entity types

3.7 WHEN the `toCaver` converter processes person results from search THEN the system SHALL CONTINUE TO return the correct `language` value (this converter already maps `source.language` directly and is not affected by the bug)

ClemRz:
# Search Language Field Fix — Bugfix Design

## Overview

The advanced-search endpoint returns `language: null` (or `undefined` / missing) for all entity types because the converter layer assumes language is derived from a `names` Array (database shape), but Typesense hit documents carry language as a flat string field. The fix adds a fallback path in `getMainLanguage()`, patches `toOrganization`'s inline derivation, and adds the missing `language` field to the document Typesense schema and sync query.

## Glossary

- **Bug_Condition (C)**: The source object has no `names` Array (i.e., it is a Typesense hit document) and the converter attempts to derive `language` from `source.names`
- **Property (P)**: The converter returns the correct language string from `source.language` when `source.names` is absent
- **Preservation**: When `source.names` IS an Array (database path), the existing derivation logic must continue to work identically
- **getMainLanguage()**: Utility in `api/services/mapping/utils.js` that extracts the language of the main name from a source object
- **toOrganization**: Converter in `api/services/mapping/converters.js` that maps organization data; derives language inline via `Array.isArray(source.names)`
- **toDocument**: Converter in `api/services/mapping/converters.js` that maps document data; currently has no `language` field mapping for search results
- **Typesense hit document**: The flat JSON object returned by Typesense search; has `language` as a top-level string, no `names` Array

## Bug Details

### Bug Condition

The bug manifests when the advanced-search endpoint converts Typesense hit documents into API response objects. The converters assume language is nested inside a `names` Array (the database shape), but Typesense hit documents have a flat `language` string field. This affects three distinct code paths:

1. `getMainLanguage(source)` — used by `toEntrance`, `toSimpleCave`, `toMassif`, `toSimpleMassif`, `toSimpleEntrance`, `toSimpleOrganization`, `toDeletedEntity`
2. `toOrganization` — derives language inline via `Array.isArray(source.names) ? source.names.find(…)?.language : undefined`
3. `toDocument` — never maps a `language` field at all; additionally, the document Typesense schema is missing the `language` field definition

**Formal Specification:**
```
FUNCTION isBugCondition(source, converterName)
  INPUT: source of type Object, converterName of type String
  OUTPUT: boolean

  IF converterName IN ['toEntrance', 'toSimpleCave', 'toMassif', 'toSimpleMassif',
                        'toSimpleEntrance', 'toSimpleOrganization', 'toDeletedEntity']
    RETURN NOT (source.names instanceof Array)
           AND typeof source.language === 'string'
  END IF

  IF converterName === 'toOrganization'
    RETURN NOT Array.isArray(source.names)
           AND typeof source.language === 'string'
  END IF

  IF converterName === 'toDocument'
    RETURN typeof source.language === 'string'
           // toDocument never maps language regardless of source shape
  END IF

  RETURN false
END FUNCTION
```

### Examples

- **Entrance from search**: `{ id: "148336", name: "Piqueur (Gouffre du)", language: "fra" }` → `getMainLanguage()` returns `null` because `source.names` is not an Array. Expected: `"fra"`
- **Entrance from DB**: `{ id: 148336, names: [{ name: "Piqueur", isMain: true, language: "fra" }] }` → `getMainLanguage()` returns `"fra"`. This path must remain unchanged.
- **Organization from search**: `{ id: "42", name: "Spéléo-Club", language: "fra" }` → `toOrganization` sets `language: undefined` because `Array.isArray(source.names)` is false. Expected: `"fra"`
- **Document from search**: `{ id: "100", title: "Karst Report", language: "eng" }` → `toDocument` returns no `language` field at all. Expected: `"eng"` (once schema is updated)
- **Cave from search**: `{ id: "500", name: "Grotte de Lascaux", language: "fra" }` → `toSimpleCave` calls `getMainLanguage()` which returns `null`. Expected: `"fra"`

## Expected Behavior

### Preservation Requirements

**Unchanged Behaviors:**
- When `source.names` IS an Array (database path), `getMainLanguage()` must continue to find the main name's language from the Array
- When `source.names` IS an Array, `toOrganization` must continue to derive `language` and `nameId` from the Array
- `toDocument`'s existing `mainLanguage` and `languages` fields (derived from DB joins) must remain unchanged
- The `toCaver` converter already maps `source.language` directly and must not be affected
- All individual entity endpoints (`GET /api/v1/entrances/:id`, etc.) must continue returning correct language values
- Field-search faceting by language must continue to work for all entity types
- The `getMainName()` utility already handles both paths (search `name` vs DB `names` Array) and must not be affected

**Scope:**
All inputs where `source.names` IS an Array should be completely unaffected by this fix. This includes:
- Individual entity endpoint responses (database queries with populated `names`)
- Any converter call where the source comes from Waterline ORM
- The `toCaver` converter (already maps `source.language` directly)

## Hypothesized Root Cause

Based on code analysis, the root causes are confirmed (not hypothesized):

1. **`getMainLanguage()` has no fallback**: Unlike `getMainName()` which falls back to `source.name` when `source.names` is not an Array, `getMainLanguage()` simply returns `null` when `source.names` is not an Array. It never checks `source.language`.

2. **`toOrganization` inline derivation has no fallback**: The expression `Array.isArray(source.names) ? source.names.find((name) => name.isMain)?.language : undefined` explicitly returns `undefined` when `source.names` is not an Array, ignoring `source.language`.

3. **Document Typesense schema missing `language` field**: The document entity's Typesense schema in `api/dbSync/entities/document.js` does not define a `language` field. The sync query joins `j_document_language` to get `languages` (an array of language IDs) but never extracts a single `language` field for the Typesense document. Even if the converter were fixed, there would be no `language` value in the Typesense hit document to map.

4. **`toDocument` never maps `language` for search**: The `toDocument` converter maps `mainLanguage` (from DB `source.mainLanguage?.id`) and `languages` (from DB `source.languages`), but has no mapping for a flat `language` field that would come from Typesense.

## Correctness Properties

Property 1: Bug Condition — Search results return language from flat field

_For any_ source object where `source.names` is not an Array and `source.language` is a string, the fixed `getMainLanguage()` function SHALL return `source.language` instead of `null`.

**Validates: Requirements 2.1, 2.2, 2.3**

Property 2: Preservation — Database sources continue using names Array

_For any_ source object where `source.names` IS an Array, the fixed `getMainLanguage()` function SHALL produce the same result as the original function, preserving the existing derivation from `source.names.find(name => name.isMain)?.language`.

**Validates: Requirements 3.1, 3.2, 3.3**

Property 3: Bug Condition — Organization search results return language

_For any_ source object where `source.names` is not an Array and `source.language` is a string, the fixed `toOrganization` converter SHALL return `source.language` as the `language` field instead of `undefined`.

**Validates: Requirements 2.4**

Property 4: Preservation — Organization database sources continue using names Array

_For any_ source object where `source.names` IS an Array, the fixed `toOrganization` converter SHALL produce the same `language` and `nameId` values as the original converter.

**Validates: Requirements 3.4**

Property 5: Bug Condition — Document search results include language

_For any_ document source object from Typesense that has a `language` string field, the fixed `toDocument` converter SHALL include that `language` value in the output.

**Validates: Requirements 2.5**

Property 6: Preservation — Document database sources unchanged

_For any_ document source object from the database, the fixed `toDocument` converter SHALL continue to return `mainLanguage` and `languages` fields identically to the original converter.

**Validates: Requirements 3.5**

## Fix Implementation

### Changes Required

**File**: `api/services/mapping/utils.js`

**Function**: `getMainLanguage`

**Specific Changes**:
1. **Add fallback to `source.language`**: When `source.names` is not an Array, return `source.language ?? null` instead of just `null`. This mirrors how `getMainName()` already falls back to `source.name`.

```js
getMainLanguage: (source) => {
  if (!(source.names instanceof Array)) return source.language ?? null;
  return source.names.find((name) => name.isMain)?.language ?? null;
},
```

---

**File**: `api/services/mapping/converters.js`

**Function**: `toOrganization`

**Specific Changes**:
2. **Add fallback for `language` derivation**: Change the inline language expression to fall back to `source.language` when `source.names` is not an Array.

```js
language: Array.isArray(source.names)
  ? source.names.find((name) => name.isMain)?.language
  : source.language,
```

3. **Add `language` to `toDocument` output**: Add a `language` field mapping that reads from `source.language` for search results. This field is separate from the existing `mainLanguage` / `languages` fields which serve the database path.

---

**File**: `api/dbSync/entities/document.js`

**Specific Changes**:
4. **Add `language` field to Typesense schema**: Add `{ name: 'language', type: 'string', facet: true, optional: true, sort: true }` to the document schema's `fields` array.

5. **Extract primary language in sync query**: Join `j_document_language` to extract the first language ID as the `language` field. The join already exists for the `languages` array; the `importFormater` needs to extract the first element as `language`.

6. **Update `importFormater`**: Set `d.language = d.languages?.[0] ?? undefined` so the flat `language` field is populated from the first language in the array before the document is indexed.

## Testing Strategy

### Validation Approach

**Goal**: Surface counterexamples that demonstrate the bug BEFORE implementing the fix. Confirm the root cause analysis by observing failures on unfixed code.

**Test Plan**: Write unit tests that call `getMainLanguage()`, `toOrganization()`, and `toDocument()` with Typesense-shaped source objects (flat `language` string, no `names` Array). Run on unfixed code to observe `null`/`undefined`/missing values.

**Test Cases**:
1. **getMainLanguage with flat source**: Call `getMainLanguage({ language: "fra" })` — expect `null` on unfixed code (will fail assertion for `"fra"`)
2. **toOrganization with flat source**: Call `toOrganization({ id: "1", name: "Test", language: "fra" })` — expect `language: undefined` on unfixed code
3. **toDocument with flat source**: Call `toDocument({ id: "1", title: "Test", language: "eng" })` — expect no `language` field on unfixed code
4. **toSimpleCave with flat source**: Call `toSimpleCave({ id: "1", name: "Cave", language: "ita" })` — expect `language: null` on unfixed code

**Expected Counterexamples**:
- `getMainLanguage({ language: "fra" })` returns `null` instead of `"fra"`
- `toOrganization` returns `language: undefined` for Typesense-shaped input
- Confirmed cause: no fallback to `source.language` when `source.names` is absent

### Fix Checking

**Goal**: Verify that for all inputs where the bug condition holds, the fixed functions produce the expected behavior.

**Pseudocode:**
```
FOR ALL source WHERE NOT (source.names instanceof Array) AND typeof source.language === 'string' DO
  result := getMainLanguage_fixed(source)
  ASSERT result === source.language
END FOR

FOR ALL source WHERE NOT Array.isArray(source.names) AND typeof source.language === 'string' DO
  result := toOrganization_fixed(source)
  ASSERT result.language === source.language
END FOR
```

### Preservation Checking

**Goal**: Verify that for all inputs where the bug condition does NOT hold, the fixed functions produce the same result as the original functions.

**Pseudocode:**
```
FOR ALL source WHERE source.names instanceof Array DO
  ASSERT getMainLanguage_original(source) === getMainLanguage_fixed(source)
END FOR

FOR ALL source WHERE Array.isArray(source.names) DO
  ASSERT toOrganization_original(source).language === toOrganization_fixed(source).language
  ASSERT toOrganization_original(source).nameId === toOrganization_fixed(source).nameId
END FOR
```

**Testing Approach**: Property-based testing with fast-check is recommended for preservation checking because:
- It generates many random `names` Arrays with varying structures (empty arrays, arrays with no main name, arrays with main name)
- It catches edge cases like `names: []` or `names: [{ isMain: false }]`
- It provides strong guarantees that the database path is unchanged

**Test Plan**: Observe behavior on UNFIXED code first for database-shaped inputs, then write property-based tests capturing that behavior.

**Test Cases**:
1. **getMainLanguage preservation**: Generate random `names` Arrays and verify fixed function returns same result as original
2. **toOrganization preservation**: Generate random organization sources with `names` Arrays and verify `language` and `nameId` are identical
3. **toDocument preservation**: Verify `mainLanguage` and `languages` fields are unchanged for database-shaped document sources

### Unit Tests

- Test `getMainLanguage` with Typesense-shaped input (flat `language` string)
- Test `getMainLanguage` with database-shaped input (`names` Array with main name)
- Test `getMainLanguage` with edge cases: no `language` field, empty `names` Array, `names` Array with no main entry
- Test `toOrganization` with Typesense-shaped input
- Test `toOrganization` with database-shaped input
- Test `toDocument` with Typesense-shaped input including `language`
- Test document `importFormater` extracts first language correctly

### Property-Based Tests

- Generate random ISO 639-3 language codes and verify `getMainLanguage` returns them from flat sources
- Generate random `names` Arrays (with and without main entries) and verify preservation of original behavior
- Generate random organization sources in both shapes and verify correct language derivation
- Generate mixed inputs (some with `names` Array, some without) and verify the correct branch is taken

### Integration Tests

- Test advanced-search endpoint returns non-null `language` for entrance results
- Test advanced-search endpoint returns non-null `language` for cave results
- Test advanced-search endpoint returns non-null `language` for massif results
- Test advanced-search endpoint returns non-null `language` for organization results
- Test advanced-search endpoint returns `language` for document results (after schema update)
- Test individual entity endpoints continue returning correct `language` values

Paul-AUB (changes requested):
# Code Review — commit `0eafb8f2` `fix(search): return language field from Typesense hit documents in advanced-search`

## Context

When the search API returns results from Typesense, each hit is a flat document (e.g. `{ id, title, language: "fra", ... }`). The converters (`toSimpleCave`, `toSimpleOrganization`, `toOrganization`, `toDocument`) were originally written for DB-shaped objects that carry a `names: Array` from which the language is derived. Because the Typesense shape has no `names` array, `getMainLanguage` was returning `null` and the `language` field was lost in the API response.

The fix correctly adds a fallback to `source.language` throughout the converter chain.

---

## Issues

### 1. No-op `?? undefined` — [`api/dbSync/entities/document.js:205`](api/dbSync/entities/document.js#L205)

```js
d.language = d.languages?.[0] ?? undefined;
```

`d.languages?.[0]` already evaluates to `undefined` when the array is absent or empty. The `?? undefined` fallback has no effect (the nullish coalescing operator only substitutes for `null` or `undefined`, so it adds nothing here). The intent is unclear — if the goal is to keep `undefined` for an absent value, write:

```js
d.language = d.languages?.[0];
```

If the intent was to default to `null` (to make the absence explicit for serialization):

```js
d.language = d.languages?.[0] ?? null;
```

---

### 2. `language` added to `query_by` is questionable — [`api/dbSync/entities/document.js`](api/dbSync/entities/document.js#L313)

```js
query_by: 'title,description,language',
```

Language values are ISO 639-3 codes (`"fra"`, `"eng"`, `"deu"`, …). Including them in `query_by` means a full-text search query can accidentally match on these opaque codes. For example, a user searching for `"eng"` (as part of a word) could spuriously boost documents whose language is `"eng"`. These codes carry no semantic weight for users.

Language is correctly declared as a `facet`, which is exactly what enables filtering. It should not be in `query_by`. Recommendation: remove `language` from `query_by` and rely on facet-based filtering instead.

---

### 3. Dual response shape for `toDocument` — [`api/services/mapping/converters.js:246-248`](api/services/mapping/converters.js#L246)

After this fix, `toDocument` produces different shapes depending on the source:

| Source shape | `mainLanguage` | `languages` | `language` |
|---|---|---|---|
| DB (Waterline ORM) | `"fra"` | `["fra", "eng"]` | `undefined` |
| Typesense hit | `undefined` | `undefined` | `"fra"` |

This duality is silent — no comment documents it, and callers downstream must handle both cases without knowing which they received. At minimum, this should be documented with a comment. Ideally, `toDocument` could normalize the output so that `language` (or `mainLanguage`) is always populated regardless of source shape.

---

### 4. No test for `toSimpleOrganization` — [`test/integration/1_services/SearchLanguageField.property.test.js`](test/integration/1_services/SearchLanguageField.property.test.js)

`toSimpleOrganization` also calls `getMainLanguage(source)` and was affected by the same bug. It is transitively covered because `getMainLanguage` is tested directly in `1a`, but the explicit contract for `toSimpleOrganization` with a Typesense source is absent. Adding a `1e` case (analogous to `1b`–`1d`) would complete the test matrix and protect against future regressions if `toSimpleOrganization` is ever refactored to not use `getMainLanguage`.

---

## Positive notes

- The core fix is minimal and correctly placed. Centralizing the fallback in `getMainLanguage` (`utils.js`) means all converters that call it benefit automatically — no need to touch each one individually (except `toOrganization`, which inlines the logic instead of calling `getMainLanguage`).
- Using `fast-check` for property-based testing is a good fit here: the language field must hold for _any_ valid ISO 639-3 code, not just a handful of hand-picked examples.
- The "Preservation" suite (Property 2) is well-structured and ensures the DB path is not regressed.
- The commit scope is tight — only the affected files are modified, with no unrelated cleanup.

---

## Summary

| Severity | Finding |
|---|---|
| Nit | `?? undefined` is a no-op in `importFormater` |
| Medium | `language` should not be in `query_by` (identifier, not searchable text) |
| Low | Dual response shape for `toDocument` is undocumented and fragile |
| Low | `toSimpleOrganization` lacks an explicit property test case |

ClemRz:
Thanks for the thorough review @Paul-AUB — all four items were legit. Here's what I addressed in the amended commit:

1. **No-op `?? undefined`** — Simplified to `d.language = d.languages?.[0];`. The nullish coalescing was indeed redundant.

2. **`language` in `query_by`** — Removed. ISO 639-3 codes are opaque identifiers, not user-searchable text. Facet-based filtering is the correct mechanism.

3. **Dual response shape for `toDocument`** — Normalized `language` so it's always populated regardless of source shape. For DB sources it now falls back through `source.language ?? source.mainLanguage?.id ?? source.languages?.[0]?.id ?? null`. Added an inline comment documenting the derivation. The `mainLanguage` and `languages` fields remain unchanged for backward compatibility.

4. **Missing `toSimpleOrganization` test** — Added test case `1e` covering `toSimpleOrganization` with a flat Typesense source. Also updated `2c` to assert that `language` is now populated for DB-shaped document sources too.

All 8 property tests pass, lint is clean.

Paul-AUB (approved):

PR #1529, opened 2026-04-04
fix(cave): can't remove isDiving from a cave

### LINKED ISSUE #1514: Can't remove `isDiving` from a cave
**Describe the bug**
Cave not updated when trying to remove `isDiving`

**To Reproduce**
PUT https://api.grottocenter.org/api/v1/caves/149381 with isDiving=false
Response is the object with isDiving=true

Subsequent querying GET https://api.grottocenter.org/api/v1/caves/149381 also gives isDiving=true

**Expected behavior**
Response is the object with isDiving=false

ClemRz:
## 🤔 What

Fix cave update endpoint silently ignoring falsy values like `isDiving=false`.

Closes #1514

## 🤷‍♂️ Why

The `PUT /api/v1/caves/:id` controller used JavaScript truthy checks (`if (value)`) to decide whether to include fields in the update payload. This meant any falsy value — `false`, `0`, `""` — was silently dropped and never persisted.

The reported symptom was that setting `isDiving=false` had no effect, but the same bug also affected `depth`, `length`, `temperature`, `latitude`, and `longitude` (setting any of them to `0` would be ignored).

## 🔍 How

Replaced all six truthy checks in `api/controllers/v1/cave/update.js` with `!== undefined` checks, which correctly distinguish "parameter not provided" from "parameter provided with a falsy value."

```diff
- if (newLatitude) updatedFields.latitude = newLatitude;
+ if (newLatitude !== undefined) updatedFields.latitude = newLatitude;
```

Same pattern applied to `longitude`, `depth`, `length`, `temperature`, and `isDiving`.

## 🧪 Testing

All 1864 existing tests pass. To verify manually:

```
PUT /api/v1/caves/:id  { "isDiving": false }
# Response should now reflect isDiving: false
```

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/api/controllers/v1/cave/update.js
+++ b/api/controllers/v1/cave/update.js
@@ -25,12 +25,12 @@ module.exports = async (req, res) => {
     // dateReviewed will be updated automaticly by the SQL historisation trigger
   };
 
-  if (newLatitude) updatedFields.latitude = newLatitude;
-  if (newLongitude) updatedFields.longitude = newLongitude;
-  if (newDepth) updatedFields.depth = newDepth;
-  if (newLength) updatedFields.caveLength = newLength;
-  if (newTemperature) updatedFields.temperature = newTemperature;
-  if (newIsDiving) updatedFields.isDiving = newIsDiving;
+  if (newLatitude != null) updatedFields.latitude = newLatitude;
+  if (newLongitude != null) updatedFields.longitude = newLongitude;
+  if (newDepth != null) updatedFields.depth = newDepth;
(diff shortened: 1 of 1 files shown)


Paul-AUB (dismissed):
Pre-approved, but please check what would happen in the DB if the value `null` is passed to the API at it would not more be filtered. It is suitable if it doesn't break the DB.

ClemRz:
Thank you @Paul-AUB for the comment. I updated the code to skip on both `null` and `undefined`.

PR #1531, opened 2026-04-06
fix(search): return 400 instead of 500 for unknown entity names

ClemRz:
## 🤔 What

- Fix a 500 crash in `POST /api/v1/field-search` when an unrecognized entity name is provided (e.g. `"cave"` instead of `"caves"`)
- Apply the same guard to `POST /api/v1/advanced-search` which had the same vulnerability

## 🤷‍♂️ Why

`SearchService.fieldSearch` and `collectionSearch` return `null` for entity names not present in the Typesense collection registry. Both controllers accessed properties on the null result (e.g. `r.found`) without a guard, causing a `TypeError: Cannot read properties of null (reading 'found')` and a 500 response.

Production log:
```
TypeError: Cannot read properties of null (reading 'found')
    at module.exports [as v1/search/field-search] (field-search.js:29:22)
```

Triggered by a request with `"entity": "cave"` — the valid name is `"caves"`.

## 🔍 How

Added a null check after the search service call in both controllers. When the result is null, the endpoint now returns a `400 Bad Request` with a message listing the valid entity names, e.g.:

```
Unknown entity "cave". Valid entities are: organizations, cavers, massifs, caves, entrances, documents
```

## 🧪 Testing

```bash
# Should return 400 with valid entity list
curl -X POST http://localhost:1337/api/v1/field-search \
  -d '{"entity":"cave","field":"name","query":"test"}'

# Should work normally
curl -X POST http://localhost:1337/api/v1/field-search \
  -d '{"entity":"caves","field":"name","query":"test"}'
```

### DIFF
--- a/api/controllers/v1/search/advanced-search.js
+++ b/api/controllers/v1/search/advanced-search.js
@@ -9,11 +9,13 @@ module.exports = async (req, res) => {
   let matchAllFields = req.param('matchAllFields') ?? true;
   if (!matchAllFields || matchAllFields === 'false') matchAllFields = false;
 
+  const entity = req.param('entity') ?? '';
+
   let results;
   try {
     results = await SearchService.collectionSearch({
       query: req.param('query'),
-      entity: req.param('entity') ?? '',
+      entity,
       sort: req.param('sort'),
       filter: req.param('filter') ?? {},
       isLogicalCompareAnd: !!matchAllFields,
@@ -23,6 +25,13 @@ module.exports = async (req, res) => {
   } catch (error) {
     if (handleTypesenseError(res, error)) return undefined;
     throw error;
+  }
+
+  if (!results) {
+    const validEntities = SearchService.allEntitiesKeys.join(', ');
+    return res.badRequest(
+      `Unknown entity "${entity}". Valid entities are: ${validEntities}`
+    );
   }
 
   return ControllerService.treatAndConvert(
--- a/api/controllers/v1/search/field-search.js
+++ b/api/controllers/v1/search/field-search.js
@@ -34,6 +34,14 @@ module.exports = async (req, res) => {
     throw error;
   }
 
+  if (!r) {
+    const validEntities = SearchService.allEntitiesKeys.join(', ');
+    res.badRequest(
+      `Unknown entity "${entity}". Valid entities are: ${validEntities}`
+    );
+    return;
+  }
+
   const out = {
     totalDistinct: r.found,
     totalDocuments: r.found_docs,


Paul-AUB (changes requested):
# Code Review — commit `3a93e2fe` (`origin/fix/search-null-entity-crash`)

**fix(search): return 400 instead of 500 for unknown entity names**

---

## Summary

The fix is correct in intent: guard against `null` returned by `SearchService.collectionSearch` / `fieldSearch` when an unrecognized entity name is supplied, and respond 400 instead of crashing with a 500.

---

## Issues Found

### 🔴 Critical — `SearchService.allEntitiesKeys` is not exported

**File:** `api/controllers/v1/search/advanced-search.js:29` and `api/controllers/v1/search/field-search.js:36`

```js
const validEntities = SearchService.allEntitiesKeys.join(', ');
```

`allEntitiesKeys` is a **module-level constant** in `SearchService.js` (line 18) but is **never added to `module.exports`**. Accessing it via `SearchService.allEntitiesKeys` returns `undefined`, so calling `.join(', ')` on it will throw:

```
TypeError: Cannot read properties of undefined (reading 'join')
```

This means the null-guard itself would trigger a new 500 — the very error it was designed to prevent.

**Fix:** either export `allEntitiesKeys` from `SearchService.js`:
```js
module.exports = {
  ...
  allEntitiesKeys,
};
```
or inline the valid entities list in the controllers without depending on the private constant.

---

### 🟡 Minor — `field-search.js` null-guard placed after `collectionSearch` already exits but before its own `try/catch`

Looking at the current `field-search.js`, the `fieldSearch` call (line 19) has **no try/catch** wrapping it, unlike `advanced-search.js`. If `fieldSearch` throws (network error, Typesense down, etc.) it will bubble up unhandled. This pre-existed the commit but the added null-check block sits in code that has this gap — worth fixing in the same pass.

---

### 🟢 Positive points

- Extracting `entity` to a variable before the `try` block (`advanced-search.js`) so it remains in scope for the null-check afterward is clean and correct.
- The error message format (`Unknown entity "X". Valid entities are: …`) is user-friendly and actionable.
- Commit message is clear, explains root cause, and maps well to the diff.

---

## Verdict

**Needs fix before merge.** The critical missing export of `allEntitiesKeys` will cause the error handler to throw instead of responding, leaving the 500 regression in place.

ClemRz (commented):
## Review Response

### 🔴 `allEntitiesKeys` not exported — false positive

`allEntitiesKeys` **is** exported. The bottom of `SearchService.js` reads:

```js
module.exports = {
  allEntities,
  allEntitiesKeys,
  // ...
};
```

It was already present in `module.exports` on `develop` before this PR. `SearchService.allEntitiesKeys` resolves correctly.

### 🟡 `field-search.js` missing try/catch — false positive

The `fieldSearch` call was already wrapped in a try/catch before this PR (you can verify on `develop`). The null-guard was added **after** the existing try/catch block, which is the correct placement.

Both controllers already had proper error handling; this PR only added the null-check for the return value, which is the minimal fix needed.

Paul-AUB (approved):
My bad, I used a stale file during the CR. All good

PR #1532, opened 2026-04-06
chore(deps): upgrade all dependencies to latest

ClemRz:
## 🤔 What

Upgrade all project dependencies to their latest versions.

## 🤷‍♂️ Why

Keep dependencies current for security patches, bug fixes, and performance improvements. Reduces the gap between current and latest versions to make future upgrades easier.

### Semver-compatible updates (via `npm update`)
- `@aws-sdk/client-ses` 3.1000.0 → 3.1025.0
- `@commitlint/cli` + `config-conventional` 20.4.2 → 20.5.0
- `@natlibfi/marc-record-serializers` 11.0.1 → 11.0.2
- `express-rate-limit` 8.2.1 → 8.3.2
- `fast-check` 4.5.3 → 4.6.0
- `lint-staged` 16.3.1 → 16.4.0
- `multer` 2.1.0 → 2.1.1
- `nyc` 17.1.0 → 18.0.0
- `sinon` 21.0.1 → 21.0.3
- `typesense` 3.0.1 → 3.0.5

### Major version bump
- `ejs` 4.0.1 → 5.0.1 — only used directly in `api/helpers/send-email.js` via `ejs.renderFile()`. Sails keeps its own internal ejs@3. The `renderFile` API is unchanged.

### Overrides cleanup
- Removed `fast-xml-parser` override — AWS SDK and Azure SDK now both resolve to 5.5.8+ natively, so the override was no longer forcing anything.
- Kept `diff`, `qs`, `validator`, and `serialize-javascript` overrides (still needed to patch vulnerable transitive versions).

### Test fix
- Fixed a flaky property-based test (`BlacklistService - Property 3: UPSERT idempotence`) that failed intermittently due to Node.js/PostgreSQL clock skew when comparing timestamps captured before `revoke()` with the DB's `NOW()`.

### Intentionally skipped
- ESLint stays at v8 — `eslint-config-airbnb-base` does not yet support ESLint's flat config (v9+).

## 🧪 Testing

Standard test suite: `PORT=1338 npm run test`

1960 tests passing. No regressions from the upgrade.

## 📸 Previews

N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "@natlibfi/marc-record-serializers": "^11.0.1",
         "archiver": "^7.0.1",
         "argon2": "^0.44.0",
-        "ejs": "^4.0.1",
+        "ejs": "^5.0.1",
         "express-rate-limit": "^8.2.1",
         "jsonwebtoken": "^9.0.3",
         "mime-types": "^3.0.2",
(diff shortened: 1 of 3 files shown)


Paul-AUB (approved):
LGTM

Issue #1449, opened 2026-02-11
Signup email verification


ClemRz:
# Requirements Document

## Introduction

The Grottocenter API signup flow (`POST /api/v1/signup`) currently creates accounts without verifying that the user owns or can access the provided email address. This leads to accounts created with typos or emails the user does not control. This feature adds email verification to the signup flow by leveraging the existing but unused `activated`, `activationCode`, and `mailIsValid` fields on the TCaver model. After signup, the user receives a verification email containing a unique link. The account remains inactive until the user confirms their email by visiting that link.

Additionally, the system must handle legacy users already present in the database whose accounts predate the verification flow. A SQL migration marks all existing users as inactive with valid email, and the resend-verification endpoint becomes the entry point for legacy users to activate.

## Glossary

- **Signup_Controller**: The Sails action at `api/controllers/v1/auth/sign-up.js` that handles `POST /api/v1/signup` requests
- **Verification_Controller**: The new Sails action that handles `GET /api/v1/verify-email` requests to confirm email ownership
- **Resend_Controller**: The new Sails action that handles `POST /api/v1/resend-verification` requests to resend the verification email
- **Auth_Service**: The service at `api/services/AuthService.js` responsible for password hashing and authentication
- **Email_Helper**: The Sails helper at `api/helpers/send-email.js` that sends emails via AWS SES using EJS templates
- **TCaver**: The Waterline model representing a user account, including `activated`, `activationCode`, and `mailIsValid` fields
- **Activation_Code**: A cryptographically random token stored in `TCaver.activationCode` and sent to the user via email for verification
- **Verification_Email**: An email sent to the user after signup containing a link with the Activation_Code
- **Legacy_Migration**: A SQL migration script that sets `activated = false` for all existing TCaver records in the database, preserving `mailIsValid = true`
- **Forgot_Password_Controller**: The Sails action at `api/controllers/v1/account/forgot-password.js` that handles `POST /api/v1/forgotPassword` requests to initiate password reset

### Requirement 1: Generate and Store Activation Code on Signup

**User Story:** As a new user, I want the system to generate a unique verification token when I sign up, so that my email ownership can be verified.

#### Acceptance Criteria

1. WHEN a valid signup request is received, THE Signup_Controller SHALL generate a cryptographically random Activation_Code of 64 hexadecimal characters
2. WHEN a valid signup request is received, THE Signup_Controller SHALL store the Activation_Code in the `activationCode` field of the new TCaver record
[...]

PR #1534, opened 2026-04-16
1449 signup email verification

dawoldo:
## 🤔 What

Email verification post signup.

- Generate and store activation code on signup.
- Send Verification email on signup with a activation link.
- Block Login and password reset endpoints for unverified accounts.
- Legacy user migration script.
- A new verification email view.

[The issue](https://github.com/GrottoCenter/grottocenter-api/issues/1449#issue-3928866877)

## 🤷‍♂️ Why

The Grottocenter API signup flow (POST /api/v1/signup) currently creates accounts without verifying that the user owns or can access the provided email address. This leads to accounts created with typos or emails the user does not control. This feature adds email verification to the signup flow by leveraging the existing but unused activated, activationCode, and mailIsValid fields on the TCaver model. After signup, the user receives a verification email containing a unique link. The account remains inactive until the user confirms their email by visiting that link.

Additionally, the system must handle legacy users already present in the database whose accounts predate the verification flow. A SQL migration will mark  all existing users as inactive with valid email, and the resend-verification endpoint becomes the entry point for legacy users to activate.

## 🔍 How

- Added ```verify-email``` and ```resend-verification-email``` controllers and their respective endpoints.
- ```sendVerificationEmail()``` and ```generateActivationCode``` methods added to the AuthService to respect SOLID principles (since both the signup, resend-verification controllers may need to use that code block).
- Updated ```tcaver.json``` fixtures to ensure mock users are activated by default so the existing test suite doesn't fail on standard endpoints.

## 🧪 Testing

- Provide a valid email during signup; verify the activationCode is saved to the database and the email is dispatched.
- Attempt to log into an unverified account: ensure it stops the flow and returns the expected 401 Unauthorized.
- Attempt to trigger a password reset on an unverified account: ensure it is successfully blocked.
- Click the activation link from the email: ensure the database updates ```activated = true``` and ```activationCode = ''```.

### Deployment

This migration handles the transition of legacy users to the new verification flow by deactivating their status while temporarily marking their existing emails as valid.

#### Pre-Flight Check

Run this query to estimate the impact and ensure the ```backup_user_states``` table doesn't already exist from a failed previous attempt:

```SELECT count(*) FROM t_caver WHERE activated = true OR mail_is_valid = false;```

#### Execution

Run the migration script ```4_02_2026_04_16_legacy_user_migration.sql```. It is recommended to run this within a transaction block:
```
BEGIN;
-- [Content of 4_02_2026_04_16_legacy_user_migration.sql]
COMMIT;
```

#### Verification

Confirm that the number of rows in ```backup_user_states``` matches the count from Step 1:
```SELECT count(*) FROM backup_user_states;```

### Rollback procedure

In the event that legacy users are unable to access the system or the verification flow triggers unexpected errors, follow these steps to restore original user states.

- Execute the following SQL to join the backup table back to the main table:

```BEGIN;

UPDATE t_caver t
SET activated = b.activated,
    mail_is_valid = b.mail_is_valid
FROM backup_user_states b
WHERE t.id = b.id;

COMMIT;
```

- Once you have verified that users have regained their original status, drop the temporary backup table:
```DROP TABLE backup_user_states; ```

### ⚠️ Important Notes

- This rollback will overwrite any manual activations performed by users between the deployment and the rollback.
- The ```backup_user_states``` table will persist in the grottoce database until manually dropped.

## 📸 Previews

N/A

### DIFF
--- a/api/controllers/v1/account/forgot-password.js
+++ b/api/controllers/v1/account/forgot-password.js
@@ -26,6 +26,15 @@ module.exports = async (req, res) => {
     return res.ok();
   }
 
+  // instructs the user to verify their email before asking for a password reset
+  if (!userFound.activated) {
+    return res.unauthorized({
+      status: 'NotVerified',
+      message:
+        'Your account is not verified yet. Please check your email for the verification link.',
+    });
+  }
+
   // Generate reset password token
   const token = TokenService.issue(
     {
--- a/api/controllers/v1/auth/resend-verification-email.js
+++ b/api/controllers/v1/auth/resend-verification-email.js
@@ -1,0 +1,52 @@
+const util = require('util');
+const AuthService = require('../../../services/AuthService');
+
+const setTimeoutP = util.promisify(setTimeout);
+
+module.exports = async (req, res) => {
+  await setTimeoutP(500); // Basic brute force prevention
+
+  const email = req.param('email');
+
+  if (!email) {
+    return res.badRequest('You must provide an email address.');
+  }
+
+  const user = await TCaver.findOne({ mail: email.toLowerCase() });
+
+  if (!user) {
+    return res.ok();
+  }
+
+  // If the user is already activated, return success without sending an email
+  if (user.activated) {
+    return res.ok();
+  }
+
+  if (!user.mailIsValid) {
+    return res.badRequest(
+      'Your email address is marked as invalid. Please update your email address or contact us.'
+    );
+  }
+
+  const activationCode = AuthService.generateActivationCode();
+
+  try {
+    await TCaver.updateOne({ id: user.id }).set({ activationCode });
+  } catch (err) {
+    sails.log.error(
+      `Failed to update activation code for user ${user.id}:`,
+      err
+    );
+    return res.serverError('An error occurred while updating your account.');
+  }
+
+  try {
+    // Attempt to send the verification email
+    await AuthService.sendVerificationEmail(user, activationCode, req.i18n);
+  } catch (err) {
+    // Errors are already logged in AuthService.
+  }
+
+  return res.ok();
+};
--- a/api/controllers/v1/auth/sign-up.js
+++ b/api/controllers/v1/auth/sign-up.js
@@ -36,6 +36,9 @@ module.exports = async (req, res) => {
     return res.conflict('Email or nickname is already used.');
   }
 
+  // Only generate the activation code after every other check passes
+  const activationCode = AuthService.generateActivationCode();
+
   try {
     // Rely on the ORM for the rest of the input validation
     const newCaver = await TCaver.create({
@@ -45,10 +48,24 @@ module.exports = async (req, res) => {
       name: req.param('name') === '' ? null : req.param('name'),
       nickname,
       password: await AuthService.createHashedPassword(password),
+      activated: false,
+      mailIsValid: false,
+      activationCode,
       surname: req.param('surname') === '' ? null : req.param('surname'),
     }).fetch();
 
     await CaverService.updateInSearch(newCaver);
+
+    try {
+      await AuthService.sendVerificationEmail(
+        newCaver,
+        activationCode,
+        req.i18n
+      );
+    } catch (_) {
+      // Errors are already logged in AuthService.
+      // We catch them here to ensure the signup itself isn't considered a failure if sending the email fails.
+    }
   } catch (_) {
     return res.badRequest();
   }
--- a/api/controllers/v1/auth/verify-email.js
+++ b/api/controllers/v1/auth/verify-email.js
@@ -1,0 +1,37 @@
+module.exports = async (req, res) => {
+  const token = req.param('token');
+
+  if (!token) {
+    return res.badRequest('You must provide an activation token.');
+  }
+
+  const caver = await TCaver.findOne({ activationCode: token });
+
+  if (!caver) {
+    return res.notFound(
+      'Activation token is invalid or has already been used.'
+    );
+  }
+
+  const updates = {
+    activationCode: null,
+    mailIsValid: true,
+  };
+
+  if (!caver.activated) {
+    updates.activated = true;
+  }
+
+  try {
+    await TCaver.updateOne({ id: caver.id }).set(updates);
+  } catch (err) {
+    sails.log.error(`Failed to activate user ${caver.id}:`, err);
+    return res.serverError('An error occurred during account activation.');
+  }
+
+  if (caver.activated) {
+    return res.ok({ message: 'Account was already verified.' });
+  }
+
+  return res.ok({ message: 'Account successfully verified.' });
+};
--- a/api/services/AuthService.js
+++ b/api/services/AuthService.js
@@ -1,5 +1,6 @@
 const argon2 = require('argon2');
 const util = require('util');
+const crypto = require('crypto');
 
 const setTimeoutP = util.promisify(setTimeout);
 
@@ -16,6 +17,8 @@ const authenticateResult = {
   SUCCESS: 'SUCCESS',
   MISMATCH: 'MISMATCH',
   MUST_RESET: 'MUST_RESET',
+  NOT_VERIFIED: 'NOT_VERIFIED',
+  BANNED: 'BANNED',
 };
 module.exports.authenticateResult = authenticateResult;
 
@@ -34,12 +37,57 @@ async function authenticate(email, password) {
   );
 
   if (!user) return { status: authenticateResult.MISMATCH };
+
   if (!user.password?.startsWith('$argon2'))
-    return { status: authenticateResult.MUST_RESET };
+    return { status: authenticateResult.MUST_RESET, user };
 
   const isHashMatch = await argon2.verify(user.password, password);
   if (!isHashMatch) return { status: authenticateResult.MISMATCH };
 
+  if (user.banned) return { status: authenticateResult.BANNED, user };
+
+  if (!user.activated) return { status: authenticateResult.NOT_VERIFIED, user };
+
   return { status: authenticateResult.SUCCESS, user };
 }
 module.exports.authenticate = authenticate;
+
+/**
+ * Generate a cryptographically random activation code
+ * @returns {String}
+ */
+function generateActivationCode() {
+  return crypto.randomBytes(32).toString('hex');
+}
+module.exports.generateActivationCode = generateActivationCode;
+
+/**
+ * Send a verification email to a user
+ * @param {Object} user
+ * @param {String} token
+ * @param {Object} i18n
+ */
+async function sendVerificationEmail(user, token, i18n) {
+  const verifyLink = `${sails.config.custom.baseUrl}/ui/verify-email?token=${token}`;
+
+  try {
+    await sails.helpers.sendEmail.with({
+      allowResponse: false,
+      emailSubject: 'Verify your email address',
+      i18n,
+      recipientEmail: user.mail,
+      viewName: 'verifyEmail',
+      viewValues: {
+        recipientName: user.nickname,
+        verifyLink,
+      },
+    });
+  } catch (err) {
+    sails.log.error(
+      `Failed to send verification email for user ${user.nickname} (${user.mail}):`,
+      err
+    );
+    throw err;
+  }
+}
+module.exports.sendVerificationEmail = sendVerificationEmail;
--- a/config/locales/es.json
+++ b/config/locales/es.json
@@ -32,5 +32,9 @@
 	"validated": "validado",
 	"We received a request to reset the password for your account.": "Recibimos una solicitud para restablecer la contraseña de su cuenta.",
 	"You are receiving this email because of your subscription to the country %s.": "Está recibiendo este correo debido a su suscripción al país %s.",
-	"You are receiving this email because of your subscription to the massif %s.": "Está recibiendo este correo debido a su suscripción al macizo %s."
+	"You are receiving this email because of your subscription to the massif %s.": "Está recibiendo este correo debido a su suscripción al macizo %s.",
+	"Thank you for creating your Grottocenter account.": "Gracias por crear su cuenta en Grottocenter.",
+    "Please verify your email address by clicking the link below.": "Por favor, verifica su dirección de correo electrónico haciendo clic en el enlace de abajo.",
+    "If you didn't create an account, you can ignore this email.": "Si no ha creado una cuenta, puede ignorar este mensaje.",
+    "Verify my email address": "Verificar mi correo electrónico"
 }
--- a/config/locales/fr.json
+++ b/config/locales/fr.json
@@ -32,5 +32,10 @@
 	"validated": "validé(e)",
 	"We received a request to reset the password for your account.": "Nous avons reçu une demande de réinitialisation du mot de passe de votre compte.",
 	"You are receiving this email because of your subscription to the country %s.": "Vous recevez cet email car vous êtes abonné au pays %s.",
-	"You are receiving this email because of your subscription to the massif %s.": "Vous recevez cet email car vous êtes abonné au massif %s."
+	"You are receiving this email because of your subscription to the massif %s.": "Vous recevez cet email car vous êtes abonné au massif %s.",
+	"Thank you for creating your Grottocenter account.": "Merci d'avoir créé votre compte Grottocenter.",
+	"Please verify your email address by clicking the link below.": "Veuillez vérifier votre adresse mail en cliquant sur le lien ci-dessous.",
+	"If you didn't create an account, you can ignore this email.": "Si vous n'avez pas créé de compte, vous pouvez ignorer cet mail.",
+	"Verify my email address": "Vérifier mon adresse mail",
+	"Verify your email address": "Verify your email address"
 }
--- a/sql/4_02_2026_04_16_legacy_user_migration.sql
+++ b/sql/4_02_2026_04_16_legacy_user_migration.sql
@@ -1,0 +1,9 @@
+\c grottoce;
+
+-- Marks all legacy users' as emails as inactive (so they undergo the new verification flow) 
+-- while ensuring their existing emails are treated as valid initially.
+
+UPDATE t_caver
+SET activated = false,
+    mail_is_valid = true
+WHERE activated = true OR mail_is_valid = false;
--- a/api/controllers/v1/auth/login.js
+++ b/api/controllers/v1/auth/login.js
@@ -13,6 +13,13 @@ module.exports = async (req, res) => {
   }
 
   const result = await AuthService.authenticate(email, password);
+
+  if (result.status === AuthService.authenticateResult.MISMATCH) {
+    return res.unauthorized({
+      status: 'Mismatch',
+      message: 'Invalid email or password.',
+    });
+  }
   if (result.status === AuthService.authenticateResult.MUST_RESET) {
     return res.unauthorized({
       status: 'MustReset',
@@ -20,30 +27,30 @@ module.exports = async (req, res) => {
     });
   }
 
-  if (result.status !== AuthService.authenticateResult.SUCCESS) {
+  if (result.status === AuthService.authenticateResult.NOT_VERIFIED) {
+    return res.unauthorized({
+      status: 'NotVerified',
+      message:
+        'Your account is not verified yet. Please check your email for the verification link.',
+    });
+  }
+
+  if (result.status === AuthService.authenticateResult.BANNED) {
+    // Update login metadata for a banned connection attempt (unit tests require it)
+    CommonService.query(
+      `UPDATE t_caver SET date_last_connection = NOW(), connection_counter = connection_counter + 1 WHERE id = $1`,
+      [result.user.id]
+    ).catch((e) =>
+      sails.log.error('Failed to update login metadata:', e.message)
+    );
+
     return res.unauthorized({
       status: 'Mismatch',
       message: 'Invalid email or password.',
     });
   }
 
-  // Update login metadata (fire-and-forget)
-  CommonService.query(
-    `UPDATE t_caver SET date_last_connection = NOW(), connection_counter = connection_counter + 1 WHERE id = $1`,
-    [result.user.id]
-  ).catch((e) =>
-    sails.log.error('Failed to update login metadata:', e.message)
-  );
-
-  // Reject banned cavers — return generic 401 to hide ban status
-  if (result.user.banned) {
-    sails.log.warn(`Banned caver ${result.user.id} attempted login`);
-    return res.unauthorized({
-      status: 'Mismatch',
-      message: 'Invalid email or password.',
-    });
-  }
-
+  // If we get to this code block, authenticateResult is necessarily SUCCESS
   const token = TokenService.issue(
     {
       id: result.user.id,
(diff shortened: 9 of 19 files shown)

### CODE REVIEW

File: sql/4_02_2026_04_16_legacy_user_migration.sql
Comment (ClemRz) [inline sql/4_02_2026_04_16_legacy_user_migration.sql:1]: When I wipe out the local env to force the execution of the SQL scripts I get this error:

```
$ npm run dev:clean && docker logs grotto-postgres -f

...

/usr/local/bin/docker-entrypoint.sh: running /docker-entrypoint-initdb.d/4_2026_04_16_legacy_user_migration.sql
2026-04-16 13:25:23.132 UTC [170] ERROR:  relation "t_caver" does not exist at character 8
2026-04-16 13:25:23.132 UTC [170] STATEMENT:  UPDATE t_caver
        SET activated = false,
            mail_is_valid = true
        WHERE activated = true OR mail_is_valid = false;
psql:/docker-entrypoint-initdb.d/4_2026_04_16_legacy_user_migration.sql:7: ERROR:  relation "t_caver" does not exist
LINE 1: UPDATE t_caver
               ^
```

The error is because you need to select the db at the beginning of the script, you can look the other files for examples:
```
\c grottoce;
```

Also please rename the file with `4_02_2026_04_16_legacy_user_migration.sql‎` to give room for more scripts in the `4_` section.

File: config/locales/fr.json
Comment (ClemRz) [inline config/locales/fr.json:1]: Please could you also include translations for English and Spanish?

File: api/controllers/v1/account/forgot-password.js [hunk 1]
```diff
@@ -26,6 +26,14 @@ module.exports = async (req, res) => {
     return res.ok();
   }
 
+  // instructs the user to verify their email before asking for a password reset
+  if (!userFound.activated) {
+    return res.unauthorized({
```
ClemRz, line 31: In the login.js you also use `status: 'NotVerified',` and here you didn't, is there a reason?

api/controllers/v1/auth/resend-verification-email.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  // If the user is already activated, return success without sending an email
+  if (user.activated) {
+    return res.ok();
+  }
+
+  if (!user.mailIsValid) {
+    return res.badRequest(
+      'Your email address is marked as invalid. Please update your email address.'
```
ClemRz: `... Please update your email address or contact us.'`

api/controllers/v1/auth/verify-email.js
```diff
@@ -0,0 +1,30 @@
...
+  const token = req.param('token');
+
+  if (!token) {
+    return res.badRequest('You must provide an activation token.');
+  }
+
+  const caver = await TCaver.findOne({ activationCode: token });
+
+  if (!caver) {
+    return res.notFound('Activation code is invalid or has already been used.');
```
ClemRz: `'Activation code is ...'` let's keep the wording consistent: `'Activation token is ...'`

api/controllers/v1/auth/verify-email.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  if (caver.activated) {
+    return res.ok({ message: 'Account is already verified.' });
```
ClemRz: `'Account was already verified.'`

ClemRz (changes requested):

ClemRz (changes requested):
## Review against `signup-email-verification/requirements.md`

Good work overall — the core flow is solid and the code is clean. I'm flagging a few issues that range from spec compliance gaps to potential bugs.

### Requirements Coverage

| Requirement | Status | Notes |
|---|---|---|
| R1: Generate & store activation code on signup | ⚠️ Partial | Code is 64 hex chars ✅, `activated=false` ✅, `activationCode` stored ✅, but `mailIsValid` is **not set to `false`** — see inline comment |
| R2: Send verification email on signup | ✅ Pass | Email sent after creation, failure is caught and logged without failing signup |
| R3: Verify email via activation link | ✅ Pass | Sets `activated=true`, `mailIsValid=true`, clears code, handles already-verified and not-found cases |
| R4: Block login for unverified accounts | ✅ Pass | `NOT_VERIFIED` status returned with instructive message |
| R5: Resend verification email | ✅ Pass | Generates new code, sends email, doesn't reveal email existence, handles already-activated |
| R6: Legacy user migration | ⚠️ Issues | Missing `\c grottoce;` prefix, not idempotent — see inline comments |
| R7: Block password reset for unverified | ✅ Pass | Returns 401 with verification message, no reset email sent |

### Missing: Tests

There are **no integration tests** for any of the new endpoints (`verify-email`, `resend-verification-email`) or for the modified behavior in `login` and `forgot-password`. The project conventions require tests for all modifications.

### Missing: Swagger / OpenAPI

The two new endpoints (`POST /api/v1/verifyEmail`, `POST /api/v1/resendVerificationEmail`) are not documented in `assets/swaggerV1.yaml`. Per project conventions, the Swagger spec must be updated when adding endpoints.

### Route method mismatch with requirements

The requirements spec defines `GET /api/v1/verify-email` for the Verification_Controller. The PR implements `POST /api/v1/verifyEmail`. A GET would be more appropriate here since the user clicks a link from their email — browsers follow links with GET requests. Worth discussing whether this should be GET or POST, but either way the route naming uses camelCase (`verifyEmail`) while the controller file uses kebab-case (`verify-email`), which is fine for Sails but inconsistent with the spec's expectation.

### Minor issues

- `config/locales/fr.json` and `test/fixtures/tcaver.json` lost their trailing newlines (files should end with `\n`)
- The `sendVerificationEmail` function swallows its error after logging — the callers also wrap it in try/catch, so the outer catch blocks will never trigger. This is harmless but the comment "Errors are already logged in AuthService" in the callers is misleading since the function never throws.

api/controllers/v1/auth/sign-up.js
```diff
@@ -45,10 +48,24 @@ module.exports = async (req, res) => {
       name: req.param('name') === '' ? null : req.param('name'),
       nickname,
       password: await AuthService.createHashedPassword(password),
+      activated: false,
+      activationCode,
+      // mailIsValid: false, if we keep this property as false by default they can never verify later
```
ClemRz: **Requirement 1, AC 4** states: *"THE Signup_Controller SHALL set the `mailIsValid` field to `false` on the new TCaver record."*

The concern in the comment is valid — if `mailIsValid` defaults to `false` in the model, then it is technically being set. But the intent of the requirement is explicit: set it to `false` on signup, then flip it to `true` on verification. The current code relies on the model default, which works, but the commented-out line with the rationale comment is confusing. Either:
1. Explicitly set `mailIsValid: false` (matches the spec literally), or
2. Remove the commented-out line entirely if you are relying on the model default — the comment as-is reads like a TODO/concern rather than a deliberate decision.

sql/4_02_2026_04_16_legacy_user_migration.sql
```diff
@@ -0,0 +1,7 @@
+-- Marks all legacy users' as emails as inactive (so they undergo the new verification flow) 
+-- while ensuring their existing emails are treated as valid initially.
+
+UPDATE t_caver
+SET activated = false,
+    mail_is_valid = true
+WHERE activated = true OR mail_is_valid = false;
```
ClemRz, line 9: Please add a deployment procedure and a rollback procedure to the PR description.

api/controllers/v1/auth/verify-email.js
ClemRz, line 1: ### Timing-based token enumeration security flaw:

The `findOne({ activationCode: token })` lookup returns `notFound` for invalid tokens but `ok` for valid ones. An attacker could brute-force activation codes by measuring response differences. Since the codes are 64 hex chars (256 bits of entropy), this is not a practical attack vector, but worth noting.

More importantly: the `activationCode` column should have a database index if it does not already, since every verification request queries by it.

api/services/AuthService.js
```diff
@@ -40,6 +42,47 @@ async function authenticate(email, password) {
...
+      viewName: 'verifyEmail',
+      viewValues: {
+        recipientName: user.nickname,
+        verifyLink,
+      },
+    });
+  } catch (err) {
+    sails.log.error(
+      `Failed to send verification email for user ${user.nickname} (${user.mail}):`,
+      err
```
ClemRz, line 84: The function catches its own error and only logs it. This means the `try/catch` blocks wrapping calls to `sendVerificationEmail` in both `sign-up.js` and `resend-verification-email.js` will never catch anything from this function. The comments saying "Errors are already logged in AuthService" are accurate, but the outer try/catch is dead code for email errors. Consider re-throwing the error after logging so callers can react.

ClemRz (changes requested):
### Missing: Integration tests

No tests were added for:
- `POST /api/v1/verifyEmail` — happy path, invalid token, already-verified
- `POST /api/v1/resendVerificationEmail` — happy path, unknown email, already-activated
- `POST /api/v1/login` — unverified account returns 401 with `NotVerified` status
- `POST /api/v1/forgotPassword` — unverified account returns 401
- `POST /api/v1/signup` — new account has `activated=false` and `activationCode` set

The project requires tests for all modifications.

ClemRz (changes requested):
Good progress since the first round. The `\c grottoce;` prefix, the `mailIsValid: false` explicit set, the wording fixes, the `throw err` in `sendVerificationEmail`, the `NotVerified` status in `forgot-password`, the "or contact us" message, the DB index, and the file rename are all addressed. Tests and Swagger were added too. Below are the remaining issues I found.

### 1. Security: `NOT_VERIFIED` check happens after password verification in `authenticate()`

**File:** `api/services/AuthService.js`, line 47

The `activated` check runs *after* `argon2.verify`. This means an attacker can distinguish between "wrong password" (fast `MISMATCH`) and "correct password but not verified" (`NOT_VERIFIED`) by observing the response body. It leaks whether a password guess is correct for unverified accounts.

Move the `activated` check **before** the password check, right after the user is found:

```js
if (!user) return { status: authenticateResult.MISMATCH };
if (!user.activated) return { status: authenticateResult.NOT_VERIFIED, user };
if (!user.password?.startsWith('$argon2'))
  return { status: authenticateResult.MUST_RESET };
// ...argon2.verify...
```

This way an unverified account always gets `NOT_VERIFIED` regardless of the password, and the timing is consistent.

---

### 2. Security: login updates metadata and checks ban *after* returning `NOT_VERIFIED`

**File:** `api/controllers/v1/auth/login.js`, lines 24–31 vs 38–52

The `NOT_VERIFIED` early return on line 27 means:
- The `connection_counter` / `date_last_connection` update on line 38 never runs for unverified users — fine.
- But the **banned check** on line 47 also never runs. A banned + unverified user gets `NotVerified` instead of `Mismatch`, which leaks their ban status (or rather, hides it — they'd be told to verify instead of being silently rejected).

The banned check should happen before or alongside the verified check. Consider checking `result.user.banned` before `NOT_VERIFIED` in the controller, or handling it inside `authenticate()`.

---

### 3. Bug: `resend-verification-email` returns 200, Swagger documents 204, tests expect 204

**File:** `api/controllers/v1/auth/resend-verification-email.js`

The controller calls `res.ok()` without data in all success paths (lines 14, 19, 40). Per `api/responses/ok.js`, `res.ok()` without data returns **204**. That matches the tests and Swagger. However, `res.ok()` on line 14 (user not found) and line 19 (already activated) also return 204, which is correct for anti-enumeration.

But there's a subtle issue: on line 19, `res.ok()` is called without data → 204. On line 40 (after sending email), `res.ok()` is also called without data → 204. This is fine. **However**, if the `sendVerificationEmail` call throws (e.g., SES failure), the `catch` block on line 38 swallows the error silently and still returns 204. The user gets no indication the email wasn't sent. Consider logging a warning in the catch block — the comment says "Errors are already logged in AuthService" but the `updateOne` on line 33 could also fail, and *that* error is not logged anywhere.

Suggestion:
```js
} catch (err) {
  sails.log.error('Failed to resend verification email:', err);
}
```

---

### 4. `sign-up.js` still has a stale concern comment

**File:** `api/controllers/v1/auth/sign-up.js`, line 48

```js
// concern: if we set the isMailValid property as false on signup by default said user can never verify later.
```

This comment is now stale — `mailIsValid` is explicitly set to `false` on the next line, and the `verify-email` controller sets it to `true` on verification. The concern has been resolved. Remove the comment to avoid confusion.

---

### 5. `mailIsValid` model default vs signup intent mismatch

**File:** `api/models/TCaver.js`, line 103

The model `defaultsTo: true` for `mailIsValid`. The signup controller explicitly sets `mailIsValid: false`. This works, but it means any *other* code path that creates a TCaver without explicitly setting `mailIsValid` will default to `true` — which is the wrong default for the new verification flow. Consider changing the model default to `false` so the safe default applies everywhere. This would require checking that the legacy migration still works (it explicitly sets `mail_is_valid = true`, so it would be unaffected).

---

### 6. Swagger: `verify-email` and `resend-verification-email` paths missing `/api/v1` prefix

**File:** `assets/swaggerV1.yaml`, lines 3257 and 3289

The paths are defined as `'/verify-email'` and `'/resend-verification-email'`, but the actual routes are `/api/v1/verify-email` and `/api/v1/resend-verification-email`. Check whether the Swagger spec has a `servers` or `basePath` that prepends `/api/v1` — if it does, this is fine. If not, the paths won't match the actual API.

---

### 7. Swagger: `verify-email` documents token as query param, but controller reads from body

**File:** `assets/swaggerV1.yaml` vs `api/controllers/v1/auth/verify-email.js`

The Swagger spec says `in: query` for the `token` parameter. The controller uses `req.param('token')`, which in Sails reads from body, query, or route params. The tests send it in the body (`.send({ token: '...' })`). Since this endpoint is meant to be triggered from an email link (click), the token would naturally come as a query parameter. The implementation supports both, but the Swagger and tests should be consistent. Same issue for `resend-verification-email` — Swagger says `in: query` for `email`, tests send it in the body.

---

### 8. `verify-email` should clear `activationCode` to `null`, not empty string

**File:** `api/controllers/v1/auth/verify-email.js`, line 24

```js
activationCode: '', // We clear the code from the db
```

The model defines `activationCode` as `allowNull: true`. Setting it to an empty string means `findOne({ activationCode: '' })` would match all verified users. This is not exploitable because the controller also checks `caver.activated` and returns early, but it's cleaner to set it to `null`:

```js
activationCode: null,
```

This also prevents the edge case where someone sends an empty string as a token — `findOne({ activationCode: '' })` would match a verified user.

---

### 9. Email subject is not translated

**File:** `api/services/AuthService.js`, line 77

```js
emailSubject: 'Verify your email address',
```

The `send-email` helper wraps the subject with `i18n.__()` (line 101 of `send-email.js`: `Data: \`Grottocenter - ${i18n.__(emailSubject)}\``), so this string *will* be translated if a translation key exists. But the key `"Verify your email address"` is not present in any of the locale files (`en.json`, `fr.json`, `es.json`). Add it, or the subject will fall back to the English string for all locales.

---

### 10. Test: `verify-email` happy path doesn't assert the DB state

**File:** `test/integration/4_routes/Auth.test.js`, lines 175–180

The "should return 200 for valid token" test only checks the status code. It should also verify that the user's `activated` is now `true` and `activationCode` is cleared:

```js
it('should return 200 for valid token and activate the user', async () => {
  await supertest(sails.hooks.http.app)
    .post('/api/v1/verify-email')
    .send({ token: 'valid_activation_code' })
    .expect(200);

  const user = await TCaver.findOne({ mail: 'verify_email@test.com' });
  should(user.activated).be.true();
  should(user.mailIsValid).be.true();
  should(user.activationCode).be.oneOf([null, '']);
});
```

---

### 11. Test: no test for signup setting `activated: false` and `activationCode`

The previous review flagged this. The signup tests verify status codes but don't assert that the created user has `activated: false` and a non-empty `activationCode`. Add an assertion in the "Minimal data" or "Complete data" test:

```js
const created = await TCaver.findOne({ mail: 'newtest@newtest.com' });
should(created.activated).be.false();
should(created.activationCode).be.a.String().and.not.empty();
```

config/locales/es.json
```diff
@@ -32,5 +32,9 @@
 	"validated": "validado",
 	"We received a request to reset the password for your account.": "Recibimos una solicitud para restablecer la contraseña de su cuenta.",
 	"You are receiving this email because of your subscription to the country %s.": "Está recibiendo este correo debido a su suscripción al país %s.",
-	"You are receiving this email because of your subscription to the massif %s.": "Está recibiendo este correo debido a su suscripción al macizo %s."
+	"You are receiving this email because of your subscription to the massif %s.": "Está recibiendo este correo debido a su suscripción al macizo %s.",
+	"Thank you for creating your Grottocenter account.": "Gracias por crear tu cuenta en Grottocenter.",
```
ClemRz, line 36: `tu` -> `su`

config/locales/es.json
```diff
@@ -32,5 +32,9 @@
...
+    "Please verify your email address by clicking the link below.": "Por favor, verifica tu dirección de correo electrónico haciendo clic en el enlace de abajo.",
```
ClemRz, line 37: `verifica tu` -> `verifique su`

config/locales/es.json
```diff
@@ -32,5 +32,9 @@
...
+    "If you didn't create an account, you can ignore this email.": "Si no has creado una cuenta, puedes ignorar este correo electrónico.",
```
ClemRz, line 38: `Si no ha creado una cuenta, puede ignorar este mensaje.`

ClemRz (changes requested):

ClemRz (changes requested):
Good progress since the first round. The previous inline comments have been addressed: `\c grottoce;` added to the migration, wording fixes applied, `mailIsValid` explicitly set to `false` on signup, `sendVerificationEmail` now re-throws, Swagger docs added, tests added, and the `BANNED` status was moved into `AuthService.authenticate`. A few things still need attention.

### Security: authenticate check order leaks account existence

In `AuthService.authenticate`, the banned and activated checks now run **before** password verification:

```js
if (!user) return MISMATCH;
if (user.banned) return BANNED;        // ← before password check
if (!user.activated) return NOT_VERIFIED; // ← before password check
if (!user.password?.startsWith('$argon2')) return MUST_RESET;
const isHashMatch = await argon2.verify(user.password, password);
```

This means an attacker can distinguish "this email exists and is banned" from "this email doesn't exist" without knowing the password — the response comes back faster (no argon2 verify) and with a different status. The previous code checked banned **after** password verification, which was the correct approach for hiding account existence.

Suggested fix: move the `banned` and `activated` checks after `argon2.verify` succeeds, same as the original code did for banned. The `MUST_RESET` check is fine before verify since it's about password format, not account status.

```js
if (!user) return MISMATCH;
if (!user.password?.startsWith('$argon2')) return MUST_RESET;
const isHashMatch = await argon2.verify(user.password, password);
if (!isHashMatch) return MISMATCH;
if (user.banned) return BANNED;
if (!user.activated) return NOT_VERIFIED;
return SUCCESS;
```

### Security: activation code not cleared on already-verified path

In `verify-email.js`, when `caver.activated` is already `true`, the controller returns early without clearing the `activationCode`:

```js
if (caver.activated) {
  return res.ok({ message: 'Account was already verified.' });
}
```

The `findOne({ activationCode: token })` matched, so the code is still in the database. If the user was activated through some other path (admin action, migration), the old activation code remains valid forever. Consider clearing it on this path too, or at minimum document why it's acceptable.

### Bug: `resend-verification-email` swallows DB update errors

In `resend-verification-email.js`, the `try/catch` wraps both the DB update and the email send:

```js
try {
  await TCaver.updateOne({ id: user.id }).set({ activationCode });
  await AuthService.sendVerificationEmail(user, activationCode, req.i18n);
} catch (err) {
  // Errors are already logged in AuthService.
}
```

If `TCaver.updateOne` fails, the error is silently swallowed and the user gets a 204 success. The comment says "Errors are already logged in AuthService" but a DB error wouldn't come from AuthService. The email error is correctly logged and re-thrown by `sendVerificationEmail`, but the DB error has no logging at all. Consider either:
- Separating the two operations so only the email error is swallowed, or
- Adding a `sails.log.error` in the catch block for non-email errors

### Bug: `login.js` updates metadata for non-SUCCESS results

In `login.js`, the login metadata update (fire-and-forget) runs for any result that isn't `MISMATCH` or `MUST_RESET`:

```js
if (result.status === MISMATCH) return ...;
if (result.status === MUST_RESET) return ...;

// This runs for NOT_VERIFIED and BANNED too:
CommonService.query(`UPDATE t_caver SET date_last_connection = NOW(), ...`);

if (result.status === NOT_VERIFIED) return ...;
if (result.status === BANNED) return ...;
```

This increments `connection_counter` and updates `date_last_connection` for unverified and banned users who haven't actually logged in. Move the metadata update to after all the guard clauses, right before issuing the token.

### Swagger: `verify-email` documents token as query param but controller reads from body

The Swagger spec defines the token as `in: query`, but the test sends it via `.send({ token: ... })` (request body) and the controller uses `req.param('token')` which reads from both. The Swagger should match the primary expected usage. Since this is a POST endpoint, `requestBody` would be more consistent with how `resend-verification-email` is documented. Alternatively, if the intent is for the front-end to call this from a link click (which would be a query param), then GET would be more appropriate as noted in the previous review.

### Swagger: `forgotPassword` response code mismatch

The Swagger documents `204` for success, but `forgot-password.js` calls `return res.ok()` which returns 200 (with no data it returns 204 via the custom `ok.js` response — actually, looking at `ok.js`, it returns 204 when `data` is falsy). The banned path also calls `res.ok()` with no data. But the success path at the bottom calls `res.ok()` after sending the email — also no data. So 204 is correct. However, the `401` for unverified accounts returns data (`{ status, message }`), which is fine. Just confirming this is consistent.

### Locales: `fr.json` has untranslated "Verify your email address"

```json
"Verify your email address": "Verify your email address"
```

This is the email subject line and it's left in English in the French locale. Same issue in `es.json` — the key `"Verify your email address"` is missing entirely, so it will fall back to the English key. The Spanish locale should have a translation like `"Verificar su dirección de correo electrónico"`, and the French one should be `"Vérifiez votre adresse mail"`.

### Minor: missing trailing newline in `verifyEmail.ejs`

The file doesn't end with a newline character. The `.editorconfig` specifies `insert_final_newline = true`.

### Minor: `resend-verification-email` TODO comment

```js
// Might need a timeout between requests
```

This is a valid concern — without rate limiting, the endpoint can be used to spam verification emails. Worth either implementing it now (even a simple in-memory cooldown) or converting this to a tracked issue. As-is, the comment reads like a forgotten TODO.

### Test: signup assertions in `after()` hook

In `Auth.test.js`, the signup test suite asserts on the created user's properties inside the `after()` cleanup hook:

```js
after(async () => {
  const created = await TCaver.findOne({ mail: 'newtest@newtest.com' });
  should(created.activated).be.false();
  should(created.activationCode).be.a.String().and.not.empty();
  // ... destroy
});
```

If these assertions fail, the `destroyOne` calls below them won't execute, leaving test data behind and potentially causing cascading failures in subsequent test runs. Move the assertions into a dedicated `it()` block and keep `after()` purely for cleanup.

ClemRz (approved):
In the PR description you mentioned a table named `backup_user_states` but I think that you are not showing how to create it.
Also you are not mentioning the addition of the index on the `t_caver` table.

ClemRz:
Before doing any merge please make sure you:
- squash all your commits into one
- run the migrations (I can do it for you, just let me know)

PR #1536, opened 2026-04-21
feat(ses): detect bounced emails via suppression list polling

### LINKED ISSUE #1535: Detect bounced/complained emails via SES suppression list polling
## Problem

Grottocenter uses AWS SES for sending emails (password reset, notifications, email verification). When an email is sent to an invalid address or triggers a spam complaint, SES signals this **asynchronously** — the `SendEmailCommand` response only confirms the message was accepted for delivery, not that it was actually delivered.

Currently, the application has no mechanism to detect these delivery failures. This means:

- Emails continue to be sent to addresses that will never receive them
- The `mail_is_valid` field on `t_caver` is never updated based on actual delivery outcomes
- Administrators have no visibility into which caver accounts have invalid email addresses
- For upcoming email verification at registration, users with invalid addresses will never receive their verification link, and the system won't proactively mark the email as invalid — it will just silently expire

## Proposed Solution

Implement a polling mechanism that periodically queries the **SES account-level suppression list** — a native SES feature (enabled by default) that records addresses causing hard bounces or spam complaints. A scheduled Node.js task cross-references this list with `t_caver.mail` and marks matches as invalid.

This approach requires **no additional AWS infrastructure** (no SNS topic, no public webhook endpoint) and follows the existing `setTimeout`-based scheduling pattern used for Typesense sync.

As part of this work, the existing SES v1 SDK (`@aws-sdk/client-ses`) will be migrated to v2 (`@aws-sdk/client-sesv2`) to consolidate all SES operations under a single package.

An admin-only API endpoint will also be added to list cavers with invalid emails.

ClemRz:
## 🤔 What

- Migrate from `@aws-sdk/client-ses` (v1) to `@aws-sdk/client-sesv2` (single SDK for both email sending and suppression list access)
- Add a daily background poller that fetches the SES account-level suppression list, cross-references it with `t_caver.mail`, and marks bounced/complained addresses as invalid (`mail_is_valid = false`)
- Add `GET /api/v1/cavers/invalid-mail` admin-only endpoint to list affected cavers
- Closes #1535

## 🤷‍♂️ Why

SES signals delivery failures (hard bounces, spam complaints) asynchronously — the `SendEmailCommand` response only confirms the message was accepted, not delivered. Until now, the application had no mechanism to detect these failures, meaning emails continued to be sent to invalid addresses and the `mail_is_valid` field was never updated based on actual delivery outcomes.

The SES account-level suppression list (already active, capturing BOUNCE and COMPLAINT) provides this information with zero additional AWS infrastructure — no SNS topic, no public webhook endpoint.

## 🔍 How

**SDK migration** (`config/awsSes.js`, `api/helpers/send-email.js`):
- Replaced `SESClient` with `SESv2Client`, restructured `SendEmailCommand` params (`Message` → `Content.Simple`, `Source` → `FromEmailAddress`)

**Suppression Service** (`api/services/SesSuppressionService.js`):
- `fetchSuppressedEmails()` — paginated `ListSuppressedDestinationsCommand` call, returns lowercase email array
- `markCaversAsInvalid()` — single bulk `UPDATE t_caver SET mail_is_valid = false WHERE LOWER(mail) = ANY($1) AND mail_is_valid = true`

**Poller** (`api/sesSuppressionPoller/sesSuppressionPoller.js`):
- `setTimeout` loop registered in `config/bootstrap.js`, runs daily at 3 AM UTC (offset from dbSync's 2 AM)
- Gracefully skips when AWS credentials are missing (local dev)
- Catches and logs errors without crashing or skipping future cycles

**Admin endpoint** (`api/controllers/v1/caver/get-invalid-mail.js`):
- Mirrors `get-banned.js` pattern — `tokenAuth` policy, admin check, `toListCaver` converter

## 🧪 Testing

5 property-based tests (fast-check, 100 runs each):
1. SES v2 email param construction preserves content
2. Pagination collects all suppressed destinations
3. Suppressed email matching marks correct cavers (case-insensitive)
4. Marking is idempotent (second call updates 0 rows)
5. `getMsUntilNextExec` targets next day at 3 AM UTC

9 example-based tests covering: empty/single-page SES responses, no-match emails, missing credentials, API/DB errors, empty suppression list, logging, and the admin endpoint (401/403/200 cases).

```bash
PORT=1338 npm run test
```

## 📸 Previews

N/A — backend-only changes, no UI impact.


ClemRz:
Note: snuck in an unrelated commit — 6ca41508 `chore(countries): adds Kosovo`

Paul-AUB (changes requested):
# Code Review — Commits e7b6a9bf & c3901d82

**Author:** Clément Ronzon  
**Reviewed:** 2026-04-29

---

## Commit  `c3901d82` · feat(ses): detect bounced emails via suppression list polling (#1535)

### What it does

Migrates the AWS SES integration from `@aws-sdk/client-ses` (v1 API) to `@aws-sdk/client-sesv2` and adds a daily background poller that:

1. Fetches the account-level SES suppression list (bounces + complaints).
2. Cross-references it with `t_caver.mail` and marks matching cavers as `mailIsValid = false`.
3. Exposes a new admin-only `GET /api/v1/cavers/invalid-mail` endpoint.

### Strengths

- **SDK migration is clean.** The old `ses: '2010-12-01'` client option (which SESv2 does not support) is correctly removed. All field renames (`Source` → `FromEmailAddress`, `Message` → `Content.Simple`) match the SES v2 spec.
- **Efficient bulk SQL update.** `SesSuppressionService.markCaversAsInvalid` uses a single parametrised `UPDATE … WHERE LOWER(mail) = ANY($1::text[]) AND mail_is_valid = true`, which avoids N+1 queries and skips already-invalid rows.
- **Pagination is handled correctly.** The `do…while (nextToken)` pattern in `fetchSuppressedEmails` properly exhausts all SES pages.
- **Timer scheduling is robust.** `registerPoller()` re-arms the timer *before* calling `pollSuppressionList()`, so a poll failure cannot orphan the scheduler.
- **Test coverage is solid.** Property-based tests (fast-check) cover `getMsUntilNextExec` across 100 random timestamps. Integration tests cover the route at 401/403/200 and the service for empty inputs, pagination stubs, and DB no-match cases.
- **Route ordering is correct.** `/api/v1/cavers/invalid-mail` is registered before `/:id`, preventing the literal string "invalid-mail" from being matched as a caver ID.

### Issues

#### Medium — `get-invalid-mail` has no pagination

```js
// api/controllers/v1/caver/get-invalid-mail.js:14
const caversWithInvalidEmail = await TCaver.find({ mailIsValid: false });
```

`TCaver.find` with no limit returns the entire result set in memory. For a large bounced-email population this could produce a very large response. For an admin diagnostic endpoint this is likely acceptable short-term, but adding a `limit` / `skip` or at least a `count` header would make it production-safe.

#### Minor — `fetchSuppressedEmails` has no guard against runaway pagination

```js
// api/services/SesSuppressionService.js:14-30
do { ... nextToken = response.NextToken; } while (nextToken);
```

There is no maximum-page or timeout guard. An unexpectedly large suppression list (or a misbehaving AWS endpoint returning perpetual tokens) would loop indefinitely. Consider adding a `maxPages` cap with a warning log.

#### Minor — `getMsUntilNextExec` always schedules for the *next* calendar day

```js
// api/sesSuppressionPoller/sesSuppressionPoller.js:5-10
d.setUTCDate(d.getUTCDate() + 1);
d.setUTCHours(3, 0, 0, 0);
```

If the server restarts at 02:59 UTC, the first run will be scheduled ~24 hours later rather than ~1 minute later. This is a deliberate design choice (avoids immediate execution on startup), but it means a deploy just before 03:00 UTC delays the first suppression sync by a full day. Consider logging the scheduled wake-up time so operators are aware.

#### Minor — `ReplyToAddresses` silently removed

The SES v1 params had an explicit `ReplyToAddresses: []`. The SES v2 params drop it entirely. Omitting the field is equivalent for AWS (defaults to no reply-to address), but the removal is a silent behavioural change for anyone inspecting the outgoing email headers. A comment noting this is intentional would help future readers.

#### Minor — No new SQL migration for `mail_is_valid` column

The `mail_is_valid` column on `t_caver` is referenced but no migration script is added in this commit. If it pre-exists from an earlier migration, this is fine — but if the column is new, a migration file is missing. Verify that the column already exists in the baseline schema.

### Summary

The feature is well-structured and the test coverage is above average for this codebase. The main concern to address before high-traffic production use is the unbounded response in `get-invalid-mail`. The rest of the findings are minor polish items.

ClemRz (commented):
Thanks for the thorough review Paul!

**Medium — `get-invalid-mail` has no pagination**

Intentionally mirrors `get-banned`, which has the same unbounded `TCaver.find()`. Adding pagination to one but not the other would be inconsistent. Happy to open a follow-up issue to add pagination to both if the team wants it.

**Minor — `fetchSuppressedEmails` has no guard against runaway pagination**

Good catch — added a `MAX_PAGES = 100` cap with a warning log. Also added a test that verifies the guard stops infinite pagination.

**Minor — `getMsUntilNextExec` always schedules for the next calendar day**

This follows the exact same pattern as `dbSync.js` (`getMsUntilNextExec` → always next day, re-arm before execute). Changing one without the other would create an inconsistency. The "restart at 02:59" edge case applies equally to dbSync, which has been running this way in production without issue.

**Minor — `ReplyToAddresses` silently removed**

Added a comment explaining the intentional omission.

**Minor — No new SQL migration for `mail_is_valid` column**

The column already exists in production — `mail_is_valid bool NOT NULL DEFAULT true` is in the `t_caver` table. No migration needed.

Paul-AUB (approved):

PR #1537, opened 2026-04-22
test(perf): parallelize test suite and fix stability (5x speedup)

ClemRz:
## 🤔 What

Parallelizes the test suite across multiple shards, each with its own database, and fixes all known sources of test flakiness. Net result: ~5x faster wall time and deterministic green runs.

- Replace 33 HTTP-hitting property tests with ~70 conventional `it()` tests
- Add a parallel test runner with folder-based sharding and auto-detected CPU count
- Add a PostgreSQL template database snapshot that rebuilds automatically when fixtures, models, or SQL change
- Reuse the Sails HTTP server in supertest instead of ephemeral listen/close cycles
- Add live progress display with percentage, `q` to abort, top 3 slowest tests
- Add `glob` and `pg` to devDependencies (were transitive only)

## 🤷‍♂️ Why

The test suite took ~2 minutes sequentially and suffered from frequent flaky failures:

- **Socket hang ups**: `supertest(app)` spins up an ephemeral HTTP server per request. Under parallel load, rapid listen/close cycles cause connection corruption and 404s on valid routes.
- **Property tests over HTTP**: `fc.asyncProperty` with supertest means 100 iterations × HTTP round-trips, exhausting sockets and causing `ECONNRESET`.
- **Slow bootstrap**: Every run dropped all tables, loaded fixtures, and ran migrations from scratch (~30s).
- **No parallelism**: Tests ran sequentially in a single process despite being independent by domain.

### Parallel runner (`scripts/parallel-test.js`)

- Discovers test files via glob, groups route tests by domain folder (e.g., `Cavers/`, `Caves/`), and packs groups into N shards using greedy bin-packing.
- Affinity groups keep cross-dependent folders together (Caves + Documents + Entrances).
- Each shard gets its own PostgreSQL database cloned from a template (`CREATE DATABASE ... TEMPLATE`), taking ~100ms per clone.
- Shards run as child processes with isolated ports and database URLs.
- A custom Mocha reporter (`test/progress-reporter.js`) sends pass/fail/total events to the parent via fd 3 for live progress.
- Press `q` to abort mid-run — kills all shards and prints a partial summary using the progress tracker's live counts.
- Shard cleanup is wrapped in `try/finally` to prevent orphaned databases on crashes.

### Template snapshot (`scripts/snapshot-test-db.js`)

- Lifts Sails with `migrate: drop`, loads fixtures, runs custom SQL, then marks the database as a PostgreSQL template.
- Staleness detection compares mtime of fixture files, models, and SQL migrations against a stamp file. Stale → auto-rebuild before the parallel run.

### Supertest server reuse (`test/bootstrap.test.js`)

- After Sails lifts, `sails.hooks.http.app` is swapped to `sails.hooks.http.server` (the already-listening HTTP server instance).
- When supertest receives a server object instead of a function, it skips `http.createServer()` + `listen(0)` + `close()` and connects directly to the existing port.
- This eliminates all ephemeral server flakiness across 150+ test files with zero per-file changes.

### Property test migration

- HTTP property tests converted to conventional `it()` tests with representative inputs covering the same boundaries.
- Service-layer property tests (with sinon stubs, no HTTP) are preserved — they're fast and stable.
- Blacklist test sleep (1.1s `setTimeout`) replaced by backdating the revocation timestamp.

### Production code changes (minimal)

- `AuthService.js`: brute-force delay and argon2 cost read from `sails.config.custom` instead of hardcoded. Defaults in `config/custom.js` preserve existing behavior. `config/env/test.js` zeroes them out.
- `Cave update afterUpdate hook`: fixed broken Waterline association restore (`updateOne` + `replaceCollection` instead of chained `.set()`).

## 🧪 Testing

```shell
npm run test                    # Parallel (default) — ~25s wall time
npm run test -- --bail          # Fail fast
npm run test -- --grep "Auth"   # Filter by pattern
npm run test -- --shards 4      # Override shard count
npm run test:sequential         # Single-process fallback for debugging
npm run coverage                # Coverage (sequential, slower)
```

Shard output is written to `test/shard-N.tmp` — paths are shown in the results table. On failure, inspect with `cat test/shard-0.tmp`.

## 📸 Previews

Normal run:
```
6 shards, 261 files
  ⟳ 74% (1490/2014)  q to abort

══════════════════════════════════════════════════════════════════════
RESULTS
══════════════════════════════════════════════════════════════════════
  ✅ Shard 0 (non-routes): 1055 pass, 0 fail (23s) → test/shard-0.tmp
  ✅ Shard 1 (Caves+Documents+Entrances): 185 pass, 0 fail (19s) → test/shard-1.tmp
  ...
──────────────────────────────────────────────────────────────────────
  Total: 2014 passing, 0 failing
  Wall time: 25.2s
══════════════════════════════════════════════════════════════════════

SLOWEST TESTS
  1. 1.8s — RelevanceService moveRelevance() should swap exactly two entities
  2. 1.1s — BlacklistService loadCache round trip
  3. 1.0s — CoordinatesSnapshotService fault condition
```

Aborted run (press `q`):
```
══════════════════════════════════════════════════════════════════════
RESULTS (aborted — partial)
══════════════════════════════════════════════════════════════════════
  Total: 847 passing, 0 failing (of 2014 total)
  Wall time: 12.3s
══════════════════════════════════════════════════════════════════════
```


Paul-AUB (changes requested):
# Code Review — Commit `0dd2845`

**Author:** Clément Ronzon  
**Date:** 2026-04-22  
**Subject:** `test(perf): parallelize test suite and fix stability (5x speedup)`

---

## Overview

This commit introduces a full overhaul of the test infrastructure. The core ideas are:

1. **Parallel sharding** — a custom Node.js runner (`scripts/parallel-test.js`) splits tests by domain folder, each shard running against its own PostgreSQL database cloned from a shared template.
2. **Template snapshotting** — `scripts/snapshot-test-db.js` seeds a database once and marks it as a PostgreSQL `TEMPLATE` database. Subsequent runs clone it in ~300 ms instead of rebuilding from scratch every time.
3. **Stability fixes** — 33 property-based (fast-check) HTTP tests replaced with ~70 conventional `it()` tests to eliminate socket hang-ups caused by ephemeral `listen/close` cycles.
4. **Configurable auth delays** — brute-force delay and argon2 cost are now Sails config values, overridable per environment, so tests don't wait 500 ms per login.

---

## Code Quality & Style

### Strengths

- **`scripts/parallel-test.js`** is well-structured with clear section dividers and single-responsibility functions. The `withClient` helper properly disposes the `pg.Client` via `try/finally`.
- **`scripts/snapshot-test-db.js`** correctly terminates active connections before `DROP DATABASE` to avoid "database is being accessed by other users" errors.
- **Staleness detection** using `mtime` on a stamp file is simple and pragmatic. The `SNAPSHOT_SOURCES` list covers fixtures, fixture order, models, and SQL migrations — the right set.
- **Bin-packing** (greedy, largest-first) is a sound heuristic for distributing test groups across shards without needing any profiling data.
- **`try/finally` shard cleanup** (`dropShardDb`) in `main()` prevents orphaned `_shard_N` databases even when the runner crashes.
- Credentials in the PostgreSQL connection URL are redacted before logging (`replace(/\/\/.*@/, '//***@')`).
- The `config/env/test.js` override for argon2 is correctly scoped to `hash()` calls only; fixture passwords pre-hashed at higher cost are unaffected by this change.

### Concerns

#### 1. `withClient` duplication

`withClient` is defined independently in both `scripts/parallel-test.js` and `scripts/snapshot-test-db.js` with slightly different signatures (one accepts `connectionUrl`, the other uses a module-level `maintenanceUrl`). This is a minor DRY issue and could drift if one implementation is updated without the other.

**Suggestion:** Extract to a shared `scripts/test-db-utils.js` module. Low priority for now since both are small.

#### 2. Hardcoded `AFFINITIES` in the sharding logic

```js
const AFFINITIES = [
  ['Caves', 'Documents', 'Entrances'],
];
```

This is a cross-dependency workaround discovered empirically. It works, but it's fragile: adding a new domain that has a data dependency on Caves or Documents requires manually updating this array. There is no test or assertion to catch the regression if someone forgets.

**Suggestion:** Add a comment explaining *which specific fixture or test* creates this dependency (e.g., "Cave tests reference `Documents.isValidated`, set by `Documents/validate`"). The commit message mentions it, but the code does not — this is exactly the kind of non-obvious invariant that warrants an inline note.

#### 3. Port allocation strategy

```js
const BASE_PORT = 1340;
// ...
PORT: String(BASE_PORT + i),
```

Ports are statically allocated starting at 1340 and incremented per shard. If another process occupies one of these ports on the developer's machine, the corresponding shard will fail to lift Sails with a cryptic error.

**Suggestion:** Consider detecting a free port range before spawning shards, or at minimum print the port number in the shard boot failure output.

#### 4. `runSnapshot` spawns `node` by name

```js
const child = spawn('node', ['scripts/snapshot-test-db.js', '--seed'], { ... });
```

`process.execPath` should be used here instead of the bare string `'node'`, for consistency with how `MOCHA_BIN` is resolved below:

```js
const child = spawn(process.execPath, args, { ... });
```

This prevents edge cases where `node` in `PATH` differs from the Node version that launched the runner (e.g., in `nvm` environments).

#### 5. Progress reporter `sort` inside `addSlow` is called on every event

```js
addSlow(ms, title) {
  slowTests.push({ ms, title });
  slowTests.sort((a, b) => b.ms - a.ms);
  if (slowTests.length > 3) slowTests.length = 3;
},
```

Sorting on every `addSlow` call is fine for the current load (dozens of tests), but a simple insertion into a 3-element sorted array would be O(1) instead of O(n log n). Not a real-world concern, but easy to improve if the array ever grows.

#### 6. `--bail` across shards is per-shard, not global

Passing `--bail` to each shard mocha process stops *that shard* on the first failure, but the other shards continue running. This is the expected behaviour for a parallel runner, but it may surprise users expecting `--bail` to abort the entire suite.

**Suggestion:** Document this in README or the `--bail` CLI help output, or add a global bail signal: once one shard exits with failure, send `SIGTERM` to the remaining shards.

---

## Correctness

- Replacing `calledOnce` with `called` in `RecentChangeWindow.test.js` is correct: the original assertion was too strict for async race conditions.
- Backdating the revocation timestamp in the blacklist clock-skew fix is safer than `sleep()` — it removes non-determinism.
- The batch SQL query in `snapshot-test-db.js` (`[...].join('\n')`) is passed as a single `CommonService.query` call. This relies on the underlying adapter executing the joined queries as one transaction or at least one round-trip. Verify the adapter supports multi-statement strings without splitting them.

---

## Performance

- The 5x speedup claim is plausible: parallelism alone across 4–6 CPUs gives 4–6x, partially offset by template clone overhead (~300 ms) and cross-shard affinity grouping reducing effective parallelism.
- `argon2Options: { memoryCost: 1024, timeCost: 1, parallelism: 1 }` in test environment is appropriately minimal and isolated from production config.
- The mtime-based staleness check uses `fs.statSync` in a loop, which is synchronous. At the scale of this codebase (a few hundred fixture/model files) this is negligible.

---

## Test Coverage

- The conversion from fast-check property tests to conventional `it()` tests **reduces** the number of input combinations explored per run. Property tests exercise random edge cases across many runs; conventional tests only cover the explicitly written cases.
- For the Geoloc and Search tests this is a reasonable trade (eliminating socket instability is worth more than exhaustive property exploration for route-level tests). But the property tests for `typesense-error-handling` and `typesense-preservation` have been deleted and replaced with a smaller fixed set — make sure the replacement tests cover the failure modes the property tests found historically (if any).
- The ban/unban property tests (`Cavers/ban-unban.property.test.js`) are retained and substantially reworked rather than deleted, which is the right call for permission-sensitive logic.

---

## Security

- `authBruteForceDelay: 0` in `config/env/test.js` is correctly scoped to `NODE_ENV=test`. The production default remains 500 ms. No risk of accidental test config leaking to production.
- Shard database names include `_shard_N` suffix. These are created and dropped within the same test run. No credentials are stored in the shard DBs beyond what the template contains.
- The `q` key abort handler reads from stdin in raw mode. Ctrl-C (`key[0] === 3`) is also handled, which is correct. No injection risk.

---

## Summary

| Area | Assessment |
|---|---|
| Architecture | Excellent — template + sharding is the right approach for this scale |
| Code quality | Good — well-organized, handles edge cases (cleanup, abort, credential masking) |
| `node` vs `process.execPath` in `runSnapshot` | Minor bug — should use `process.execPath` |
| Hardcoded AFFINITIES | Fragile — needs an inline explanation of the concrete dependency |
| `--bail` cross-shard behaviour | Underdocumented — users may be surprised |
| Property test reduction | Acceptable trade-off, but worth a comment noting coverage intent |
| Overall risk | Low — changes are test-only except for the configurable auth delay (production default is unchanged) |

**Verdict: OK with minor comments.** The `process.execPath` fix in `runSnapshot` is the only change worth blocking on; the rest are suggestions.

ClemRz (commented):
Thanks for the thorough review Paul — all 7 points addressed in the latest push. Here's the breakdown:

### 1. `process.execPath` in `runSnapshot` (blocking)

Fixed. `spawn('node', ...)` → `spawn(process.execPath, ...)` so the snapshot subprocess uses the same Node binary that launched the runner. Consistent with how `MOCHA_BIN` is already resolved.

### 2. Hardcoded `AFFINITIES` — inline explanation

Expanded the comment to document the concrete dependencies:
- **Caves ↔ Documents**: Cave tests reference fixture documents (`addDocument`, `unlinkDocument`, `create with documents:[1,2]`) and depend on `Documents/validate` having set `isValidated`.
- **Caves ↔ Entrances**: Entrance tests create/update entrances linked to caves from fixtures; cave deletion tests verify cascade on associated entrances.

Also added a note telling future contributors to update the array when adding dependent domains.

### 3. `--bail` cross-shard behaviour — documented

Added a note in README under the `--bail` section explaining that it stops the failing shard only, not all shards, and pointing to `npm run test:sequential -- --bail` for true global fail-fast. Verified with an intentional test failure: shard 3 bailed after 1 failure while the other 5 shards ran to completion.

### 4. `withClient` duplication — extracted

Created `scripts/test-db-utils.js` with shared `withClient(connectionUrl, fn)` and `disconnectAll(client, dbName)`. Both `parallel-test.js` and `snapshot-test-db.js` now import from this module. The `pg` direct dependency is removed from `parallel-test.js`.

### 5. Port allocation — port shown in failure output

Failed shards now display `[port NNNN]` in the results table. Example from the test run with the intentional failure:
```
❌ Shard 3 (...): 217 pass, 1 fail (22s) [port 1343] → test/shard-3.tmp
```
Passing shards omit the port to keep the output clean.

### 6. `addSlow` sort optimization

Both `parallel-test.js` (`createProgress().addSlow`) and `test/progress-reporter.js` now use sorted insertion into the 3-element array instead of push + sort on every call.

### 7. Property test replacement — coverage intent documented

Both `typesense-error-handling.test.js` and `typesense-preservation.test.js` now have expanded JSDoc headers documenting:
- Which property test file they replace
- What input partitions they cover (boundary values, auth errors, 5xx, validation)
- The explicit coverage trade-off and why boundary values are sufficient (the underlying code uses simple range checks like `>= 400 && < 500`)

Paul-AUB (approved):

Issue #1450, opened 2026-02-11
Massif sensitive setting


ClemRz:
# Introduction

This feature extends the Grottocenter API to support marking entire massifs as sensitive, cascading that sensitivity to all caves and entrances within the massif polygon, and automatically marking new caves/entrances added inside a sensitive massif as sensitive. This protects location data for ecologically or culturally sensitive cave areas at the massif level, rather than requiring individual entrance-level marking.

Sensitivity designations must be based on applicable legislation, not ideological or subjective reasons. Only Administrators can mark or unmark massifs as sensitive.

# User stories

- As an Administrator, I want to mark a massif as sensitive based on applicable legislation, so that I can protect the location data of all caves and entrances within that massif area while ensuring designations are legally justified.
- As an Administrator, I want all entrances within a sensitive massif to be automatically marked as sensitive, so that I do not have to update each entrance individually.
- As a system administrator, I want only administrators to be able to remove the sensitive flag from a massif, so that sensitive designations are not accidentally reversed.
- As a data contributor, I want new entrances added inside a sensitive massif to be automatically marked as sensitive, so that location data is protected without manual intervention.
- As a Moderator, I want changes to massif sensitivity to be tracked in the history table, so that I can audit who changed the sensitivity status and when.

# Key rules

- Only Administrators can set or unset `isSensitive` on a massif (403 for all other roles).
- The `isSensitive` field is only visible in the massif response for Administrators; it is omitted for non-admin users.
- Marking a massif as sensitive cascades `isSensitive = true` to all entrances within the massif polygon (via PostGIS `ST_Contains`).
- Removing sensitivity from a massif does NOT automatically unmark individual entrances.
- New entrances created inside a sensitive massif are automatically marked as sensitive.
- Sensitivity changes are tracked in the HMassif history table.
- The sensitive status is NOT searchable or filterable in Typesense — search behavior does not change.

⚠️ This ticket is meant to be broken down into several sub-tickets ⚠️

ClemRz:
@dawoldo I detailed the requirements

PR #1539, opened 2026-04-24
1450 massif sensitive setting

dawoldo:
## 🤔 What

This PR implements a sensitivity management system for massifs, allowing administrators to protect location data for all entrances within a massif area.

- **Model Updates**: Added `isSensitive` boolean flag to `TMassif` and `HMassif` models (defaults to `false`).
- **New API Endpoints**:
    - `POST /api/v1/massifs/:id/mark-sensitive`: Marks a massif as sensitive and propagates the status to all geographically contained entrances. Returns a summary count of affected entrances.
    - `POST /api/v1/massifs/:id/unmark-sensitive`: Removes the sensitive status from a massif record.
    - `GET /api/v1/massifs/:id/preview-sensitive`: Previews the number of entrances that would be affected by marking the massif as sensitive.
- **Auto-Inheritance**: Enforced automatic sensitive marking for any new entrance created within the polygon of a sensitive massif (supporting both standard creation and CSV imports).
- **Security & Privacy**: 
    - Restricted all sensitivity-modifying actions to users within the administrator group.
    - Implemented a privacy gate in the `toMassif` converter to hide the `isSensitive` field from non-admin users.
    - Decoupled sensitivity toggling from the general `PUT /massifs/:id` update route to prevent accidental changes.
- **Auditing**: Sensitivity updates are now captured in the `HMassif` history table, including the `reviewer` ID and timestamp.
- **Documentation**: Updated Swagger documentation to reflect the new routes and response shapes.

[Link to issue: #1450](https://github.com/GrottoCenter/grottocenter-api/issues/1450#issue-3928987374)

## 🤷‍♂️ Why

To protect the location data of caves and entrances within sensitive geographic areas. This feature provides Administrators with the tools to manage data protection at scale (per-massif) rather than forcing manual updates for every individual entrance, while maintaining an audit trail for these changes.

## 🔍 How

- **Spatial Logic**: Used PostGIS `ST_Contains` for geographic containment checks to link entrances to their respective massifs based on coordinates.
- **Service Layer**: Centralized propagation and validation logic in `MassifService.js` and `EntranceService.js`.
- **Mapping Layer**: Utilized the `meta.hasCompleteViewRight` flag in the converter layer to handle conditional field visibility (Privacy Gate).
- **Controller Refactoring**: Moved sensitivity logic out of the general `update.js` controller and into specific, authorized actions to ensure strict access control.

## 🧪 Testing

- **Automated Tests**:
    - Added `mark-sensitive.test.js` and `preview-sensitive.test.js`.
    - Updated `find.test.js` to assert that `isSensitive` is omitted for non-admins.
    - Updated `update.test.js` to ensure the general update route no longer handles the sensitive flag.
- **Verification**: Verified that history records are correctly created in the `h_massif` table upon status changes.

### DIFF
--- a/api/controllers/v1/massif/create.js
+++ b/api/controllers/v1/massif/create.js
@@ -43,13 +43,16 @@ module.exports = async (req, res) => {
     );
   }
 
+  const rawData = MassifService.getConvertedDataFromClientRequest(req);
+
   // Launch creation request using transaction: it performs a rollback if an error occurs
   const newMassif = await sails.getDatastore().transaction(async (db) => {
     const cleanedData = {
       author: req.token.id,
       dateInscription: new Date(),
-      documents: req.body.documents ? req.body.documents : [],
+      documents: rawData.documents ? rawData.documents : [],
       geogPolygon: wkt,
+      isSensitive: rawData.isSensitive,
     };
 
     const massif = await TMassif.create(cleanedData)
@@ -80,6 +83,10 @@ module.exports = async (req, res) => {
 
     return massif;
   });
+
+  if (newMassif.isSensitive) {
+    await MassifService.setSensitivity(newMassif.id, true);
+  }
 
   const newMassifPopulated = await MassifService.getPopulatedMassif(
     newMassif.id
--- a/api/controllers/v1/massif/create.js
+++ b/api/controllers/v1/massif/create.js
@@ -5,6 +5,7 @@ const NotificationService = require('../../../services/NotificationService');
 const RecentChangeService = require('../../../services/RecentChangeService');
 const { toMassif } = require('../../../services/mapping/converters');
 const { validateNameLength } = require('../../../utils/nameValidation');
+const RightService = require('../../../services/RightService');
 
 // eslint-disable-next-line consistent-return
 module.exports = async (req, res) => {
@@ -45,6 +46,15 @@ module.exports = async (req, res) => {
   }
 
   const rawData = MassifService.getConvertedDataFromClientRequest(req);
+
+  const isAdmin = RightService.hasGroup(
+    req.token.groups,
+    RightService.G.ADMINISTRATOR
+  );
+
+  if (rawData.isSensitive !== undefined && !isAdmin) {
+    return res.forbidden('Only administrators can set the sensitivity status.');
+  }
 
   // Launch creation request using transaction: it performs a rollback if an error occurs
   const newMassif = await sails.getDatastore().transaction(async (db) => {
@@ -88,7 +98,8 @@ module.exports = async (req, res) => {
   if (newMassif.isSensitive) {
     const updatedEntranceIds = await MassifService.setSensitivity(
       newMassif.id,
-      true
+      true,
+      req.token.id
     );
     await Promise.all(
       updatedEntranceIds.map(async (id) => {
--- a/api/controllers/v1/massif/create.js
+++ b/api/controllers/v1/massif/create.js
@@ -1,5 +1,6 @@
 const ControllerService = require('../../../services/ControllerService');
 const MassifService = require('../../../services/MassifService');
+const EntranceService = require('../../../services/EntranceService');
 const NotificationService = require('../../../services/NotificationService');
 const RecentChangeService = require('../../../services/RecentChangeService');
 const { toMassif } = require('../../../services/mapping/converters');
@@ -85,7 +86,16 @@ module.exports = async (req, res) => {
   });
 
   if (newMassif.isSensitive) {
-    await MassifService.setSensitivity(newMassif.id, true);
+    const updatedEntranceIds = await MassifService.setSensitivity(
+      newMassif.id,
+      true
+    );
+    await Promise.all(
+      updatedEntranceIds.map(async (id) => {
+        const populated = await EntranceService.getPopulatedEntrance(id);
+        if (populated) await EntranceService.updateInSearch(populated);
+      })
+    );
   }
 
   const newMassifPopulated = await MassifService.getPopulatedMassif(
--- a/api/controllers/v1/massif/create.js
+++ b/api/controllers/v1/massif/create.js
@@ -57,50 +57,57 @@ module.exports = async (req, res) => {
   }
 
   // Launch creation request using transaction: it performs a rollback if an error occurs
-  const newMassif = await sails.getDatastore().transaction(async (db) => {
-    const cleanedData = {
-      author: req.token.id,
-      dateInscription: new Date(),
-      documents: rawData.documents ? rawData.documents : [],
-      geogPolygon: wkt,
-      isSensitive: rawData.isSensitive,
-    };
-
-    const massif = await TMassif.create(cleanedData)
-      .fetch()
-      .usingConnection(db);
-
-    // Name
-    await TName.create({
-      author: req.token.id,
-      dateInscription: new Date(),
-      isMain: true,
-      language: req.body.descriptionAndNameLanguage.id,
-      massif: massif.id,
-      name: req.body.name,
-    }).usingConnection(db);
-
-    // Description
-    if (req.body?.description) {
-      await TDescription.create({
+  const { newMassif, updatedEntranceIds } = await sails
+    .getDatastore()
+    .transaction(async (db) => {
+      const cleanedData = {
         author: req.token.id,
-        body: req.body.description,
         dateInscription: new Date(),
-        massif: massif.id,
-        language: req.body.descriptionAndNameLanguage.id,
-        title: req.body.descriptionTitle,
-      }).usingConnection(db);
-    }
+        documents: rawData.documents ? rawData.documents : [],
+        geogPolygon: wkt,
+        isSensitive: rawData.isSensitive,
+      };
 
-    return massif;
-  });
+      const massif = await TMassif.create(cleanedData)
+        .fetch()
+        .usingConnection(db);
+
+      // Name
+      await TName.create({
+        author: req.token.id,
+        dateInscription: new Date(),
+        isMain: true,
+        language: req.body.descriptionAndNameLanguage.id,
+        massif: massif.id,
+        name: req.body.name,
+      }).usingConnection(db);
+
+      // Description
+      if (req.body?.description) {
+        await TDescription.create({
+          author: req.token.id,
+          body: req.body.description,
+          dateInscription: new Date(),
+          massif: massif.id,
+          language: req.body.descriptionAndNameLanguage.id,
+          title: req.body.descriptionTitle,
+        }).usingConnection(db);
+      }
+
+      let affectedEntranceIds = [];
+      if (massif.isSensitive) {
+        affectedEntranceIds =
+          await MassifService.propagateSensitivityToEntrances(
+            massif.id,
+            req.token.id,
+            db
+          );
+      }
+
+      return { newMassif: massif, updatedEntranceIds: affectedEntranceIds };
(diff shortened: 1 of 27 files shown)


ClemRz (changes requested):
#### 1. `setSensitivity` called without `reviewerId` in propagation test

In `MassifSensitivityPropagation.test.js`, the test calls:

```js
await MassifService.setSensitivity(massifId, true);
// and
await MassifService.setSensitivity(massifId, false);
```

The third parameter `reviewerId` is missing. This means `TMassif.updateOne().set({ reviewer: undefined })` is called, which in Waterline will either be ignored or set to null. The test passes because it doesn't assert on the reviewer, but this masks a real-world scenario where the reviewer should always be set. The test should pass a valid reviewer ID.

#### 2. `mark-sensitive.js` controller has no try/catch around the service call

The `mark-sensitive` controller calls `MassifService.setSensitivity()` and `EntranceService.getPopulatedEntrance()` without any error handling. If the spatial query fails (e.g., massif has no polygon), the error will bubble up as an unhandled 500. The `unmark-sensitive` and `preview-sensitive` controllers have the same issue. Compare with the existing controller pattern in the project which uses try/catch with `res.serverError()`.

#### 3. `create.js` calls `setSensitivity` after the transaction, creating a spurious history entry

In `create.js`, the massif is created inside a transaction, but then `setSensitivity` is called outside it:

```js
const newMassif = await sails.getDatastore().transaction(async (db) => {
  // ... create massif
});
if (newMassif.isSensitive) {
  const updatedEntranceIds = await MassifService.setSensitivity(
    newMassif.id, true, req.token.id
  );
}
```

`setSensitivity` calls `TMassif.updateOne()` which will trigger the history trigger again, creating a spurious history entry for a massif that was just created. The massif was already created with `isSensitive: true` — calling `setSensitivity` again updates it to the same value but fires the trigger. This also means the `reviewer` and `dateReviewed` get set on a brand-new massif, which is semantically wrong (it was just created, not reviewed).

The entrance propagation logic is fine to run outside the transaction, but the massif update inside `setSensitivity` is redundant and harmful here.

#### 4. Swagger Massif schema missing `isSensitive` property

The `Massif` component schema in `swaggerV1.yaml` (line ~5974) does not include `isSensitive`. The new endpoint response schemas reference `$ref: '#/components/schemas/Massif'`, but that schema doesn't document the field. It should be added as:

```yaml
isSensitive:
  type: boolean
  description: Whether the massif is marked as sensitive (only visible to administrators)
```

#### 5. `coerceBool` duplicated in `MassifService.js` and `EntranceService.js`

The same `coerceBool` function is defined in both services. This should be extracted to a shared utility (e.g., `api/utils/coerceBool.js`). The implementations are also slightly different — `EntranceService` uses `typeof value === 'string' ? value === 'true' : Boolean(value)` while `MassifService` uses `value === 'true' || value === true`. The `EntranceService` version is more robust (handles `'false'` string correctly via `Boolean`), though both work for the expected inputs.

#### 6. `countUnsensitiveEntrances` silently returns 0 on error

The `countUnsensitiveEntrances` method catches all errors and returns 0. For the `preview-sensitive` endpoint, this means an admin could see "0 entrances would be affected" when in reality the query failed. This is misleading. At minimum, the error should be logged.

#### 7. `isPointInSensitiveMassif` silently returns `false` on error

Same pattern — if the spatial query fails, the entrance won't be auto-marked as sensitive. This is a fail-open behavior. For a security-sensitive feature, this should at least log the error so operators can detect the failure.

#### 8. Test fixtures for massif IDs 100 and 101 have no `names` association

The fixture massifs don't have names, which means `getPopulatedMassif` will return them with empty names. This is fine for the service-level tests but could cause issues if route tests try to use these fixtures with endpoints that expect a name.

#### 9. `preview-sensitive.test.js` is thin

The test only checks 403 for user and 404 for non-existing massif. There's no test for the happy path (admin gets a count). The `mark-sensitive` and `unmark-sensitive` tests have happy-path coverage, but `preview-sensitive` doesn't verify the actual count returned.

#### 10. `update.test.js` sends `isSensitive: true` but asserts it stays `false`

This is actually correct behavior (the update controller strips `isSensitive`), but the test could use a comment explaining that this is intentional — verifying that the general update route ignores the sensitivity flag.

#### 11. `mark-sensitive.js` doesn't use `ControllerService.treatAndConvert`

The `mark-sensitive` controller returns `res.ok({ count, massif })` directly, while `unmark-sensitive` uses `ControllerService.treatAndConvert`. The project convention (per the steering rules) is to use `ControllerService.treat` or `ControllerService.treatAndConvert` for success paths. The `mark-sensitive` controller should follow the same pattern, though the custom response shape (with `count`) makes this slightly awkward.

#### 12. No `401` response documented in Swagger for the new endpoints

The new endpoints require `tokenAuth` but only document `403` and `404`. They should also document `401 Unauthorized` for missing/invalid tokens, consistent with other protected endpoints.

ClemRz (changes requested):
## NFR: Observability

### Massifs becoming sensitive

When a massif is updated to become sensitive, as a system administrator I'd like to see the lis of entrance IDs that will be converted from non-sensitive to sensitive printed in the logs.

### Entrance automatically becoming sensitive

When an entrance is created inside the boundaries of a sensitive massif, as a system administrator I'd like to see a mention in the logs indicating that it'll be automatically marked as sensitive.

ClemRz (changes requested):
### Still Open

10. **`create.js` calls `setSensitivity` after the transaction** — Still present. The code at lines 68–79 of `create.js` creates the massif with `isSensitive: true` inside the transaction, then calls `setSensitivity(newMassif.id, true, ...)` outside it. This updates the massif to the same value, fires the history trigger, and sets `reviewer`/`dateReviewed` on a brand-new record. The entrance propagation is fine, but the `TMassif.updateOne` inside `setSensitivity` is redundant and creates a spurious history entry. The fix would be to extract the entrance propagation logic into its own method (e.g., `propagateSensitivityToEntrances`) and call that from `create.js` instead of `setSensitivity`.

11. **`update.test.js` comment** — Not addressed. The test sends `isSensitive: true` and asserts it stays `false`, which is correct behavior, but a brief comment explaining the intent would help future readers. (Minor.)

12. **Fixture massifs 100/101 have no `names` association** — Partially addressed. The fixtures do include `names: [100]` and `names: [101]`, and corresponding `tname.json` entries exist. This looks correct.

### Observability NFR — Not Addressed

ClemRz requested:

- When a massif becomes sensitive, log the list of entrance IDs converted from non-sensitive to sensitive.
- When an entrance is auto-marked sensitive at creation time, log a mention.

Neither log statement has been added. In `setSensitivity`, the `entranceIds` array is available right before the update — a `sails.log.info(...)` there would satisfy the first request. In `createEntrance`, after the `isPointInSensitiveMassif` check resolves to `true`, a similar log would satisfy the second.

---

## New Issues Found

### A. `toMassif` called without `meta` in search results (bug)

In `converters.js` line 713:

```js
else if (_type === 'massifs') data = c.toMassif(item.document);
```

`meta` is not passed, so `isSensitive` will always be `undefined` in search results — even for admins. Every other converter in `toSearchResult` receives `meta`. This should be:

```js
else if (_type === 'massifs') data = c.toMassif(item.document, meta);
```

### B. `setSensitivity` JSDoc is incomplete

The JSDoc comment only documents `@param {number} reviewerId` but the function signature is `setSensitivity(massifId, isSensitive, reviewerId)`. The first two parameters are undocumented.

### C. `isSensitive: undefined` leaks into JSON when `meta` is absent

In `toMassif`, when `meta?.hasCompleteViewRight` is not `true`, the field is set to `undefined`. In JavaScript, `JSON.stringify` strips `undefined` properties, so this works for HTTP responses. But if the object is used internally before serialization (e.g., in tests or logging), the property exists with value `undefined`, which could cause confusion. A cleaner pattern would be to conditionally spread:

```js
...(meta?.hasCompleteViewRight === true && { isSensitive: source.isSensitive }),
```

This is a minor style point — the current approach works correctly for the HTTP path.

### D. `mark-sensitive` response shape differs from `unmark-sensitive`

`mark-sensitive` returns `{ count, massif }` via `ControllerService.treat`, while `unmark-sensitive` returns just the massif via `ControllerService.treatAndConvert`. This asymmetry means the client needs to handle two different response shapes for what are conceptually paired operations. Consider whether `unmark-sensitive` should also return a `{ count: 0, massif }` wrapper for consistency, or document the difference clearly.

### E. No idempotency guard on `mark-sensitive`

If a massif is already sensitive and an admin calls `mark-sensitive` again, `setSensitivity` will:

1. Update the massif (setting `isSensitive` to the same value, creating a history entry)
2. Find zero entrances to propagate (they're already sensitive)
3. Return `count: 0`

This isn't harmful, but it creates unnecessary history entries. A simple guard at the top of `setSensitivity` (or in the controller) would avoid this:

```js
if (massif.isSensitive === isSensitive) return [];
```

The same applies to `unmark-sensitive`.

### F. Entrance propagation uses `TEntrance.update` without history

When `setSensitivity` bulk-updates entrances via `TEntrance.update({ id: entranceIds }).set({ isSensitive: true })`, this is a Waterline update that bypasses any PostgreSQL trigger on `t_entrance` (assuming there's an `h_entrance` history trigger similar to `h_massif`). If entrance sensitivity changes should be audited, this needs attention. If not, it's fine but worth a conscious decision.

### G. Race condition between `create.js` transaction and `setSensitivity`

In `create.js`, the massif is created inside a transaction, then `setSensitivity` runs outside it. If the server crashes between the transaction commit and `setSensitivity`, the massif exists with `isSensitive: true` but the entrances haven't been propagated. This is a narrow window, but for a security feature, it's worth noting. The entrance propagation could be included in the transaction, or a background job could reconcile.

### H. SQL migration: `CREATE OR REPLACE FUNCTION` replaces the entire trigger function

The migration at `sql/9_02_2026_04_23_massif_is_sensitive.sql` uses `CREATE OR REPLACE FUNCTION histo_update_massif()` which replaces the entire function body. If a future migration also modifies this function, the last one wins. This is standard practice for trigger functions, but the new function body should be verified against the current production version to ensure no other columns were recently added that might be dropped from the INSERT. Worth a manual check.

ClemRz (changes requested):
### 1. observability log in `createEntrance` doesn't include the entrance ID

In `EntranceService.createEntrance`, the log message is:

```
'Entrance auto-marked sensitive at creation because its coordinates lie within a sensitive massif.'
```

At this point in the code, the entrance hasn't been created yet (the log is before the transaction), so the ID isn't available, think about moving it.

### 2. `isSensitive: undefined` in `toMassif` for non-admins

Still uses `undefined` rather than conditional spread. Works correctly for HTTP responses (`JSON.stringify` strips it), but the property key exists on the object. Minor style preference — not blocking.

### 3. Swagger `unmark-sensitive` response schema doesn't reflect the new `{ count, massif }` shape

The Swagger spec for `unmark-sensitive` still shows:

```yaml
schema:
  $ref: '#/components/schemas/Massif'
```

But the controller now returns `{ count: 0, massif: {...} }`, matching `mark-sensitive`. The Swagger should be updated to use the same `{ count, massif }` object schema as `mark-sensitive`.

### 4. `update.test.js` `afterEach` is placed before `before`

In `update.test.js`, the `afterEach(() => { sinon.restore(); })` block is placed before the `before(async () => { ... })` block. Mocha executes hooks in declaration order within their scope, so this works, but conventionally `before` comes first. Cosmetic only.

ClemRz (approved):
Excellent work!

PR #1541, opened 2026-04-26
fix(entrance): populate point_geom on INSERT, not just UPDATE

ClemRz:
## 🤔 What

Adds a `BEFORE INSERT` trigger on `t_entrance` to populate `point_geom` from `latitude`/`longitude` at creation time.

- New trigger function `set_entrance_point_geom()` in `sql/0_triggers.sql`
- Test database bootstrap updated to create the trigger before fixtures are loaded

## 🤷‍♂️ Why

The `point_geom` column was added retroactively to support spatial queries (massif containment, geolocation, data quality views). A backfill migration populated it for existing rows, and the existing `histo_update_entrance` trigger keeps it in sync on UPDATE. But no `BEFORE INSERT` trigger was added, so newly created entrances had `NULL` in `point_geom` until their first edit.

This caused new entrances to be invisible to all spatial queries — they wouldn't appear in massif cave/entrance lists, statistics, or geolocation results until manually updated.

## 🔍 How

A dedicated trigger function is used rather than modifying `histo_update_entrance()`, because that function references `OLD.*` values which don't exist on INSERT. The new trigger is placed alongside the other entrance triggers in `0_triggers.sql`.

For the test database (which is created by Waterline's `migrate: drop` and doesn't run `sql/*.sql`), the trigger is created in `test/customSQL.js` and executed during bootstrap before fixtures are loaded, so fixture entrances also get `point_geom` set automatically.

## 🧪 Testing

- `npm run dev:clean` to rebuild Docker containers from scratch, then verify `docker logs grotto-postgres` shows no errors
- Create a new entrance via the API and confirm it appears in its massif's `GET /api/v1/massifs/:id` response immediately (without needing an update first)
- `npm run test` — existing tests pass; the test bootstrap now creates the trigger so fixture entrances get `point_geom` on insert

### DIFF
--- a/scripts/snapshot-test-db.js
+++ b/scripts/snapshot-test-db.js
@@ -57,6 +57,9 @@ function runFullBootstrap() {
         async bootstrap() {
           await CommonService.query(customSQL.ALTER_MASSIF_COLUMN_GEOG_POLYGON);
           await CommonService.query(customSQL.ALTER_ENTRANCE_COLUMN_POINT_GEOM);
+          await CommonService.query(
+            customSQL.CREATE_ENTRANCE_POINT_GEOM_INSERT_TRIGGER
+          );
         },
       },
       // eslint-disable-next-line consistent-return
--- a/sql/0_triggers.sql
+++ b/sql/0_triggers.sql
@@ -438,8 +438,8 @@ end if;
 --on insert la valeur de la date de modification dans t_entrance
 NEW.date_reviewed := now();
 
--- Keep the point geometry up to date
-NEW.point_geom := ST_SetSRID(ST_MakePoint(NEW.longitude, NEW.latitude), 4326);
+-- point_geom is now maintained by the set_entrance_point_geom trigger
+-- (BEFORE INSERT OR UPDATE), so no duplication here.
 RETURN NEW;
 END;
 $$ LANGUAGE plpgsql;
@@ -497,6 +497,24 @@ CREATE OR REPLACE TRIGGER histo_delete_entrance BEFORE DELETE ON t_entrance FOR 
 --trigger qui exécute la sauvegarde des derniers changements lors d'un CREATE ou d'un UPDATE
 --------------------------------------------------------
 CREATE OR REPLACE TRIGGER last_change_entrance BEFORE INSERT OR UPDATE ON t_entrance FOR EACH ROW EXECUTE PROCEDURE change_entrance();
+--
+-- Keep point_geom in sync with latitude/longitude on every INSERT or UPDATE.
+-- This single trigger replaces the previous duplication between
+-- histo_update_entrance (UPDATE-only, no NULL guard) and a separate
+-- INSERT-only trigger. NULL coordinates produce a NULL point_geom.
+--------------------------------------------------------
+CREATE OR REPLACE FUNCTION set_entrance_point_geom() RETURNS trigger AS $$
+BEGIN
+  IF NEW.longitude IS NOT NULL AND NEW.latitude IS NOT NULL THEN
+    NEW.point_geom := ST_SetSRID(ST_MakePoint(NEW.longitude, NEW.latitude), 4326);
+  ELSE
+    NEW.point_geom := NULL;
+  END IF;
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE OR REPLACE TRIGGER set_entrance_point_geom_trigger BEFORE INSERT OR UPDATE ON t_entrance FOR EACH ROW EXECUTE PROCEDURE set_entrance_point_geom();
 --
 --
 ------------------------------------------------------------
--- a/sql/8_02_2026_04_25_entrance_insert_trigger.sql
+++ b/sql/8_02_2026_04_25_entrance_insert_trigger.sql
@@ -1,0 +1,105 @@
+\c grottoce;
+
+-- Consolidate point_geom maintenance into a single BEFORE INSERT OR UPDATE
+-- trigger, replacing the previous UPDATE-only logic in histo_update_entrance()
+-- and the INSERT-only set_entrance_geom_on_insert trigger.
+--
+-- Changes:
+-- 1. set_entrance_point_geom() now handles both INSERT and UPDATE with a NULL guard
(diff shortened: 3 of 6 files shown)


Paul-AUB (changes requested):
# Code Review — commit af5e39e6f412aa032b4343a91995a58c7571ddd9

**Author:** Clément Ronzon  
**Date:** 2026-04-25  
**Commit message:** `fix(entrance): add BEFORE INSERT trigger to populate point_geom`

---

## Overview

This commit fixes a silent data bug: when a new entrance is created, the `point_geom` geometry column (used by every spatial query — massif containment, geolocation, data quality views) was left `NULL` because the only trigger that sets it, `histo_update_entrance`, fires `BEFORE UPDATE` only. The fix adds a dedicated `BEFORE INSERT` trigger `set_entrance_geom_on_insert` backed by a new function `set_entrance_point_geom()`. The test bootstrap is updated so the test database gets the same trigger.

The fix is correct and targets a real, impactful bug. The changes are small and focused. The points below are mostly about gaps that remain after the fix.

---

## Critical issue

### Missing deployment migration

`0_triggers.sql` is the base schema, executed only when standing up a fresh database. The project applies incremental changes via numbered migration files (e.g. `8_01_2026_02_18_populate_entrance_point_geom.sql`). **Existing deployments will never receive this trigger** unless a migration file is added.

A file like `8_02_2026_04_25_entrance_insert_trigger.sql` containing the function and trigger DDL is required. Without it, the bug remains open on every deployed instance.

---

## Code quality

### Duplicated geometry expression

The expression `ST_SetSRID(ST_MakePoint(NEW.longitude, NEW.latitude), 4326)` now exists in two places:

- `sql/0_triggers.sql:442` — inside `histo_update_entrance()`, unconditionally on every UPDATE
- `sql/0_triggers.sql` (new code) — inside `set_entrance_point_geom()`, guarded by a NULL check

These two copies will diverge over time (different SRID, different coordinate order, etc.). The cleanest fix would be to have `histo_update_entrance()` call `set_entrance_point_geom()` directly, or to merge the logic into a single `BEFORE INSERT OR UPDATE` trigger, eliminating the duplication entirely.

### NULL handling inconsistency

The new `set_entrance_point_geom()` function correctly guards against NULL lat/lon:

```sql
IF NEW.longitude IS NOT NULL AND NEW.latitude IS NOT NULL THEN
  NEW.point_geom := ST_SetSRID(ST_MakePoint(...), 4326);
END IF;
```

The existing `histo_update_entrance()` does **not** guard (line 442 in `0_triggers.sql`):

```sql
NEW.point_geom := ST_SetSRID(ST_MakePoint(NEW.longitude, NEW.latitude), 4326);
```

On UPDATE with NULL coordinates, this sets `point_geom` to an invalid/NULL geometry rather than leaving the existing value. This is a pre-existing bug, but the new code makes the inconsistency visible and is the right moment to fix it.

---

## Testing

### No targeted regression test

The test bootstrap now creates the trigger in the test database, which is correct. However, no test verifies the actual new behavior: that `point_geom` is populated immediately after an INSERT.

The existing spatial tests (`MassifCaveSpatial.property.test.js`, `GeoLocService.property.test.js`) rely on pre-seeded fixture data and would not catch a broken INSERT trigger. A simple integration test that:

1. inserts a row into `t_entrance` with known lat/lon
2. reads back `point_geom`
3. asserts it equals `ST_SetSRID(ST_MakePoint(lon, lat), 4326)`

would lock in this fix as a true regression guard.

---

## Minor

- `EXECUTE PROCEDURE` is deprecated since PostgreSQL 11 in favor of `EXECUTE FUNCTION`. The new code is consistent with the rest of the file, but worth tracking as tech debt.
- The `CREATE OR REPLACE TRIGGER` syntax requires PostgreSQL 14+. Again consistent with existing usage in the file (line 499).

---

## Summary

| Severity | Issue |
|----------|-------|
| **Critical** | No migration file — existing databases never get the trigger |
| **Medium** | Duplicated geometry expression in two trigger functions |
| **Medium** | NULL-guard inconsistency between INSERT and UPDATE paths |
| **Low** | No dedicated regression test for the INSERT trigger behavior |
| **Info** | `EXECUTE PROCEDURE` is deprecated |

The fix itself is sound and necessary. The blocker before merging is the missing migration file.

ClemRz (commented):
Thanks for the thorough review Paul — all five points addressed in `6c8c1b3`.

**Critical — Missing migration file**: Added `sql/8_02_2026_04_25_entrance_insert_trigger.sql`. It recreates both `set_entrance_point_geom()` (with the NULL guard and ELSE branch) and `histo_update_entrance()` (without the duplicated `point_geom` line), drops the old INSERT-only trigger, and creates the new `BEFORE INSERT OR UPDATE` one.

**Medium — Duplicated geometry expression**: Eliminated. `point_geom` maintenance now lives exclusively in `set_entrance_point_geom()`, which fires `BEFORE INSERT OR UPDATE`. The line in `histo_update_entrance()` is replaced with a comment pointing to the single source of truth. Went with a dedicated trigger rather than merging into `histo_update_entrance()` because that function references `OLD.*` for the history INSERT, which doesn't exist on INSERT operations.

**Medium — NULL handling inconsistency**: The trigger function now has an explicit `ELSE NEW.point_geom := NULL` branch, so NULL coordinates consistently produce a NULL geometry on both INSERT and UPDATE.

**Low — Regression test**: Added a test in `create.test.js` that creates an entrance via the API, then queries `point_geom` directly with `ST_X`/`ST_Y`/`ST_SRID` to assert it was populated on INSERT. Also fixed a gap in `scripts/snapshot-test-db.js` — its bootstrap was missing the `CREATE_ENTRANCE_POINT_GEOM_INSERT_TRIGGER` call, so the parallel test template was built without the trigger (caused the new test to fail in parallel mode while passing in single-shard runs).

**Info — `EXECUTE PROCEDURE`**: Left as-is for consistency with the rest of `0_triggers.sql`. Agreed it's worth a future cleanup pass across the whole file.

The trigger was also renamed from `set_entrance_geom_on_insert` to `set_entrance_point_geom_trigger` to reflect that it now covers both INSERT and UPDATE. `test/customSQL.js` updated to match.

Paul-AUB (approved):

PR #1542, opened 2026-04-29
fix(organization): coerce empty-string coordinates to null

ClemRz:
## 🤔 What

- Fix 500 error on `PUT /api/v1/organizations/:id` when `latitude` or `longitude` are sent as empty strings
- Add `coerceToNumeric` utility to convert empty/whitespace strings to `null` for numeric DB columns
- Apply the same fix to `CaveService` and the cave update controller (same latent bug)

## 🤷‍♂️ Why

The frontend sends `latitude: ""` and `longitude: ""` when a user clears the coordinate fields. The `TGrotto` model declares these as `type: 'string'` but the underlying PostgreSQL column is `numeric(24,20)`. Waterline passes the empty string straight to the DB adapter, which triggers a `22P02: invalid input syntax for type numeric: ""` error and returns a 500 to the user.

The `createGrotto` path had an ad-hoc `delete` workaround for this, but the update path did not. The cave endpoints had the same latent issue.

## 🔍 How

- New `api/utils/coerceToNumeric.js` — mirrors the existing `coerceToInt` pattern. Converts empty and whitespace-only strings to `null`, passes everything else through unchanged.
- `GrottoService.getConvertedDataFromClientRequest` — wraps `latitude`/`longitude` with `coerceToNumeric` so both create and update paths are covered at the extraction layer.
- `GrottoService.createGrotto` — replaces the inline `delete` hack with `coerceToNumeric` as a safety net for direct callers.
- `CaveService.getConvertedDataFromClient` — same treatment for cave latitude/longitude.
- `api/controllers/v1/cave/update.js` — applies `coerceToNumeric` when assigning coordinates to `updatedFields`.

## 🧪 Testing

- Added `test/integration/2_utils/coerceToNumeric.test.js` — unit tests for the utility plus integration tests verifying `GrottoService` and `CaveService` coerce correctly.
- Added integration test in `test/integration/4_routes/Organization/update.test.js` that reproduces the exact production scenario: sending empty-string coordinates and asserting a 200 with `null` values instead of a 500.
- All 1700 existing tests pass.

### DIFF
--- a/api/controllers/v1/cave/update.js
+++ b/api/controllers/v1/cave/update.js
@@ -1,5 +1,6 @@
 const ControllerService = require('../../../services/ControllerService');
 const coerceToInt = require('../../../utils/coerceToInt');
+const coerceToNumeric = require('../../../utils/coerceToNumeric');
 const CaveService = require('../../../services/CaveService');
 const NotificationService = require('../../../services/NotificationService');
 const { toCave } = require('../../../services/mapping/converters');
@@ -26,8 +27,10 @@ module.exports = async (req, res) => {
     // dateReviewed will be updated automaticly by the SQL historisation trigger
   };
 
-  if (newLatitude != null) updatedFields.latitude = newLatitude;
-  if (newLongitude != null) updatedFields.longitude = newLongitude;
+  if (newLatitude != null)
+    updatedFields.latitude = coerceToNumeric(newLatitude);
+  if (newLongitude != null)
+    updatedFields.longitude = coerceToNumeric(newLongitude);
   if (newDepth != null) updatedFields.depth = coerceToInt(newDepth);
   if (newLength != null) updatedFields.caveLength = coerceToInt(newLength);
   if (newTemperature != null) updatedFields.temperature = newTemperature;
--- a/api/services/CaveService.js
+++ b/api/services/CaveService.js
@@ -1,5 +1,6 @@
 const CommonService = require('./CommonService');
 const coerceToInt = require('../utils/coerceToInt');
+const coerceToNumeric = require('../utils/coerceToNumeric');
 const DocumentService = require('./DocumentService');
 const DescriptionService = require('./DescriptionService');
 const NameService = require('./NameService');
@@ -99,8 +100,8 @@ module.exports = {
     depth: coerceToInt(req.param('depth')),
     documents: req.param('documents'),
     isDiving: req.param('isDiving'),
-    latitude: req.param('latitude'),
-    longitude: req.param('longitude'),
+    latitude: coerceToNumeric(req.param('latitude')),
+    longitude: coerceToNumeric(req.param('longitude')),
     caveLength: coerceToInt(req.param('length')),
     massif: req.param('massif'),
     temperature: req.param('temperature'),
--- a/api/services/GrottoService.js
+++ b/api/services/GrottoService.js
@@ -5,6 +5,7 @@ const NameService = require('./NameService');
 const NotificationService = require('./NotificationService');
 const GeocodingService = require('./GeocodingService');
 const RecentChangeService = require('./RecentChangeService');
+const coerceToNumeric = require('../utils/coerceToNumeric');
 
 module.exports = {
   // Extract everything from a request body except id
@@ -14,8 +15,8 @@ module.exports = {
     country: req.body?.country?.id ?? null,
     county: req.param('county'),
     customMessage: req.param('customMessage'),
-    latitude: req.param('latitude'),
-    longitude: req.param('longitude'),
+    latitude: coerceToNumeric(req.param('latitude')),
+    longitude: coerceToNumeric(req.param('longitude')),
     mail: req.param('mail'),
     postalCode: req.param('postalCode'),
     region: req.param('region'),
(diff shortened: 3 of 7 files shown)


Paul-AUB (changes requested):
# Code Review — Commit 92abc62b

**Commit:** `92abc62b29b25e248e19d4b8ba647293da58f735`  
**Author:** Clément Ronzon  
**Date:** 2026-04-29  
**Subject:** `fix(organization): coerce empty-string coordinates to null before DB update`

---

## Overview

This fix addresses a PostgreSQL `22P02` error ("invalid input syntax for type numeric") triggered when the frontend sends empty strings for `latitude` / `longitude` on `PUT /api/v1/organizations/:id`. The root cause is that HTML form inputs can submit `""` instead of `null`, and the DB adapter passes that string literally to Postgres.

The solution introduces a `coerceToNumeric` utility (mirroring the existing `coerceToInt`) that converts empty/whitespace-only strings to `null`, and applies it consistently across all four write paths:
- `GrottoService.getConvertedDataFromClientRequest`
- `GrottoService.createGrotto`
- `CaveService.getConvertedDataFromClient`
- `cave/update.js` controller

---

## Code Quality

### Strengths

- **Clean utility design.** `coerceToNumeric` is small, pure, and side-effect-free. The JSDoc clearly states what it does and what it intentionally does not do (further validation is delegated downstream).
- **Consistent pattern.** Mirrors `coerceToInt` in naming and structure, making it easy to discover and reason about.
- **Ad-hoc hack removed.** The old `delete cleanedData.latitude` / `delete cleanedData.longitude` block in `createGrotto` was fragile and surprising; replacing it with the coercion utility is a clear improvement.
- **No scope creep.** The change is tightly scoped to the bug — no unrelated cleanup or refactoring.

### Issues / Suggestions

#### 1. `coerceToNumeric` does not validate that the value is actually numeric

The function passes non-numeric strings (e.g. `"abc"`, `"12.3.4"`) through unchanged. This is documented intentionally ("letting Waterline / the DB adapter handle further validation"), and the tests even assert this behaviour. However, this means the same `22P02` error will still occur for a non-empty, non-numeric string input.

**Recommendation:** Either add a `parseFloat` / `isNaN` guard, or document explicitly why upstream validation is trusted to prevent this case. The current JSDoc does partially address this, but a note in the test file would help future developers understand the design boundary.

#### 2. Double coercion in `createGrotto`

`getConvertedDataFromClientRequest` already applies `coerceToNumeric` before returning `cleanedData`. Then `createGrotto` applies it again on the same `cleanedData`. The second call is a no-op (since `null` → `null` and a valid numeric string → unchanged), but it is redundant and could confuse a reader.

```js
// In createGrotto — cleanedData.latitude is already coerced here:
cleanedData.latitude = coerceToNumeric(cleanedData.latitude);
```

**Recommendation:** Either trust that the caller (which controls `cleanedData`) has already coerced the values, or add a comment explaining why the defensive re-coercion is intentional (e.g. "createGrotto can be called directly with raw data not going through getConvertedDataFromClientRequest").

#### 3. `eslint-disable-next-line no-param-reassign` still needed

The two `no-param-reassign` suppressions remain in `createGrotto`. This is a side-effect of mutating `cleanedData` rather than creating a local copy. Not a blocker, but worth noting for a future refactor toward immutability.

#### 4. Test file location is slightly misleading

`test/integration/2_utils/coerceToNumeric.test.js` is placed in the integration test tree, but the `coerceToNumeric()` unit tests are pure unit tests (no I/O, no DB, no Sails lifting). The service-level tests (`GrottoService`, `CaveService`) also run without a DB connection and are therefore unit tests.

Only the route-level test in `4_routes/Organization/update.test.js` is a true integration test.

**Recommendation:** Move `coerceToNumeric.test.js` to a `test/unit/` directory, or at minimum rename it to make the distinction clear.

---

## Test Coverage

- **Unit tests for the utility:** comprehensive — covers `null`, `undefined`, empty string, whitespace variants, numeric strings, numbers, and non-numeric strings.
- **Service-level contract tests:** cover both `GrottoService.getConvertedDataFromClientRequest` and `CaveService.getConvertedDataFromClient` for the three key cases (empty, valid, null).
- **Integration test:** covers the original bug scenario end-to-end (empty strings → `200` response with `null` coordinates).
- **Gap:** No integration test for the cave update path (`PUT /api/v1/caves/:id` with empty coordinates). The fix is applied there in code but not exercised by any new test.

---

## Security

No security concerns. The change narrows the attack surface slightly by preventing a crash path that could leak DB error details in non-production environments.

---

## Summary

| | |
|---|---|
| **Verdict** | ✅ Approve with minor notes |
| **Blocking issues** | None |
| **Non-blocking notes** | Double coercion in `createGrotto`, test file location, missing cave-update integration test |
| **Risk** | Low — the utility is pure and the change is additive |

ClemRz (commented):
Thanks for the thorough review! Pushed an update addressing the feedback:

**1. Non-numeric string validation** — Added a comment in the test file documenting the design boundary: `coerceToNumeric` intentionally only handles blank→null coercion; actual numeric validation is delegated to Waterline / the DB adapter.

**2. Double coercion in `createGrotto`** — Added a comment explaining the defensive re-coercion is intentional: `createGrotto` can be called directly with raw data that hasn't gone through `getConvertedDataFromClientRequest`.

**3. `no-param-reassign` suppressions** — Kept as-is since the defensive re-coercion is intentional (see above). Agreed it's worth revisiting in a future immutability refactor.

**4. Missing cave update integration test** — Added a test for `PUT /api/v1/caves/:id` with empty-string coordinates. Note: the assertion checks the DB record directly since `latitude`/`longitude` are deprecated on `TCave` and not exposed by the `toCave` converter.

All 2071 tests pass.

Paul-AUB (approved):

PR #1555, opened 2026-05-02
fix(massif): make description optional in massif creation

### LINKED ISSUE #1553: Make Massif description optional in Massif creation
# [API] `description` should not be required when creating a massif

## Summary

The `POST /api/v1/massifs` endpoint currently requires `description` and `descriptionTitle` as mandatory parameters. This constraint is too strict — a massif should be creatable without a description, which can be added later.

## Current behavior

In `api/controllers/v1/massif/create.js`, both fields are listed in `requiredParams`:

```js
const requiredParams = [
  'name',
  'description',           // ← should be optional
  'descriptionAndNameLanguage',
  'descriptionTitle',      // ← should be optional
  'geogPolygon',
];
```

If either is missing, the API returns a `400 Bad Request`:
```
description,descriptionTitle parameter(s) must be provided
```

As a result, currently most of the Massifs are displayed weirly with blank description (but not counted as "no description):

Image

## Expected behavior

- `description` and `descriptionTitle` should be **optional**.
- If both are provided, the description is created as today.
- If omitted, the massif is created without a description (which can be added later via the existing description endpoints).

## Fix

Remove `'description'` and `'descriptionTitle'` from `requiredParams` in `api/controllers/v1/massif/create.js`.

Also update `assets/swaggerV1.yaml` to reflect that these fields are no longer required.

ClemRz:
## 🤔 What

- Make `description` and `descriptionTitle` optional when creating a massif via `POST /api/v1/massifs`
- Add validation that `description` and `descriptionTitle` must be provided together (both or neither)
- Update Swagger spec to reflect the change
- Closes #1553

## 🤷‍♂️ Why

The `POST /api/v1/massifs` endpoint currently requires `description` and `descriptionTitle` as mandatory parameters. This is too strict — a massif should be creatable without a description, which can be added later via the existing description endpoints. This constraint was leading to massifs being created with blank descriptions just to satisfy the validation.

## 🔍 How

- Removed `description` and `descriptionTitle` from the `requiredParams` array in `api/controllers/v1/massif/create.js`
- Added a paired validation: if one of `description`/`descriptionTitle` is provided, the other must be too (returns 400 otherwise)
- The existing conditional `if (req.body?.description)` in the transaction block already handles skipping description creation when not provided
- Updated `assets/swaggerV1.yaml` to mark both parameters as `required: false`

## 🧪 Testing

- Added test: massif creation succeeds (200) without description fields
- Added test: returns 400 when `description` is provided without `descriptionTitle`
- Added test: returns 400 when `descriptionTitle` is provided without `description`
- Existing tests for complete data creation still pass

## 📸 Previews

Creating a massif without a description (local dev):

```bash
curl -s -X POST http://localhost:1337/api/v1/massifs \
  -H "Authorization: Bearer $TOKEN" \
  -d '{
    "name": "Massif de la Sainte-Baume",
    "descriptionAndNameLanguage": { "id": "fra" },
    "geogPolygon": {
      "type": "Polygon",
      "coordinates": [[[5.6, 43.3], [5.8, 43.3], [5.8, 43.4], [5.6, 43.4], [5.6, 43.3]]]
    }
  }'
```

Response (200):

```json
{
  "id": 491,
  "name": "Massif de la Sainte-Baume",
  "descriptions": [],
  "documents": [],
  "networks": [],
  "geogPolygon": "{\"type\":\"Polygon\",\"coordinates\":[[[5.6,43.3],[5.8,43.3],[5.8,43.4],[5.6,43.4],[5.6,43.3]]]}"
}
```

Partial description rejected (400):

```bash
curl -s -X POST http://localhost:1337/api/v1/massifs \
  -H "Authorization: Bearer $TOKEN" \
  -d '{
    "name": "Partial Test",
    "description": "only body, no title",
    "descriptionAndNameLanguage": { "id": "fra" },
    "geogPolygon": {
      "type": "Polygon",
      "coordinates": [[[5.6, 43.3], [5.8, 43.3], [5.8, 43.4], [5.6, 43.4], [5.6, 43.3]]]
    }
  }'
# => {"message":"description and descriptionTitle must be provided together"}
```

### DIFF
--- a/api/controllers/v1/massif/create.js
+++ b/api/controllers/v1/massif/create.js
@@ -10,13 +10,7 @@ const RightService = require('../../../services/RightService');
 // eslint-disable-next-line consistent-return
 module.exports = async (req, res) => {
   // Check params
-  const requiredParams = [
-    'name',
-    'description',
-    'descriptionAndNameLanguage',
-    'descriptionTitle',
-    'geogPolygon',
-  ];
+  const requiredParams = ['name', 'descriptionAndNameLanguage', 'geogPolygon'];
 
   let i = 0;
   const missingParamaters = [];
@@ -28,6 +22,15 @@ module.exports = async (req, res) => {
   }
   if (missingParamaters.length > 0) {
     return res.badRequest(`${missingParamaters} parameter(s) must be provided`);
+  }
+
+  // If a description is provided, its title must also be provided (and vice versa)
+  const hasDescription = !!req.param('description');
+  const hasDescriptionTitle = !!req.param('descriptionTitle');
+  if (hasDescription !== hasDescriptionTitle) {
+    return res.badRequest(
+      'description and descriptionTitle must be provided together'
+    );
   }
 
   // Validate name length
--- a/assets/swaggerV1.yaml
+++ b/assets/swaggerV1.yaml
@@ -4344,14 +4344,14 @@ paths:
       parameters:
         - name: description
           in: query
-          required: true
-          description: Description of the massif
+          required: false
+          description: Description of the massif. If provided, descriptionTitle must also be provided.
           schema:
             type: string
         - name: descriptionTitle
           in: query
-          required: true
-          description: Description title of the massif
+          required: false
+          description: Description title of the massif. If provided, description must also be provided.
           schema:
             type: string
         - name: descriptionAndNameLanguage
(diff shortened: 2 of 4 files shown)


Paul-AUB (approved):
LGTM !

PR #1556, opened 2026-05-02
feat(qualitySort): add sort and order query params to with-quality endpoints

### LINKED ISSUE #1512: Sort/order param for */with-quality/* endpoints
Yes : https://github.com/GrottoCenter/grottocenter-front/issues/1245

I want to have a `sort` parameter to be processed **before** pagination, to be able to display in the front the quality in decreasing order (at least).

Seems reasonable to have also `order` asc or desc, and maybe on what data we sort.

N/a

**Additional context**
N/a

ClemRz:
## 🤔 What

- Add server-side `sort` and `order` query parameters to the three `*/with-quality/*` endpoints
- Sorting is applied at the SQL level **before** pagination (`ORDER BY` before `LIMIT`/`OFFSET`)
- Closes #1512

Affected endpoints:
- `GET /api/v1/entrances/with-quality/massifs/:id`
- `GET /api/v1/entrances/with-quality/countries/:id`
- `GET /api/v1/entrances/with-quality/countries/:countryId/regions/:regionId`

## 🤷‍♂️ Why

The front-end currently sorts only the current page of results client-side (see [grottocenter-front#1245](https://github.com/GrottoCenter/grottocenter-front/issues/1245)). This means page 1 doesn't necessarily show the best-quality caves — a higher-quality cave could be on page 3. Server-side sorting before pagination fixes this.

## 🔍 How

- **Shared validation utility** (`api/utils/validateSortParams.js`): extracts and validates `sort` and `order` query params against a strict allow-list of 18 columns from the `v_data_quality_compute_entrance` materialized view. Returns validated values or sends a 400 response with a descriptive error message.
- **Service layer** (`DataQualityComputeService.js`): a shared `applySort()` helper injects `ORDER BY <column> <direction>` into the SQL string before the `LIMIT` clause. The column and direction are interpolated directly (safe because allow-list validation has already occurred). Count queries are unchanged.
- **Controllers**: all three `get-entrances-data-quality.js` controllers call `validateSortParams()` and pass the result to the service.
- **Swagger**: all three endpoint paths updated with `sort` (enum of 18 columns) and `order` (enum: asc/desc, default: asc) parameter definitions, plus 400 error documentation.

**Note**: the computed `data_quality` score is calculated in JavaScript post-fetch, not stored in the materialized view, so it cannot be sorted server-side. This is a known limitation — see [comment on #1512](https://github.com/GrottoCenter/grottocenter-api/issues/1512#issuecomment-4364471871).

## 🧪 Testing

64 tests passing (27 unit/property/service + 37 integration):

- **Unit tests** (`test/integration/2_utils/validateSortParams.test.js`): 15 tests covering valid/invalid sort, valid/invalid order, case-insensitivity, defaults, edge cases
- **Property tests** (`test/integration/2_utils/validateSortParams.property.test.js`): 3 fast-check properties (100 runs each) — valid ORDER BY construction, invalid sort rejection, invalid order rejection
- **Service tests** (`test/integration/1_services/DataQualityComputeService.sort.test.js`): 9 tests verifying SQL construction for all 6 service methods
- **Integration tests** (shared via `test/integration/4_routes/_helpers/sortOrderTests.js`): 5 tests per endpoint × 3 endpoints — valid sort+order, invalid sort, invalid order, sort without order, order without sort

```bash
npm run test -- --grep "data quality|validateSortParams|DataQualityComputeService - sort"
```

## 📸 Previews

```bash
# Sort by name descending
curl 'http://localhost:1337/api/v1/entrances/with-quality/countries/FR?sort=entrance_name&order=desc&limit=3'

# Invalid sort column → 400
curl 'http://localhost:1337/api/v1/entrances/with-quality/countries/FR?sort=hacked_column'
# → {"message":"Invalid sort column 'hacked_column'. Valid columns: entrance_name, ..."}

# Invalid order → 400
curl 'http://localhost:1337/api/v1/entrances/with-quality/countries/FR?sort=entrance_name&order=sideways'
# → {"message":"Invalid order value 'sideways'. Must be 'asc' or 'desc'"}
```

### DIFF
--- a/api/controllers/v1/country/get-entrances-data-quality.js
+++ b/api/controllers/v1/country/get-entrances-data-quality.js
@@ -4,6 +4,11 @@ const {
   toQualityDataEntrance,
 } = require('../../../services/mapping/converters');
 const { toListFromController } = require('../../../services/mapping/utils');
+const {
+  SORTABLE_COLUMNS_COUNTRY,
+  VALIDATION_ERROR,
+  validateSortParams,
+} = require('../../../utils/validateSortParams');
 
 module.exports = async (req, res) => {
   // Get entrances and the informations associated at each entrance
@@ -11,11 +16,18 @@ module.exports = async (req, res) => {
   const limit = Math.min(parseInt(req.param('limit', 50), 10), 1000);
   const offset = Math.max(parseInt(req.param('offset', 0), 10), 0);
 
+  const sortResult = validateSortParams(req, res, SORTABLE_COLUMNS_COUNTRY);
+  if (sortResult === VALIDATION_ERROR) return null;
+  const sort = sortResult ? sortResult.sort : null;
+  const order = sortResult ? sortResult.order : null;
+
   const [entrancesInformationToCompute, totalCount] = await Promise.all([
     DataQualityComputeService.getEntrancesWithQualityByCountry(
       countryId,
       limit,
-      offset
+      offset,
+      sort,
+      order
     ),
     DataQualityComputeService.getEntrancesWithQualityByCountryCount(countryId),
   ]);
--- a/api/controllers/v1/massif/get-entrances-data-quality.js
+++ b/api/controllers/v1/massif/get-entrances-data-quality.js
@@ -4,6 +4,11 @@ const {
   toQualityDataEntrance,
 } = require('../../../services/mapping/converters');
 const { toListFromController } = require('../../../services/mapping/utils');
+const {
+  SORTABLE_COLUMNS,
+  VALIDATION_ERROR,
+  validateSortParams,
+} = require('../../../utils/validateSortParams');
 
 module.exports = async (req, res) => {
   // Get entrances and the informations associated at each entrance
@@ -11,11 +16,18 @@ module.exports = async (req, res) => {
   const limit = Math.min(parseInt(req.param('limit', 50), 10), 1000);
   const offset = Math.max(parseInt(req.param('offset', 0), 10), 0);
 
+  const sortResult = validateSortParams(req, res, SORTABLE_COLUMNS);
+  if (sortResult === VALIDATION_ERROR) return null;
+  const sort = sortResult ? sortResult.sort : null;
+  const order = sortResult ? sortResult.order : null;
+
   const [entrancesInformationToCompute, totalCount] = await Promise.all([
     DataQualityComputeService.getEntrancesWithQualityByMassif(
       massifId,
       limit,
-      offset
+      offset,
+      sort,
+      order
     ),
     DataQualityComputeService.getEntrancesWithQualityByMassifCount(massifId),
   ]);
--- a/api/controllers/v1/region/get-entrances-data-quality.js
+++ b/api/controllers/v1/region/get-entrances-data-quality.js
@@ -4,6 +4,11 @@ const {
   toQualityDataEntrance,
 } = require('../../../services/mapping/converters');
 const { toListFromController } = require('../../../services/mapping/utils');
+const {
+  SORTABLE_COLUMNS_COUNTRY,
+  VALIDATION_ERROR,
+  validateSortParams,
+} = require('../../../utils/validateSortParams');
 
 module.exports = async (req, res) => {
   const { countryId } = req.params;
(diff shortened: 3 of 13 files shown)


Paul-AUB (changes requested):
# Issue: feat(qualitySort) — bugs and risks in sort/order implementation

**Commit:** `1d76cac0e7128c56537789e4d0f53e4d1399bcaf`
**Author:** Clément Ronzon
**Endpoints affected:** `GET /entrances/with-quality/massifs/:id`, `.../countries/:id`, `.../regions/:id`

---

## Bug: Sorting silently dropped when no pagination params are provided

In `api/services/DataQualityComputeService.js`, when `limit === null`, the query template has `LIMIT $2 OFFSET $3` stripped **before** `applySort` runs. `applySort` does `sql.replace('LIMIT', ...)`, which is a no-op on the already-stripped string — the `ORDER BY` clause is never injected and no error is raised.

**Reproducer:** call any of the three endpoints with `?sort=entrance_name` but without `limit`/`offset` params — results are returned unsorted.

**Fix:** apply `applySort` before stripping the LIMIT clause:

```js
// In each getEntrancesWithQualityBy* method:
let query = applySort(GET_ENTRANCES_WITH_QUALITY_BY_MASSIF, sort, order);
if (limit === null || offset === null) {
  query = query.replace('LIMIT $2 OFFSET $3', '');
}
```

---

## Risk: Cross-endpoint column availability — `massif_name`

`SORTABLE_COLUMNS` in `api/utils/validateSortParams.js` is shared across all three endpoints, but the country and region query strings use explicit `SELECT` column lists that do **not** include `massif_name`. The allow-list only prevents SQL injection — it does not verify that the column exists in the view for the specific query being run.

Sorting by `massif_name` on the country or region endpoints will produce a PostgreSQL runtime error.

**Fix options:**
- Use per-endpoint allow-lists, or
- Confirm `v_data_quality_compute_entrance` exposes `massif_name` for all query paths and add explicit integration tests for `?sort=massif_name` on the country and region endpoints.

---

## Code smell: String sentinel `'error'` return value

`validateSortParams` returns the string `'error'` to signal that a 400 has already been sent. A caller that forgets the `=== 'error'` check will treat it as a truthy success value and attempt to destructure `.sort` / `.order` from it.

**Suggestion:** export the sentinel as a named constant, or use `null` with a separate `errorSent` flag, or throw — anything that fails loudly rather than silently.

---

## Minor: Brittle `replace('LIMIT', ...)` in `applySort`

`applySort` (line 21 of `DataQualityComputeService.js`) replaces the first literal occurrence of the word `LIMIT`. This works today but would silently mis-position `ORDER BY` if a subquery or CTE with its own `LIMIT` is introduced in future. A named placeholder in the template strings (e.g. `/*{{ORDER_BY}}*/`) would be more explicit and future-proof.

---

## Minor: Swagger parameter duplication

The `sort` and `order` parameter blocks are copy-pasted verbatim into all three endpoint definitions in `assets/swaggerV1.yaml`. If the allowed column list ever changes, all three blocks must be updated in sync. Extracting them as `components/parameters` with `$ref` would eliminate the duplication.

ClemRz:
Thanks for the thorough review @Paul-AUB — all points addressed. Here's what changed:

### 1. BUG: Sorting silently dropped without pagination params ✅

Root cause: `applySort` relied on finding the literal `LIMIT` keyword, but when `limit=null` the LIMIT clause was stripped *before* `applySort` ran, making it a no-op.

Fix: introduced separate `_NO_LIMIT` query templates that still contain the `ORDER_BY_PLACEHOLDER` (see point 4), so sorting works regardless of whether pagination is present. Added service-level tests that verify `ORDER BY` appears in the SQL even when `limit=null`.

### 2. RISK: `massif_name` not available on country/region endpoints ✅

The country and region queries use explicit `SELECT` lists that don't include `massif_name`.

Fix: added `SORTABLE_COLUMNS_COUNTRY` (17 columns, excluding `massif_name`). Country and region controllers now pass it to `validateSortParams`. The massif endpoint keeps the full 18-column list. Integration tests verify that `?sort=massif_name` returns 400 on country/region endpoints.

### 3. CODE SMELL: String sentinel `'error'` ✅

Replaced with `VALIDATION_ERROR = Symbol.for('validateSortParams.error')`. Callers compare against the exported Symbol — can't be confused with a truthy string.

### 4. MINOR: Brittle `replace('LIMIT', ...)` ✅

Replaced with a named placeholder `/*{{ORDER_BY}}*/` that `applySort` substitutes. Won't collide with SQL keywords in subqueries or CTEs.

### 5. MINOR: Swagger duplication ✅

Extracted three shared parameters in `components/parameters`: `qualitySortAll` (18 columns, for massif), `qualitySortCountry` (17 columns, for country/region), and `qualityOrder`. All three endpoints now use `$ref`.

---

All 2133 tests pass, 0 lint errors.

Paul-AUB (approved):

PR #1558, opened 2026-05-02
feat(notification): notify document author on validation and rejection

### LINKED ISSUE #1557: feat(notification): notify document author on validation and rejection
## Summary

When a moderator validates or rejects a document, the submitting user should be notified — both in-app and optionally by email. Currently, the `VALIDATE` notification is only sent to geographic subscribers (country/massif/region), and rejection sends no notification at all.

Related front-end issues:
- GrottoCenter/grottocenter-front#1243
- GrottoCenter/grottocenter-front#988

### What exists today

- `NotificationService.notifySubscribers()` sends in-app notifications (+ optional email) to users subscribed to the geographic area of the entity. It does **not** notify the document author.
- `PUT /api/v1/documents/:id/validate` and `PUT /api/v1/documents/validate` (batch) call `notifySubscribers` on acceptance but do **nothing** on rejection.
- `TNotification` already has a `document` field.
- `NOTIFICATION_TYPES` includes `VALIDATE` but has no `REJECT` type.
- `TCaver.sendNotificationByEmail` exists as a global boolean opt-in — there is no per-event-type granularity yet.
- The `toCaver` converter explicitly excludes `mail` from the response (RGPD comment), so email is never returned on the profile endpoint.

### 1. Add a `REJECT` notification type

- Add a `REJECT` entry to `NOTIFICATION_TYPES` in `NotificationService`.
- Insert the corresponding row in `t_notification_type` via a SQL migration.
- Add a fixture entry for tests.

### 2. Notify the document author on validation decisions

- In `validate.js` and `multiple-validate.js`, after the document is updated:
  - On **acceptance**: create a `TNotification` for the document's `author` with type `VALIDATE` (in addition to the existing subscriber notifications).
  - On **rejection**: create a `TNotification` for the document's `author` with type `REJECT`.
- The notifier is the moderator (`req.token.id`).
- Skip the author notification if the moderator is also the author (self-validation).

### 3. Send email notification to the document author

- When creating the author notification, if the author has `sendNotificationByEmail === true`, send an email.
- For rejections, the email should include the `validationComment` (rejection reason). This will require either a new email template or an extension of the existing `notification.ejs` template to conditionally display a rejection comment.

### 4. Per-event-type email notification preferences (optional / stretch)

The current `sendNotificationByEmail` is a single boolean. Issue #1243 describes per-event-type preferences (e.g., opt out of email for document validation but keep it for other events). This could be:
- A new `t_notification_preference` table with `(caver_id, notification_type_id, email_enabled)` rows.
- Or a JSON column on `t_caver`.

This is a nice-to-have for this issue. The simpler approach is to use the existing global `sendNotificationByEmail` flag for now and add granularity later.

## Acceptance criteria

- [ ] When a document is accepted, the submitting user receives an in-app notification of type `VALIDATE`.
- [ ] When a document is rejected, the submitting user receives an in-app notification of type `REJECT`.
- [ ] The rejection notification includes or references the `validationComment`.
- [ ] If the author has opted into email notifications, they receive an email for both acceptance and rejection.
- [ ] The rejection email includes the rejection reason (`validationComment`).
- [ ] Batch validation (`multiple-validate.js`) also triggers author notifications.
- [ ] The author is not notified if they are the same person as the moderator.
- [ ] Existing geographic subscriber notifications remain unchanged.

## Technical notes

- The `notifySubscribers` function is designed for geographic fan-out. Author notification is a different pattern (single targeted recipient), so it may be cleaner as a separate helper (e.g., `notifyAuthor`) rather than shoehorning it into `notifySubscribers`.
- The `notification.ejs` email template currently expects `subscriptionName` and `subscriptionType` — these don't apply to author notifications, so the template needs a conditional path or a new template.

ClemRz:
## 🤔 What

Notify the document author when a moderator validates or rejects their document.

- Add a `REJECT` notification type (constant, SQL migration, fixture)
- Add `notifyAuthor()` helper in `NotificationService` for targeted author notifications
- Integrate into both `validate.js` (single) and `multiple-validate.js` (batch) controllers
- Update `notification.ejs` email template to distinguish author notifications from geographic subscriber notifications, with rejection emails including the moderator's comment
- Fix pre-existing bug: `multiple-validate.js` passed `res` instead of `req` to `updateSearchAndNotify`

Closes #1557
Related front-end issues: GrottoCenter/grottocenter-front#1243, GrottoCenter/grottocenter-front#988

## 🤷‍♂️ Why

When a moderator validates or rejects a document, the submitting author had no way of knowing unless they checked manually. The existing `notifySubscribers` flow only notifies geographic subscribers (country/massif/region) on acceptance, and rejection triggered no notification at all.

## 🔍 How

- `notifyAuthor()` is a separate helper from `notifySubscribers()` because author notification is a single targeted recipient, not a geographic fan-out. It resolves the document's author, skips self-notification (moderator === author), creates a `TNotification` record, and optionally sends an email if the author has `sendNotificationByEmail` enabled.
- The email template uses an `isAuthorNotification` flag to switch between author context ("you are the author of this document") and the existing subscription context. Rejection emails include the `validationComment` rendered with `<%= %>` (HTML-escaped) to prevent XSS.
- Both controllers call `notifyAuthor` with `.catch()` so notification failures never break the validation response.

**Behavioral change:** The single-validate endpoint now returns 200 on rejection (previously returned 404 because `populatedDoc` was `undefined` on the rejection path). The document is now populated for both paths since `notifyAuthor` needs it.

**Front-end action needed:** The notification center UI needs to be updated to handle the new `REJECT` notification type (display it with appropriate messaging and styling, distinct from `VALIDATE`).

## 🧪 Testing

- 5 property-based tests (self-notification skip, notifier identity, email opt-in gate, rejection comment inclusion, template context distinction)
- 7 integration tests for `notifyAuthor()` service
- 3 template rendering tests
- 4 route tests for `validate.js` (accept, reject, self-skip, geographic subscriber non-regression)
- 4 route tests for `multiple-validate.js` (accept, reject, self-skip, mixed batch)

Run `npm run test` — all 2098 tests pass.

## 📸 Previews
Screenshot 2026-05-02 at 3 58 37 p m
Screenshot 2026-05-02 at 3 57 27 p m
Screenshot 2026-05-02 at 3 50 39 p m

### DIFF
--- a/api/controllers/v1/document/multiple-validate.js
+++ b/api/controllers/v1/document/multiple-validate.js
@@ -85,6 +85,8 @@ async function updateSearchAndNotify(req, documentId, userId) {
     NotificationService.NOTIFICATION_TYPES.VALIDATE,
     NotificationService.NOTIFICATION_ENTITIES.DOCUMENT
   );
+
+  return document;
 }
 
 module.exports = async (req, res) => {
@@ -121,6 +123,7 @@ module.exports = async (req, res) => {
   const documentIds = documentChanges.map((e) => e.id);
   const foundDocuments = await TDocument.find({ id: documentIds });
 
+  // Sequential to preserve partial-success semantics per document
   for (const document of foundDocuments) {
     const change = documentChanges.find((d) => d.id === document.id);
     const isAModifiedDoc = !!document.modifiedDocJson;
@@ -131,6 +134,24 @@ module.exports = async (req, res) => {
         document.id,
         change.validationComment,
         req.token.id
+      );
+      // eslint-disable-next-line no-await-in-loop
+      const rejectedDoc = await DocumentService.getPopulatedDocument(
+        document.id
+      );
+      // eslint-disable-next-line no-await-in-loop
+      await NotificationService.notifyAuthor(
+        req,
+        rejectedDoc,
+        req.token.id,
+        NotificationService.NOTIFICATION_TYPES.REJECT,
+        change.validationComment
+      ).catch((err) =>
+        sails.log.error(
+          'Document multiple-validate notifyAuthor error',
+          document,
+          err
+        )
       );
       continue; // eslint-disable-line no-continue
     }
@@ -153,13 +174,35 @@ module.exports = async (req, res) => {
     }
 
     // eslint-disable-next-line no-await-in-loop
-    await updateSearchAndNotify(res, document.id, req.token.id).catch((err) =>
+    const populatedDoc = await updateSearchAndNotify(
+      req,
+      document.id,
+      req.token.id
+    ).catch((err) => {
       sails.log.error(
         'Document multiple validate updateSearchAndNotify error',
         document,
         err
-      )
-    );
+      );
+      return null;
+    });
+
+    if (populatedDoc) {
+      // eslint-disable-next-line no-await-in-loop
+      await NotificationService.notifyAuthor(
+        req,
+        populatedDoc,
+        req.token.id,
+        NotificationService.NOTIFICATION_TYPES.VALIDATE,
+        change.validationComment
+      ).catch((err) =>
+        sails.log.error(
+          'Document multiple-validate notifyAuthor error',
+          document,
+          err
+        )
+      );
+    }
   }
 
   return res.ok();
--- a/api/controllers/v1/document/validate.js
+++ b/api/controllers/v1/document/validate.js
@@ -35,9 +35,9 @@ module.exports = async (req, res) => {
     validator: req.token.id,
   });
 
-  let populatedDoc;
+  const populatedDoc = await DocumentService.getPopulatedDocument(id);
+
   if (isValidated) {
-    populatedDoc = await DocumentService.getPopulatedDocument(id);
     await DocumentService.updateInSearch(populatedDoc);
 
     await NotificationService.notifySubscribers(
@@ -48,6 +48,19 @@ module.exports = async (req, res) => {
       NotificationService.NOTIFICATION_ENTITIES.DOCUMENT
     );
   }
+
+  // Notify author for both acceptance and rejection
+  await NotificationService.notifyAuthor(
+    req,
+    populatedDoc,
+    req.token.id,
+    isValidated
+      ? NotificationService.NOTIFICATION_TYPES.VALIDATE
+      : NotificationService.NOTIFICATION_TYPES.REJECT,
+    validationComment
+  ).catch((err) =>
+    sails.log.error('Document validate notifyAuthor error', err)
+  );
 
   const params = {
     controllerMethod: 'DocumentController.validate',
--- a/api/services/NotificationService.js
+++ b/api/services/NotificationService.js
@@ -21,6 +21,7 @@ const NOTIFICATION_TYPES = {
   UPDATE: 'UPDATE',
   VALIDATE: 'VALIDATE',
   RESTORE: 'RESTORE',
+  REJECT: 'REJECT',
 };
 
 async function removeOlderNotifications() {
@@ -68,6 +69,7 @@ const sendNotificationEmail = async (
     [NOTIFICATION_TYPES.UPDATE]: 'updated',
     [NOTIFICATION_TYPES.VALIDATE]: 'validated',
     [NOTIFICATION_TYPES.RESTORE]: 'restored',
+    [NOTIFICATION_TYPES.REJECT]: 'rejected',
   };
 
   const actionVerb = actionVerbMap[notificationType];
@@ -133,6 +135,8 @@ const sendNotificationEmail = async (
         recipientName: user.nickname,
         subscriptionName: user.subscriptionName,
         subscriptionType: user.subscriptionType,
+        isAuthorNotification: user.isAuthorNotification || false,
+        validationComment: user.validationComment || null,
       },
     })
     .intercept('sendSESEmailError', () => {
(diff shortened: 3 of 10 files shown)


Paul-AUB (changes requested):
# Issue: feat(notification) — bugs and risks in author notification implementation

**Commit:** `450de79c5b67775f85cb87e7d8a000fac83c5eb8`
**Author:** Clément Ronzon
**Feature:** Notify document author on validation and rejection

---

## Overview

Adds author notifications (in-app + optional email) when a document is validated or rejected. Introduces `NotificationService.notifyAuthor`, a new `REJECT` notification type, a SQL migration, updates both validate controllers, and extends the email template. Test coverage is thorough (unit, property-based, and integration).

---

## Bug: Null crash if REJECT migration has not been applied

In `NotificationService.notifyAuthor` ([NotificationService.js:522-531](api/services/NotificationService.js#L522-L531)):

```js
const notificationTypeRecord = await TNotificationType.findOne({
  name: notificationType,
});
await TNotification.create({
  notificationType: notificationTypeRecord.id, // TypeError if record is null
  ...
});
```

If `sql/10_01_add_reject_notification_type.sql` has not been applied, `findOne` returns `null` and the next line throws `TypeError: Cannot read properties of null (reading 'id')`. The surrounding `try/catch` swallows it silently (returns `false`, logs a generic message), leaving the moderator with no feedback and the author unnotified.

**Fix:** add a guard before accessing `.id`:

```js
if (!notificationTypeRecord) {
  throw new Error(`Notification type '${notificationType}' not found in DB — migration may not have run`);
}
```

---

## Risk: `getPopulatedDocument` called twice on the same document

In `validate.js` ([validate.js:39-51](api/controllers/v1/document/validate.js#L39-L51)), both branches of the `if (isValidated)` block call `DocumentService.getPopulatedDocument(id)` independently. The result could be fetched once before the branch:

```js
// Current — fetches twice
if (isValidated) {
  populatedDoc = await DocumentService.getPopulatedDocument(id); // call 1
  ...
} else {
  populatedDoc = await DocumentService.getPopulatedDocument(id); // call 2
}
```

In the rejection path of `multiple-validate.js` ([multiple-validate.js:138-145](api/controllers/v1/document/multiple-validate.js#L138-L145)), `getPopulatedDocument` is called again inline as a function argument, adding a third DB round-trip for a document that was already fetched at the start of the loop iteration.

Neither is a correctness issue, but both are unnecessary DB calls on a likely hot path.

---

## Code smell: Inline `await` inside function argument

In the rejection branch of `multiple-validate.js` ([multiple-validate.js:138-145](api/controllers/v1/document/multiple-validate.js#L138-L145)):

```js
await NotificationService.notifyAuthor(
  req,
  await DocumentService.getPopulatedDocument(document.id), // inline await
  ...
)
```

Assign the result to a variable first — this makes the code easier to read, allows the value to be reused, and prevents the result being silently swallowed if the outer `.catch()` only catches the `notifyAuthor` error, not the `getPopulatedDocument` error.

---

## Minor: `no-await-in-loop` suppression without explanation

The `/* eslint-disable no-await-in-loop */` / `/* eslint-enable */` comments in `multiple-validate.js` suppress the rule without explaining why sequential execution is needed here. Add a brief comment (e.g., "sequential to preserve partial-success semantics per document") so the next reader doesn't have to re-derive the reasoning.

---

## What's Good

- **Resilient by design:** `notifyAuthor` is fire-and-forget — errors are caught and logged without aborting the validate response. This is the right call for a secondary side effect.
- **Self-notification guard:** `if (authorId === moderatorId) return true` correctly skips notifications when the moderator is also the author.
- **Email template is XSS-safe:** `<%= validationComment %>` uses EJS's HTML-escaping output tag, not the unescaped `<%-`.
- **Comprehensive tests:** unit, property-based (fast-check), and integration tests cover the new `notifyAuthor` method and route-level behavior well.
- **Migration + fixture in sync:** `sql/10_01_add_reject_notification_type.sql` and `test/fixtures/tnotificationtype.json` are both updated consistently.

---

## Summary

| Severity | Finding |
|---|---|
| **Bug** | `notificationTypeRecord` can be `null` if migration not applied → silent `TypeError` |
| **Risk** | `getPopulatedDocument` called 2–3× for the same document per request |
| **Code smell** | Inline `await` in function argument in rejection path of `multiple-validate.js` |
| **Minor** | `no-await-in-loop` suppression lacks explanation |

ClemRz:
Thanks for the thorough review @Paul-AUB — all four items were legitimate. Addressed in the amended commit:

**Bug: Null crash if migration not applied** — Added a guard before accessing `notificationTypeRecord.id`. If `findOne` returns `null`, it now throws a descriptive error (`Notification type '...' not found in DB — migration may not have run`) instead of a cryptic `TypeError`.

**Risk: `getPopulatedDocument` called multiple times** — In `validate.js`, hoisted the `getPopulatedDocument` call above the `if (isValidated)` branch so it's fetched once and reused for both search update and author notification. In `multiple-validate.js`, extracted the inline `await` in the rejection path to a `rejectedDoc` variable, eliminating the extra DB round-trip and the readability issue (also addresses the **code smell** item).

**Minor: `no-await-in-loop` suppression without explanation** — Added a comment above the loop: `// Sequential to preserve partial-success semantics per document`. Also replaced the `/* eslint-disable/enable */` block comments with per-line `eslint-disable-next-line` comments for consistency with the rest of the file.

All 2101 tests pass.

Paul-AUB (approved):

PR #1223, opened 2023-07-19
fix(#1157): update request rate limiting

Clm-Roig:
Close #1157

- Rate limiting is not using database rights anymore.
- Added limiting for leaders. 

A lot of code is copy / pasted. I tried to refactor it but I'm not satisfied enough. How can I improve it ?

**EDIT:** Ok I found a way. The package allows the **max** (number of requests) value to be computed using a function (instead of a number). I will deduce the rate limiting depending on the user groups.

PR #1563, opened 2026-05-04
refactor(rateLimit): introduce tiered rate limiting by role

### LINKED ISSUE #1157: [API] Limiter le nombre de requêtes par unité de temps
Les Apis sont utilisées pour télécharger l'ensemble des données de Grottocenter.
Cela conduit à une surcharge de notre infrastructure, à une augmentation des coûts et à un ralentissement des réponses de Grottocenter pour l'ensemble des utilisateurs

Mise en place d'une limitation dans l'utilisation des Apis, selon les règles ci-dessous qui ont fait l'objet d'une discussion au sein du CA de l'assocation Wikicaves
image

**A prendre en compte**
Si c'est possible ne mettre en place cette restriction que lorsqu'il sera possible de télécharger les données avec une autre méthode selon ce qui est décrit ici https://github.com/GrottoCenter/grottocenter-api/issues/1118

ClemRz:
## 🤔 What

Replaces the binary skip/limit rate limiting with graduated, role-based rate limits. Consolidates the two separate DELETE rate limiters into one.

Supersedes #1223.

- Closes #1157

## 🤷‍♂️ Why

The previous rate limiter had two problems:

1. **Flat limits** — everyone below Moderator got the same cap (100 req/30s), with Moderators and Admins skipping entirely. No middle ground for Leaders or regular authenticated users.
2. **Too-short window** — a 30-second window with 100 requests is easy to hit on the homepage alone (~25 API calls per load). A 10-minute window with tiered limits better matches real usage patterns.

Issue #1157 requested tiered limits per role to prevent bulk data scraping while keeping the site usable for normal visitors.

## 🔍 How

**`config/rateLimit/rateLimiter.js`** — Rewrote with two middleware:

| Middleware | Window | Visitor | User | Leader | Moderator | Admin |
|---|---|---|---|---|---|---|
| `generalRateLimit` | 10 min | 200 | 400 | 1,000 | 10,000 | skip |
| `deleteRateLimit` | 1 hour | 1 | 1 | 20 | 20 | skip |

Uses `express-rate-limit`'s `max` as a function to return different limits based on `req.token.groups`. All limits are overridable via environment variables (`VISITOR_RATE_LIMIT`, `USER_RATE_LIMIT`, etc.) for production tuning.

**`config/http.js`** — Replaced the three middleware entries (`generalRateLimit`, `userDeleteRateLimit`, `moderatorDeleteRateLimit`) with two (`generalRateLimit`, `deleteRateLimit`).

**`test/integration/2_utils/rateLimiter.test.js`** — Rewrote tests to cover all tiers. Tests use low env-var overrides to avoid socket exhaustion from too many supertest requests.

## 🧪 Testing

```shell
npm run test -- --grep "Rate Limiter"   # 12 tests
npm run test                             # Full suite — 2081 passing
```

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/config/http.js
+++ b/config/http.js
@@ -31,8 +31,7 @@ module.exports.http = {
   middleware: {
     // Requests limiter configuration
     generalRateLimit: rateLimiter.generalRateLimit,
-    userDeleteRateLimit: rateLimiter.userDeleteRateLimit,
-    moderatorDeleteRateLimit: rateLimiter.moderatorDeleteRateLimit,
+    deleteRateLimit: rateLimiter.deleteRateLimit,
 
     /** *************************************************************************
      *                                                                          *
@@ -46,8 +45,7 @@ module.exports.http = {
       'corsHeaders',
       'parseAuthToken',
       'generalRateLimit',
-      'userDeleteRateLimit',
-      'moderatorDeleteRateLimit',
+      'deleteRateLimit',
       'responseTimeLogger',
       'requestLogger',
       'fileMiddleware',
--- a/config/rateLimit/rateLimiter.js
+++ b/config/rateLimit/rateLimiter.js
@@ -1,130 +1,125 @@
 const rateLimit = require('express-rate-limit');
 const RightService = require('../../api/services/RightService');
 
+// 10-minute window for general requests
+const RATE_LIMIT_WINDOW_MS = 10 * 60 * 1000;
+// 1-hour window for DELETE requests
+const DELETE_RATE_LIMIT_WINDOW_MS = 60 * 60 * 1000;
+
+// General request limits per 10-minute window, by role.
+// Each tier can be overridden via environment variables for tuning in production.
+// We use ?? so that an explicit `0` is honoured (e.g. to fully block a tier).
+const parseLimit = (envValue, fallback) => {
+  if (envValue === undefined) return fallback;
+  const n = Number(envValue);
+  return Number.isNaN(n) ? fallback : n;
+};
+
+const VISITOR_MAX = parseLimit(process.env.VISITOR_RATE_LIMIT, 200);
+const USER_MAX = parseLimit(process.env.USER_RATE_LIMIT, 400);
+const LEADER_MAX = parseLimit(process.env.LEADER_RATE_LIMIT, 1000);
+const MODERATOR_MAX = parseLimit(process.env.MODERATOR_RATE_LIMIT, 10000);
+// Administrators are not rate-limited (skip returns true)
+
+// DELETE request limits per 1-hour window, by role
+const USER_DELETE_MAX = parseLimit(process.env.USER_DELETE_RATE_LIMIT, 1);
+const DELETE_MAX = parseLimit(process.env.DELETE_RATE_LIMIT, 20);
+// Administrators are not rate-limited on DELETEs either
+
+const isTestOrDev = () =>
+  process.env.NODE_ENV === 'test' || process.env.NODE_ENV === 'development';
+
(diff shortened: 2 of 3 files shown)


urien:
Merci pour cette belle proposition
Je crois que les modérateurs peuvent supprimer certaines choses mais pas les leaders. le volume de suppression pour les modérateurs est adapté. 
le fait de ne pas mettre de limite pour les admin est plus simple quand il y a des imports en masse

Est ce que la mise en place de la messagerie privée et d'une gestion plus fine des notifications va augmenter les besoins ? ce sera un point de vigilance lorsqu'on mettra en place ces services

Paul-AUB (dismissed):
**Pre-approved, only non blocker nits:**

# Code Review — Commit 882a1bc

**Author:** Clément Ronzon  
**Date:** 2026-05-04  
**Closes:** #1157  
**Files changed:** `config/http.js`, `config/rateLimit/rateLimiter.js`, `test/integration/2_utils/rateLimiter.test.js`

---

## Overview

This commit replaces a binary skip/limit rate-limiting strategy with a graduated, role-aware tier system. It also consolidates the two separate delete middlewares (`userDeleteRateLimit` + `moderatorDeleteRateLimit`) into a single `deleteRateLimit`. The motivation is sound: the old system was either "unlimited" (admins + moderators) or "limited at a flat cap" (everyone else), which provided no flexibility and had several latent bugs.

---

## Positive Changes

- **Fixes a real typo bug:** the old `generalRateLimit` config read `process.env.RATE_LIMIT_WINDOWS` (with an S) when the truthy branch was taken, so the env-var override never actually worked. This is now corrected.
- **Eliminates duplicated `require.cache` clearing** in tests via the `freshRateLimiter()` helper — cleaner and less error-prone.
- **Proper env-var backup/restore** in the test `before`/`after` hooks prevents environment leakage between test runs.
- **Well-structured test coverage** for each role tier (visitor, user, moderator, admin) and both middleware types.
- **`isTestOrDev()` helper** centralises environment detection and removes repeated inline checks.

---

## Issues and Concerns

### 4. Falsy-zero edge case in env-var parsing

**File:** [config/rateLimit/rateLimiter.js](config/rateLimit/rateLimiter.js)

```js
const VISITOR_MAX = Number(process.env.VISITOR_RATE_LIMIT) || 200;
```

If `VISITOR_RATE_LIMIT=0` is deliberately set (to fully block visitors), `Number("0") || 200` evaluates to `200`, silently ignoring the operator's intent. Use `Number(process.env.VISITOR_RATE_LIMIT) ?? 200` with a preceding validity check, or use `parseInt` with an explicit zero-check:

```js
const VISITOR_MAX =
  process.env.VISITOR_RATE_LIMIT !== undefined
    ? Number(process.env.VISITOR_RATE_LIMIT)
    : 200;
```

This is a low-priority concern since blocking all visitors via a limit of 0 is an unusual config, but it is a foot-gun.



### 7. Unauthenticated users and DELETE limit in `deleteMax`

**File:** [config/rateLimit/rateLimiter.js](config/rateLimit/rateLimiter.js)

```js
const deleteMax = (req) => {
  if (!req.token) return USER_DELETE_MAX;
  ...
};
```

Unauthenticated users hitting a DELETE endpoint get `USER_DELETE_MAX` (1/hour). If DELETE routes are all auth-guarded at the controller level, this is a dead code path and the `if (!req.token)` branch never fires. If any DELETE routes are public, returning `USER_DELETE_MAX` is correct but deserves a comment. Worth verifying with a quick grep across route policies.

PR #1564, opened 2026-05-05
fix(entrance): resolve permanent delete FK violations and preserve audit history

ClemRz:
## 🤔 What

Fixes permanent entrance/cave deletion which was failing with FK violations and taking 20+ seconds when it did succeed.

- Resolves FK constraint violations (`h_description_t_entrance2_fk`, `h_comment_t_comment`, `j_caver_cave_explorer_t_cave_fk`, etc.) that caused 500 errors on permanent delete
- Reduces response time from ~20s to ~150ms (100x improvement)
- Preserves all `h_*` history rows for auditability instead of deleting them
- Adds structured audit logging with affected entity IDs
- Adds SQL migration to drop history table FK constraints that don't exist in production

## 🤷‍♂️ Why

On May 3rd, a `DELETE /api/v1/entrances/26035?isPermanent=1` took 20.7 seconds. Investigation revealed:
1. **Performance**: 30+ sequential DB operations with zero parallelism in the cascade delete
2. **Correctness**: Multiple pre-existing FK violations that caused 500 errors for entrances with `exit` references, cave junction tables, or history rows
3. **Auditability**: History rows were being destroyed (or silently failing to destroy due to Waterline's composite PK bug), leaving no audit trail

Production data confirmed 305 orphaned `h_*` rows already exist — proving the FK constraints were never enforced in practice.

## ⚖️ FK Drop Trade-offs

The migration drops all foreign key constraints on `h_*` tables that reference `t_entrance(id)` or `t_cave(id)`.

**What we lose:**
- PostgreSQL will no longer reject inserts into `h_*` tables with a bogus `id_entrance` or `id_cave` value. A programming error could theoretically create a dangling reference without the DB catching it.

**Why this is acceptable:**
- History rows are only created by application code when a user edits an entity. At that moment, the referenced entrance/cave always exists — there's no user-facing flow that inserts an `h_` row pointing to a non-existent parent.
- Production already has 305 orphaned `h_*` rows (parent `t_` row deleted, `h_` rows retained), proving these FKs were never enforced in practice.
- Direct SQL manipulation (admin scripts, migrations) could create dangling references, but this is already the case in production.

**What we gain:**
- Permanent deletes work correctly without FK violations
- Full audit trail preserved in `h_*` tables after permanent deletion
- Schema aligns with production's actual behavior
- No more silent Waterline failures on history models

## 🔍 How

**Performance:**
- Parallelize the 5 independent `subEntityDelete` calls via `Promise.all`
- Parallelize soft-delete phase (search index + recent change)
- Make `NotificationService.notifySubscribers` fire-and-forget (don't block response on email delivery)

**FK fixes:**
- SQL migration drops all `h_* → t_entrance/t_cave` FK constraints (parent FKs + cross-entity FKs)
- Use raw SQL for history table updates (Waterline silently fails on composite-PK models)
- Clean up `exit` FK references on `t_description`, `t_rigging`, `t_comment` before hard delete
- Clean up cave junction tables (`j_caver_cave_explorer`, `j_grotto_cave_explorer`, `j_grotto_cave_partner`) using raw SQL since the cave is already soft-deleted when `permanentlyDeleteCave` runs

**Auditability:**
- History rows (`h_entrance`, `h_location`, `h_description`, `h_rigging`, `h_history`, `h_comment`, `h_name`, `h_cave`) are intentionally preserved
- `NameService.permanentDelete` no longer destroys `h_name` rows
- `CaveService.permanentlyDeleteCave` no longer destroys `h_cave` rows
- Nullable FK columns on h_ tables are nulled/reassigned; NOT NULL columns (like `h_location.id_entrance`) are left intact since the FK constraint is dropped

**Code quality:**
- Extract `subEntityDelete` to module scope (removes `no-inner-declarations` eslint-disable)
- Cast `entranceId` to `Number` at the top of the controller
- Consolidate redundant `exitTarget`/`entrTarget` into single `target` variable
- Add structured JSON audit log: `Permanent delete entrance 20: {"action":"delete","entranceId":20,"locations":[2699],...}`

## 🧪 Testing

- Run the SQL migration on a local DB: `npm run dev:clean` then verify with `docker exec grotto-postgres psql -U root -d grottoce -c "SELECT conname FROM pg_constraint WHERE conrelid = 'h_description'::regclass AND contype = 'f';"`
- Test permanent delete of an entrance with sub-entities: `DELETE /api/v1/entrances/:id?isPermanent=1`
- Verify h_ rows are preserved: `SELECT * FROM h_entrance WHERE id = :entranceId`
- New test: `npm run test -- --grep "Auditability"` covers all entity types + junction tables + cave cascade

## 📸 Previews

Before: `DELETE /api/v1/entrances/26035?isPermanent=1` → 20,713ms (or 500 error)
After: `DELETE /api/v1/entrances/20?isPermanent=1` → 148ms with full audit trail preserved

### DIFF
--- a/api/controllers/v1/cave/delete.js
+++ b/api/controllers/v1/cave/delete.js
@@ -33,13 +33,15 @@ module.exports = async (req, res) => {
     await TCave.destroyOne({ id: caveId }); // Soft delete
     cave.isDeleted = true;
 
-    await CaveService.deleteInSearch(caveId);
-    await RecentChangeService.setDeleteRestoreAuthor(
-      'delete',
-      'cave',
-      caveId,
-      req.token.id
-    );
+    await Promise.all([
+      CaveService.deleteInSearch(caveId),
+      RecentChangeService.setDeleteRestoreAuthor(
+        'delete',
+        'cave',
+        caveId,
+        req.token.id
+      ),
+    ]);
   }
 
   const deletePermanently = !!req.param('isPermanent');
@@ -60,7 +62,8 @@ module.exports = async (req, res) => {
     await CaveService.permanentlyDeleteCave(cave, shouldMergeInto, mergeIntoId);
   }
 
-  await NotificationService.notifySubscribers(
+  // Fire-and-forget: don't block the response on subscriber notifications
+  NotificationService.notifySubscribers(
     req,
     cave,
     req.token.id,
@@ -68,7 +71,11 @@ module.exports = async (req, res) => {
       ? NotificationService.NOTIFICATION_TYPES.PERMANENT_DELETE
       : NotificationService.NOTIFICATION_TYPES.DELETE,
     NotificationService.NOTIFICATION_ENTITIES.CAVE
-  );
+  ).catch((err) => {
+    sails.log.error(
+      `Failed to notify subscribers for cave ${caveId}: ${err.message}`
+    );
+  });
 
   return ControllerService.treatAndConvert(
     req,
--- a/api/controllers/v1/entrance/delete.js
+++ b/api/controllers/v1/entrance/delete.js
@@ -6,6 +6,55 @@ const { toEntrance } = require('../../../services/mapping/converters');
 const NameService = require('../../../services/NameService');
 const CaveService = require('../../../services/CaveService');
 const RecentChangeService = require('../../../services/RecentChangeService');
+const CommonService = require('../../../services/CommonService');
+
+/**
+ * Hard-delete rows matching `criteria` from `model`.
+ *
+ * Waterline uses a two-phase destroy for models with `is_deleted`:
+ *   1st destroy() → sets is_deleted = true (soft delete)
+ *   2nd destroy() → removes the row (hard delete, since it's already soft-deleted)
+ *
+ * This helper encapsulates that convention so the intent is explicit and
+ * any future Waterline behaviour change only needs fixing in one place.
+ */
+async function hardDestroy(model, criteria) {
+  await model.destroy(criteria); // Soft delete (is_deleted = true)
+  await model.destroy(criteria); // Hard delete (removes row)
+}
+
+/**
+ * Delete sub-entities linked to an entrance.
+ * History (h_) rows are intentionally NOT touched — they are preserved
+ * for auditability with FK references nulled via raw SQL separately.
+ *
+ * IMPORTANT: When adding a new sub-entity type here, also add a matching
+ * h_ UPDATE in the raw SQL block below (search for "h_location" to find it).
+ */
+async function subEntityDelete(
+  entrance,
+  subEntitiesKey,
+  notificationKey,
+  model,
+  entranceId,
+  shouldMergeInto,
+  mergeIntoEntity
+) {
+  const subEntities = entrance[subEntitiesKey];
+  if (subEntities.length === 0) return;
+
+  await TNotification.destroy({
+    [notificationKey]: subEntities.map((e) => e.id),
+  });
+
+  if (shouldMergeInto) {
+    await model.update({ entrance: entranceId }).set({
+      entrance: mergeIntoEntity.id,
+    });
+  } else {
+    await hardDestroy(model, { entrance: entranceId });
+  }
+}
 
 module.exports = async (req, res) => {
   const hasRight = RightService.hasGroup(
@@ -16,8 +65,7 @@ module.exports = async (req, res) => {
     return res.forbidden('You are not authorized to delete a entrance.');
   }
 
-  // Check if entrance exists and if it's not already deleted
-  const entranceId = req.param('id');
+  const entranceId = Number(req.param('id'));
   const entrance = await EntranceService.getPopulatedEntrance(entranceId);
   if (!entrance) {
     return res.notFound({ message: `Entrance of id ${entranceId} not found.` });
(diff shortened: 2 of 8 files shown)


Paul-AUB (changes requested):
# Code Review — Commit `97585335`

**Author:** Clément Ronzon  
**Date:** 2026-05-05  
**Subject:** `fix(entrance): resolve permanent delete FK violations and preserve audit history`

---

## Overview

This commit fixes a critical bug where permanently deleting an entrance (or cave) would crash with FK constraint violations. The root cause was a combination of:

1. `h_*` (history) tables having FK constraints pointing at their parent `t_*` table, meaning a parent row couldn't be hard-deleted while history rows existed.
2. Waterline silently failing to operate on models with composite PKs (e.g., `h_name`, junction tables), leaving orphaned references.

The fix also introduces a deliberate **audit preservation** policy: `h_*` rows are no longer deleted during permanent deletes — they are kept for traceability, and only their FK references are nulled or reassigned.

---

## What Changed

| File | Change |
|---|---|
| `api/controllers/v1/entrance/delete.js` | Full rewrite of permanent delete logic |
| `api/controllers/v1/cave/delete.js` | Parallelization, fire-and-forget notifications |
| `api/services/CaveService.js` | Fix junction table cleanup, audit logging, preserve h_ rows |
| `api/services/NameService.js` | Remove `HName.destroy()` (was silently a no-op), document intent |
| `sql/1_2026_05_05_drop_history_parent_fk.sql` | Migration dropping blocking FK constraints |
| `test/customSQL.js` + `test/bootstrap.test.js` | Apply FK migration in test environment |
| `test/integration/4_routes/Entrances/delete-auditability.test.js` | New comprehensive auditability test |

---

## Positive Points

### Correctness
- The FK violations are correctly identified and fixed. The `ALTER TABLE … DROP CONSTRAINT IF EXISTS` migration is safe and idempotent.
- Using raw SQL for composite-PK models is the right call — Waterline's silent failure on these was a ticking time bomb.
- The `entranceId = Number(req.param('id'))` type coercion is a genuine bug fix; raw query params are strings and this was likely causing subtle mismatches in SQL comparisons.

### Performance
- Replacing sequential `await` chains with `Promise.all` for independent operations is a meaningful improvement. The `~100x` claim in the commit message is optimistic, but the speedup is real for entrances with many sub-entities.

### Auditability policy
- Preserving `h_*` rows is the right decision for a geodata platform. The explicit comments explaining *why* `HName.destroy()` is skipped (`// h_name rows are intentionally preserved`) make the intent clear rather than looking like dead code.

### Test coverage
- The new test is thorough: it sets up every sub-entity type, inserts matching `h_*` rows via raw SQL, triggers the delete, and then asserts both that `t_*` rows are gone **and** that `h_*` rows survive. Junction table cleanup is also verified.

### Notifications
- Fire-and-forget with a `.catch()` log is the correct pattern for non-blocking side effects. The previous `await` on notification was silently blocking all responses.

---

## Issues and Suggestions

### 2. The merge path silently skips `h_*` FK reassignment for some tables (potential bug)
In `subEntityDelete`, the merge branch only updates the `t_` model:
```js
await model.update({ entrance: entranceId }).set({ entrance: mergeIntoEntity.id });
```
The corresponding `h_` FK reassignment is handled separately in the main controller via raw SQL (`UPDATE h_description SET id_entrance = $1 WHERE id_entrance = $2`, etc.). This split logic is fragile — if a new sub-entity type is added to `subEntityDelete` without a matching raw SQL block in the controller, the `h_` FK will be left pointing at the deleted entrance, potentially causing future issues.

**Suggestion:** Either consolidate the merge logic (raw SQL for both `t_` and `h_` in the same place), or add a comment coupling `subEntityDelete` additions to the raw SQL block.

### 3. `h_location` FK is not reassigned on merge (potential bug)
The raw SQL block nulls/reassigns `h_description`, `h_rigging`, `h_comment`, `h_history`, and their `exit` variants — but **`h_location`** is absent:
```sql
-- Missing:
UPDATE h_location SET id_entrance = $1 WHERE id_entrance = $2
```
The `h_location` table likely has a FK on `id_entrance`. If the constraints are dropped by the migration this may not error immediately, but the orphaned reference remains inconsistent.

### 4. Test `after()` does not clean up `h_*` rows
```js
after(async () => {
  await TDocument.destroy({ id: document?.id });
  // ... t_ cleanups only
});
```
The `h_*` rows inserted in `before()` are never removed. Since Waterline's `migrate:drop` recreates the schema between runs this is not a blocker for CI, but it could cause unexpected state in long-lived test databases and makes the `after()` not truly idempotent.

### 5. Double-destroy pattern is implicit
```js
await model.destroy({ entrance: entranceId }); // Soft delete (is_deleted = true)
await model.destroy({ entrance: entranceId }); // Hard delete (removes row)
```
This pattern relies on Waterline's internal lifecycle behaviour (first call sets `is_deleted = true`, second call removes the row) and is not documented in Waterline itself. The comments here are helpful, but if Waterline's behaviour ever changes this will silently break. Consider wrapping this into a named helper (e.g., `hardDestroy(model, criteria)`) with the explanation in one place.

### 7. SQL migration file should have a rollback section
`sql/1_2026_05_05_drop_history_parent_fk.sql` has no rollback. Since `DROP CONSTRAINT` is irreversible without the original constraint definition, teams doing point-in-time recovery or rollback won't know how to undo this. Documenting the constraint definitions (or adding a commented-out `ADD CONSTRAINT` block) would help future operators.

---


## Summary

This is a well-motivated and largely well-executed fix for a real production bug. The audit preservation policy is sound, and the test coverage is notably good for this type of infrastructure change. The main concern is the **missing `h_location` FK reassignment** in the merge path, which looks like an omission rather than a deliberate choice and should be verified against the actual schema before merging.

ClemRz:
Thanks for the thorough review Paul! All points addressed in `19490021`:

**# 2 (Split logic fragility):** Added a coupling comment on both `subEntityDelete` and the raw SQL block, making it explicit that adding a new sub-entity type to one requires a matching update in the other.

**# 3 (Missing `h_location` FK reassignment):** Good catch — this was indeed an omission. Added `UPDATE h_location SET id_entrance = $1 WHERE id_entrance = $2` to the raw SQL block. Also added `h_name` reassignment for completeness since it has an `id_entrance` column too.

**# 4 (Test `after()` not cleaning h_ rows):** Added raw SQL `DELETE` statements for all h_ rows created in `before()`.

**# 5 (Double-destroy pattern):** Extracted a `hardDestroy(model, criteria)` helper at the top of the controller with a JSDoc explaining the two-phase convention. `NameService.permanentDelete` also got an expanded comment.

**# 7 (SQL migration rollback):** Added a commented-out rollback section at the end of the migration file with all the `ADD CONSTRAINT` statements and a note about needing to clean orphaned rows first.

Paul-AUB (approved):

PR #1565, opened 2026-05-06
feat(account): consolidate endpoints and add locale-aware emails

ClemRz:
## 🤔 What

Consolidates scattered account-related endpoints into a unified API surface and adds locale-aware email sending.

- Add `GET /api/v1/account` — returns authenticated user's private data (mail, language, mailIsValid, sendNotificationByEmail, etc.)
- Add `PATCH /api/v1/account` — single endpoint for self-service updates (name, surname, nickname, email, password, language, sendNotificationByEmail)
- Restrict `PUT /api/v1/cavers/:caverId` to admin-only (profile management of other cavers)
- Add optional `language` parameter to the sign-up endpoint
- Add `locale` input to the `send-email` helper for recipient-language emails
- Wire notification emails to use the recipient's preferred language
- Remove orphaned `change-email`, `change-alert-for-news`, `change-language` controllers
- Silence pre-existing `func-names` lint warnings in property test files

## 🤷‍♂️ Why

The account API had multiple overlapping endpoints for the same mutations (`PUT /cavers/:id` could change email/password, while `PATCH /account/email` did the same). This created confusion about which endpoint to use and inconsistent authorization rules.

Additionally, the `id_language` field on `t_caver` was hardcoded to `'000'` at sign-up and never used for email communication — 58,945 cavers had this placeholder value. Emails were sent in the request's locale rather than the recipient's preferred language.

## 🔍 How

**Endpoint consolidation:**
- `PATCH /api/v1/account` validates each field individually (email format via ORM, password length, language existence in `t_language`, nickname uniqueness, sendNotificationByEmail as boolean) and applies all changes in a single `updateOne` call.
- `PUT /api/v1/cavers/:caverId` now rejects non-admin callers with 403 upfront. Only `name`, `nickname`, `surname`, `organizations` are updatable.
- `PATCH /api/v1/account/password` remains separate for the unauthenticated reset-token flow.

**Locale-aware emails:**
- The `send-email` helper accepts an optional `locale` (ISO 639-1 code). When provided, it temporarily switches the i18n locale before rendering the template and subject, then restores it.
- Guards against i18n objects that don't support `setLocale` (e.g., stubs in tests, the global `sails.hooks.i18n`).
- `NotificationService` resolves each subscriber's language FK to its `part1` code before sending.
- `forgot-password` passes the user's locale directly instead of mutating `req`.

## 🧪 Testing

All tests pass. New test coverage:
- `GET /api/v1/account` — 200 with correct fields (including sendNotificationByEmail), 401 without token
- `PATCH /api/v1/account` — name/surname update, nickname update, nickname conflict (409), empty nickname (400), clear name with empty string, unknown properties (400), empty body (400), valid/invalid language, sendNotificationByEmail toggle (true/false/non-boolean rejection)
- `PUT /api/v1/cavers/:caverId` — admin success, non-admin rejection (403), unknown fields (400)

Manual curl testing performed against local dev DB covering all roles and edge cases.

## 📸 Previews

N/A — backend-only changes.


urien:
C'est une très belle proposition
Merci @ClemRz

Paul-AUB (changes requested):
# Code Review — commit aa80207685f6bcb647b9feaf4a2970d7e09713fc

**Author:** Clément Ronzon  
**Date:** 2026-05-05  
**Subject:** `feat(account): consolidate account endpoints and add locale-aware emails`

---

## Overview

This commit consolidates account management into two unified endpoints (`GET /api/v1/account` and `PATCH /api/v1/account`), removes orphaned controllers, restricts `PUT /api/v1/cavers/:caverId` to admin-only, and implements locale-aware email sending so recipients receive emails in their preferred language.

---

## Positive Points

- **Good API design**: Consolidating multiple narrow PATCH routes (`/account/email`, `/account/alertForNews`) into a single `PATCH /account` with an explicit allowlist of updatable fields is cleaner and easier to maintain.
- **Allowlist validation**: The `UPDATABLE_PROPERTIES` array in `update.js` correctly rejects unknown properties before doing anything else.
- **Empty-string coercion**: Clearing `name` and `surname` via empty string (`'' → null`) matches the pattern already established in other controllers (e.g., entrance, massif).
- **Locale restore**: The send-email helper saves and restores the previous i18n locale, showing awareness of shared state.
- **Test coverage**: `Account.test.js` was significantly expanded to cover the new endpoints.
- **Swagger updated**: Both new endpoints are documented with request/response schemas.

---

## Issues and Suggestions

### Critical

#### 1. Concurrency / shared-state bug in `send-email.js`

```js
// send-email.js
i18n.setLocale(locale);
const emailHtml = await ejs.renderFile(...);   // <-- async gap
i18n.setLocale(previousLocale);               // restores AFTER the await
```

`i18n` is a shared singleton. If two emails are sent concurrently (e.g., a notification and a verification email), the locale set by the first call can be overwritten by the second before the first's `renderFile` finishes, producing emails in the wrong language. The subject line (`subjectText`) is computed after `renderFile`, which compounds this risk.

**Fix**: pass the locale directly to EJS rendering as a view variable and use a locale-scoped translation helper, or clone/scope the i18n object per call rather than mutating shared state.

#### 2. TOCTOU race condition on nickname uniqueness (`update.js:66–72`)

```js
const existingCaver = await TCaver.findOne({ nickname: req.body.nickname });
if (existingCaver && existingCaver.id !== req.token.id) {
  return res.conflict('This nickname is already used.');
}
updates.nickname = req.body.nickname;
// ... later
await TCaver.updateOne({ id: req.token.id }).set(updates);
```

Two concurrent PATCH requests with the same new nickname can both pass the `findOne` check before either writes. The DB should have a unique constraint on `nickname` — the catch block should handle the resulting constraint violation and return 409 instead of 400. Currently the `catch` only handles `UsageError`/`E_INVALID_NEW_RECORD`, which may not map to a unique constraint violation.

**Fix**: catch the uniqueness DB error specifically and return `res.conflict()`, or rely solely on the DB constraint and remove the application-level check.

---

---

### Medium

#### 5. Email validation does not verify format (`update.js:27–30`)

```js
if (!req.body.email) {
  return res.badRequest('You must provide an email.');
}
updates.mail = req.body.email.toLowerCase();
```

An email like `"notanemail"` passes this check. The old `change-email.js` relied on the ORM for format validation (via `try/catch`). The new controller does not — any string is accepted and passed to `TCaver.updateOne`, which may or may not enforce a format constraint depending on the model definition.

**Fix**: either validate the format here (regex or a dedicated helper) or ensure the ORM model enforces it and the `catch` block surfaces the resulting error as a 400.

#### 6. Defensive `hasLocaleSupport` check hides misconfiguration (`send-email.js`)

```js
const hasLocaleSupport =
  typeof i18n.getLocale === 'function' &&
  typeof i18n.setLocale === 'function';
```

If `i18n` somehow lacks these methods, the locale is silently not applied and the email is sent in the wrong language. This should be an invariant, not a graceful degradation. Failing loudly (or asserting) would surface misconfiguration immediately rather than producing subtle i18n bugs in production.

#### 7. Language lookup duplicated across multiple call sites

The pattern of resolving `user.language` (a FK id) to `lang.part1` is now copy-pasted in three places:

- `api/controllers/v1/auth/resend-verification-email.js`
- `api/controllers/v1/auth/sign-up.js`
- `api/services/NotificationService.js`

This should be extracted into a shared helper (e.g., `LanguageService.getLocale(languageId)`) to avoid divergence.

---

### Low

#### 8. Typo in existing code not fixed: `propretiesUpdatable` (`caver/update.js:4`)

The variable is still named `propretiesUpdatable` (transposed letters). Since this file was touched, it's a minor opportunity to fix the spelling (`propertiesUpdatable`). Not a blocker.

#### 9. `eslint-disable` comments in property tests

```js
/* eslint-disable func-names */
```

Added to 5 test files to silence a pre-existing lint warning. This is acceptable as a short-term workaround but the underlying issue (anonymous function names in fast-check properties) should eventually be fixed rather than suppressed.

#### 10. `find.js` returns `language` as a raw FK id, not a resolved object

```js
return res.ok({
  ...
  language: caver.language,   // ISO 639-3 id (e.g. "fra") or "000"
  ...
});
```

The Swagger spec documents this as a string, which is correct, but clients may expect a resolved language object (name, part1, etc.) like other endpoints that return caver data. Consider whether consumers of this endpoint need more than the id.

ClemRz:
# Review Fixes Applied

Thanks for the thorough review @Paul-AUB! Here's what was addressed:

## Critical

### 1. Concurrency / shared-state bug in `send-email.js` — Fixed

Replaced the shared singleton mutation with a **fresh `i18n-2` instance per email call**, each scoped to the target locale. No shared mutable state, no interleaving risk. Templates now receive a `__` translation helper and `locale` as explicit view variables instead of the full `i18n` object.

`devMode` is enabled only in development (`process.env.NODE_ENV === 'development'`) so missing keys get auto-added locally but never pollute locale files in test/production.

### 2. TOCTOU race on nickname uniqueness — Fixed

Added `E_UNIQUE` handling in the catch block to return `res.conflict()` if the DB rejects a duplicate. The application-level pre-check remains for a friendlier error message on the happy path.

## Medium

### 5. Email validation — Fixed

Added a basic format regex (`/^[^\s@]+@[^\s@]+\.[^\s@]+$/`) before passing to the ORM. The ORM's `isEmail` constraint and SES provide additional layers.

### 6. `hasLocaleSupport` defensive check — Removed

No longer relevant — the helper now creates its own `i18n-2` instance rather than depending on a passed-in object.

### 7. Language lookup duplication — Fixed

Extracted `LanguageService.getLocale(languageId)` and replaced the 3 duplicated patterns in `resend-verification-email.js`, `sign-up.js`, and `NotificationService.js`.

## Low

### 8. Typo `propretiesUpdatable` — Fixed

Renamed to `propertiesUpdatable` in `caver/update.js`.

### 9. `eslint-disable` in property tests — Kept as-is

Acceptable workaround per the review.

### 10. `find.js` returns language as raw FK — Kept as-is

This is consistent with the rest of the codebase. The `toCaver` converter in `converters.js` returns `language: source.language` (the raw ISO 639-3 FK id, e.g. `"fra"`), and every other endpoint that returns caver data does the same. The front-end expects this format. Returning a populated object would break the existing contract.

## Additional improvements

- **Dead `req` parameter removed** from `sendNotificationEmail`, `notifySubscribers`, and `notifyAuthor` (+ all 45 call sites). It was no longer used after the i18n refactor.
- **`i18n-2` pinned** to exact version `0.7.3` in `package.json`.
- **`find.js` uses `ControllerService.treat`** for the success path (consistency with project conventions).
- **Email `<title>` tag** now uses the translated subject line instead of a hardcoded English string.
- **Footer includes** consistently pass both `__` and `locale`.
- **All 14 locale files** synced — missing translations added for the 8 new keys (ar, bg, ca, de, el, es, fr, he, id, it, ja, nl, pt, ro).
- **Junk entries** cleaned from `fr.json` (auto-generated by `i18n-2` during test runs).

Paul-AUB (dismissed):

Paul-AUB:
# Code Review — feat(account): consolidate account endpoints and add locale-aware emails

Commit: `f5c427bfbff1ce4b3c3f0d52f863a519c49a7736`  
Author: Clément Ronzon  
Date: 2026-05-05

---

## Issues to fix

### Blocker

#### 3. Email change without resetting `mailIsValid`
**File:** `api/controllers/v1/account/update.js:38`

Changing `mail` without setting `mailIsValid = false` and re-sending a verification email allows a user to replace their address with an unverified one while retaining verified status. The old `change-email` controller likely handled this — replicate that behaviour here.

**To do:**
- Add `mailIsValid: false` to `updates` when `email` is changed
- Trigger a re-send of the verification email (as in `AuthService.sendVerificationEmail`)

---

### Security

#### 1. No current-password confirmation before a password change
**File:** `api/controllers/v1/account/update.js:41-52`

A stolen token allows silently changing the password. Require `currentPassword` in the request body and verify it via `AuthService` before applying the change.

#### 2. No email uniqueness pre-check before the DB write
**File:** `api/controllers/v1/account/update.js:29-38`

Duplicate email is caught by `E_UNIQUE` (line 115), but an explicit pre-check (mirroring the nickname check at line 85) would give a clearer error message and avoid relying on the `catch` block for an expected condition.

---

### Correctness

#### 5. `res.ok()` returns 200 but tests expect 204
**Files:** `api/controllers/v1/account/update.js:124` / `test/integration/4_routes/Account.test.js`

The controller calls `res.ok()` (200) while tests assert `.expect(204)`. Verify whether Sails resolves `res.ok()` to 204 in this context — if not, fix either the controller (`res.noContent()`) or the tests.

#### 4. `locale` may be `undefined` on sign-up without a language
**File:** `api/controllers/v1/auth/sign-up.js`

When no `languageParam` is provided, `locale` stays `undefined` and the email helper falls back silently to `sails.config.i18n.defaultLocale`. Prefer an explicit default at the call site rather than relying on the helper's implicit fallback.

#### 6. Race condition on nickname uniqueness check
**File:** `api/controllers/v1/account/update.js:85-91`

The `findOne` + `updateOne` is not atomic. Two concurrent requests with the same nickname can both pass the pre-check, then one hits `E_UNIQUE` with a generic message. Acceptable behaviour but the error message is less precise than the pre-check message.

---

### Code quality

#### 8. `LanguageService.getLocale` hits the DB on every call
**File:** `api/services/LanguageService.js:15`

Languages rarely change. A simple in-memory cache (or the ORM query cache if enabled) would reduce redundant queries, especially in `NotificationService` which calls it per notification.

#### 9. Deletion of `NotificationService` tests
**File:** `test/integration/1_services/NotificationService.test.js`

The file is deleted (not just modified to silence lint warnings). Confirm this is intentional and that coverage is maintained elsewhere.

PR #1566, opened 2026-05-06
1451 private messaging

dawoldo:
## 🤔 What

This PR completes the implementation of the Private Messaging System. Key features include:

- **Messaging API**: Endpoints for sending messages, listing active/archived conversations, and fetching paginated message history.
- **Unread Messages Management**: Endpoints to fetch unread counts for both active and archived conversations.
- **Archive Logic**: Independent archival states per participant with automated unread count increments.
- **Notification Preferences**: A new unified API (`GET/PATCH /api/v1/account/notifications`) to manage all email communication settings in one place.
- **Email Integration**: Automated email alerts for new messages (respecting user preferences) with direct linking to conversations.

## 🤷‍♂️ Why

The goal is to provide a secure and private communication channel for cavers on the platform while ensuring:
- **Privacy**: PII (like emails and real names) is never exposed in message responses.
- **User Control**: Users can opt-out of email alerts for private messages independently of platform news.
- **Consistency**: All messaging endpoints follow a standardized, structured error format for better frontend integration.

## 🔍 How

- **Access Control**: Enforced strict participant-only access at the controller level using `req.token.id`, with no administrative bypasses to preserve privacy.
- **Security**: Implemented eligibility checks on replies—preventing messages to banned or inactive users even in existing threads.
- **Error Handling**: Introduced `sails.helpers.formatMessagingError` to standardize `code`, `message`, and `reference_id` across all domain-specific errors.
- **Data Integrity**: Added the `send_message_notification_by_email` field to the `TCaver` model with a `true` default value.

## 🧪 Testing

- **Integration Tests**: New tests added to `test/integration/4_routes/Messages.test.js` covering:
    - Message sending and automatic conversation creation.
    - PII stripping verification.
    - Archival/Unarchival state persistence.
    - Eligibility checks for replies to banned users.
- **Manual Verification**: Verified email dispatching logic and `conversationLink` generation in `NotificationService`.
- **Linting**: All changes passed `npm run lint:fix`.

### Structured Error Response (Requirement 9)
```json
{
  "code": "E_AUTHORIZATION",
  "message": "You cannot message this user (banned or inactive).",
  "metadata": {},
  "reference_id": "req-trace-12345"
}
```

### New Message Email Template 
The email includes the sender's nickname and a direct link to the conversation:

```CaveExplorer88 sent you a new message on Grottocenter. Click here to view the conversation```

### CODE REVIEW

**Review by ClemRz — CHANGES_REQUESTED** [review]
## Critical Issues

### 1. Missing `is_deleted` check in `getEligibleRecipient` (Requirement 1, AC 7)

The requirements define an Eligible_Recipient as a caver who is "NOT banned AND NOT is_deleted." However, `MessageService.getEligibleRecipient` only checks `banned`, `activated`, and `login`:

```javascript
if (caver.banned || !caver.activated || !caver.login) {
```

It never checks `caver.isDeleted`. A soft-deleted caver would pass the eligibility check and could receive messages. This violates Requirement 1 AC 7: "WHEN a Sender attempts to send a message to a caver ID that does not exist or to a soft-deleted (is_deleted) caver, THE Messaging_System SHALL reject the request and return a not-found error."

### 2. Eligibility definition mismatch with requirements

The requirements define Eligible_Recipient as:

> `(mail_is_valid OR NOT activated) AND password IS NOT NULL AND NOT banned AND NOT is_deleted`

The code checks `!caver.login` (checking login instead of password). Per the glossary, Non_User_Cavers are identified by `password IS NULL`, not `login IS NULL`. The code should check `!caver.password` instead of `!caver.login`.

Additionally, the requirements explicitly state that users who "signed up but have not yet confirmed their email" ARE eligible. The code rejects them with `!caver.activated`.

### 3. Missing messaging indexes in `test/customSQL.js`

Per the steering rules: "When adding or modifying an index in any SQL file under `sql/`, always add the same `CREATE INDEX IF NOT EXISTS` statement to the `INDEX_OPTIMIZATION_MIGRATION` block in `test/customSQL.js`."

The PR adds three indexes in `sql/0_tables.sql`:

```sql
CREATE INDEX idx_j_participant_caver ON j_participant(id_caver);
CREATE INDEX idx_t_message_conversation ON t_message(id_conversation, date_sent DESC);
CREATE INDEX idx_t_message_sender ON t_message(id_caver_sender);
```

These are **not** added to `test/customSQL.js`. The test DB won't have these indexes, potentially causing slow queries and flaky test timeouts.

---

## Medium Issues

### 4. `JParticipant` model missing `primaryKey` definition

The SQL defines a composite primary key `(id_conversation, id_caver)`, but the Waterline model has no `primaryKey` attribute. Waterline defaults to `id` as the primary key. Without it, operations like `JParticipant.updateOne(...)` may not work correctly since Waterline expects an `id` column. This could cause silent failures or unexpected behavior.

### 5. Archive/Unarchive returns `res.ok()` (200) but tests expect 204

The `archive.js` and `unarchive.js` controllers call `return res.ok()` which returns HTTP 200. However, the test asserts `.expect(204)`. Either the tests are wrong or the controllers should use `return res.status(204).send()`. Since there's no response body, 204 is semantically correct — the controllers should be fixed.

### 6. `conversationLink` is always empty string in email notification (Requirement 7, AC 4)

```javascript
// TODO: construct conversation link when the frontend is worked on
const conversationLink = '';
```

Requirement 7 AC 4 states: "THE Messaging_System SHALL include a link to the Conversation in the email notification body." The email template conditionally renders the link (`<% if (conversationLink) { %>`), so with an empty string, no link is ever shown. At minimum, a placeholder URL pattern like `${baseUrl}/messages/${conversationId}` should be used.

### 7. Notification email subject is not localized

The `emailSubject` is hardcoded as `'New Message'` in `NotificationService.notifyMessageRecipient`. While the body uses `i18n.__()`, the subject line doesn't go through the i18n system. It should be `req.i18n.__('New Message')`.

### 8. `notifyMessageRecipient` uses `await` but errors are swallowed silently

The controller does `await NotificationService.notifyMessageRecipient(...)`. If the email sending takes time, this blocks the response to the user. The comment says "Errors are handled inside the service to avoid affecting message creation" — but the `await` still delays the response. Consider using fire-and-forget (`NotificationService.notifyMessageRecipient(...).catch(...)` without `await`) or at least document why blocking is acceptable.

---

## Minor Issues

### 9. Inconsistent error codes

- Archive/unarchive use `'E_FORBIDDEN'` when the user isn't a participant
- Create uses `'E_AUTHORIZATION'` for the same scenario
- The requirements spec says "E_AUTHORIZATION" for authorization errors (Req 9 AC 2)

Pick one and be consistent. `E_AUTHORIZATION` aligns with the requirements.

### 10. `formatMessagingError` naming is misleading for non-messaging controllers

The helper is used in `get-notifications.js` and `update-notifications.js` which are account controllers, not messaging controllers. The name `formatMessagingError` implies it's messaging-specific. Consider renaming to `formatStructuredError` or similar.

### 11. `list-conversations` doesn't use structured error format for server errors

```javascript
return res.serverError('An error occurred while listing conversations.');
```

While `list-archived-conversations.js` uses the structured format. This is inconsistent with Requirement 9.

### 12. `count-unread` controller doesn't use structured error format

```javascript
return res.serverError('An error occurred while counting unread messages.');
```

Should use `sails.helpers.formatMessagingError(...)` per Requirement 9.

### 13. Test uses `setTimeout` for ordering

The archived conversations sort test uses `setTimeout(resolve, 1000)` to ensure different timestamps. This makes the test slow and potentially flaky. A better approach would be to directly set `archivedAt` via a native query.

---

## Requirements Coverage

| Requirement | Status | Notes |
|---|---|---|
| Req 1 (Send Message) | ⚠️ Partial | Missing `is_deleted` check, wrong eligibility field (`login` vs `password`), rejects unverified users incorrectly |
| Req 2 (Read Messages) | ✅ Covered | Pagination, mark-as-read, PII stripping all implemented |
| Req 3 (Unread Count) | ✅ Covered | Active/archived split implemented |
| Req 4 (Archive) | ⚠️ Partial | AC 6 (increment unread for archived) relies on existing logic but no explicit test |
| Req 5 (PII Protection) | ✅ Covered | `formatParticipant` strips PII consistently |
| Req 6 (Authorization) | ✅ Covered | tokenAuth on all endpoints, participant-only access |
| Req 7 (Email Notification) | ⚠️ Partial | No conversation link, subject not localized |
| Req 8 (Notification Prefs) | ✅ Covered | Unified endpoint, partial updates, default true |
| Req 9 (Error Handling) | ⚠️ Partial | Inconsistent — some controllers use structured format, others don't |

dawoldo:
> ### 1. Missing `is_deleted` check in `getEligibleRecipient` (Requirement 1, AC 7)
> 
> The requirements define an Eligible_Recipient as a caver who is "NOT banned AND NOT is_deleted." However, `MessageService.getEligibleRecipient` only checks `banned`, `activated`, and `login`:
> 
> ```js
> if (caver.banned || !caver.activated || !caver.login) {
> ```
> 
> It never checks `caver.isDeleted`. A soft-deleted caver would pass the eligibility check and could receive messages. This violates Requirement 1 AC 7: "WHEN a Sender attempts to send a message to a caver ID that does not exist or to a soft-deleted (is_deleted) caver, THE Messaging_System SHALL reject the request and return a not-found error."

The Caver model has no ```is_deleted``` property. Are we supposed to be implementing soft deletes as part of this story(which would imply a database migration)? If we need to add it, please update the issue's description to explicitely say so.

ClemRz (changes requested):
## What was fixed since last review

- ✅ **Issue number 1** — `is_deleted` doesn't exist on `t_caver`. Eligibility logic `(mail_is_valid OR NOT activated) AND password IS NOT NULL AND NOT banned` is correctly implemented. Out of scope.
- ✅ **Issue number 2** — Eligibility now checks `!caver.password` (not `!caver.login`), and unactivated users are allowed
- ✅ **Issue number 3** — Indexes added to `test/customSQL.js`
- ✅ **Issue number 5** — Archive/unarchive now return 204
- ✅ **Issue number 8** — Notification is fire-and-forget (no `await`)
- ✅ **Issue number 9** — Error codes are now consistent (`E_AUTHORIZATION`)
- ✅ **Issue number 11** — `list-conversations` uses structured error format
- ✅ **Issue number 12** — `count-unread` uses structured error format
- ✅ **Issue number 13** — `setTimeout` reduced to 100ms

---

## Previously raised — still unaddressed

### 1. `JParticipant` schema breaks project conventions (prev number 4)

The `j_participant` table stores mutable state (`state`, `archived_at`) directly on the junction row. This breaks the project's pattern where all `j_*` tables are pure link tables with no extra columns, no direct Waterline queries, and no need for a serial `id`.

The problem this causes: the global `config/models.js` injects a default `id` attribute into every model. Since `j_participant` has no `id` column in SQL, Waterline operations like `JParticipant.updateOne(...)` and `JParticipant.findOne(...)` work in tests (`migrate: 'drop'` creates the `id` column) but may fail in production (`migrate: 'safe'`, no `id` column exists).

**Recommended fix:** Extract archive state into a dedicated table with a proper serial PK:

```sql
CREATE TABLE t_conversation_archive (
    id serial NOT NULL,
    id_conversation int4 NOT NULL,
    id_caver int4 NOT NULL,
    archived_at timestamp NOT NULL DEFAULT now(),
    CONSTRAINT t_conversation_archive_pk PRIMARY KEY (id),
    CONSTRAINT t_conversation_archive_conversation_fk FOREIGN KEY (id_conversation) REFERENCES t_conversation(id) ON DELETE CASCADE,
    CONSTRAINT t_conversation_archive_caver_fk FOREIGN KEY (id_caver) REFERENCES t_caver(id) ON DELETE CASCADE,
    CONSTRAINT t_conversation_archive_unique UNIQUE (id_conversation, id_caver)
);
```

This gives you:

- `j_participant` stays a pure link table (consistent with every other `j_*` table) — drop the `state` and `archived_at` columns, drop the `participant_state` enum
- Archive state has its own Waterline model (`TConversationArchive`) with a proper `id` PK — `findOne`, `create`, `destroyOne` all work natively
- Archive = `TConversationArchive.create(...)`, unarchive = `TConversationArchive.destroyOne(...)`, check = `TConversationArchive.count(...)`
- Easier to extend later (e.g., `muted_at`, `pinned_at`) without bloating the junction

- ✅ **Issue number 6** — `conversationLink` is empty with an explicit TODO. Acceptable deferral.

- ✅ **Issue number 7** — Email subject not localized, but consistent with existing codebase pattern (`'Notification'`). Not an issue.

### 2. `formatMessagingError` naming (prev number 10)

Still used in account controllers (`get-notifications.js`, `update-notifications.js`). The name implies messaging-specific scope. Low priority but worth a rename to `formatStructuredError` for clarity.

---

## New issues in this push

### 3. `markAsRead` silently unarchives conversations (Critical)

`MessageService.markAsRead` unconditionally resets participant state to `'active'` and clears `archivedAt`:

```js
markAsRead: async (conversationId, readerId) => {
  await JParticipant.update({
    conversation: conversationId,
    caver: readerId,
  }).set({ state: 'active', archivedAt: null });
  // ...
}
```

This is called every time `getMessages` is invoked. If a user archives a conversation then opens it to re-read, the archive state is silently destroyed. This contradicts the "independent archival states per participant" design.

**Suggestion:** If following the recommendation in item 1 (separate `t_conversation_archive` table), this problem disappears by design — `markAsRead` would only update `t_message.date_read` and have no reason to touch archive state. If keeping the current schema, remove the `JParticipant.update(...)` call from `markAsRead` and leave it as a message-only operation.

### 4. Duplicate `Conversation` schema in Swagger

`Conversation` is defined twice in `swaggerV1.yaml` (~line 5788 and ~line 6445) with different shapes. The second silently overrides the first. Remove the first definition (the one with `participants`/`messages` arrays).

### 5. Spurious entry in `fr.json`

Last line adds `"Nouveau message": "Nouveau message"` — a French-to-French key that serves no purpose. The i18n key should be the English string, which is already present as `"New Message": "Nouveau message"`. Also, the file lost its trailing newline.

### 6. Notification email uses sender's locale

`notifyMessageRecipient` uses `req.i18n` (the sender's request locale) rather than resolving the recipient's preferred language. However, the existing `sendNotificationEmail` in the same file follows the same pattern — so this is consistent with the codebase. Low priority unless the project decides to fix it globally.

### 7. Swagger response codes don't match implementation

- `list-conversations` and `list-archived-conversations` return `206` (via `treatAndConvert` → `partialContent`), Swagger documents `200`.
- `archive`/`unarchive` return `204`, Swagger documents `200`.

### 8. No message-specific rate limiting

`POST /api/v1/messages` is covered by the general rate limit (100 req/30s per IP), but has no message-specific throttle. A user could send ~100 messages in 30 seconds. Consider a stricter per-user limit for this endpoint (e.g., 10 messages/minute).

ClemRz (changes requested):
Good progress since the last review. The `t_conversation_archive` extraction, the `formatStructuredError` rename, the Swagger response code fixes, and the `fr.json` cleanup all address the previously raised issues cleanly.

---

### Issues still present

#### 1. `listConversations` result mapping drops `archivedAt` (Medium)

The SQL query selects `tca.archived_at as "archivedAt"` and the Swagger documents it on the `Conversation` schema, but the JS mapping in `MessageService.listConversations` never includes it in the returned object:

```js
return result.rows.map((row) => ({
  id: row.id,
  dateInscription: row.dateInscription,
  lastMessage: ...,
  unreadCount: ...,
  otherParticipant: ...,
  // ← archivedAt is missing
}));
```

The archived conversations endpoint will never surface `archivedAt` to the client despite the Swagger promising it.

**Fix:** Add `archivedAt: row.archivedAt || null` to the mapping.

---

#### 2. Swagger missing `404` on `/messages/conversations/{id}/unarchive` (Minor)

The `unarchive` controller returns `res.notFound(...)` when the conversation doesn't exist, but the Swagger spec for that endpoint only documents `204`, `401`, `403`, and `500`. The `archive` endpoint correctly documents `404`.

---

#### 3. `test/customSQL.js` missing sequence reset for `t_conversation_id_seq` and `t_message_id_seq` (Medium)

The PR adds `SELECT SETVAL('public.t_conversation_archive_id_seq', ...)` but does not add equivalent resets for `t_conversation_id_seq` and `t_message_id_seq`. Since the tests create `TConversation` and `TMessage` records via Waterline (which uses the serial PK), the sequences should be reset alongside the other tables to avoid ID conflicts if fixtures are ever added for these tables.

---

#### 4. `setTimeout` in sort test is still timing-dependent (Low — previously raised)

```js
await new Promise((resolve) => {
  setTimeout(resolve, 100);
});
```

100ms is better than 1000ms, but it's still a race against the DB clock. A more robust approach: insert the archive row directly via `TConversationArchive.create(...)` with an explicit `archivedAt` value set 1 second apart, bypassing the need for real-time delay. This eliminates any flakiness on slow CI runners.

ClemRz (changes requested):
Good work addressing the issues from the previous reviews. The `archivedAt` mapping fix, the Swagger 404 on unarchive, the sequence resets, and the deterministic sort test all look correct.

A few remaining items:

---

### 1. `JParticipant` model will get a phantom `id` column in tests (Low — informational)

Since `config/models.js` injects a default `id` attribute into every model and the test DB uses `migrate: 'drop'`, Waterline will create an `id` serial column on `j_participant` in tests. In production (`migrate: 'safe'`), that column doesn't exist. This is fine *today* because no code calls `JParticipant.findOne()` or `JParticipant.updateOne()` — all participant queries go through raw SQL. But it's a latent trap: if someone later adds a Waterline call on `JParticipant`, it'll pass tests and fail in prod.

Consider adding a comment in the model file to warn future contributors, or explicitly set `primaryKey: 'conversation'` (even if imperfect) to signal that this model doesn't follow the standard `id`-based pattern.

Not blocking.

---

### 2. `body` column type mismatch between SQL and Waterline model (Low)

SQL defines `body text NOT NULL` (unlimited length), but the Waterline model declares `maxLength: 5000`. The controller already validates the 5000-char limit before insertion, so the `maxLength` on the model is redundant but harmless. However, if someone bypasses the controller (e.g., a service call), Waterline will reject bodies > 5000 chars even though the DB would accept them. This is probably fine as defense-in-depth, just noting the discrepancy.

---

### 3. `getMessages` sorts ASC but pagination skips from the beginning (Low)

```js
.sort('dateSent ASC')
.skip(skip)
.limit(limit)
```

With ASC sort + skip=0, the first page returns the *oldest* messages. Most messaging UIs show the newest messages first (or load the latest page by default). The frontend will need to either request the last page first or reverse the array. This is a UX/API design choice — just flagging it in case it wasn't intentional.

---

### 4. No `idLanguage` default in `createEligibleUser` test helper

The test helper sets `idLanguage: 'eng'` by default, but some test calls override it (e.g., `notification-preferences.test.js` uses `'000'`). The `TCaver` model likely requires `idLanguage`. This is fine, just noting that `'eng'` may not match a real fixture language ID — if the test DB has FK constraints on `id_language`, this could fail. Verify that `'eng'` exists in the `t_language` fixture.

ClemRz (changes requested):
### Issues (Must Fix)

- **[sql/0_tables.sql:1393]** The `t_message` table defines `id_caver_sender int4 NOT NULL` but the FK constraint uses `ON DELETE SET NULL`. These are contradictory — if a caver row is ever deleted, PostgreSQL will attempt to set `id_caver_sender` to NULL, which violates the NOT NULL constraint and causes the DELETE to fail with an error. Either change the column to allow NULLs (`id_caver_sender int4`) or change the FK action to `ON DELETE CASCADE` or `ON DELETE RESTRICT` depending on the desired behavior. Given that messages should be preserved even if a caver is deleted (Requirement 2 AC 7), the recommended fix is to allow NULLs:
  ```suggestion
  id_caver_sender int4,
  ```
  And update the Waterline model (`TMessage.js`) accordingly: change `allowNull: false` to remove it or set it to `true` on `caverSender`. The `formatParticipant` helper already handles null participants gracefully (returns `'Deleted User'`).

- **[config/locales/en.json]** Missing trailing newline. The `.editorconfig` requires `insert_final_newline = true`, but `en.json` ends without one (the diff shows `\ No newline at end of file`). This will cause lint/CI warnings and noisy diffs in future PRs.

- **[api/controllers/v1/message/create.js:44-56]** When `conversationId` is provided but the conversation doesn't exist in the DB (e.g., invalid ID), the raw SQL query returns 0 rows and the controller returns 403 ("You are not a participant"). This is misleading — a 404 would be more appropriate for a non-existent conversation. Consider adding a `TConversation.count({ id: conversationId })` check before the participant query (similar to what `archive.js`, `unarchive.js`, and `get-messages.js` already do) to distinguish "conversation doesn't exist" (404) from "you're not a participant" (403).

- **[api/services/MessageService.js:167-175]** The `getMessages` method sorts `DESC` then reverses the array in JS. This works but is slightly confusing for pagination: with `skip=0`, the client gets the *newest* page of messages (reversed to chronological order). This means to load older messages, the client must increment `skip` — but the total count doesn't change, so the client can't easily tell which "page" they're on in a traditional sense. This is a valid design choice for chat UIs (load latest first, scroll up for history), but it's worth documenting this behavior in the Swagger spec or a code comment so frontend developers understand the pagination semantics.

- **[api/services/NotificationService.js:221]** The `conversationLink` TODO is acceptable as a deferral, but consider at minimum constructing a placeholder URL pattern (e.g., `${sails.config.custom.frontendUrl}/messages/${conversationId}`) so the email template renders something useful once the frontend route exists. This avoids a second backend deployment just to enable the link.

- **[api/controllers/v1/message/archive.js / unarchive.js]** Both controllers use `res.status(204).send()` which is a raw Express response. While this pattern exists elsewhere in the codebase, the project conventions prefer Sails custom responses. The `res.ok()` response already returns 204 when called without data (see `api/responses/ok.js` line 21: `const httpCode = data ? 200 : 204`). Consider using `return res.ok()` instead for consistency with the response logging middleware.

- **[test/integration/4_routes/Account/notification-preferences.test.js:13]** The test re-declares `const TCaver = sails.models.tcaver` inside the `before` hook, but `TCaver` is already a global in this project (declared in `.eslintrc.js`). This is unnecessary and inconsistent with the `Messages.test.js` file which uses the global directly.

### Nitpicks (Optional)

- **[api/controllers/v1/message/archive.js:1 / unarchive.js:7]** The `require` for `MessageService` uses a relative path (`'../../../services/MessageService'`). Services are globals in Sails.js — you could use `MessageService.isParticipant(...)` directly without the require. However, since the `create.js` and `get-messages.js` controllers also use explicit requires, this is consistent within the PR. Just noting that the project convention for services is to use them as globals.

- **[config/locales/en.json]** The new key `"Verify your email address": "Verify your email address"` appears unrelated to the messaging feature. If it's needed for another feature, it should be in a separate commit/PR to keep changes focused.

- **[api/services/MessageService.js:17-27]** The `findExistingConversation` query uses a subquery pattern. A slightly more efficient approach for exactly-two-participant conversations would be a self-join: `SELECT p1.id_conversation FROM j_participant p1 JOIN j_participant p2 ON p1.id_conversation = p2.id_conversation WHERE p1.id_caver = $1 AND p2.id_caver = $2 LIMIT 1`. Minor optimization, not blocking.

ClemRz (dismissed):
Good progress since the last round. The SQL schema is now consistent (nullable `id_caver_sender` with `ON DELETE SET NULL`), the `en.json` trailing newline is fixed, and the `TMessage` model aligns with the DB definition.

- **[config/locales/fr.json:41]** The `"Verify your email address": "Verify your email address"` entry is untranslated — it's English text in the French locale file. It also appears unrelated to the messaging feature. Consider either providing the French translation (`"Verify your email address": "Vérifiez votre adresse email"`) or removing it from this PR if it belongs to a different feature.

- **[api/controllers/v1/message/list-conversations.js:1-2, list-archived-conversations.js:1-2, count-unread.js:1]** These controllers use explicit `require('../../../services/MessageService')` and `require('../../../services/ControllerService')`, while `archive.js`, `unarchive.js`, `get-messages.js`, and `create.js` use the globals directly (which are declared in `.eslintrc.js`). This is inconsistent within the same PR. Since the globals are already registered, consider removing the `require` statements from these three files for consistency.

- **[api/services/NotificationService.js:222]** The `conversationLink` TODO is acceptable as a deferral, but consider constructing a placeholder URL pattern (e.g., `${sails.config.custom.frontendUrl || 'https://grottocenter.org'}/ui/messages/${conversationId}`) so the email template renders something useful once the frontend route exists. This avoids a second backend deployment just to enable the link.

### Nitpicks (Optional)

- **[config/locales/en.json:41]** The `"Verify your email address": "Verify your email address"` key appears unrelated to the messaging feature. If it's needed for another feature, it would be cleaner in a separate commit to keep changes focused.

- **[api/services/NotificationService.js:207]** Minor: double space in the comment `// TODO: construct conversation link when the  frontend is worked on` → should be single space.

ClemRz (changes requested):
### Issues (Must Fix)

- **[config/routes.js:70, config/policies.js:35]** This PR adds a route (`PATCH /api/v1/account/email → v1/account/change-email`) and a policy entry (`v1/account/change-email: 'tokenAuth'`) for a controller that doesn't exist (`api/controllers/v1/account/change-email.js`). Sails will throw a 500 error at startup or when the route is hit. Either remove these two lines. It seems to be a bad merge conflict resolution.

- **[api/services/NotificationService.js:226]** The conversation link uses `sails.config.custom.frontendUrl` which is not defined anywhere in the project configuration. The project uses `sails.config.custom.baseUrl` (defined in `config/custom.js` and `config/env/production.js`). This will produce a broken URL like `undefined/ui/messages/123` in every notification email.
  ```suggestion
  const conversationLink = `${sails.config.custom.baseUrl}/ui/messages/${conversationId}`;
  ```

- **[api/services/NotificationService.js:219]** The `notifyMessageRecipient` function iterates over all participants (excluding the sender) and sends emails. For a 1-on-1 conversation this is always exactly one recipient, but the `Promise.all(result.rows.map(...))` pattern adds unnecessary complexity. A simpler approach would be to take the first (and only) row directly, which also makes the intent clearer.

- **[api/controllers/v1/message/create.js:193-200]** The `NotificationService.notifyMessageRecipient` call passes `req` as the first argument, but the service only uses it for... nothing actually. The `req` parameter is declared but never used inside `notifyMessageRecipient`. Consider removing it from the function signature and the call site to avoid confusion. The email templates evolved since the first creation of this PR.

- **[test/integration/4_routes/Messages.test.js:56-62]** The `after` hook uses `await TMessage.destroy({})` which destroys ALL messages in the test database, not just the ones created by this test suite. In a parallel test runner, this could interfere with other shards. Consider scoping the cleanup to only the records created during this test (e.g., by tracking conversation IDs and deleting messages by conversation).

- **[api/models/TMessage.js:30-32]** The `caverSender` attribute has no `allowNull` setting, which means Waterline defaults to `allowNull: false`. However, the SQL schema defines `id_caver_sender int4` (nullable) with `ON DELETE SET NULL`. If a caver is deleted, PostgreSQL will set this to NULL, but Waterline validation would reject a NULL value on subsequent reads/updates. Consider adding `allowNull: true` explicitly to match the DB schema.

- **[sql/0_tables.sql:1393]** Minor inconsistency: the constraint line uses a tab character for indentation (`\tCONSTRAINT t_message_t_caver_fk`) while all other lines in the same CREATE TABLE use spaces. This doesn't affect functionality but is inconsistent with the rest of the file.

### Nitpicks (Optional)

- **[config/locales/fr.json:48]** The `"Verify your email address": "Vérifiez votre adresse mail"` entry appears unrelated to the messaging feature. If it belongs to a different PR/feature, consider keeping changes focused.

- **[api/services/MessageService.js:1]** `const CommonService = require('./CommonService')` — `CommonService` is already declared as a global in `.eslintrc.js`. The explicit require is unnecessary (though harmless). The same applies to the `require` of `MessageService` in `test/integration/1_services/NotificationService.test.js:4`.

- **[test/integration/4_routes/Account/notification-preferences.test.js]** This test file duplicates coverage already present in `Account.test.js` (which also tests GET and PATCH `/api/v1/account/notifications`). Consider consolidating into one location to avoid maintaining parallel test suites for the same endpoints.

ClemRz (dismissed):
Good progress since the last round. The ghost `change-email` route/policy is removed, `baseUrl` is now correctly used for the conversation link, the `TMessage.caverSender` association is properly nullable, the `NotificationService` no longer takes an unused `req` parameter, and the duplicate notification-preferences test file has been consolidated into `Account.test.js`. All previously raised critical issues have been addressed.

- **[api/controllers/v1/message/create.js:87-100]** When `conversationId` is provided and the other participant's eligibility check fails with `E_FORBIDDEN`, the controller returns 403 with code `E_AUTHORIZATION`. This is correct per the requirements, but the error *message* from `getEligibleRecipient` says "Recipient is not eligible for private messaging" — which leaks information about the other participant's account status (banned/inactive) to the sender. Consider using a generic message like "You cannot send messages in this conversation" to avoid revealing why the other participant is ineligible.

- **[api/controllers/v1/message/get-messages.js:18]** `parseInt(req.param('skip', 0), 10)` — if the user passes a non-numeric string like `skip=abc`, `parseInt('abc', 10)` returns `NaN`, which would cause `TMessage.find().skip(NaN)` to behave unpredictably. Consider adding a `|| 0` fallback: `const skip = parseInt(req.param('skip'), 10) || 0;`. The same pattern applies to `limit` in all paginated controllers (`list-conversations.js`, `list-archived-conversations.js`, `get-messages.js`).

- **[api/services/MessageService.js:155-162]** The `getMessages` method calls `markAsRead` before fetching messages. If `markAsRead` fails (e.g., DB connection issue), the entire request fails and the user sees no messages. Consider swapping the order — fetch messages first, then mark as read — so that a mark-as-read failure doesn't prevent the user from reading their messages. Alternatively, wrap `markAsRead` in a try/catch that logs but doesn't propagate.

- **[test/integration/4_routes/Messages.test.js:56-75]** The `after` hook cleans up conversations by querying `j_participant` for the sender. However, tests like "should allow sending to an unactivated user" and "should return 403 if recipient is activated but has invalid email" create temporary users and destroy them in `finally` blocks — but any conversations created with those users leave orphaned `j_participant` and `t_conversation` rows (the FK cascade on `j_participant` is on `id_caver`, so deleting the caver removes the participant row, but the conversation itself remains). Consider tracking those conversation IDs for cleanup, or relying on the sender-based participant query in `after` (which should catch them since the sender is always a participant).

- **[config/locales/fr.json:47]** The `"Verify your email address": "Vérifiez votre adresse mail"` entry appears unrelated to the messaging feature. If it belongs to a different PR/feature, consider keeping changes focused to reduce noise in the diff.

### Nitpicks (Optional)

- **[sql/0_tables.sql:1388]** The `body` column uses `varchar(5000)` while PostgreSQL convention for longer text fields is `text` (no performance difference). `varchar(5000)` is fine and matches the explicit business rule, just noting the alternative.

- **[views/emailTemplates/new-message.ejs:30]** The template doesn't end with a newline character. The `.editorconfig` specifies `insert_final_newline = true`.

- **[api/services/MessageService.js:95]** The `listConversations` query computes unread counts for ALL of the caver's conversations in a subquery, not just the ones being returned on the current page. For users with many conversations this is slightly wasteful, but unlikely to be a real performance issue given the messaging context. Could be optimized later with a correlated subquery if needed.

ClemRz (dismissed):
All previously raised critical issues have been addressed. The architecture is clean — `t_conversation_archive` is properly separated, `JParticipant` has the composite PK warning comment, eligibility logic matches the requirements, indexes are mirrored in `test/customSQL.js`, and the ghost `change-email` route/policy is gone.

- **[api/controllers/v1/message/get-messages.js:52-55]** The controller calls `MessageService.getMessages(conversationId, skip, limit, caverId)` and then separately calls `MessageService.countMessages(conversationId)`. Inside `getMessages`, `markAsRead` is called first. If `markAsRead` succeeds but the subsequent `TMessage.find()` fails, the messages are marked as read without the user ever seeing them. Consider swapping the order inside `getMessages` — fetch first, then mark as read — so a fetch failure doesn't silently consume unread state.

- **[api/services/MessageService.js:95]** The `listConversations` query computes unread counts via a subquery that scans ALL messages for the caver (not just the conversations on the current page). For users with many conversations and messages, this could become slow. A correlated subquery scoped to the returned conversations would be more efficient, though this is unlikely to be a real issue at current scale.

- **[api/services/NotificationService.js:210]** The `notifyMessageRecipient` function doesn't fall back to a default locale when `LanguageService.getLocale` returns `undefined` (e.g., if the recipient has no language set). Other notification functions in the same file (via `sendNotificationEmail`) fall back to `sails.config.i18n.defaultLocale`. Consider adding `|| sails.config.i18n.defaultLocale` after the `getLocale` call to avoid passing `undefined` as the locale to the email helper.

- **[test/integration/4_routes/Messages.test.js:56-75]** The `after` hook cleanup queries `j_participant` for the sender's conversations. If a test creates a conversation between two *other* users (not the sender), those won't be cleaned up. Currently all tests use `sender` as a participant so this is fine, but it's fragile for future test additions. Consider also cleaning up conversations tracked in `manuallyCreatedConversationIds` that might not have the sender as a participant (the `otherConvo` tests create conversations with no participants at all).

- **[test/integration/1_services/NotificationService.test.js:4]** `const MessageService = require('../../../api/services/MessageService')` — `MessageService` is declared as a global in `.eslintrc.js`. The explicit require is unnecessary (though harmless and consistent with the existing `CaveService` require pattern in the same file).

### Nitpicks (Optional)

- **[config/locales/fr.json:47, config/locales/en.json:47]** The `"Verify your email address"` key appears in the diff but exists in all other locale files already — it seems to have been re-sorted alphabetically as part of this PR's locale changes. Not an issue, just noting it's unrelated to messaging.

- **[sql/3_14_mock_conversations.sql]** The mock data file is well-structured and provides good test coverage for pagination scenarios. Consider adding a brief comment at the top explaining which caver IDs map to which roles (e.g., `-- Caver 1 = Admin, 2 = Moderator, 3 = User, etc.`) for future maintainability.

- **[api/services/MessageService.js:17-27]** The `findExistingConversation` query uses a self-join pattern (as suggested in a previous review). Clean implementation.

ClemRz (approved):
All previously raised critical issues have been addressed. The architecture is solid — `t_conversation_archive` is properly separated, eligibility logic matches the requirements, indexes are mirrored in `test/customSQL.js`, locale fallback is in place, and `getMessages` now fetches before marking as read.

- **[api/services/MessageService.js:95]** The `listConversations` query computes unread counts via a subquery that scans ALL messages for the caver (not just the conversations on the current page). For users with many conversations and messages, this could become slow. A correlated subquery scoped to the returned conversations would be more efficient, though this is unlikely to be a real issue at current scale.

- **[api/controllers/v1/message/create.js:87-100]** When `conversationId` is provided and the other participant's eligibility check fails with `E_FORBIDDEN`, the error message from `getEligibleRecipient` is replaced with "You cannot send messages in this conversation" — which is good for privacy. However, when it fails with `E_NOT_FOUND`, the raw `err.message` ("Recipient not found") is passed through. This leaks information about the other participant's account state (deleted). Consider using a generic message like "You cannot send messages in this conversation" for both cases.

- **[test/integration/4_routes/Messages.test.js:56-75]** The `after` hook cleanup is much improved with the `ANY($1)` query across all test cavers. One edge case: the `otherConvo` conversations created via `TConversation.create({}).fetch()` have no participants at all (they're used to test "not a participant" scenarios). These are tracked in `manuallyCreatedConversationIds` which is good, but if a test fails before pushing to that array, the conversation would be orphaned. Minor — not blocking.

- **[config/locales/en.json, config/locales/fr.json]** The `"%s (%s) has been %s on Grottocenter."` key was moved from the top of the file to an alphabetically-sorted position. This is a cosmetic re-sort that creates noise in the diff but doesn't affect functionality. Just noting it for awareness.

### Nitpicks (Optional)

- **[api/services/MessageService.js:1]** `CommonService` is used as a global throughout the file (no `require` statement), which is correct per project conventions. However, `MessageService` is explicitly required in `test/integration/1_services/NotificationService.test.js:4` despite being a global. Harmless but inconsistent.

- **[sql/3_14_mock_conversations.sql]** The mock data file is well-structured. Consider adding a brief comment at the top mapping caver IDs to roles (e.g., `-- Caver 1 = Admin, 2 = Moderator, 3 = User, etc.`) for future maintainability.

PR #1567, opened 2026-05-07
Fix/1450 massif sensitive setting

dawoldo:
## 🤔 What

At the request of Frédéric, this PR would push a fix to expose the isSensitive status for massifs to all users and refactors the corresponding integration tests to take into account this new behavior

## 🔍 How

**API Change:** Updated the toMassif converter to always return the isSensitive property, removing any previous access restrictions.
**Test Refactoring:** Consolidated the find.test.js suite by merging overlapping describe blocks and field verification lists for the GET /api/v1/massifs/:id endpoint.

## 🧪 Testing

N/A

## 📸 Previews

N/A

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -73,6 +73,14 @@ module.exports = {
     'prefer-destructuring': ['error', { object: true, array: false }],
     'prettier/prettier': 'error',
     'import/no-unresolved': ['error', { ignore: ['^uuid$'] }],
+    'no-unused-vars': [
+      'error',
+      {
+        argsIgnorePattern: '^_',
+        varsIgnorePattern: '^_',
+        ignoreRestSiblings: true,
+      },
+    ],
   },
   plugins: ['prettier', 'import'],
 };
--- a/api/controllers/v1/massif/mark-sensitive.js
+++ b/api/controllers/v1/massif/mark-sensitive.js
@@ -3,7 +3,6 @@ const MassifService = require('../../../services/MassifService');
 const EntranceService = require('../../../services/EntranceService');
 const ControllerService = require('../../../services/ControllerService');
 const { toMassif } = require('../../../services/mapping/converters');
-const { getMetaFromRequest } = require('../../../services/mapping/utils');
 
 module.exports = async (req, res) => {
   const isAdmin = RightService.hasGroup(
@@ -28,13 +27,12 @@ module.exports = async (req, res) => {
   // Idempotency: skip if already sensitive
   if (massif.isSensitive) {
     const updatedMassif = await MassifService.getPopulatedMassif(massifId);
-    const meta = getMetaFromRequest(req);
     return ControllerService.treat(
       req,
       null,
       {
         count: 0,
-        massif: toMassif(updatedMassif, meta),
+        massif: toMassif(updatedMassif),
       },
       { controllerMethod: 'MassifController.mark-sensitive' },
       res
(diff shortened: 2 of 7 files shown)


ClemRz (changes requested):
#### 1. Dead `meta` parameter still passed by callers (minor cleanup missed)

The `toMassif` signature was changed from `(source, meta)` to `(source)`, but several callers still pass `meta`:

- `api/controllers/v1/massif/mark-sensitive.js` — `toMassif(updatedMassif, meta)` (×2)
- `api/controllers/v1/massif/unmark-sensitive.js` — `toMassif(updatedMassif, meta)` (×2)
- `api/services/mapping/converters.js` line ~591 — `convertIfObject(source.massif, c.toMassif, { meta })`
- `api/services/mapping/converters.js` line ~711 — `c.toMassif(item.document, meta)`

JavaScript silently ignores extra arguments, so this isn't a bug — but it leaves dead code that will confuse future readers. The `getMetaFromRequest` import in `mark-sensitive.js` and `unmark-sensitive.js` is now only used for this dead parameter.

**Suggestion:** Remove the unused `meta` argument from all call sites and drop the now-unnecessary `getMetaFromRequest` import in `mark-sensitive.js` and `unmark-sensitive.js` (or do it in a follow-up).

#### 2. `ControllerService.treatAndConvert` still passes `meta` to the converter

In `ControllerService.js` line ~82:
```js
const meta = getMetaFromRequest(req);
return res.ok(converter(found, meta));
```

This means `toMassif` is called with `meta` as a second arg from the `find.js` controller path. Again harmless since the param is now ignored, but worth noting for consistency.

#### 3. Test coverage: no unauthenticated assertion that `isSensitive` is actually present with a meaningful value

The refactored test checks `should(massif).have.properties(MASSIF_PROPERTIES)` which confirms the key exists, but doesn't assert the value is a boolean or matches the fixture. A one-liner like `should(massif.isSensitive).be.a.Boolean()` would strengthen confidence.

ClemRz (commented):
### Previous feedback resolution

1. **Dead `meta` parameter** — ✅ Fixed. The `getMetaFromRequest` import and all dead `meta` arguments have been removed from `mark-sensitive.js` and `unmark-sensitive.js`. The `convertIfObject` call for `toMassif` in the notification converter also no longer passes `{ meta }`.

2. **`ControllerService.treatAndConvert` still passing `meta`** — ✅ Addressed via `converter.length` arity check (see concern below).

3. **Boolean assertion on `isSensitive`** — ✅ Added: `should(massif.isSensitive).be.a.Boolean()`.

---

### New concern: `converter.length` arity check is fragile

In `ControllerService.js`:

```js
converter.length > 1 ? converter(found, meta) : converter(found)
```

This relies on `Function.length` to decide whether to pass `meta`. While it works today, it's a brittle contract:

- If someone adds a **default parameter** to a converter (e.g., `(source, options = {})`), `Function.length` stays at 1 because JS doesn't count params with defaults. That converter would silently stop receiving `meta`.
- If someone wraps a converter in an arrow with rest params (`(...args) => converter(...args)`), length becomes 0.
- It's an implicit, non-obvious contract — a future contributor won't know that the number of declared parameters controls runtime behavior unless they read `ControllerService.js` carefully.

**Suggestion:** This works for the immediate goal (stop passing `meta` to `toMassif`), but consider adding a brief code comment explaining the convention, e.g.:

```js
// Converters that declare a second parameter receive request metadata (meta).
// Single-param converters (e.g., toMassif) are called without meta.
```

This is a minor nit — not a blocker.

---

### Test changes look good

- The consolidated `forEach` for 404 cases is cleaner and equivalent.
- Removing the admin-specific `isSensitive` test correctly reflects the new behavior (visible to all).
- `MASSIF_PROPERTIES` now includes `isSensitive` and `networks`, and the shape assertion is solid.

### One small issue in the test spy

```js
const converterSpy = sinon.spy((data, meta) => data); // eslint-disable-line no-unused-vars
```

The eslint-disable comment suppresses the `no-unused-vars` warning for `meta`. This is fine functionally — the spy needs `length === 2` to exercise the `converter.length > 1` branch — but it would be clearer to name it `_meta` (conventional "intentionally unused" prefix) instead of using the disable comment. Very minor.

---

### Verdict

The PR is in good shape after the revisions. The only substantive point is the fragility of the `converter.length` approach, which deserves at minimum a clarifying comment. Everything else is clean.

ClemRz (approved):

PR #1573, opened 2026-05-08
feat(geocoding): offline country resolution + async enrichment queue

### LINKED ISSUE #1570: bug(api): [23502] tentrance - null id_country on create
**Describe the bug**
When creating an entrance via `POST /api/v1/entrances`, the API attempts to insert a row into `t_entrance` with a null `id_country`. The root cause is that the reverse geocoding call to OSM Nominatim fails (returns HTML instead of JSON, likely due to rate limiting or service unavailability), and `EntranceService` does not handle this failure gracefully — it proceeds with the database insert without a country code, violating the NOT NULL constraint.

**To Reproduce**
Based on the trace logs:
1. Authenticate as any user
2. Send `POST /api/v1/entrances` with valid coordinates (e.g., latitude=0.351045, longitude=22.148438)
3. The Nominatim reverse geocoding service returns an error (HTML content type instead of JSON)
4. The service proceeds to create the entrance without resolving the country
5. PostgreSQL rejects the insert due to the NOT NULL constraint on `id_country`

**Expected behavior**
The API should handle geocoding failures gracefully:
- Either retry the geocoding request
- Or return a meaningful error to the client indicating the country could not be determined
- Or require the client to provide the country explicitly when geocoding fails

**Error Details**
- Occurrences: 4
- Time Range: 2026-05-08T00:11:50Z to 2026-05-08T15:21:55Z
- Representative Trace ID: `019e082e-5300-721f-9fd0-e613ad200ad4`
- Stack Trace:
```
at Object.during (/home/site/wwwroot/api/services/EntranceService.js:197:43)
at _makeCallToDuringFn (sails-hook-orm/lib/datastore-method-utils/help-run-transaction.js:88:48)

Preceding error in trace:
osmNominatimReverse 0.351045 22.148438 Error: Content type is not json: text/html; charset=utf-8
    at ClientRequest.<anonymous> (/home/site/wwwroot/api/services/GeocodingService.js:41:13)
```

**Additional context**
- Service: `api/services/EntranceService.js` (line 197), `api/services/GeocodingService.js` (line 41)
- Model: `tentrance`
- DB Error Code: `23502` (not-null violation)
- Column: `id_country`
- All 4 occurrences came from the same user testing entrance creation with various coordinates
- Coordinates in failing rows: (0,0), (0.351045, 22.148438), (44.654216, 4.306641), (5.438297, 22.851563)

ClemRz:
## 🤔 What

Replace all synchronous Nominatim reverse-geocoding calls with a two-prong architecture:

1. **Offline country resolution** via `@rapideditor/country-coder` — synchronous, no network dependency
2. **Async enrichment queue** via `pg-boss` (PostgreSQL-backed) — for region, county, city, iso_3166_2

Resolves #1570

## 🤷‍♂️ Why

Entrance creation fails with a 500 when Nominatim is rate-limited (HTTP 429) or unable to geocode coordinates. The `id_country` NOT NULL constraint at the DB level causes the insert to fail when geocoding returns no result.

The root cause: a single bulk CSV import can exhaust Nominatim's 1 req/s rate limit, poisoning all subsequent entrance creation requests for other users.

## 🔍 How

**Prong 1 — CountryResolverService (synchronous):**
- Uses `@rapideditor/country-coder` for offline point-in-polygon lookup
- Hierarchy-based resolution: tries most-specific territory code first (e.g., `RE` for Réunion), falls back to sovereign (e.g., `FR`)
- Validates against an in-memory cache of `t_country` ISO codes (loaded at startup)
- Falls back to `'00'` (Undefined) when no country can be determined — allows admin moderation later

**Prong 2 — EnrichmentQueueService (async):**
- `pg-boss` PostgreSQL-backed job queue — no new infrastructure (uses existing PG)
- Worker processes at 1 req/s (Nominatim rate limit compliance)
- Exponential backoff retry on 429/network errors, max 5 attempts
- Job payload is `{ entityId, entityType, traceId }` — coordinates read from DB at processing time
- Stale enrichment fields cleared immediately on coordinate change
- Trace ID propagated from original request to background job logs

**Call sites modified:**
- `EntranceService.createEntrance` — country resolved offline, enrichment enqueued
- `entrance/update.js` — country re-resolved on coord change, stale fields cleared, enrichment enqueued
- `GrottoService.createGrotto` — sync Nominatim removed, enrichment enqueued
- `organization/update.js` — sync Nominatim removed, stale iso_3166_2 cleared, enrichment enqueued

**Infrastructure:**
- Queue starts/stops with the application (bootstrap.js)
- pg-boss auto-creates its schema in a `pgboss` namespace — no SQL migration needed
- Graceful shutdown via `sails.on('lower', ...)`

## 🧪 Testing

- Unit tests for CountryResolverService (hierarchy resolution, cache, edge cases)
- Property-based test for hierarchy resolution invariant (fast-check, 100 runs)
- Unit tests for EnrichmentQueueService (enqueue, processEntrance, processOrganization, error handling)
- Updated all existing tests that stubbed GeocodingService.reverse

To test manually:
1. `npm run dev:up && npm run dev`
2. Create/update an entrance with coordinates — observe enrichment logs
3. Try ocean coordinates (0, 0) — should get country `00`, no enrichment enqueued
4. Try overseas territory (Réunion: -21.1, 55.5) — should resolve to `RE` or `FR`

## 📸 Previews

```
info: [019e096c-...] Enrichment job enqueued: entrance 36776
info: [019e096c-...] Enrichment job started: entrance 36776
info: [019e096c-...] Enrichment job completed: entrance 36776 (region=La Réunion, iso_3166_2=FR-RE)
info: [019e096c-...] Enrichment job done: entrance 36776 452ms
```


urien:
Est ce qu'il serait pertinent de proposer d'autres apis que Nominatim quand cette api ne répond pas ? Ou que les réponses sont partielles...
Je viens de faire une demande à Google pour avoir un crédit de 250 $ mensuel pour utiliser les apis de Google Map.

Paul-AUB (dismissed):
This is a well-designed solution. The two-pronged approach — synchronous offline country resolution + async enrichment queue — is the right architecture for this problem. The `singletonKey` deduplication, traceId propagation through background jobs, the `global.grottoAsyncLocalStorage` singleton, and the property-based test coverage all reflect careful thinking.

1. **[config/bootstrap.js:50-57]** If `EnrichmentQueueService.start()` throws _after_ `sails.enrichmentBoss` is assigned (e.g., `createQueue` or `work` fails), the boss object is truthy but un-started. Subsequent `enqueue()` calls will bypass the "queue not initialized" warning path and instead throw from `boss.send()` on a broken boss — each error is caught by the caller's `.catch()`, so entrance creation is safe, but you get one noisy error per enqueue instead of one clean warning at startup. Fix: reset `sails.enrichmentBoss = null` inside the catch block so the graceful degradation path works as designed.

2. **[config/bootstrap.js:60-62]** `sails.on('lower', async () => { await stop(); })` — Node.js `EventEmitter` does not await async listeners. Under normal process termination, `stop()` (which waits up to 30 seconds for in-flight jobs to drain) may never complete before the process exits. pg-boss will recover on next startup, but worker/connection cleanup is skipped. Consider using `sails.config.beforeShutdown` or a synchronous-style listener with an explicit timeout callback if graceful job drainage is important.

3. **[EnrichmentQueueService.js:54-58]** The PR description says "Worker processes at 1 req/s (Nominatim rate limit compliance)" but the code enforces no such rate. `newJobCheckInterval: 1000` only controls poll frequency — if a backlog of jobs exists, pg-boss dispatches them as fast as the worker loop allows. The current single-process deployment makes this an emergent property of response time, but horizontal scaling (multiple API instances) would break the assumption. Worth documenting as a single-instance constraint, or adding an explicit per-job delay.

4. **[EnrichmentQueueService.js:149-178]** Narrow race condition: if coordinates change _while_ a job is executing (between the `TEntrance.findOne` read and the `TEntrance.updateOne` write), the job writes enrichment data geocoded from coordinates that are no longer current. `singletonKey` prevents duplicate _pending_ jobs but doesn't block a new job from being enqueued once the current one is running. The outcome is transient stale data until the second job completes — acceptable for enrichment metadata, but worth noting as a known eventual-consistency gap.

### Nitpicks (Optional)

5. **[.gitignore:138-140]** The `# Error triage agent timestamp` / `.last-error-triage` entry appears unrelated to the geocoding queue feature. Keeping unrelated changes out of focused PRs helps bisect and revert.

6. **[EnrichmentQueueService.js:54-58]** `newJobCheckInterval: 1000` polls PostgreSQL every second even when the queue is idle. Since enrichment latency is not user-facing, a higher interval (5 000–10 000 ms) would reduce background DB load at no meaningful cost.

7. **[package-lock.json]** `@rapideditor/country-coder@5.6.1` declares `"engines": { "bun": ">=1.3" }` with no Node.js constraint. This is a quirk of the package's metadata and does not affect runtime (the dynamic ESM import works on Node.js ≥12), but it may produce warnings with strict engine validation in CI.

PR #1575, opened 2026-05-12
Fix/1449 signup email verification

dawoldo:
## 🤔 What
Implemented a secure email verification workflow for both account registration and email changes.
- **Pending Email Storage**: Added `pendingMail` attribute to the `TCaver` model to store unverified email addresses.
- **Account Verification Enhancement**: Updated the `verify-email` controller to support both initial account activation and email change confirmation.
- **Secure Email Change**: Revamped the `change-email` controller to initiate a verification flow instead of immediate updates.
- **Conflict Management**: Implemented logic to prevent email collisions across both primary and pending email fields.
- **Cancellation Support**: Users can now cancel a pending email change by re-submitting their current email address.

## 🤷‍♂️ Why
Previously, changing an email address was immediate and unverified, which posed several risks:
1. **Security**: An attacker with temporary access to an account could change the email address and take full control without verification.
2. **Account Lockout**: If a user made a typo when changing their email, they could lose access to their account since the primary login identifier was updated immediately.
3. **Data Integrity**: There was no verification that the user actually owned the new email address.

This implementation ensures that the user's login identifier remains unchanged until the new email is verified, and ownership is proven through a secure token.

## 🔍 How
- **Model Layer**: Modified `api/models/TCaver.js` to include a `pendingMail` field.

```
BEGIN;
-- 1. Add the pending_mail column to store unverified email addresses
ALTER TABLE public.t_caver ADD COLUMN pending_mail varchar(50);
-- 2. Add an index to optimize uniqueness checks across both mail and pending_mail
CREATE INDEX t_caver_pending_mail_idx ON public.t_caver (pending_mail);
COMMIT;
```

- **Workflow**:
    - When a user requests an email change, the new address is stored in `pendingMail`.
    - `mailIsValid` is set to `false`, and an `activationCode` is generated and sent via email.
    - The user continues to log in using their original `mail`.
    - Upon clicking the verification link, the `verify-email` controller promotes `pendingMail` to `mail` and clears the pending state.
- **Conflict Detection**: Updated queries to check both `mail` and `pendingMail` fields across all users, ensuring that an email address cannot be claimed even as a pending change by another user.
- **Identity Retention**: The system now explicitly checks if the new email is identical to the current one, and if so, interprets it as a request to cancel any existing pending changes.

## 🧪 Testing
Comprehensive integration tests have been added in `test/integration/4_routes/Account/change-email.test.js` covering:
- **Success Paths**: Email change initiation and subsequent verification.
- **Conflict Handling**: Verification that neither primary nor pending emails can be duplicated.
- **Login Persistence**: Ensuring login works with the old email while a change is pending.
- **Cancellation**: Verifying that submittting the current email clears the pending state.
- **Edge Cases**: Handling scenarios where a pending email is taken by another user before verification is completed.


ClemRz (changes requested):
### Issues (Must Fix)

- **[sql/9_03_2026_05_08_add_caver_pending_mail.sql:1]** The migration file is missing the required `\c grottoce;` prefix (needed for Docker init compatibility) and lacks `IF NOT EXISTS` for idempotency. Additionally, the PR body describes an index (`CREATE INDEX t_caver_pending_mail_idx ON public.t_caver (pending_mail)`) that is absent from the actual migration file. Since the `change-email` controller queries `pendingMail` for conflict detection, this index is important for performance.
  ```suggestion
  \c grottoce;

  ALTER TABLE t_caver ADD COLUMN IF NOT EXISTS pending_mail VARCHAR(50) DEFAULT NULL;
  CREATE INDEX IF NOT EXISTS t_caver_pending_mail_idx ON t_caver (pending_mail);
  ```

- **[test/customSQL.js]** Per project conventions, any index added in `sql/` must also be added to the `INDEX_OPTIMIZATION_MIGRATION` block in `test/customSQL.js` so the test database has the same indexes as production. The `t_caver_pending_mail_idx` index is missing here.
  ```suggestion
  CREATE INDEX IF NOT EXISTS t_caver_pending_mail_idx ON t_caver (pending_mail);
  ```

- **[api/controllers/v1/auth/verify-email.js:22-25]** The conflict check at verification time only looks at `mail` of other users but doesn't exclude the current user and doesn't check other users' `pendingMail`. This is inconsistent with the more thorough check in `change-email.js`. If another user has the same email as their `pendingMail`, the verification would succeed and create a state where two users have the same email (one as `mail`, one as `pendingMail`). Consider aligning with the pattern in `change-email.js`:
  ```suggestion
  const alreadyInUse = await TCaver.findOne({
    id: { '!=': caver.id },
    or: [{ mail: caver.pendingMail }, { pendingMail: caver.pendingMail }],
  });
  ```

- **[api/controllers/v1/auth/verify-email.js:44-48]** There's a TOCTOU (time-of-check-time-of-use) race condition between the conflict check and the update. If two users verify at the exact same time with the same target email, both could pass the check and one update would violate the DB unique constraint on `mail`. Consider wrapping the check + update in error handling that catches the unique constraint violation from PostgreSQL (error code `23505`) and returns `res.conflict()` gracefully, rather than letting it bubble up as a 500.

- **[api/controllers/v1/account/change-email.js:31-34]** Same TOCTOU concern here — between the `findOne` conflict check and the `updateOne`, another user could claim the same email. The DB unique constraint on `mail` protects the final commit in `verify-email`, but there's no unique constraint on `pending_mail`. Two users could end up with the same `pendingMail` value if requests arrive concurrently. Consider adding a unique constraint on `pending_mail` (with a partial index excluding NULLs) or accepting this as a low-risk edge case since only one can ultimately verify.

- **[api/controllers/v1/account/change-email.js:3-5]** The controller uses `req.param('email')` which reads from both query string and body. The Swagger spec documents it as `in: query`, but the tests send it in the request body. Consider documenting it as a `requestBody` in the Swagger spec to match actual usage, or at minimum noting that both are accepted.

- **[test/integration/4_routes/Account/change-email.test.js:155]** The test assumes the fixture password is `'testtest'`. If this changes, the test will break silently. Consider importing the password from a shared constant or adding a comment referencing where this value is defined in fixtures.

### Nitpicks (Optional)

- **[api/controllers/v1/account/change-email.js:67]** The final success path returns `res.ok()` (which yields 204 since no data is passed). Consider returning a message like `res.ok({ message: 'Verification email sent.' })` for consistency with the cancellation path that returns a message body. This would also make the Swagger 200 response more accurate.

- **[assets/swaggerV1.yaml:3148-3149]** There are two consecutive blank lines after the `'409'` response. Minor formatting issue.

ClemRz (changes requested):
### Issues (Must Fix)

- **[sql/9_03_2026_05_08_add_caver_pending_mail.sql:1]** The migration file is missing the required `\c grottoce;` prefix (needed for Docker init compatibility) and lacks `IF NOT EXISTS` for idempotency. The PR body also describes an index (`CREATE INDEX t_caver_pending_mail_idx`) that is absent from the actual migration file. Since the `update.js` controller queries `pendingMail` for conflict detection, this index matters for performance.
  ```suggestion
  \c grottoce;

  ALTER TABLE t_caver ADD COLUMN IF NOT EXISTS pending_mail VARCHAR(50) DEFAULT NULL;
  CREATE INDEX IF NOT EXISTS t_caver_pending_mail_idx ON t_caver (pending_mail);
  ```

- **[test/customSQL.js]** Per project conventions, any index added in `sql/` must also be added to the `INDEX_OPTIMIZATION_MIGRATION` block in `test/customSQL.js` so the test database has the same indexes. The `t_caver_pending_mail_idx` index is missing here.

- **[api/controllers/v1/auth/verify-email.js:23-26]** The conflict check at verification time only looks at `{ mail: caver.pendingMail }` without excluding the current user and without checking other users' `pendingMail`. This is inconsistent with the more thorough check in `update.js`. Two specific problems:
  1. If the user's own `mail` happens to equal their `pendingMail` (shouldn't happen normally, but defensive coding), the query would find the user themselves and incorrectly return 409.
  2. If another user has the same value as their `pendingMail`, the verification would succeed, creating a state where two users have the same email (one as `mail`, one as `pendingMail`).

  Consider aligning with the pattern in `update.js`:
  ```suggestion
  const alreadyInUse = await TCaver.findOne({
    id: { '!=': caver.id },
    or: [{ mail: caver.pendingMail }, { pendingMail: caver.pendingMail }],
  });
  ```

- **[api/controllers/v1/account/update.js:87-88]** The controller fetches `caver` a second time at line 87 (`const caver = await TCaver.findOne({ id: req.token.id })`) even though `currentCaver` was already fetched at line 43 for the email flow. When the email field is present, this results in two identical DB queries. Consider reusing `currentCaver` or restructuring so the fetch only happens once. More importantly, if the email field is NOT present in the request body, `currentCaver` is never fetched, but `caver` is — so the two fetches serve different code paths. This is fine functionally but the naming (`currentCaver` vs `caver`) is confusing since they hold the same data.

- **[api/controllers/v1/auth/verify-email.js:53-58]** The `notifyEmailChanged` call is fire-and-forget but has no `.catch()` handler. If `notifyEmailChanged` rejects (despite the service's documentation saying it never throws), the unhandled rejection would crash the process in Node.js. Consider adding `.catch()` for safety:
  ```suggestion
  if (caver.pendingMail) {
    AccountNotificationService.notifyEmailChanged({
      oldEmail: caver.mail,
      nickname: caver.nickname,
      languageId: caver.language,
    }).catch((err) => {
      sails.log.error('Failed to send email change notification:', err);
    });
    return res.ok({ message: 'Email successfully changed.' });
  }
  ```

- **[test/integration/4_routes/Account/update-notifications.test.js:33-39]** The `afterEach` restores `mail` and `password` but doesn't restore `pendingMail`, `activationCode`, or `mailIsValid`. Since the tests now set these fields (via the PATCH + verify-email flow), stale state could leak between tests if a test fails mid-way. Consider:
  ```suggestion
  afterEach(async () => {
    notifyEmailChangedStub.restore();
    notifyPasswordChangedStub.restore();
    await TCaver.updateOne({ id: 3 }).set({
      mail: originalEmail,
      password: originalPasswordHash,
      pendingMail: null,
      activationCode: null,
      mailIsValid: true,
    });
  });
  ```

- **[assets/swaggerV1.yaml]** The `PATCH /account` endpoint now returns `200` with a message body when cancelling a pending email change, but the Swagger spec only documents `204` as the success response. Consider adding a `200` response to document this new behavior, or changing the cancellation response to `204` for consistency.

- **[api/controllers/v1/auth/verify-email.js:44-48]** There's a TOCTOU (time-of-check-time-of-use) race condition between the conflict check and the update. If two users verify at the exact same time targeting the same email, both could pass the check. The DB unique constraint on `mail` protects the final commit, but the resulting error would bubble up as a 500 rather than a clean 409. Consider wrapping the update in error handling that catches PostgreSQL unique constraint violations (error code `23505`) and returns `res.conflict()` gracefully.

### Nitpicks (Optional)

- **[assets/swaggerV1.yaml:3244-3245]** There are two consecutive blank lines after the `'409'` response definition. Minor formatting issue.

- **[test/integration/4_routes/Account/change-email.test.js:155]** The test hardcodes `'testtest'` as the fixture password. Consider using `AuthTokenService.TEST_PASSWORD` (which is already imported and used in `update-notifications.test.js`) for consistency and resilience to fixture changes.

ClemRz (dismissed):

- **[sql/9_03_2026_05_08_add_caver_pending_mail.sql:4]** The file is missing a trailing newline. Most editors and linters expect files to end with a newline character (the diff shows `\ No newline at end of file`). This is a minor issue but can cause noise in future diffs.

- **[api/controllers/v1/account/update.js:195-199]** The `notifyPasswordChanged` call is fire-and-forget without a `.catch()` handler. While the service documentation says it never throws (it catches internally), the `sendVerificationEmail` call a few lines above correctly adds `.catch()` for safety. Consider adding the same pattern here for consistency and defense-in-depth:
  ```suggestion
  AccountNotificationService.notifyPasswordChanged({
    email: caver.mail,
    nickname: caver.nickname,
    languageId: caver.language,
  }).catch((err) => {
    sails.log.error('Failed to send password change notification:', err);
  });
  ```

- **[assets/swaggerV1.yaml:3415-3416]** The `200` response description for `/verify-email` says "Account successfully verified or Account was already verified" but the endpoint now also returns `200` with `{ message: 'Email successfully changed.' }` when a pending email change is committed. Consider updating the description to mention this third outcome, e.g.: "Account successfully verified, email change committed, or account was already verified."

- **[assets/swaggerV1.yaml:3429-3430]** There's a blank line between the `'409'` and `'500'` response entries. While valid YAML, it's inconsistent with the rest of the spec where responses are listed without gaps. This was flagged in the previous review.

### Nitpicks (Optional)

- **[assets/swaggerV1.yaml:3256-3258]** Two consecutive blank lines between the `/names/{id}` path and `/account/password` path. This was also flagged in the previous review — minor formatting inconsistency.

- **[test/integration/4_routes/Account/change-email.test.js:5]** `Fixted` is imported and instantiated (lines 7-8) but `fixtures` is only used once (line 55) to find another user's email. Since the test already queries `TCaver` directly elsewhere, consider using `await TCaver.find({ mail: { '!=': userMail } }).limit(1)` instead and removing the `Fixted` dependency from this file. This would make the test more resilient to fixture changes.

ClemRz (dismissed):

- **[assets/swaggerV1.yaml:3430]** The `500` response description for `/verify-email` still says "An error occurred during account activation" but the controller now returns "An error occurred during email verification." Consider updating the Swagger description to match the actual error message.

- **[api/controllers/v1/account/update.js:40-42]** The email regex `^[^\s@]+@[^\s@]+\.[^\s@]+$` is more permissive than the Waterline `isEmail` validation on the model (which uses a stricter validator). This means a request could pass the controller check but fail at the ORM layer with a `UsageError`. This isn't a bug (the `UsageError` is caught and returns 400), but it could produce a less helpful error message ("Invalid data provided." vs "You must provide a valid email address."). Consider using a stricter regex or relying solely on the model validation and catching the `UsageError` with a specific message.

- **[api/controllers/v1/account/update.js:73]** The `mailIsValid = false` update is set unconditionally when an email change is initiated. If the user's `mailIsValid` was already `false` (e.g., from a previous pending change that was never verified), this is harmless but redundant. More importantly, if the user has other pending operations that depend on `mailIsValid` being `true` (e.g., password reset), setting it to `false` here could block those flows until verification completes. This is likely intentional (security-wise it makes sense), but worth documenting in a code comment since it's a deliberate design choice.

### Nitpicks (Optional)

- **[test/integration/4_routes/Account/change-email.test.js:4]** `TokenService` is imported from `../../../../api/services/TokenService` and used only to decode the JWT in the `before()` hook. Since `AuthTokenService` already provides the user ID in other test files (or you could use `jwt.decode()` which doesn't verify), this deep import into `api/services/` creates a tighter coupling between tests and internals. Not blocking, just a note for future refactoring.

- **[test/integration/0_models/TCaver.test.js:28]** The model test creates a caver without a `password` field. While `password` is `allowNull: true` in the model, production code always sets a password on signup. Consider adding `password: 'hashed'` to make the test fixture more realistic, though this is purely cosmetic.

ClemRz (approved):

- **[api/controllers/v1/account/update.js:40]** The email regex `^[^\s@]+@[^\s@]+\.[^\s@]+$` is more permissive than Waterline's `isEmail` validator on the model. For example, `foo@bar..baz` passes this regex but fails the model validation, resulting in a less helpful `UsageError` ("Invalid data provided.") instead of the controller's friendlier message. This isn't a bug (the error is still caught), but consider tightening the regex or adding a comment explaining the intentional layered validation.

- **[api/controllers/v1/auth/verify-email.js:62-71]** The `if (caver.pendingMail)` check after the `try/catch` block relies on the pre-update snapshot of `caver` (fetched at line 10). This works correctly today, but it's subtle — a future refactor that re-fetches `caver` after the update would break the notification logic silently. Consider adding a brief comment like `// caver still holds pre-update state; pendingMail indicates this was an email-change verification` to make the intent explicit for future readers.

- **[api/controllers/v1/account/update.js:73]** Setting `mailIsValid = false` when an email change is initiated means the user's account is in a "mail not valid" state until they verify. If other parts of the system gate functionality on `mailIsValid` (e.g., password reset, notifications), this could have unintended side effects. The design seems intentional (security-wise it makes sense to mark the account as having an unverified email state), but a brief code comment explaining this deliberate choice would help future maintainers.

### Nitpicks (Optional)

- **[test/integration/0_models/TCaver.test.js:32]** The model test creates a caver without a `password` field. While `password` is `allowNull: true` in the model, production code always sets a password on signup. Adding `password: 'hashed'` would make the test fixture slightly more realistic, though this is purely cosmetic.

- **[test/integration/4_routes/Account/change-email.test.js:4]** `TokenService` is imported via `require('../../../../api/services/TokenService')` through `AuthTokenService` which already provides the user ID. The deep import into `api/services/` creates tighter coupling between tests and internals. Not blocking — just a note for future refactoring if the test helper API evolves.

PR #1576, opened 2026-05-12
feat(account): centralize auth helpers and enforce strong password policy

ClemRz:
## 🤔 What

Addresses the review action items from PR #1565 and enforces a strong password validation policy across all password-setting endpoints.

- Cache `LanguageService.getLocale` to avoid repeated DB queries
- Add explicit email uniqueness pre-check in the account update endpoint
- Require current password verification before allowing a password change
- Centralize password validation in `AuthService.validatePassword()` with strict strength requirements
- Enforce the new policy on sign-up, change-password, and account update

## 🤷‍♂️ Why

1. **Cache**: Languages never change at runtime — querying `TLanguage.findOne()` on every notification email is wasteful.
2. **Email uniqueness**: The nickname field had an explicit pre-check with a specific error message, but email relied on the generic `E_UNIQUE` catch block. Now both return clear 409 messages.
3. **Current password**: Changing a password without verifying the old one is a security gap. Now the endpoint requires `currentPassword` and returns 403 on mismatch.
4. **Password policy**: The previous minimum was 8 characters with no complexity requirements. The new policy enforces:
   - At least 12 characters
   - At least one uppercase letter
   - At least one lowercase letter
   - At least one digit
   - At least one special character (`!@#$%^&*()_+-=[]{}|;:'",.<>?/~``)

## 🔍 How

- `AuthService.js`: Extracted `verifyPassword(hash, plain)` and `validatePassword(password)` helpers. The validation function returns `{ valid, message }` so controllers can relay specific feedback to the user. Also added `generateActivationCode()` and `sendVerificationEmail()` to centralize auth concerns.
- `account/update.js`: Added email uniqueness pre-check (409). Added `currentPassword` verification before password change (400 if missing, 403 if wrong). Replaced inline length check with `AuthService.validatePassword()`.
- `account/change-password.js`: Replaced inline length check with `AuthService.validatePassword()`.
- `auth/sign-up.js`: Replaced inline length check with `AuthService.validatePassword()`.
- `LanguageService.js`: Added an in-memory `Map` cache for locale lookups.
- `swaggerV1.yaml`: Added `currentPassword` field and `403` response to the PATCH `/account` endpoint docs.

## 🧪 Testing

All 2209 tests pass. New/updated tests:

- `test/integration/1_services/LanguageService.test.js` — cache behavior
- `test/integration/4_routes/Account.test.js` — email conflict (409), missing currentPassword (400), wrong currentPassword (403), password lacking special char (400), successful update
- `test/integration/4_routes/Account/change-password.test.js` — added cases for missing uppercase and missing special character
- Updated all test passwords across sign-up, change-password, and ban/unban tests to comply with the new policy

## 📸 Previews

N/A — backend-only changes.

### DIFF
--- a/api/controllers/v1/account/change-password.js
+++ b/api/controllers/v1/account/change-password.js
@@ -3,8 +3,6 @@ const jwt = require('jsonwebtoken');
 const AuthService = require('../../../services/AuthService');
 const TokenService = require('../../../services/TokenService');
 
-const PASSWORD_MIN_LENGTH = 8;
-
 // eslint-disable-next-line consistent-return
 module.exports = async (req, res) => {
   const password = req.param('password');
@@ -12,10 +10,9 @@ module.exports = async (req, res) => {
   if (!password) {
     return res.badRequest('You must provide a password.');
   }
-  if (password.length < PASSWORD_MIN_LENGTH) {
-    return res.badRequest(
-      `Your password must be at least ${PASSWORD_MIN_LENGTH} characters long.`
-    );
+  const validation = AuthService.validatePassword(password);
+  if (!validation.valid) {
+    return res.badRequest(validation.message);
   }
 
   if (req.token) {
--- a/api/controllers/v1/account/update.js
+++ b/api/controllers/v1/account/update.js
@@ -1,9 +1,11 @@
 const AuthService = require('../../../services/AuthService');
 const CaverService = require('../../../services/CaverService');
 
-const PASSWORD_MIN_LENGTH = 8;
-
+// Fields that can be updated via this endpoint. `currentPassword` is handled
+// as an out-of-band credential (required when changing password) and is not
+// persisted — it is listed here only so it passes the unknown-property guard.
 const UPDATABLE_PROPERTIES = [
+  'currentPassword',
   'email',
   'language',
   'name',
@@ -35,6 +37,10 @@ module.exports = async (req, res) => {
     if (!/^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(normalizedEmail)) {
       return res.badRequest('You must provide a valid email address.');
     }
+    const existingEmail = await TCaver.findOne({ mail: normalizedEmail });
+    if (existingEmail && existingEmail.id !== req.token.id) {
+      return res.conflict('This email is already used.');
+    }
     updates.mail = normalizedEmail;
   }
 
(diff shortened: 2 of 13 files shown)


Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[api/controllers/v1/account/update.js:39-43]** Email change still does not reset `mailIsValid`. When `email` is updated, `mailIsValid` should be set to `false` and a new verification email should be sent (via `AuthService.sendVerificationEmail`). This was flagged as a **blocker** in the PR #1565 review and is not addressed here. A user can silently replace their email address with an unverified one while retaining full verified status, which is a trust/identity integrity issue.

   Minimum fix:
   ```js
   updates.mail = normalizedEmail;
   updates.mailIsValid = false;
   // After updateOne succeeds, call AuthService.sendVerificationEmail(updatedCaver, token, locale)
   ```

2. **[api/controllers/v1/account/update.js:6-15]** `currentPassword` is listed in `UPDATABLE_PROPERTIES`, so a request body like `{ currentPassword: 'secret', language: 'fra' }` passes the unknown-property guard and silently discards `currentPassword`. This is harmless but surprising. Consider stripping `currentPassword` from `UPDATABLE_PROPERTIES` and handling it as an out-of-band credential rather than an updatable field, or document in the code why it is included.

3. **[test/integration/1_services/LanguageService.test.js:26-33]** The "should return cached value on subsequent calls" test only checks that two sequential calls return equal values — it does not verify that the DB was not queried twice. Since this is an integration test without mocking, a spy on `TLanguage.findOne` (if the framework permits) would make the test a true cache-hit assertion. As written, the test would pass even if the cache were absent.

### Nitpicks (Optional)

4. **[api/services/LanguageService.js:22]** `lang?.part1 || undefined` — the `|| undefined` is intentional (guards against empty string) but its purpose is non-obvious. A short inline comment or replacing with an explicit ternary (`lang?.part1 ? lang.part1 : undefined`) would clarify the intent for future readers.

ClemRz:
The must fix item will be handled by PR #1575

ClemRz:
@Paul-AUB I added the enforcement of the new password policy.

Paul-AUB (changes requested):
Good iteration overall — suggestions #2, #3, and nitpick #4 from the previous review have all been addressed cleanly (comment on `currentPassword` in `UPDATABLE_PROPERTIES`, sinon-based cache assertion, and the `|| undefined` guard comment). Password policy, `currentPassword` verification, and email uniqueness pre-check are well implemented and tested.

### Issues (Must Fix)

Will be done in another PR, ignore. All good here.

2. **[api/controllers/v1/account/update.js:51-57]** The new password strength validation (`AuthService.validatePassword`) runs before the `currentPassword` presence check. This allows an authenticated user to probe the policy rules (length, uppercase, special char, etc.) without proving they know their current password. Since a valid JWT is required, the risk is limited, but moving the `currentPassword` check (lines 55–58) to before the `validatePassword` call would confirm identity before revealing policy details.

Paul-AUB (dismissed):

Paul-AUB (approved):

PR #1578, opened 2026-05-13
feat(entrance): expose data quality score in search and geoloc endpoints

### LINKED ISSUE #1577: Expose entrance data quality in advanced search and geoloc endpoints
# Expose entrance data quality in advanced search and geoloc endpoints

## Context

GrottoCenter computes a data quality score (0–100) for each entrance, reflecting how complete and up-to-date its content is across categories such as location, description, documents, rigging, history, and comments. This score is already used in dedicated data-quality endpoints (per country, massif, region).

The frontend needs to surface this score in two additional contexts:

---

## 1. Advanced search — entrances

**Current state**: The `POST /api/v1/advanced-search` endpoint (with `entity=entrances`) does not return the data quality score for each entrance result, and does not support filtering or sorting by it.

**Need**:
- Return the data quality score as a field in each entrance result.
- Allow the client to filter results by data quality with a minimum threshold (`quality >= value`) and/or a maximum threshold (`quality <= value`).
- Allow the client to sort results by data quality (ascending/descending), alongside existing sort options.

---

## 2. GET /geoloc/entrances

**Current state**: The `GET /api/v1/geoloc/entrances` endpoint returns a lightweight entrance payload for map rendering. It does not include the data quality score.

**Need**:
- Return the data quality score as a field in each entrance object of the response.

---

## Why this matters

These two surfaces are the main places where users browse and discover entrances. Exposing data quality there allows:
- users to assess how reliable an entry is at a glance,
- filtering or sorting to prioritise well-documented entrances,
- motivating contributors by making quality visible in discovery flows.

ClemRz:
## 🤔 What

Expose the entrance data quality score (0–100) in two high-traffic API surfaces:
- **Advanced search** (`POST /api/v1/advanced-search` with `entity=entrances`)
- **Geoloc map** (`GET /api/v1/geoloc/entrances`)

This allows the frontend to display quality indicators, and users to filter/sort entrances by data completeness.

Closes #1577 

## 🤷‍♂️ Why

The data quality score was already computed via the `v_data_quality_compute_entrance` materialized view but only available through dedicated quality endpoints. Making it visible in discovery flows (search results, map markers) helps users assess entrance reliability at a glance and motivates contributors by surfacing quality gaps.

## 🔍 How

- **Typesense schema**: Added `dataQuality` (int32, sortable, optional) to the entrances collection
- **dbSync**: Batch-joins quality data from the materialized view in `processRows`, computes score via `getQualityData`
- **EntranceService.updateInSearch**: Queries the materialized view on individual document refresh
- **Converter**: Passes through the integer score in `toEntrance` for search results
- **Filter normalization**: New `normalizeDataQualityFilter` utility handles `[min, max]`, null bounds, clamping to 0–100, and non-numeric removal
- **Sort validation**: Entity-specific validation returns 400 when `dataQuality` sort is used on non-entrance entities
- **Geoloc SQL**: `LEFT JOIN LATERAL` on the materialized view (limit 1, ordered by `id_massif ASC`) with a new composite index `(id_entrance, id_massif)`
- **Swagger**: Documented `dataQuality` field, filter, sort options, and added missing description for the `quality` (size_coef) field

## 🧪 Testing

- Property-based tests for `normalizeDataQualityFilter` (7 properties, 100 runs each)
- Route tests for filter normalization, sort validation, and response shape
- Service tests for `EntranceService.updateInSearch` and `GeoLocService.getEntrancesMap`
- Unit tests for dbSync `processRows` quality join logic

Run: `npm run test`

## 📸 Previews

```bash
# Geoloc endpoint — dataQuality field present
curl 'http://localhost:1337/api/v1/geoloc/entrances?sw_lat=40&sw_lng=0&ne_lat=50&ne_lng=10'
# → { "id": 14, "name": "Gournier", ..., "quality": 10, "dataQuality": 71 }

# Advanced search — sort by quality desc
curl -X POST http://localhost:1337/api/v1/advanced-search \
  -d '{"query":"*","entity":"entrances","sort":"dataQuality:desc","size":3}'
# → Leicasse (80), Souffleur (80), Garrel (76)

# Filter by quality range
curl -X POST http://localhost:1337/api/v1/advanced-search \
  -d '{"query":"*","entity":"entrances","filter":{"dataQuality":[40,100]}}'

# Sort validation — 400 on non-entrance entity
curl -X POST http://localhost:1337/api/v1/advanced-search \
  -d '{"query":"*","entity":"massifs","sort":"dataQuality:desc"}'
# → 400: "The dataQuality sort field is only valid for entity=entrances."
```

### DIFF
--- a/api/controllers/v1/search/advanced-search.js
+++ b/api/controllers/v1/search/advanced-search.js
@@ -4,20 +4,42 @@ const {
   handleTypesenseError,
 } = require('../../../services/TypesenseErrorService');
 const { toSearchResult } = require('../../../services/mapping/converters');
+const normalizeDataQualityFilter = require('../../../utils/normalizeDataQualityFilter');
+
+// Sort fields that are only valid for specific entities
+const ENTITY_SPECIFIC_SORT_FIELDS = {
+  dataQuality: ['entrances'],
+};
 
 module.exports = async (req, res) => {
   let matchAllFields = req.param('matchAllFields') ?? true;
   if (!matchAllFields || matchAllFields === 'false') matchAllFields = false;
 
   const entity = req.param('entity') ?? '';
+  const sort = req.param('sort');
+
+  // Validate entity-specific sort fields
+  if (sort) {
+    const sortFields = sort.split(',').map((s) => s.trim().split(':')[0]);
+    for (const sortField of sortFields) {
+      const allowedEntities = ENTITY_SPECIFIC_SORT_FIELDS[sortField];
+      if (allowedEntities && !allowedEntities.includes(entity)) {
+        return res.badRequest(
+          `The ${sortField} sort field is only valid for entity=${allowedEntities.join(', ')}.`
+        );
+      }
+    }
+  }
+
+  const filter = normalizeDataQualityFilter(req.param('filter') ?? {});
 
   let results;
   try {
     results = await SearchService.collectionSearch({
       query: req.param('query'),
       entity,
-      sort: req.param('sort'),
-      filter: req.param('filter') ?? {},
+      sort,
+      filter,
       isLogicalCompareAnd: !!matchAllFields,
       page: req.param('page') ?? 1,
       size: req.param('size') ?? 10,
--- a/api/dbSync/entities/entrance.js
+++ b/api/dbSync/entities/entrance.js
@@ -3,6 +3,8 @@ const {
   NON_INDEXED_BOOLEAN_FIELDS,
   computeDateLastModif,
 } = require('../../../config/constants/entrance');
+const { getQualityData } = require('../../utils/computeEntranceDataQuality');
+const CommonService = require('../../services/CommonService');
 
 function average(arr) {
   if (arr.length === 0) return null;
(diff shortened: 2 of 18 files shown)


Paul-AUB (dismissed):
Well-structured PR with thorough test coverage and a clean requirements/design document. The separation of concerns across the dbSync, service, and controller layers is well done. A few things worth addressing below.

1. **[api/dbSync/entities/entrance.js ~line 144]** The `joinMany` call for quality data has no ORDER BY, so for entrances belonging to multiple massifs `row.qualityData[0]` is non-deterministic. Both `getEntrancesMap` and `updateInSearch` consistently pick `ORDER BY id_massif ASC LIMIT 1`, but the dbSync full-resync path may index a different score for those entrances. Consider sorting `row.qualityData` by `id_massif` before using `[0]`, or using a `DISTINCT ON (id_entrance)` subquery in the main dbSync SQL rather than relying on `joinMany`.

2. **[api/controllers/v1/search/advanced-search.js ~line 20]** The entity-specific sort validation only inspects the first sort field via `sort.split(':')[0]`. A multi-field sort string like `_text_match:desc,dataQuality:asc` would place `dataQuality` past the first `:`, silently bypassing the 400 guard on non-entrance entities. Consider splitting on `,` first and checking each `field:direction` pair.

### Nitpicks (Optional)

3. **[api/services/mapping/converters.js ~line 411]** The `else` branch was changed from `result.dataQuality = null` to `result.dataQuality = source.dataQuality ?? null`. In practice this is harmless, but it quietly passes through any non-null, non-number value (e.g. an unexpected object) rather than normalising it to `null` as before. If the intent is "integer from Typesense or null", the `else if` + `else` logic already covers that — the `?? null` fallback in the final branch is redundant.

4. **[assets/swaggerV1.yaml ~line 5993]** Placing `nullable: true` at the property level alongside `oneOf` is valid in OpenAPI 3.0 per spec, but many validators and code generators (e.g. openapi-generator, redoc strict mode) do not handle it correctly and will reject `null` as matching none of the `oneOf` variants. Consider wrapping the `DataQuality` `$ref` in an inline schema with `nullable: true`, or adding `nullable: true` to the integer variant as well.

Paul-AUB (approved):

2. **[api/services/mapping/converters.js:374–375]** When `source.massifs` is a non-empty array from Typesense, it bypasses `toSimpleMassif` entirely. The indexed shape is `{ id, name }`, while `toSimpleMassif` produces `{ id, name, language, isDeleted }`. If any search-result consumer reads `massif.language` or `massif.isDeleted`, those fields will be `undefined`. The dbSync massif query already selects the main name; adding the language of that name row and hard-coding `isDeleted: false` (deleted massifs are already filtered) would close the gap without extra queries.

3. **[api/services/EntranceService.js:359–387]** The quality query and the massif query are sequential `await` calls but are fully independent. Wrapping them in `Promise.all` would halve the round-trip latency for every `updateInSearch` 
invocation.

**==> Seems particularly relevant**

### Nitpicks (Optional)

4. **[test/integration/1_services/EntranceService.test.js:431–452]** The second test stubs `CommonService.query` to return quality-row data for _all_ calls. The massif query (`m.id`, `n.name`) therefore receives quality-row columns and produces `entrance.massifs = [{ id: undefined, name: undefined }]`. The test only asserts `dataQuality`, so it passes, but `updateDocument` is handed garbage massifs. Using `stub.onFirstCall().resolves(qualityData).onSecondCall().resolves({ rows: [] })` (or vice-versa) would make the test self-documenting and catch future ordering bugs.

Paul-AUB (changes requested):
The Promise.all to gain timing seems relevant

ClemRz:

**Fixed (from your CHANGES_REQUESTED review):**

1. **Promise.all for parallel queries** — Wrapped the independent quality and massif queries in `EntranceService.updateInSearch` with `Promise.all` to halve round-trip latency.

2. **Massifs shape consistency** — Added `language` and `isDeleted: false` to the massif objects produced by both `dbSync/entities/entrance.js` and `EntranceService.updateInSearch`, so the indexed shape matches what `toSimpleMassif` produces. Updated the Typesense schema accordingly.

3. **Test stub ordering** — Replaced the single `CommonService.query` stub with per-query stubs using `sinon.match(/v_data_quality_compute_entrance/)` and `sinon.match(/t_massif/)` so each query gets the correct mock response. Added assertions for `massifs` in both tests.

All tests pass (2247 passing), lint clean.

Paul-AUB (approved):
LGTM

PR #1582, opened 2026-05-15
fix(auth): harden password handling and add security headers

### LINKED ISSUE #1579: Security: Harden Password Handling on API Side
## Summary

**Type:** Security / Enhancement
**Severity:** Medium
**Related front-end issue:** [GrottoCenter/grottocenter-front#1317](https://github.com/GrottoCenter/grottocenter-front/issues/1317)

The front-end issue [GrottoCenter/grottocenter-front#1317](https://github.com/GrottoCenter/grottocenter-front/issues/1317) identifies that passwords are transmitted as plaintext JSON to the API. While the API already redacts sensitive fields in error logs (via `api/utils/sanitize.js`), there are additional hardening measures the API should implement to reduce the attack surface.

---

### What's already in place ✅

- **Body redaction in error logs:** The `responseTimeLogger` middleware in `config/http.js` uses `sanitize(req.body)` which replaces `password`, `token`, `authorization`, `secret`, `apikey`, and `api_key` fields with `[REDACTED]` before logging on 500 errors.
- **Argon2 hashing:** Passwords are hashed with Argon2 before storage.
- **HTTPS in production:** The production deployment enforces TLS.

### What's missing ❌

1. **No HSTS headers** — The API does not send `Strict-Transport-Security` headers to prevent TLS downgrade attacks.
2. **No audit of all logging paths** — While the `responseTimeLogger` redacts bodies, other code paths (e.g., `sails.log.error` in controllers/services catching exceptions) may inadvertently log request data containing passwords.
3. **No request body size limit on auth endpoints** — Auth endpoints don't enforce a maximum body size, which could be exploited for DoS via oversized payloads.
4. **No rate limiting specific to auth endpoints** — Verify that login/signup/password-change endpoints have appropriate rate limiting to mitigate brute-force attacks (partially mitigated if already configured in `config/rateLimit/`).

---

## Affected Endpoints

| Method | Route | Controller |
|--------|-------|-----------|
| POST | `/api/v1/login` | `v1/auth/login` |
| POST | `/api/v1/signup` | `v1/auth/sign-up` |
| PATCH | `/api/v1/account` | `v1/account/update` |
| PATCH | `/api/v1/account/password` | `v1/account/change-password` |
| POST | `/api/v1/forgotPassword` | `v1/account/forgot-password` |

---

### 1. Add HSTS header

Add `Strict-Transport-Security: max-age=31536000; includeSubDomains` to all responses in production. This can be done via Express middleware in `config/http.js`.

### 2. Audit all logging paths for credential leakage

Review all `sails.log.*` calls in auth-related controllers and services to ensure no code path logs `req.body` or password-containing objects without going through `sanitize()`.

### 3. Enforce request body size limits on auth endpoints

Configure a reasonable body size limit (e.g., 1 KB) for `/login`, `/signup`, and `/account/password` to prevent abuse.

### 4. Verify rate limiting on auth endpoints

Confirm that the rate limiting configuration in `config/rateLimit/` covers all auth endpoints with appropriate thresholds (e.g., 5 login attempts per minute per IP).

---

## Medium-term: Client-side pre-hashing

The medium-term work to support client-side password pre-hashing (coordinated migration with the front-end) has been split into a dedicated issue: **#1581**

---

## Acceptance Criteria

- [ ] HSTS header is sent on all production responses
- [ ] No auth-related controller or service logs password-containing objects without redaction
- [ ] Auth endpoints have explicit body size limits
- [ ] Rate limiting is confirmed active on all auth endpoints

ClemRz:
## 🤔 What

Security hardening for password-handling endpoints and general API responses.

- Add HSTS, X-Content-Type-Options, X-Frame-Options security headers to all responses
- Add auth-specific rate limiting (10 requests per 15-minute window per IP) on login, signup, forgot-password, and change-password endpoints
- Add 1 KB body size limit on auth endpoints to prevent DoS via oversized payloads
- Reduce verbosity of the authenticated-user log line (only id, nickname, group names — no iat/exp/sub noise)
- Closes #1579
- Related front-end issue: GrottoCenter/grottocenter-front#1317

## 🤷‍♂️ Why

The front-end issue identified that passwords transit as plaintext JSON. While the API already redacts credentials in error logs (via `sanitize()`), additional defense-in-depth measures were missing: no HSTS to prevent TLS downgrade, no brute-force protection specific to auth endpoints, and no payload size guard against abuse.

## 🔍 How

- **Security headers** (`config/http.js`): New `securityHeaders` middleware added early in the middleware order. Sets `Strict-Transport-Security: max-age=31536000; includeSubDomains`, `X-Content-Type-Options: nosniff`, and `X-Frame-Options: DENY`.
- **Auth rate limiter** (`config/rateLimit/rateLimiter.js` + `config/http.js`): New `authRateLimit` instance (express-rate-limit, 15-min window, 10 max). A wrapper middleware in `http.js` only applies it to auth paths. Skipped in test/dev environments. Configurable via `AUTH_RATE_LIMIT` env var.
- **Auth body limit** (`config/http.js`): `authBodyLimit` middleware checks `Content-Length` against 1 KB for auth paths and returns 413 if exceeded. Runs before the body parser so oversized payloads are rejected without buffering.
- **Log cleanup** (`config/http.js`): `parseAuthToken` now logs a compact JSON with only `id`, `nickname`, and group names instead of the full decoded token object.

## 🧪 Testing

```bash
npm run test -- --grep "Security headers|Auth body size"
```

Full suite passes (2218 tests, 0 failures).

## 📸 Previews

N/A — backend-only change, no UI impact.

### DIFF
--- a/config/http.js
+++ b/config/http.js
@@ -17,6 +17,25 @@ const TokenService = require('../api/services/TokenService');
 const logger = require('../api/utils/logger');
 const sanitize = require('../api/utils/sanitize');
 
+// Auth endpoints that receive credentials and need stricter limits.
+// Note: PATCH /api/v1/account is intentionally excluded — while it can accept
+// a password field, it is primarily a general profile-update endpoint (nickname,
+// email, language, etc.) that requires an existing valid token. The dedicated
+// password-change route (/api/v1/account/password) is the one exposed to
+// unauthenticated reset flows and is covered here.
+const AUTH_PATHS = [
+  '/api/v1/login',
+  '/api/v1/signup',
+  '/api/v1/forgotPassword',
+  '/api/v1/account/password',
+];
+
+// Maximum body size for auth endpoints (1 KB).
+// Checked against the actual parsed body after the body parser runs,
+// so it cannot be bypassed via chunked transfer encoding or a spoofed
+// Content-Length header.
+const AUTH_BODY_LIMIT = 1024;
+
 module.exports.http = {
   /** **************************************************************************
    *                                                                           *
@@ -33,6 +52,15 @@ module.exports.http = {
     generalRateLimit: rateLimiter.generalRateLimit,
     deleteRateLimit: rateLimiter.deleteRateLimit,
 
+    // Stricter rate limiter for auth endpoints only.
+    // Wraps the express-rate-limit instance so it only fires on auth paths.
+    authRateLimit(req, res, next) {
+      if (AUTH_PATHS.includes(req.path)) {
+        return rateLimiter.authRateLimit(req, res, next);
+      }
+      return next();
+    },
+
     /** *************************************************************************
      *                                                                          *
      * The order in which middleware should be run for HTTP requests.           *
@@ -43,13 +71,16 @@ module.exports.http = {
     order: [
       'traceId',
       'corsHeaders',
+      'securityHeaders',
       'parseAuthToken',
       'generalRateLimit',
       'deleteRateLimit',
+      'authRateLimit',
       'responseTimeLogger',
       'requestLogger',
       'fileMiddleware',
       'bodyParser',
+      'authBodyLimit',
       'compress',
       'poweredBy',
       'addPackageVersionHeader',
@@ -97,6 +128,39 @@ module.exports.http = {
       return next();
     },
 
+    // Security headers: HSTS, X-Content-Type-Options, X-Frame-Options
+    securityHeaders(req, res, next) {
+      // HSTS: instruct browsers to always use HTTPS (1 year, include subdomains)
+      res.set(
+        'Strict-Transport-Security',
+        'max-age=31536000; includeSubDomains'
+      );
+      // Prevent MIME-type sniffing
+      res.set('X-Content-Type-Options', 'nosniff');
+      // Prevent clickjacking — API responses should never be framed
+      res.set('X-Frame-Options', 'DENY');
+      return next();
+    },
+
+    // Stricter body size limit for auth endpoints (1 KB max).
+    // Runs after the body parser so it checks the actual parsed payload size,
+    // not the client-supplied Content-Length header (immune to chunked bypass).
+    authBodyLimit(req, res, next) {
+      if (!AUTH_PATHS.includes(req.path)) {
+        return next();
+      }
+      const bodySize = Buffer.byteLength(
+        JSON.stringify(req.body) || '',
+        'utf8'
+      );
+      if (bodySize > AUTH_BODY_LIMIT) {
+        return res.status(413).json({
+          message: 'Request body too large for this endpoint.',
+        });
+      }
+      return next();
+    },
+
     // If a bearer token is present & valid, put it in req.token.
     // If a token is present but revoked or missing iat, actively reject with 401.
     parseAuthToken: (req, res, next) => {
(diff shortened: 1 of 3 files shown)


Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[config/http.js:144-149] `authBodyLimit` is bypassable via chunked transfer encoding** — When a client omits the `Content-Length` header (e.g. by using `Transfer-Encoding: chunked`), `parseInt(req.headers['content-length'], 10)` returns `NaN`, and `NaN > AUTH_BODY_LIMIT` evaluates to `false`, so the middleware unconditionally calls `next()`. An attacker can trivially bypass the 1 KB limit by stripping that header. The test suite only exercises the happy path where `supertest` automatically sets `Content-Length`, so this gap goes undetected. Consider replacing the header check with body-parser's built-in `limit` option applied only on auth routes (e.g. `express.json({ limit: '1kb' })` wired as a path-specific middleware), which enforces the limit on the actual bytes streamed rather than trusting a client-supplied header.

2. **[config/http.js:21-26] `PATCH /api/v1/account` absent from AUTH_PATHS** — Issue #1579 explicitly lists `/api/v1/account` (the general account update endpoint) as an affected endpoint in the acceptance criteria table. It is not present in `AUTH_PATHS`, so neither the auth rate limiter nor the 1 KB body cap applies to it. If this was a deliberate choice (e.g. because that route does not accept passwords), a short comment explaining the exclusion would prevent future confusion and keep the acceptance criteria traceable.

3. **[config/rateLimit/rateLimiter.js:141-145] `authRateLimit` does not exempt administrators** — Both `generalRateLimit` and `deleteRateLimit` skip administrator accounts. `authRateLimit` has no such exemption. If an administrator triggers more than 10 login/forgot-password requests in a 15-minute window (e.g. from automated tooling or repeated sessions), they would be locked out. Locking out everyone equally may be intentional for auth endpoints; if so, a comment to that effect would make the asymmetry look deliberate rather than accidental.

4. **[config/rateLimit/rateLimiter.js:141-143] Simplify the `skip` callback** — `skip: () => { if (isTestOrDev()) return true; return false; }` can be written as `skip: () => isTestOrDev()`, consistent with how the rest of the file is written.

### Nitpicks (Optional)

5. **[test/integration/4_routes/security-headers.test.js:5-41] Four separate HTTP requests for the same endpoint** — The four header assertions in the `Security headers middleware` describe block each fire an independent `GET /api/v1/swagger.yaml` request. Moving the request into a `before` / `beforeEach` block and asserting all headers in a single response would make the suite faster and cleaner without changing coverage.

ClemRz:

**1. `authBodyLimit` bypassable via chunked transfer encoding (Must Fix)** ✅
Replaced the `Content-Length` header check with a post-parse size check: `authBodyLimit` now runs *after* the body parser and measures `Buffer.byteLength(JSON.stringify(req.body))`. This checks the actual parsed payload, not a client-supplied header, so it's immune to chunked encoding bypass.

**2. `PATCH /api/v1/account` absent from AUTH_PATHS (Should Consider)** ✅
Added a comment explaining the deliberate exclusion — that endpoint requires an existing valid token and the dedicated password-change route (`/api/v1/account/password`) is already covered.

**3. `authRateLimit` does not exempt administrators (Should Consider)** ✅
Added a JSDoc comment explaining this is intentional: auth endpoints are the primary brute-force target and an attacker doesn't have a valid token when hammering `/login`.

**4. Simplify the `skip` callback (Nitpick)** ✅
Simplified to `skip: () => isTestOrDev()`.

**5. Four separate HTTP requests in security headers test (Nitpick)** ✅
Moved the request into a `before` hook so all header assertions share a single response.

Full suite passes (2218 tests, 0 failures).

Paul-AUB (approved):

PR #1584, opened 2026-05-15
fix(caver): fix delete/merge losing data and add missing FK handling

### LINKED ISSUE #1583: Suppression des personnes : bug sur l'attachement des données à une autre personne
**Describe the bug**
Dans la base de données nous avons de nombreuses personnes qui sont déclérées plusieurs fois. cela vient des données du BBS dans lequel le nom des personnes était saisi de manière non normé.
Nous avons une fonctionnalité qui permet de supprimer une personne et de rattacher ses données a une autre personne

Cela ne fonctionne pas bien : je viens de supprimer toutes les instances de Joerg et désormais les contributions apparaissent comme associées à deleted

Image

Il y a un premier problème lié au fait que le compte supprimé était celui d'un contributeur, ce qui n'aurait pas du être possible. 
Le second problème étant que les informations associées au compte supprimées n'ont pas été rattachées au compte que j'ai indiqué

---
_Moved from GrottoCenter/grottocenter-front#1271_

ClemRz:
## 🤔 What

Fix the caver delete/merge endpoint that was silently losing data instead of reassigning it to the merge target.

- Closes #1583

## 🤷‍♂️ Why

When deleting a caver with a merge target (`entityId`), contributions were ending up as "deleted" instead of being reassigned. Two root causes:

1. The front-end sends `entityId=true` (a boolean) instead of the actual caver ID — `parseInt('true')` → `NaN` → the API silently fell through to the no-merge path, reassigning all authorship to the "deleted" placeholder caver (id 8).
2. The controller referenced a ghost `exploredEntrances` association that no longer exists (removed during the `j_entrance_caver` → `j_caver_cave_explorer` migration), and was missing FK handling for `t_crs`, `t_point`, `h_document.validator`, `t_token_blacklist`, and `j_caver_region_subscription`.

## 🔍 How

- **Validate `entityId`**: return 400 if it's not a valid numeric ID
- **Return 400 when merge target doesn't exist**: instead of silently falling through to the no-merge path
- **Remove `exploredEntrances`** references (legacy from dropped `j_entrance_caver`)
- **Add missing FK reassignment**: `TCrs`, `TPoint`, `HDocument.validator`, `TTokenBlacklist`, `subscribedToRegions`
- **Use `replaceCollection`** to always clear junction tables regardless of whether `getCaver` populates them
- **Add audit logging** on permanent delete for traceability
- **Add `TCrs` to ESLint globals**
- **Clean up fixture**: remove legacy `exploredEntrances` from `tcaver.json`

## 🧪 Testing

```bash
npm run test -- --grep "Caver features"
```

New test cases:
- Returns 400 when `entityId` is not a valid number (e.g., `entityId=true`)
- Returns 400 when merge target caver does not exist

Manual test: `DELETE /api/v1/cavers/:id?entityId=<targetId>` with a valid target ID — verify the caver is deleted and its data appears under the target.

## 📸 Previews

N/A (backend only)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -32,6 +32,7 @@ module.exports = {
     TCaver: 'readonly',
     TComment: 'readonly',
     TCountry: 'readonly',
+    TCrs: 'readonly',
     TDescription: 'readonly',
     TDocument: 'readonly',
     TDocumentDuplicate: 'readonly',
--- a/api/controllers/v1/caver/delete.js
+++ b/api/controllers/v1/caver/delete.js
@@ -1,5 +1,4 @@
 const ControllerService = require('../../../services/ControllerService');
-const NotificationService = require('../../../services/NotificationService');
 const RightService = require('../../../services/RightService');
 const { toCaver } = require('../../../services/mapping/converters');
 const CaverService = require('../../../services/CaverService');
@@ -30,22 +29,31 @@ module.exports = async (req, res) => {
     return res.forbidden('You are not authorized to delete an author.');
 
   const mergeIntoId = parseInt(req.param('entityId'), 10);
-  let shouldMergeInto = !Number.isNaN(mergeIntoId);
+  const shouldMergeInto = req.param('entityId') !== undefined;
+  if (shouldMergeInto && Number.isNaN(mergeIntoId)) {
+    return res.badRequest({
+      message: `Invalid entityId: expected a numeric caver ID, got "${req.param('entityId')}".`,
+    });
+  }
   let mergeIntoEntity;
   if (shouldMergeInto) {
     mergeIntoEntity = await TCaver.findOne(mergeIntoId)
-      .populate('exploredEntrances')
       .populate('exploredCaves')
       .populate('groups')
       .populate('subscribedToCountries')
       .populate('subscribedToMassifs')
+      .populate('subscribedToRegions')
       .populate('grottos')
       .populate('documents');
-    shouldMergeInto = !!mergeIntoEntity;
+    if (!mergeIntoEntity) {
+      return res.badRequest({
+        message: `Merge target caver of id ${mergeIntoId} not found.`,
+      });
+    }
   }
 
   // eslint-disable-next-line no-inner-declarations
-  async function reafectField(model, field, replacement = null) {
+  async function reassignField(model, field, replacement = null) {
     if (shouldMergeInto) {
       await model.update({ [field]: caverId }).set({ [field]: mergeIntoId });
     } else {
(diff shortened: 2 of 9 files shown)


Paul-AUB (changes requested):
The bug description and fix strategy are well-reasoned, and the input validation improvements (400 on `NaN` entityId, 400 on missing merge target) are solid. However, there are a few cases where the merge path will still silently drop data.

### Issues (Must Fix)

1. **[`api/controllers/v1/caver/delete.js:146`]** `caver.exploredCaves` is not in the object returned by `CaverService.getCaver()`. The service populates `exploredCaves` internally but builds derived properties `exploredEntrances`/`exploredNetworks` and omits `exploredCaves` from the return value (`api/services/CaverService.js:109-124`). On the merge path, `linkedEntitiesDeleteOrMerge('exploredCaves')` therefore always sees `sourceEntities = []`, skips the `addToCollection` call, and still runs `replaceCollection(caverId, 'exploredCaves', [])` — clearing the source's explored caves without transferring them to the target. To fix: either populate `exploredCaves` directly in the delete controller (with a separate `TCaver.findOne(caverId).populate('exploredCaves')`), or expose the raw populated array from `getCaver`.

2. **[`api/controllers/v1/caver/delete.js:150`]** `subscribedToRegions` is never populated in `CaverService.getCaver()` (neither in the populate chain nor in the returned object). Adding `linkedEntitiesDeleteOrMerge('subscribedToRegions')` without fixing this means region subscriptions will always be cleared from the source caver and never transferred to the merge target. The merge target query already populates `subscribedToRegions` (line 46) — the source query in `getCaver` needs the same.

3. **[`api/controllers/v1/caver/delete.js:145-153`]** `documents` is populated with `limit: 10` in `getCaver` (`api/services/CaverService.js:64-67`). If the source caver has more than 10 documents, the overflow documents won't appear in `caver.documents`, so only the first 10 would be transferred on the merge path — the rest are cleared by `replaceCollection` without being added to the target. The delete controller should either re-query `documents` without a limit, or use the database-level reassignment approach (e.g., `UPDATE j_caver_document SET id_caver = mergeIntoId WHERE id_caver = caverId`) rather than relying on the populated array.

4. **[`api/controllers/v1/caver/delete.js:175`]** `NotificationService.notifySubscribers` is called with `NOTIFICATION_ENTITIES.ENTRANCE` for a caver delete — looks like a copy-paste artifact from an entrance controller. Pre-existing bug, not introduced here, but given the new audit log this PR adds, it would be a natural cleanup.

5. **[`test/integration/4_routes/Cavers/delete.test.js`]** The two new tests cover error paths only. There is no integration test for the happy-path merge (e.g., `DELETE /api/v1/cavers/100?entityId=1` → 200, then verify the source's data appears under caver 1). Without a green merge-path test, the fix for issue #1583 has no regression coverage.

### Nitpicks (Optional)

6. **[`api/controllers/v1/caver/delete.js:57`]** `reafectField` misspells "reassign" — pre-existing, but cheap to clean up since this PR already touches all call-sites.

ClemRz:

**1. `exploredCaves` not in `getCaver()` return** — Fixed. Added `exploredCaves` to the return object and added `.populate('subscribedToRegions')` to the query.

**2. `subscribedToRegions` not populated** — Fixed. Same as above — now populated and returned.

**3. `documents` limited to 10** — Fixed. The delete controller now does a separate `TCaver.findOne(caverId).populate('documents')` (without limit) to get the full document list for the merge path. `getCaver()` itself is unchanged — the `limit: 10` is intentional for the public-facing endpoint and will be addressed separately with pagination.

**4. Wrong notification entity** — Fixed. Removed the `notifySubscribers` call entirely — there's no `NOTIFICATION_ENTITIES.CAVER` and notifying about an "entrance" for a caver delete was nonsensical.

**5. No happy-path merge test** — Fixed. Added an integration test that verifies `exploredCaves`, `documents`, and `subscribedToRegions` are transferred from source to target during a merge.

**6. `reafectField` misspelling** — Fixed. Renamed to `reassignField`.

Paul-AUB (approved):

PR #1587, opened 2026-05-15
fix(document): clean up entrances and bibliographic metadata on permanent delete

### LINKED ISSUE #1571: bug(api): [23503] tdocument - FK violation j_document_entrance on delete
**Describe the bug**
When permanently deleting a document via `DELETE /api/v1/documents/:id?isPermanent=1`, the API attempts to destroy the `t_document` row without first removing associated records in the `j_document_entrance` junction table. PostgreSQL rejects the delete due to the foreign key constraint `j_document_entrance_t_document_fk`.

**To Reproduce**
Based on the trace logs:
1. Authenticate as a Moderator user
2. Send `DELETE /api/v1/documents/233874?isPermanent=1`
3. The document has associated entrance links in `j_document_entrance`
4. The API tries to destroy the document directly without cascading or pre-deleting junction records
5. PostgreSQL returns a foreign key violation error

**Expected behavior**
The API should either:
- Delete associated `j_document_entrance` records before deleting the document
- Use CASCADE on the foreign key constraint
- Or return a clear 409 Conflict error explaining that the document still has linked entrances that must be unlinked first

**Error Details**
- Occurrences: 2
- Time Range: 2026-05-08T11:46:07Z to 2026-05-08T11:46:22Z
- Representative Trace ID: `019e0768-fbd3-72db-bdb8-8cfad806890a`
- Stack Trace:
```
at _afterCounting (waterline/lib/waterline/methods/destroy-one.js:167:17)
at _afterTalkingToAdapter (waterline/lib/waterline/methods/count.js:227:16)
at Object.success (sails-postgresql/lib/adapter.js:338:18)

Raw PG error:
  code: '23503'
  detail: 'Key (id)=(233874) is still referenced from table "j_document_entrance".'
  constraint: 'j_document_entrance_t_document_fk'
```

**Additional context**
- Service: Document delete handler (Waterline `destroyOne`)
- Model: `tdocument`
- DB Error Code: `23503` (foreign key violation)
- Constraint: `j_document_entrance_t_document_fk`
- Both occurrences were from the same Moderator user attempting to permanently delete documents (IDs 233873, 233874) that still had entrance associations

ClemRz:
## 🤔 What

- Fix FK violations when permanently deleting a document (`DELETE /api/v1/documents/:id?isPermanent=1`)
- Add cleanup for `j_document_entrance` and `t_bibliographic_metadata` before hard delete
- Refactor duplicated test bootstrap logic into a shared `test/seed-database.js` module
- Closes #1571

## 🤷‍♂️ Why

Permanently deleting a document failed with a 500 error due to two missing FK cleanup steps:
1. `j_document_entrance` — the `entrances` many-to-many association was not being cleared/merged, unlike all other junction tables
2. `t_bibliographic_metadata` — this table has no Waterline model but holds a FK to `t_document`, so it must be cleaned up via raw SQL

The bootstrap refactor was needed because the fix required adding `t_bibliographic_metadata` table creation to the test DB setup, and the seeding logic was duplicated between `test/bootstrap.test.js` and `scripts/snapshot-test-db.js` — making it easy to forget one (which is exactly how the original divergence happened).

## 🔍 How

**Controller fix** (`api/controllers/v1/document/delete.js`):
- Added `await linkedEntitiesDeleteOrMerge('entrances')` alongside the other junction table cleanups
- Added `sails.sendNativeQuery('DELETE FROM t_bibliographic_metadata WHERE id_document = $1', [documentId])` before the hard delete

**Test infrastructure** (`test/seed-database.js`):
- Extracted the shared Sails lift + fixture load + post-fixture SQL sequence into a single module
- Both `bootstrap.test.js` and `snapshot-test-db.js` now call `seedDatabase({ testUrl })` instead of duplicating the logic
- Added `CREATE_BIBLIOGRAPHIC_METADATA_TABLE` to `test/customSQL.js` since this table has no Waterline model
- Added `dropNonWaterlineTables()` pre-step to handle the chicken-and-egg problem where `migrate:drop` can't drop `t_document` if `t_bibliographic_metadata` already exists from a previous run

## 🧪 Testing

```bash
npm run test:snapshot   # Rebuild template (required once)
npm run test            # All 2212 tests pass
npm run test -- --grep "Document delete"  # 10 tests, including 3 new ones
```

New test cases:
- Permanently delete a document with entrance associations
- Merge entrance associations when permanently deleting into another document
- Permanently delete a document with bibliographic metadata

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/api/controllers/v1/document/delete.js
+++ b/api/controllers/v1/document/delete.js
@@ -99,6 +99,7 @@ module.exports = async (req, res) => {
     }
 
     await linkedEntitiesDeleteOrMerge('countries');
+    await linkedEntitiesDeleteOrMerge('entrances');
     await linkedEntitiesDeleteOrMerge('isoRegions');
     await linkedEntitiesDeleteOrMerge('languages');
     await linkedEntitiesDeleteOrMerge('massifs');
@@ -116,6 +117,10 @@ module.exports = async (req, res) => {
     await TDescription.destroy({ document: documentId });
 
     await HDocument.destroy({ t_id: documentId });
+    await sails.sendNativeQuery(
+      'DELETE FROM t_bibliographic_metadata WHERE id_document = $1',
+      [documentId]
+    );
     await TDocument.destroyOne({ id: documentId }); // Hard delete
   }
 
--- a/scripts/snapshot-test-db.js
+++ b/scripts/snapshot-test-db.js
@@ -31,76 +31,6 @@ const templateDbName = `${testDbName}_template`;
 url.pathname = '/postgres';
 const maintenanceUrl = url.toString();
 
-function runFullBootstrap() {
-  return new Promise((resolve, reject) => {
-    // eslint-disable-next-line global-require
-    const sails = require('sails');
-    // eslint-disable-next-line global-require, import/no-extraneous-dependencies
-    const Fixted = require('fixted');
-    // eslint-disable-next-line global-require
-    const sailsPostGreAdapter = require('sails-postgresql');
-    // eslint-disable-next-line global-require
-    const customSQL = require('../test/customSQL');
-    // eslint-disable-next-line global-require
-    const CommonService = require('../api/services/CommonService');
-    // eslint-disable-next-line global-require
-    const FIXTURE_ORDER = require('../test/fixtureOrder');
-
-    sails.lift(
-      {
-        log: { level: 'error' },
-        datastores: {
-          default: { adapter: sailsPostGreAdapter, url: testUrl },
-        },
-        models: { migrate: 'drop' },
-        csrf: false,
-        async bootstrap() {
-          await CommonService.query(customSQL.ALTER_MASSIF_COLUMN_GEOG_POLYGON);
-          await CommonService.query(customSQL.ALTER_ENTRANCE_COLUMN_POINT_GEOM);
-          await CommonService.query(
-            customSQL.CREATE_ENTRANCE_POINT_GEOM_INSERT_TRIGGER
-          );
-        },
-      },
-      // eslint-disable-next-line consistent-return
-      async (liftErr) => {
-        if (liftErr) return reject(liftErr);
-
-        const fixted = new Fixted();
-        fixted.populate(
-          FIXTURE_ORDER,
-          // eslint-disable-next-line consistent-return
-          (fixtedError) => {
-            if (fixtedError) {
-              sails.lower(() => reject(fixtedError));
-              return;
-            }
-
-            CommonService.query(
-              [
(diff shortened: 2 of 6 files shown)


Paul-AUB (dismissed):

1. **[test/seed-database.js:29-37]** `dropNonWaterlineTables` drops the table but leaves the `e_metadata_status` ENUM type behind. The `CREATE_BIBLIOGRAPHIC_METADATA_TABLE` script handles the current case correctly with `IF NOT EXISTS`. However, if the enum's values ever need to change (e.g., adding a new status), the stale type won't be refreshed between test runs because the `IF NOT EXISTS` guard will skip it. Consider adding `DROP TYPE IF EXISTS e_metadata_status` after the `DROP TABLE` line to guarantee a fully clean state:

   ```js
   await client.query('DROP TABLE IF EXISTS t_bibliographic_metadata');
   await client.query('DROP TYPE IF EXISTS e_metadata_status');
   ```

2. **[test/seed-database.js:39-93]** `seedDatabase` wraps a `.then()` chain (for `dropNonWaterlineTables`) around Sails' callback-style `lift()`, all inside a `new Promise()` constructor. This works correctly but makes error-path tracing harder. Consider awaiting `dropNonWaterlineTables` before the `new Promise()` wrapper to flatten the nesting:

   ```js
   async function seedDatabase({ testUrl }) {
     await dropNonWaterlineTables(testUrl);
     return new Promise((resolve, reject) => {
       sails.lift({ ... }, async (liftErr) => { ... });
     });
   }
   ```

### Nitpicks (Optional)

3. **[test/integration/4_routes/Documents/delete.test.js:294]** The "merge entrance" test does not assert that the source document's junction entries were cleared (only that the target gained them). The standalone "permanently delete with entrance associations" test already covers source cleanup, so this is not a gap in coverage — just a note that adding `should(await JDocumentEntrance.find({ document: doc.id })).have.length(0)` after the delete would make the merge test self-contained as a spec.

ClemRz:
Thanks for the review @Paul-AUB! All three suggestions were legitimate — I've addressed them in the amended commit:

1. **Drop stale enum type** — Added `DROP TYPE IF EXISTS e_metadata_status` after the table drop in `dropNonWaterlineTables` to guarantee a fully clean state.
2. **Flatten async nesting** — Converted `seedDatabase` to `async function` and awaited `dropNonWaterlineTables` before the `new Promise()` wrapper.
3. **Assert source cleanup in merge test** — Added `should(sourceJunctions).have.length(0)` assertion in the merge entrance test to make it self-contained.

All tests pass (2212 passing, 0 failing).

PR #1588, opened 2026-05-15
fix(entrance): make add-document endpoint idempotent

### LINKED ISSUE #1572: bug(api): [E_UNIQUE] jdocumententrance - duplicate on addToCollection
**Describe the bug**
When linking a document to an entrance via `PUT /api/v1/entrances/:entranceId/documents/:documentId`, the API calls `addToCollection` on the `jdocumententrance` junction table without first checking if the association already exists. If the document-entrance link is already present, PostgreSQL raises a uniqueness constraint violation and the API returns a 500 error.

**To Reproduce**
Based on the trace logs:
1. Authenticate as any user
2. Send `PUT /api/v1/entrances/148366/documents/233878`
3. The document 233878 is already linked to entrance 148366 in `j_document_entrance`
4. Waterline's `addToCollection` attempts to insert a duplicate row
5. PostgreSQL rejects the insert due to the unique constraint

**Expected behavior**
The API should handle the case where the association already exists:
- Either check for existence before attempting to add (idempotent PUT)
- Or catch the `E_UNIQUE` error and return a 409 Conflict with a clear message
- PUT operations should ideally be idempotent — linking an already-linked document should succeed silently

**Error Details**
- Occurrences: 2
- Time Range: 2026-05-07T20:13:23Z to 2026-05-07T20:17:19Z
- Representative Trace ID: `019e0416-6e08-7199-80ef-e5ac63addfba`
- Stack Trace:
```
at waterline/lib/waterline/methods/add-to-collection.js:424:19
at _.extend._WLModel [as _handleExec] (waterline/lib/waterline/methods/add-to-collection.js:464:9)
at Deferred.exec (parley/lib/private/Deferred.js:286:10)

Error properties:
  code: 'E_UNIQUE'
  adapterMethodName: 'createEach'
  modelIdentity: 'jdocumententrance'
```

**Additional context**
- Service: Entrance-document association handler (Waterline `addToCollection`)
- Model: `jdocumententrance`
- DB Error Code: `E_UNIQUE` (Waterline uniqueness violation)
- The request was `PUT /api/v1/entrances/148366/documents/233878` — a user attempting to link a document that was already associated
- Both occurrences came from the same user within 4 minutes, suggesting a retry after the first failure

ClemRz:
## 🤔 What

- Make `PUT /api/v1/entrances/:entranceId/documents/:documentId` idempotent
- Return 204 (success) when the document-entrance link already exists, instead of 400
- Add a try/catch around `addToCollection` to handle `E_UNIQUE` race conditions gracefully
- Update Swagger spec to document the idempotent behavior
- Closes #1572

## 🤷‍♂️ Why

When a user attempts to link a document that is already associated with an entrance, the API was returning a 500 error due to a PostgreSQL uniqueness constraint violation (`E_UNIQUE` on `jdocumententrance`). A previous fix returned 400, but PUT operations should be idempotent — repeating the same request should produce the same result without error.

## 🔍 How

1. The existing `JDocumentEntrance.count()` pre-check now returns `res.ok()` (204) instead of `res.badRequest()` when the link already exists
2. A try/catch wraps `addToCollection` to handle the race condition where a concurrent request inserts the row between the count check and the insert — also returns 204
3. Only truly new links trigger the `dateReviewed` update on the document

## 🧪 Testing

```bash
npm run test -- --grep "Entrance add document"
```

All 8 tests pass, including the updated duplicate test that now expects 204.

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/api/controllers/v1/entrance/add-document.js
+++ b/api/controllers/v1/entrance/add-document.js
@@ -19,12 +19,21 @@ module.exports = async (req, res) => {
     entrance: entranceId,
   });
   if (existingLink > 0) {
-    return res.badRequest({
-      message: `Document ${documentId} is already linked to entrance ${entranceId}.`,
-    });
+    return res.ok();
   }
 
-  await TEntrance.addToCollection(entranceId, 'documents', documentId);
+  try {
+    await TEntrance.addToCollection(entranceId, 'documents', documentId);
+  } catch (err) {
+    if (err.code === 'E_UNIQUE') {
+      sails.log.debug(
+        `Race-condition duplicate: entrance ${entranceId} / document ${documentId} link already exists.`
+      );
+      return res.ok();
+    }
+    throw err;
+  }
+
   await TDocument.updateOne(documentId).set({
     dateReviewed: new Date(),
   });
--- a/assets/swaggerV1.yaml
+++ b/assets/swaggerV1.yaml
@@ -3642,7 +3642,7 @@ paths:
       tags:
         - entrances
         - documents
-      description: Add a document to an entrance.
+      description: Add a document to an entrance. This operation is idempotent — linking an already-linked document returns 204 without error.
       parameters:
         - name: entranceId
           in: path
@@ -3657,7 +3657,7 @@ paths:
             type: string
       responses:
         '204':
-          description: Successful operation
+          description: Successful operation (also returned if the link already exists)
         '404':
           description: Could not find entrance with the entrance id you provided or could not find the document with the document id you provided.
         '403':
--- a/test/integration/4_routes/Entrances/addDocument.test.js
+++ b/test/integration/4_routes/Entrances/addDocument.test.js
@@ -79,7 +79,7 @@ describe('Entrance add document features', () => {
         await JDocumentEntrance.destroy({ document: testDocumentId });
       });
 
-      it('should return 400 when adding same document twice to same entrance', async () => {
+      it('should return 204 idempotently across two identical PUT calls', async () => {
         await supertest(sails.hooks.http.app)
           .put(
             `/api/v1/entrances/${testEntranceId}/documents/${testDocumentId}`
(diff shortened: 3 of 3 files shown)


Paul-AUB (dismissed):
Clean fix that correctly handles both the happy-path duplicate (pre-check) and the race-condition duplicate (try/catch). The approach is idiomatic for idempotent PUTs and the Swagger update keeps the contract honest.

1. **[test/integration/4_routes/Entrances/addDocument.test.js:82-101]** The PR states "only truly new links trigger the `dateReviewed` update" — that business rule isn't covered by any test. Consider adding an assertion that `dateReviewed` on the document is unchanged after the second (duplicate) PUT. This would catch a future regression where the `dateReviewed` update accidentally moved outside the new-link branch.

2. **[api/controllers/v1/entrance/add-document.js:27-31]** The caught `E_UNIQUE` is silently swallowed. A `sails.log.debug(...)` call here would surface race-condition occurrences in the log stream without impacting the response, making it easier to spot if concurrent retries become unexpectedly frequent.

### Nitpicks (Optional)

3. **[test/integration/4_routes/Entrances/addDocument.test.js:82]** The test description says "when document is already linked", but the fixture starts clean — the first PUT is actually the initial link. Both calls return 204, which is correct, but the description implies the pre-condition is an existing link. Seeding the `JDocumentEntrance` row before the first PUT (or rewording to "makes the link idempotent across two calls") would align intent with implementation.

ClemRz:
Thanks for the thorough review @Paul-AUB! All three items addressed:

1. **`dateReviewed` regression test** — Added a dedicated test asserting that `dateReviewed` remains unchanged after a duplicate PUT.
2. **`sails.log.debug` on race-condition path** — The `E_UNIQUE` catch now logs a debug message for observability.
3. **Test description rewording** — Changed from "when document is already linked" to "across two identical PUT calls" to match what the test actually does.

PR #1589, opened 2026-05-15
fix(entrance): add body validation to check-rows endpoints

### LINKED ISSUE #1585: bug(api): [TypeError] tentrance - missing body validation in check-rows
**Describe the bug**
The `POST /api/v1/entrances/check-rows` endpoint crashes with a `TypeError: Cannot read properties of undefined (reading 'entries')` when the request body does not contain a `data` property. The controller accesses `req.body.data.entries()` at line 7 without validating that `req.body.data` exists first.

**To Reproduce**
Based on the trace logs:
1. Authenticate as any user
2. Send `POST /api/v1/entrances/check-rows` with an empty body (or a body missing the `data` array)
3. The server returns a 500 error

**Expected behavior**
The API should validate that `req.body.data` is a non-empty array before iterating, and return a 400 Bad Request if the payload is malformed.

**Error Details**
- Error Type: `TypeError`
- Occurrences: 2
- Time Range: 2026-05-14T10:13:08Z to 2026-05-14T10:13:46Z
- Representative Trace ID: `019e25fa-2dd7-77c4-8f1b-d9ba6cad9df1`
- Stack Trace:
```
TypeError: Cannot read properties of undefined (reading 'entries')
    at module.exports [as v1/entrance/check-rows] (/home/site/wwwroot/api/controllers/v1/entrance/check-rows.js:7:44)
```

**Additional context**
- Controller: `api/controllers/v1/entrance/check-rows.js`
- Model: `tentrance`
- The request body was `{}` with no params or query string, suggesting the client may have failed to attach the CSV data payload
- Fix: add input validation at the top of the controller to check `req.body.data` is an array before iterating

ClemRz:
## 🤔 What

- Add input validation to `POST /api/v1/entrances/check-rows` and `POST /api/v1/documents/check-rows`
- Return 400 Bad Request when `req.body.data` is missing or not an array
- Update Swagger spec for both endpoints (add 400 response, fix request body schema)
- Add the missing `entrances/check-rows` endpoint to Swagger
- Closes #1585

## 🤷‍♂️ Why

When the request body doesn't contain a `data` property, the controller crashes with `TypeError: Cannot read properties of undefined (reading 'entries')`, returning a 500 to the client. The same vulnerability existed in both the entrance and document check-rows controllers.

## 🔍 How

1. Destructure `data` from `req.body` with a fallback to empty object for safety
2. Check `Array.isArray(data)` before the loop — return 400 with a descriptive message if invalid
3. Replace `req.body.data.entries()` with `data.entries()` using the validated local variable
4. Updated Swagger to document the 400 response and corrected the request body schema (it's `{ data: [...] }`, not a bare array)

## 🧪 Testing

```bash
npm run test -- --grep "check-rows"
```

16 tests pass — 8 for entrance, 8 for document (including 2 new validation tests each).

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/api/controllers/v1/document/check-rows.js
+++ b/api/controllers/v1/document/check-rows.js
@@ -1,10 +1,17 @@
 const { valIfTruthyOrNull } = require('../../../utils/csvHelper');
 
 module.exports = async (req, res) => {
+  const { data } = req.body || {};
+  if (!Array.isArray(data)) {
+    return res.badRequest({
+      message: 'Missing or invalid "data" property. Expected an array.',
+    });
+  }
+
   const willBeCreated = [];
   const willBeCreatedAsDuplicates = [];
   const wontBeCreated = [];
-  for (const [index, row] of req.body.data.entries()) {
+  for (const [index, row] of data.entries()) {
     const idDbImport = valIfTruthyOrNull(row.id);
     const nameDbImport = valIfTruthyOrNull(
       row['dct:rights/cc:attributionName']
--- a/api/controllers/v1/entrance/check-rows.js
+++ b/api/controllers/v1/entrance/check-rows.js
@@ -1,10 +1,17 @@
 const { valIfTruthyOrNull } = require('../../../utils/csvHelper');
 
 module.exports = async (req, res) => {
+  const { data } = req.body || {};
+  if (!Array.isArray(data)) {
+    return res.badRequest({
+      message: 'Missing or invalid "data" property. Expected an array.',
+    });
+  }
+
   const willBeCreated = [];
   const willBeCreatedAsDuplicates = [];
   const wontBeCreated = [];
-  for (const [index, row] of req.body.data.entries()) {
+  for (const [index, row] of data.entries()) {
     const idDbImport = valIfTruthyOrNull(row.id);
     const nameDbImport = valIfTruthyOrNull(
       row['dct:rights/cc:attributionName']
@@ -21,6 +28,7 @@ module.exports = async (req, res) => {
     const dnEntrance = await TEntrance.findOne({
       idDbImport,
       nameDbImport,
+      isDeleted: false,
     });
 
     if (!dnEntrance) {
--- a/assets/swaggerV1.yaml
+++ b/assets/swaggerV1.yaml
@@ -1832,16 +1832,23 @@ paths:
     post:
       tags:
         - documents
-      description: Check if the data sent are already present in Grottocenter (incomplete).
+      description: Check if the document data sent are already present in Grottocenter.
+      security:
+        - bearerAuth: []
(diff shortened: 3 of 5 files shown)


Paul-AUB (dismissed):

1. **[api/controllers/v1/document/check-rows.js:8, api/controllers/v1/entrance/check-rows.js:8]** The error message says `"Expected a non-empty array."`, but an empty array (`data: []`) is accepted and returns 200 (confirmed by the existing "should return empty arrays for empty data" test). Consider changing the message to `"Expected an array."` to avoid misleading API consumers who might think `data: []` would also return a 400.

2. **[assets/swaggerV1.yaml:3817-3858, :1831-1873]** Both the new `/entrances/check-rows` entry and the updated `/documents/check-rows` entry are missing a `security: - bearerAuth: []` declaration and a `'401': Unauthorized` response, even though both routes are protected by the `tokenAuth` policy (see `config/policies.js` lines 97 and 140). This means Swagger UI won't show the padlock icon and clients won't know they can receive a 401. Other authenticated endpoints in the file (e.g. `/account`, `/cavers/{id}/ban`) consistently include both.

3. **[api/controllers/v1/entrance/check-rows.js:29-32]** The entrance duplicate check doesn't filter by `isDeleted`:
   ```js
   const dnEntrance = await TEntrance.findOne({ idDbImport, nameDbImport });
   ```
   The document controller (line 29-33) does filter by `isDeleted: false`. A soft-deleted entrance would still be flagged as a duplicate, silently blocking re-import of data that was deleted. This is a pre-existing gap rather than introduced by this PR, but since you're adding tests for this controller it's a good moment to align the behavior.

### Nitpicks (Optional)

4. **[assets/swaggerV1.yaml:1869-1871, :3854-3856]** The `wontBeCreated.items` schema in both endpoint specs is missing `type: object`:
   ```yaml
   items:
     properties:       # ← should be: type: object\n    properties:
       line:
         type: integer
   ```
   OpenAPI does default to object when `properties` is present, but being explicit is more readable and consistent with the rest of the spec. Same pattern appears on the pre-existing document entry — worth fixing while touching the file.

ClemRz:
Thanks for the review @Paul-AUB! All 4 items were legit — addressed in the latest force-push:

1. **Error message** — Changed to `"Expected an array."` in both controllers (since `data: []` is valid).
2. **Swagger security/401** — Added `security: - bearerAuth: []` and `'401': Authentication required.` to both endpoints.
3. **`isDeleted` filter on entrance** — Added `isDeleted: false` to the `TEntrance.findOne()` query to align with the document controller.
4. **`type: object` on `wontBeCreated.items`** — Added to both endpoint specs for explicitness.

All 16 tests still pass.

PR #1590, opened 2026-05-15
fix(rss): handle upstream feed timeout gracefully

### LINKED ISSUE #1586: bug(api): [ETIMEDOUT] rss - upstream feed timeout causes 500
**Describe the bug**
The `GET /api/rss/:language` endpoint returns a 500 error when the upstream blog RSS feed (`blog-fr.grottocenter.org`) times out after 60 seconds. The rss-parser library throws a timeout error which is caught but returned as a 500 Server Error instead of being handled gracefully.

**To Reproduce**
Based on the trace logs:
1. Visit `GET /api/rss/FR` from the frontend (origin: grottocenter.org)
2. The upstream blog feed at `http://blog-fr.grottocenter.org/feeds/posts/default?alt=rss&max-results=1` is unreachable or slow
3. After 60 seconds, rss-parser throws `Request timed out after 60000ms`
4. The API returns a 500 error

**Expected behavior**
The API should handle upstream feed timeouts gracefully — either by returning a 502/504 with a user-friendly message, using a cached response, or reducing the timeout and failing faster with an appropriate status code.

**Error Details**
- Error Type: `Error: Request timed out after 60000ms`
- Occurrences: 2
- Time Range: 2026-05-13T22:13:26Z to 2026-05-13T22:13:26Z
- Representative Trace ID: `019e2365-fa3f-7258-b272-a59d4ba81859`
- Stack Trace:
```
Error: Request timed out after 60000ms
    at Timeout.<anonymous> (/home/site/wwwroot/node_modules/rss-parser/lib/parser.js:101:23)
    at listOnTimeout (node:internal/timers:605:17)
    at process.processTimers (node:internal/timers:541:7)
```

**Additional context**
- Controller: `api/controllers/v1/rss/get-feed.js`
- The upstream blog URLs use HTTP (not HTTPS) and point to Blogger-hosted feeds
- Both FR and EN feeds may be affected (this instance was FR)
- Consider: caching RSS responses, reducing timeout, returning 502 instead of 500, or serving a fallback response when the feed is unavailable

ClemRz:
## 🤔 What

- Handle upstream RSS feed timeouts and network errors gracefully
- Reduce feed timeout from 60s to 10s
- Return 502 Bad Gateway for upstream failures instead of 500
- Log errors for debugging
- Closes #1586

## 🤷‍♂️ Why

When the upstream blog RSS feed (blog-fr/en.grottocenter.org) is unreachable or slow, the API waits 60 seconds then returns a generic 500 error. This makes it look like our server is broken when it's actually the upstream feed that's unavailable.

## 🔍 How

1. Pass `{ timeout: 10000 }` to rss-parser to fail faster (10s instead of 60s)
2. In the catch block, detect upstream errors (timeout, ECONNREFUSED, ENOTFOUND, ECONNRESET) via message regex and error codes
3. Return 502 with a clear message for upstream failures, keep 500 for unexpected errors
4. Log the error with `sails.log.error` for observability

## 🧪 Testing

```bash
npm run test -- --grep "RSS endpoints"
```

5 tests pass — including 3 new tests using sinon stubs to simulate timeout, connection refused, and unexpected errors.

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/api/controllers/v1/rss/get-feed.js
+++ b/api/controllers/v1/rss/get-feed.js
@@ -6,12 +6,14 @@ const FR_RSS_FEED =
 const EN_RSS_FEED =
   'http://blog-en.grottocenter.org/feeds/posts/default?alt=rss&max-results=1';
 
+const FEED_TIMEOUT_MS = 10000;
+
 const CommonService = require('../../../services/CommonService');
 
 module.exports = async (req, res) => {
   // TODO error when language is neither FR nor EN
   const rssFeed = req.params.language === 'FR' ? FR_RSS_FEED : EN_RSS_FEED;
-  const parser = new Parser();
+  const parser = new Parser({ timeout: FEED_TIMEOUT_MS });
 
   try {
     const feed = await parser.parseURL(rssFeed);
@@ -30,9 +32,25 @@ module.exports = async (req, res) => {
 
     return res.json(result);
   } catch (err) {
+    sails.log.error(
+      `RSS feed fetch failed for ${req.params.language}: ${err.message}`
+    );
+
+    const isUpstreamError =
+      /timed\s+out/i.test(err.message) ||
+      err.code === 'ETIMEDOUT' ||
+      err.code === 'ECONNREFUSED' ||
+      err.code === 'ENOTFOUND' ||
+      err.code === 'ECONNRESET';
+
+    if (isUpstreamError) {
+      return res.status(502).json({
+        message: 'The upstream RSS feed is currently unavailable.',
+      });
+    }
+
     return res.serverError({
-      error: err,
-      message: 'An error occurred when getting the RSS feed',
+      message: 'An error occurred when getting the RSS feed.',
     });
   }
 };
--- a/test/integration/4_routes/Rss.test.js
+++ b/test/integration/4_routes/Rss.test.js
@@ -1,4 +1,6 @@
 const supertest = require('supertest');
+const sinon = require('sinon');
+const Parser = require('rss-parser');
 
 describe('RSS endpoints', () => {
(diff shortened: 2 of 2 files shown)


Paul-AUB (dismissed):

1. **[test/integration/4_routes/Rss.test.js]** The code handles `ENOTFOUND` (DNS resolution failure) but there is no corresponding test case for it. Consider adding a test alongside the `ECONNREFUSED` test to cover that branch:
   ```js
   const err = new Error('getaddrinfo ENOTFOUND blog-fr.grottocenter.org');
   err.code = 'ENOTFOUND';
   ```

2. **[api/controllers/v1/rss/get-feed.js:40]** The regex `/timed?\s*out/i` matches both `timed out` and `time out` (since `d` is optional). In practice this is harmless here because the only errors in scope are from `rss-parser` and TCP, but it may silently catch unrelated messages in a future refactor. Consider tightening to `/timed\s+out/i` (require the `d`) since the actual rss-parser message is `"Request timed out after Nms"`.

### Nitpicks (Optional)

3. **[test/integration/4_routes/Rss.test.js:65-75]** The `should return 500 for unexpected errors` test only asserts the HTTP status, not the response body. The other two 502 tests both verify `res.body.message` — applying the same pattern here would make the test suite more consistent and catch a future regression on the error message format.

ClemRz:
Thanks for the review @Paul-AUB! All 3 items addressed:

1. ✅ Added ENOTFOUND test — covers the DNS resolution failure branch
2. ✅ Tightened regex to `/timed\s+out/i` — requires the `d` and at least one space
3. ✅ Added body assertion to the 500 test — now checks `res.body.message` like the 502 tests

PR #1594, opened 2026-05-16
feat(account): notify users by email on email/password changes

### LINKED ISSUE #1569: feat(auth): notify users via email when sensitive account info changes
## Summary

Send a notification email to users when sensitive account information is changed (email address, password), so they can detect unauthorized access to their account.

## Motivation

If a user's account is compromised, the attacker could change the email or password without the legitimate user ever being notified. This leaves the user with no way to detect the breach. Sending a notification to the **previous** email address (for email changes) or the **current** email address (for password changes) gives the user a chance to react and contact an administrator.

### Email address change
- When a user requests an email change, send a notification to the **old** email address
- The email should inform the user that an email change was requested
- It should **not** include the new email address (to avoid leaking sensitive info)
- It should instruct the user to contact an administrator if they did not initiate the change

### Password change
- When a user changes their password, send a notification to their current email address
- The email should inform the user that their password was changed
- It should **not** include any password-related information
- It should instruct the user to contact an administrator if they did not initiate the change

## Security Considerations

- Never include sensitive information (new email, password, tokens) in the notification emails
- The notification should be sent **before** or **simultaneously with** the change, not after a delay
- Consider rate-limiting to avoid notification spam in edge cases

## Acceptance Criteria

- [ ] Notification email sent to the old address when email is changed
- [ ] Notification email sent to the current address when password is changed
- [ ] Emails contain no sensitive information
- [ ] Emails include clear instructions to contact an admin if the change was not initiated by the user
- [ ] Email templates added for both scenarios

ClemRz:
## 🤔 What

- Add security notification emails when a user's email address or password is changed
- Notifications are sent to the **old** email address (email change) or **current** email address (password change) so the legitimate account owner is always alerted
- Closes #1569

## 🤷‍♂️ Why

If an attacker gains access to a user's account and changes the email or password, the legitimate owner has no way to know. These notifications provide an early warning so users can contact an administrator and recover their account.

## 🔍 How

- **New service**: `AccountNotificationService` with two fire-and-forget methods (`notifyEmailChanged`, `notifyPasswordChanged`). Errors are caught and logged — never block the HTTP response.
- **Controller changes**: `account/update.js` and `account/change-password.js` call the service after successful updates (no `await`).
- **Email templates**: Two new EJS templates (`emailChanged.ejs`, `passwordChanged.ejs`) using existing header/footer partials and the `__()` i18n helper.
- **i18n**: 5 new keys added to all 15 locale files (French translated, others use English fallback).
- **Security**: No sensitive data in templates (no new email, no password, no tokens, no clickable links). Pre-update email used for simultaneous email+password changes.

## 🧪 Testing

- `npm run test -- --grep "AccountNotificationService"` — service unit tests (stubs sendEmail, verifies params, locale fallback, error handling)
- `npm run test -- --grep "Account update - notifications"` — integration tests for update controller notifications
- `npm run test -- --grep "Change password - notifications"` — integration tests for change-password controller notifications
- `npm run test -- --grep "account-notification-templates"` — EJS template rendering tests

## 📸 Previews

Screenshot 2026-05-16 at 3 51 06 p m
Screenshot 2026-05-16 at 3 49 58 p m


Paul-AUB (dismissed):

1. **[api/controllers/v1/account/change-password.js:27–32]** Missing null guard on `caver` after `findOne`. If a user is deleted in the narrow window between `TCaver.updateOne()` and `TCaver.findOne()` (e.g., an admin deletes the account while the password reset completes), `caver` is `null` and the synchronous access `caver.mail` throws a `TypeError` that is not caught anywhere in the controller. The password will already have been updated, but the response becomes 500. Consider adding a guard:
   ```js
   if (caver) {
     AccountNotificationService.notifyPasswordChanged({ ... });
   }
   ```

2. **[scripts/sort-translation.js:7]** `LANG_DIR` is hardcoded to `packages/web-app/public/lang`, which does not exist in this API repository. Running the script without explicit file arguments always exits with "No translation files found." The script only works when locale paths are passed explicitly (e.g. `node scripts/sort-translation.js config/locales/en.json`). Consider either removing the default glob entirely or updating it to `config/locales` so the script is usable as a standalone tool for this repo.

3. **[api/controllers/v1/account/update.js:50]** `TCaver.findOne()` is now hoisted unconditionally and runs on every account update, even when only non-sensitive fields like `name`, `language`, or `sendNotificationByEmail` are being changed — and neither notification will be fired. This adds a primary-key lookup for all non-email/non-password updates. Consider evaluating whether a notification will be sent before making the fetch (e.g. check `req.body.mail` or `req.body.password` up front), or accepting the trade-off and leaving a comment explaining the intentional cost.

### Nitpicks (Optional)

4. **[config/locales/*.json]** The locale sort reformats all 15 files from tab indentation to 2-space JSON. This is a large whitespace-only change bundled with the feature that will affect `git blame` and `git log` for every existing translation key. Not blocking, but worth noting for reviewers who may be confused by the diff size.

5. **[views/emailTemplates/emailChanged.ejs:2–8, views/emailTemplates/passwordChanged.ejs:2–8]** The required-values comment block omits `emailTitle`, which is automatically injected by the `send-email` helper (line 1 of each template references it via `emailHeader`). Existing templates like `forgotPassword.ejs` have the same omission, so this is consistent, but a future maintainer copying one of these as a starting point for a new template would not know the variable is available.

ClemRz:

1. **Null guard on `caver`** — Added `if (caver)` guard in `change-password.js` before calling the notification service.

2. **`LANG_DIR` default path** — Updated to `config/locales` so the script works standalone in this repo.

3. **Unconditional `findOne` in `update.js`** — Added a comment explaining the intentional trade-off (single PK lookup, negligible cost, keeps the flow simple).

4. **Locale indentation** — Updated `sort-translation.js` to detect and preserve the existing file's indentation style. Locale files are now back to tab indentation; the diff only contains the new keys + alphabetical sorting.

5. **`emailTitle` in EJS comment blocks** — Added `emailTitle` to the "Required values" documentation in all 5 email templates (`emailChanged`, `passwordChanged`, `forgotPassword`, `notification`, `verifyEmail`).

PR #1595, opened 2026-05-18
fix(search): handle string array filters as multi-value exact match

ClemRz:
## 🤔 What

- Fix `buildFilter` in `SearchService` to correctly handle string array filter values in advanced search.
- Previously, passing a string array (e.g. `"massifs.name": ["Vercors (massif du)"]`) produced invalid Typesense filter syntax.

## 🤷‍♂️ Why

The `buildFilter` function treated all arrays as numeric ranges using `[min..max]` syntax. When the front-end sends a string array for facet filtering (e.g. filtering entrances by massif name), Typesense rejected the malformed filter, causing the request to fail.

## 🔍 How

The function now distinguishes between:
- **Two-number arrays** → numeric range filter: `field:[min..max]` (existing behavior preserved)
- **String arrays** (or any other array shape) → multi-value exact match: `field:=[\`value1\`,\`value2\`]`

This matches Typesense's expected syntax for filtering on `string[]` fields.

## 🧪 Testing

```bash
# String array filter (the fix)
curl -s -X POST http://localhost:1337/api/v1/advanced-search \
  -d '{"query":"","entity":"entrances","filter":{"massifs.name":["Géoparc du Chablais"]},"matchAllFields":true,"page":1,"size":5}'

# Numeric range filter (still works)
curl -s -X POST http://localhost:1337/api/v1/advanced-search \
  -d '{"query":"","entity":"entrances","filter":{"altitude":[1000,2000]},"matchAllFields":true,"page":1,"size":5}'
```

Unit tests added for both single and multiple string array values.

### DIFF
--- a/api/services/SearchService.js
+++ b/api/services/SearchService.js
@@ -39,11 +39,20 @@ function buildFilter(filter, isLogicalCompareAnd = true) {
 
       let vFmt = trimmed;
       let operator = ':'; // Partial equal
-      if (Array.isArray(trimmed))
-        vFmt = `[${trimmed.join('..')}]`; // Range
-      else if (typeof trimmed === 'boolean' || typeof trimmed === 'number')
+      if (Array.isArray(trimmed)) {
+        if (trimmed.length === 0) return []; // Skip empty arrays
+        const isNumericRange =
+          trimmed.length === 2 && trimmed.every((el) => typeof el === 'number');
+        if (isNumericRange) {
+          vFmt = `[${trimmed.join('..')}]`; // Numeric range
+        } else {
+          // Multi-value exact match (non-numeric-range arrays)
+          operator = ':=';
+          vFmt = `[${trimmed.map((el) => `\`${String(el).replace(/`/g, '')}\``).join(',')}]`;
+        }
+      } else if (typeof trimmed === 'boolean' || typeof trimmed === 'number')
         operator = ':='; // Exact equal
-      else vFmt = `\`${trimmed}\``;
+      else vFmt = `\`${trimmed.replace(/`/g, '')}\``;
       return [`${k}${operator}${vFmt}`];
     });
   return {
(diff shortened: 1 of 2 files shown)


Paul-AUB (dismissed):

1. **[api/services/SearchService.js:43-44]** `isNumericRange` requires exactly 2 elements, so a single-element numeric array like `[5]` falls into the `else` branch and produces `` field:=[`5`] `` (backtick-wrapped). Typesense treats backtick values as string comparisons, so this would silently fail or produce wrong results on numeric fields. If single-element numeric arrays are not a valid input, consider adding an explicit guard or throwing early; if they are, they need their own branch. No test currently covers this case.

2. **[api/services/SearchService.js:48-50]** String values inside the array are interpolated into backtick literals without escaping. A user-supplied value containing a backtick (e.g. `` foo`bar ``) would produce malformed Typesense filter syntax (`` field:=[`foo`bar`] ``) and silently cause the search request to fail. The same issue exists for single-string values on line 54, so this is pre-existing — but the new code extends it to arrays. Consider stripping or escaping backtick characters before interpolation.

3. **[test/integration/1_services/SearchService.test.js]** No test covers an empty array `[]`. It passes the `null`/`undefined` filter, enters the `Array.isArray` branch, fails `isNumericRange` (length is 0), and generates `field:=[]`. Typesense will likely reject this. A test — even if just documenting the current behavior — would make the contract explicit.

### Nitpicks (Optional)

4. **[api/services/SearchService.js:48]** The comment says "e.g. string arrays for facet filters" but the `else` branch also handles any other non-range array (e.g. 3-element numeric array). "Non-numeric-range arrays" or just "multi-value arrays" would be more accurate.

ClemRz:
Thanks for the thorough review @Paul-AUB! All four points addressed in the latest push:

1. **Single-element numeric array** — now handled consistently: `[5]` produces `field:=[`5`]` (multi-value exact match with one value), which is valid Typesense syntax. No silent failure.

2. **Backtick escaping** — backtick characters are now stripped from both single string values and array elements before interpolation, preventing malformed filter syntax.

3. **Empty array guard** — `[]` is now skipped entirely (produces no filter clause) instead of generating invalid `field:=[]`.

4. **Comment wording** — updated to "non-numeric-range arrays".

Tests added for all three edge cases (empty array, single-element numeric array, backtick in values).

PR #1597, opened 2026-05-18
fix(document): use native queries to clear junction rows on permanent delete

### LINKED ISSUE #1596: bug(api): [23503] tdocument - FK violation j_document_entrance on permanent delete
**Describe the bug**
When permanently deleting a document (`DELETE /api/v1/documents/:id?isPermanent=1`), the API returns a 500 error due to a foreign key constraint violation. The `j_document_entrance` junction rows referencing the document are not properly removed before the final hard delete (`TDocument.destroyOne`).

The `linkedEntitiesDeleteOrMerge('entrances')` function calls `TDocument.updateOne(documentId).set({ entrances: [] })` to clear the junction, but this fails silently because the document has already been soft-deleted earlier in the flow (via the first `TDocument.destroyOne` which sets `is_deleted = true`). Waterline's `updateOne` likely doesn't find the soft-deleted record, so the junction rows remain, causing the FK violation on the final hard delete.

**To Reproduce**
Based on the trace logs:
1. Authenticate as a Moderator/Admin user
2. Send `DELETE /api/v1/documents/233887?isPermanent=1`
3. The document must have associated entrances in `j_document_entrance`
4. The API returns 500 instead of successfully deleting the document

**Expected behavior**
The API should delete all `j_document_entrance` junction rows before attempting the hard delete of the document, or handle the FK constraint gracefully.

**Error Details**
- Occurrences: 1
- Time Range: 2026-05-16T13:25:54Z
- Representative Trace ID: `019e30f6-f86a-7721-8185-079743303fd4`
- Stack Trace:
```
OperationalError [AdapterError]: Unexpected error from database adapter: update or delete on table "t_document" violates foreign key constraint "j_document_entrance_t_document_fk" on table "j_document_entrance"
    at _afterCounting (waterline/lib/waterline/methods/destroy-one.js:167:17)
    ...
  code: '23503',
  detail: 'Key (id)=(233887) is still referenced from table "j_document_entrance".',
  constraint: 'j_document_entrance_t_document_fk'
```

**Additional context**
- Service: `api/controllers/v1/document/delete.js`
- Model: `tdocument`
- DB Error Code: `23503` (foreign_key_violation)
- Request: `DELETE /api/v1/documents/233887?isPermanent=1`
- User: Moderator (id: 32181)
- Root cause: The `linkedEntitiesDeleteOrMerge` function uses Waterline's collection setter (`set({ entrances: [] })`) on a soft-deleted record, which may not clear the junction table rows. A direct query to delete from `j_document_entrance` where `id_document = documentId` would be more reliable.

ClemRz:
## 🤔 What

- Fix FK constraint violation (`23503`) when permanently deleting a document that has junction table associations (e.g. `j_document_entrance`)
- Closes #1596

## 🤷‍♂️ Why

When a document is permanently deleted via `DELETE /api/v1/documents/:id?isPermanent=1`, the controller first soft-deletes the record (`TDocument.destroyOne` sets `is_deleted = true`), then attempts to clear junction rows using Waterline's collection setter (`TDocument.updateOne(id).set({ entrances: [] })`).

However, `updateOne` doesn't find the soft-deleted record, so the junction rows remain intact. The final hard delete then fails with:

```
foreign key constraint "j_document_entrance_t_document_fk" on table "j_document_entrance"
```

## 🔍 How

Replaced Waterline's collection setter with direct native `DELETE` queries against all 9 junction tables (`j_document_entrance`, `j_document_country`, `j_document_iso3166_2`, `j_document_language`, `j_document_massif`, `j_document_subject`, `j_document_grotto_author`, `j_document_caver_author`, `j_document_region`).

Native queries work regardless of the document's soft-delete state, ensuring junction rows are always cleaned up before the hard delete.

The merge-into logic (`addToCollection` on the target document) is unaffected since it operates on the target, not the deleted source.

## 🧪 Testing

Added a characterization test that reproduces the exact production scenario: a non-deleted document with entrance associations is permanently deleted in a single call (triggering soft-delete then hard-delete in sequence).

```
npm run test -- --grep "Document delete"
```

## 📸 Previews

N/A — backend-only fix.

### DIFF
--- a/api/controllers/v1/document/delete.js
+++ b/api/controllers/v1/document/delete.js
@@ -75,9 +75,27 @@ module.exports = async (req, res) => {
 
     await TNotification.destroy({ document: documentId });
 
+    // Note: `regions` is intentionally excluded — it uses a separate
+    // native query below because its junction table (j_document_region)
+    // does not follow the same column naming convention.
+    const junctionTableMap = {
+      countries: 'j_document_country',
+      entrances: 'j_document_entrance',
+      isoRegions: 'j_document_iso3166_2',
+      languages: 'j_document_language',
+      massifs: 'j_document_massif',
+      subjects: 'j_document_subject',
+      authorsGrotto: 'j_document_grotto_author',
+      authors: 'j_document_caver_author',
+    };
+
     // eslint-disable-next-line no-inner-declarations
     async function linkedEntitiesDeleteOrMerge(key) {
-      if (document[key].length === 0) return;
+      if (!Array.isArray(document[key])) {
+        throw new Error(
+          `Association "${key}" was not populated on document ${documentId}`
+        );
+      }
       if (shouldMergeInto) {
         const existingEntities = mergeIntoEntity[key].map((e) => e.id);
         const entitiesToAdd = document[key]
@@ -85,12 +103,20 @@ module.exports = async (req, res) => {
           .filter((e) => !existingEntities.includes(e));
         await TDocument.addToCollection(mergeIntoId, key, entitiesToAdd);
       }
-      await TDocument.updateOne(documentId).set({ [key]: [] });
+      // Use native query to delete junction rows directly.
+      // Waterline's collection setter (set({ [key]: [] })) fails silently
+      // on soft-deleted records, leaving FK references that block hard delete.
+      await sails.sendNativeQuery(
+        `DELETE FROM ${junctionTableMap[key]} WHERE id_document = $1`,
+        [documentId]
+      );
     }
 
     if (document.files.length > 0) {
       if (shouldMergeInto) {
-        TFile.update({ document: documentId }).set({ document: mergeIntoId });
+        await TFile.update({ document: documentId }).set({
+          document: mergeIntoId,
+        });
       } else {
         await Promise.all(
           document.files.map((e) => FileService.document.delete(e))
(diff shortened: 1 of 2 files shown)


Paul-AUB (dismissed):

1. **[delete.js:90–91]** The early-return guard `if (document[key].length === 0) return` skips the native `DELETE` when the populated collection appears empty. This is safe for documents fetched through `DocumentService.getPopulatedDocument`, but if a future code path calls this function on a document whose associations weren't fully populated, the junction rows would be silently skipped. Consider removing the early return or adding an assertion so a missed populate surfaces as an error rather than a no-op.

2. **[delete.js:78–87]** `junctionTableMap` has no entry for `regions` because that table is handled via a separate native query just below. This asymmetry is easy to miss when adding a new association type — a developer might add the new key only to `junctionTableMap` (expecting it to run through `linkedEntitiesDeleteOrMerge`) but forget the separately-handled pattern, or vice versa. Consider adding a short comment next to the `regions` block noting that it is intentionally excluded from the map.

### Nitpicks (Optional)

3. **[delete.js:110]** Pre-existing: `TFile.update({ document: documentId }).set({ document: mergeIntoId })` is missing `await`, so file re-association on merge runs fire-and-forget. This is not introduced by this PR but it was exposed during the read — worth a separate fix.

4. **[delete.test.js:173–178]** The multi-line comment block inside the test body describes the fix rather than the scenario. The scenario description is useful, but the fix rationale belongs in the commit/PR rather than the test. A single-line `// Reproduces #1596: non-deleted doc + junction rows + isPermanent in one request` would be enough.

ClemRz:
Thanks for the review @Paul-AUB! All four items addressed in the latest push:

1. **Early-return guard** — Replaced with an assertion that throws if the association isn't populated (surfaces bugs instead of silently skipping).
2. **Comment for `regions` exclusion** — Added a note above `junctionTableMap` explaining why `regions` is handled separately.
3. **Missing `await` on `TFile.update`** — Fixed the fire-and-forget call.
4. **Verbose test comment** — Condensed to a single line.

PR #1603, opened 2026-05-21
fix(document): normalize massifs to IDs in validate flow

### LINKED ISSUE #1601: bug(api): [E_INVALID_VALUES_TO_SET] tdocument - massifs objects passed to updateOne on validate
**Describe the bug**
When validating documents via `PUT /api/v1/documents/validate`, the API crashes with a 500 error because the `massifs` association on the document contains full objects (`{ id, name, language }`) instead of plain IDs. Waterline's `updateOne` rejects the populated association data when it's passed through to the update operation.

The root cause is that the document record is fetched with populated `massifs`, and then the full populated object is passed back into `updateOne` without stripping association data down to IDs.

**To Reproduce**
Based on the trace logs:
1. Authenticate as a Moderator
2. Send `PUT /api/v1/documents/validate` with a body containing document IDs to validate (e.g., `{"documents": [{"id": 233899, "isValidated": "true", "validationComment": ""}]}`)
3. The document being validated must have at least one massif association (e.g., massif id 41 "Parmelan")
4. The API returns 500 instead of successfully validating the document

**Expected behavior**
The API should handle this case gracefully without returning a 500 error. Association fields like `massifs` should either be stripped before the update or mapped to an array of IDs.

**Error Details**
- Error Type: `OperationalError [UsageError]`
- Error Code: `E_INVALID_VALUES_TO_SET`
- Occurrences: 6
- Time Range: 2026-05-21T14:59:45Z to 2026-05-21T21:52:41Z
- Representative Trace ID: `019e4c86-c467-72aa-9dfd-48d5baf8a0a0`
- Stack Trace:
```
Details:
  Could not use specified `massifs`. If specified, expected `massifs` to be
  an array of ids (representing the records to associate). But instead, got:
  [ { id: '41', name: 'Parmelan (massif du)', language: 'fra' } ]

    at flaverr (node_modules/flaverr/index.js:94:15)
    at _.extend._WLModel [as _handleExec] (node_modules/waterline/lib/waterline/methods/update-one.js:138:15)
    at Deferred.exec (node_modules/parley/lib/private/Deferred.js:286:10)
```

**Additional context**
- Endpoint: `PUT /api/v1/documents/validate`
- Model: `tdocument`
- Waterline Error Code: `E_INVALID_VALUES_TO_SET`
- The same user triggered this 6 times in quick succession, likely retrying the same batch validation

ClemRz:
## 🤔 What

- Fix `E_INVALID_VALUES_TO_SET` error on `PUT /api/v1/documents/validate` when validating documents that have `massifs` stored as objects in `modifiedDocJson`
- Closes #1601

## 🤷‍♂️ Why

When a document is edited, `getConvertedDataFromClient` stores `massifs` directly from the request body without extracting IDs (unlike `authors`, `authorsGrotto`, `subjects` which all use `.map(x => x.id)`). If the front-end sends massifs as objects (`[{ id: 1, name: "..." }]`), these objects end up in `modifiedDocJson.documentData.massifs`. When a moderator later validates the document, `validateAndUpdateDocument` spreads `documentData` into `TDocument.updateOne().set()`, and Waterline rejects the objects with `E_INVALID_VALUES_TO_SET`.

## 🔍 How

Two-layer fix:

1. **`DocumentService.getConvertedDataFromClient`** — normalize `massifs` to plain IDs using `.map((m) => m.id ?? m)`, consistent with how other collection associations are handled. This prevents future `modifiedDocJson` records from storing objects.

2. **`multiple-validate.js` — `normalizeCollectionIds` helper** — before spreading `documentData` into `.set()`, normalize all collection association fields (`massifs`, `authors`, `authorsGrotto`, `subjects`, `languages`, `isoRegions`, `countries`) to plain IDs. This handles existing `modifiedDocJson` records already stored in the database with object values.

## 🧪 Testing

- Added test: "should validate document with modifiedDocJson containing massifs as objects"
- Full test suite passes (2305 tests, 0 failures)

### DIFF
--- a/api/controllers/v1/document/multiple-validate.js
+++ b/api/controllers/v1/document/multiple-validate.js
@@ -17,6 +17,28 @@ async function markDocumentValidated(
   });
 }
 
+// modifiedDocJson may pre-date this fix and contain populated objects instead of plain IDs.
+function normalizeCollectionIds(documentData) {
+  const collectionFields = [
+    'massifs',
+    'authors',
+    'authorsGrotto',
+    'subjects',
+    'languages',
+    'isoRegions',
+    'countries',
+  ];
+  const normalized = { ...documentData };
+  for (const field of collectionFields) {
+    if (Array.isArray(normalized[field])) {
+      normalized[field] = normalized[field].map((item) =>
+        typeof item === 'object' && item !== null ? (item.id ?? item) : item
+      );
+    }
+  }
+  return normalized;
+}
+
 async function validateAndUpdateDocument(
   document,
   validationComment,
@@ -31,6 +53,8 @@ async function validateAndUpdateDocument(
     newFiles,
   } = document.modifiedDocJson;
 
+  const cleanedDocumentData = normalizeCollectionIds(documentData);
+
   await sails.getDatastore().transaction(async (db) => {
     // Update associated data not handled by TDocument manually
     // Updated before the TDocument update so the last_change_document DB trigger will fetch the last updated name
@@ -40,7 +64,7 @@ async function validateAndUpdateDocument(
 
     await TDocument.updateOne(document.id)
       .set({
-        ...documentData,
+        ...cleanedDocumentData,
         modifiedDocJson: null,
         dateReviewed: new Date(),
         reviewer: reviewerId,
--- a/api/services/DocumentService.js
+++ b/api/services/DocumentService.js
@@ -93,7 +93,10 @@ module.exports = {
   getConvertedDataFromClient: async (body) => {
     // Massif will be deleted in the future (a document can be about many massifs and a massif can be the subject of many documents): use massifs
     const massif = body.massif?.id;
-    const massifs = [...(body.massifs ?? []), ...(massif ? [massif] : [])];
+    const massifs = [
+      ...(body.massifs ?? []).map((m) => m.id ?? m),
+      ...(massif ? [massif] : []),
+    ];
 
     let optionFound;
     // eslint-disable-next-line no-param-reassign
--- a/test/integration/4_routes/Documents/multiple-validate.test.js
+++ b/test/integration/4_routes/Documents/multiple-validate.test.js
@@ -161,6 +161,52 @@ describe('Document multiple-validate', () => {
       should(updated.isValidated).be.true();
       should(updated.modifiedDocJson).be.null();
     });
+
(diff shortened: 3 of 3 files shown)


Paul-AUB (dismissed):
Solid two-layer fix with a good rationale. The root cause is clearly identified and both layers (prevent future bad data + heal existing records) are correct. The integration test closes the loop nicely.

1. **[multiple-validate.js:37]** `normalizeCollectionIds` returns `item.id` with no fallback: if an object somehow lacks an `id` key (e.g., a region object stored as `{ iso: 'FR', name: 'France' }`), `item.id` resolves to `undefined`, silently dropping that association entry. The sibling fix in `DocumentService.getConvertedDataFromClient` uses `m.id ?? m` to guard against this. Consider applying the same defensive pattern here: `item.id ?? item`. (In practice the current shapes are safe, but the inconsistency is a trap for future fields added to `collectionFields`.)

2. **[multiple-validate.test.js:183]** The test constructs `modifiedDocJson` with `massifs: [{ id: 1, ... }]` and then asserts `containDeep([1])`, coupling the assertion to fixture massif ID 1. If the seed data ever changes, the test will fail in a non-obvious way. Consider creating a `TMassif` record inside the test and using its returned ID, making the test fully self-contained.

### Nitpicks (Optional)

3. **[multiple-validate.js:20-23]** The three-line block comment above `normalizeCollectionIds` describes *what* the function does (already expressed by the name, the `collectionFields` list, and the inline logic). A single-line comment calling out the non-obvious WHY ("modifiedDocJson may pre-date this fix and contain populated objects") would be more concise and consistent with the project's light-comment convention.

4. **[multiple-validate.test.js:199-202]** The new test verifies the massif association is set but doesn't assert that `descriptionData` was applied (e.g., checking `TDescription.findOne` for the updated title/body). Adding that assertion would give the test broader coverage of `validateAndUpdateDocument`'s full effect.

ClemRz:
Thanks for the review @Paul-AUB! All 4 items addressed:

1. **`normalizeCollectionIds` fallback** — now uses `(item.id ?? item)` consistent with `getConvertedDataFromClient`.
2. **Test self-containment** — creates its own `TMassif` record instead of relying on fixture ID 1.
3. **Comment conciseness** — replaced the 3-line block comment with a single-line WHY comment.
4. **`descriptionData` assertion** — test now verifies `TDescription` was updated with the expected title and body.

PR #1604, opened 2026-05-22
fix(document): use TMassif and normalize IDs in populateJSON

### LINKED ISSUE #1602: bug(api): [E_INVALID_CRITERIA] tdocument - massifs objects in where clause on find
**Describe the bug**
When fetching a single document via `GET /api/v1/documents/:id?requireUpdate=true`, the API crashes with a 500 error in `DocumentService.populateJSON` (line 402). The `massifs` field on the document contains full objects (`{ id: '41', name: 'Parmelan (massif du)', language: 'fra' }`) instead of plain IDs, and when `populateJSON` tries to use these as criteria in a `where: { id: { in: [...] } }` clause, Waterline rejects the dictionary values.

This is likely the read-side manifestation of the same root cause as the validate bug: a document's `massifs` association was corrupted or stored as objects instead of IDs, and now any code path that reads and re-queries those massifs fails.

**To Reproduce**
Based on the trace logs:
1. Authenticate as any user
2. Send `GET /api/v1/documents/233875?requireUpdate=true`
3. The document has a `massifs` field containing objects instead of plain IDs
4. `DocumentService.populateJSON` at line 402 attempts to query massifs using these objects as IDs
5. The API returns 500

**Expected behavior**
The API should handle this case gracefully without returning a 500 error. `populateJSON` should either normalize the massifs data to IDs before querying, or the underlying data should never contain objects in an ID field.

**Error Details**
- Error Type: `OperationalError [UsageError]`
- Error Code: `E_INVALID_CRITERIA`
- Occurrences: 2
- Time Range: 2026-05-21T18:47:23Z to 2026-05-21T18:48:00Z
- Representative Trace ID: `019e4bdd-b15a-7140-aeb5-fb06b0a044a8`
- Stack Trace:
```
OperationalError [UsageError]: Invalid criteria.
Details:
  Could not use the provided `where` clause. Could not filter by `id`:
  Invalid item within `in` modifier array. Does not match the declared
  data type of the corresponding attribute.
  Specified value (a dictionary: { id: '41', name: 'Parmelan (massif du)',
  language: 'fra' }) doesn't match the expected type: 'string'

    at Object.populateJSON (api/services/DocumentService.js:402:44)
    at async getModifiedDocumentData (api/controllers/v1/document/find.js:23:24)
    at async module.exports [as v1/document/find] (api/controllers/v1/document/find.js:65:16)
```

**Additional context**
- Service: `api/services/DocumentService.js` (line 402)
- Controller: `api/controllers/v1/document/find.js` (lines 23, 65)
- Model: `tdocument`
- Waterline Error Code: `E_INVALID_CRITERIA`
- Both occurrences were for document ID 233875, which appears to have corrupted massif association data
- This bug is related to the validate bug (#1601) — both stem from massifs being stored/returned as objects instead of IDs

ClemRz:
## 🤔 What

- Fix `E_INVALID_CRITERIA` error on `GET /api/v1/documents/:id?requireUpdate=true` when `modifiedDocJson.documentData` contains `massifs` (or other collection associations) as objects
- Fix wrong model usage: `populateJSON` was querying `TCountry` instead of `TMassif` for the massifs field
- Closes #1602

## 🤷‍♂️ Why

`DocumentService.populateJSON` is called when a moderator fetches a document with pending modifications (`requireUpdate=true`). It receives `documentData` from `modifiedDocJson` and queries related tables to build a populated response.

Two bugs existed:
1. **Wrong model**: `doc.massifs = await TCountry.find({ id: massifs })` — should be `TMassif`
2. **Objects instead of IDs**: `massifs` (and potentially other collection fields) stored in `modifiedDocJson` may contain objects like `[{ id: 1, name: "..." }]` rather than plain IDs. Passing these to `{ id: massifs }` causes Waterline to reject the criteria with `E_INVALID_CRITERIA`.

## 🔍 How

- Added a `toIds` helper inside `populateJSON` that normalizes any collection field to plain IDs: `(arr) => arr.map(item => item.id ?? item)`
- Applied `toIds` to all collection association queries (`authors`, `authorsGrotto`, `subjects`, `languages`, `countries`, `isoRegions`, `entrances`, `massifs`)
- Changed `TCountry.find` to `TMassif.find` for the massifs field

## 🧪 Testing

- Added test: "should return modified document with massifs as objects in modifiedDocJson"
- Full test suite passes (2306 tests, 0 failures)

### DIFF
--- a/api/services/DocumentService.js
+++ b/api/services/DocumentService.js
@@ -9,6 +9,10 @@ const {
   distantFileDownload,
 } = require('../utils/csvHelper');
 
+// Normalize a collection value to a plain ID (handles both raw IDs and objects).
+const normalizeToId = (item) =>
+  item != null && typeof item === 'object' ? (item.id ?? item) : item;
+
 module.exports = {
   async deleteInSearch(documentId) {
     await SearchService.deleteDocument('documents', documentId);
@@ -373,15 +377,18 @@ module.exports = {
       ...otherSimpleData
     } = documentData;
 
+    // Normalize collections and join the tables
+    const toIds = (arr) => arr.map(normalizeToId);
+
     // Join the tables
     const doc = { ...otherSimpleData, id: documentId };
     doc.identifierType = identifierType
       ? await TIdentifierType.findOne(identifierType)
       : null;
     doc.author = author ? await TCaver.findOne(author) : null;
-    doc.authors = authors ? await TCaver.find({ id: authors }) : [];
+    doc.authors = authors ? await TCaver.find({ id: toIds(authors) }) : [];
     doc.authorsGrotto = authorsGrotto
-      ? await TGrotto.find({ id: authorsGrotto })
+      ? await TGrotto.find({ id: toIds(authorsGrotto) })
       : [];
     doc.reviewer = reviewer ? await TCaver.findOne(reviewer) : null;
     doc.editor = editor ? await TGrotto.findOne(editor) : null;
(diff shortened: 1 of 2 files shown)


Paul-AUB (dismissed):
Clean fix for two distinct bugs in the same function. The `TCountry` → `TMassif` correction is clearly right (it was querying the wrong table entirely), and the `toIds` normalization is the correct mirror of what PR #1603 applied on the write path.

1. **[DocumentService.js:377 vs multiple-validate.js]** This PR introduces `toIds` (`item.id ?? item`) while PR #1603 introduced `normalizeCollectionIds` with the logic `typeof item === 'object' && item !== null ? item.id : item`. Both normalize collection fields to plain IDs but differ in edge-case handling: `toIds` falls back to `item` when `item.id` is `undefined`, whereas `normalizeCollectionIds` returns `undefined` in that case. Consider extracting a single shared helper (e.g., `normalizeToId` in `DocumentService` or a dedicated utils module) so the two code paths can't silently diverge on future edge cases.

2. **[find.test.js:86]** `should(res.body.massifs).be.an.Array()` passes even when the response is `massifs: []` — which would happen if massif ID 1 is absent from the fixture data, meaning the normalization ran but the DB query returned nothing. A stronger assertion (`should(res.body.massifs).not.be.empty()` or `containDeep([{ id: 1 }])`) would prove that `TMassif.find` actually returned a populated record, not just that the field exists.

### Nitpicks (Optional)

3. **[DocumentService.js:377]** The `?? []` inside `toIds` — `(arr ?? []).map(...)` — is redundant: every caller already guards with `field ? toIds(field) : []`, so `arr` is always truthy when `toIds` is called. It's harmless, but it suggests `toIds` handles the null case when it doesn't need to, which could mislead future readers about the intended ownership of the null-guard.

4. **[find.test.js:88-91]** The new test uses inline `try/finally` for cleanup while all surrounding tests in the same `describe` block rely on the outer `after` hook. Both patterns are valid, but the inconsistency is mild friction for future contributors. If the inline approach is preferred for self-contained tests (reasonable choice), a brief comment would help signal that it's intentional.

ClemRz:
Thanks for the review @Paul-AUB! All 4 items addressed:

1. **Shared helper** — extracted `normalizeToId` as a module-level function in `DocumentService.js` (exported for reuse). The local `toIds` now delegates to it. PR #1603's `normalizeCollectionIds` can adopt it after merge to avoid divergence.
2. **Stronger assertion** — test now asserts `not.be.empty()` and `containDeep([{ id: 1 }])` to prove `TMassif.find` actually returned a populated record.
3. **Removed redundant `?? []`** — callers already guard with `field ? toIds(field) : []`, so the null-guard inside `toIds` was misleading.
4. **Inline cleanup comment** — added a comment explaining the `try/finally` pattern is intentional for self-contained test data.

PR #1605, opened 2026-05-22
fix(relevance): handle null-relevance entities in sort and move

Paul-AUB:
## 🤔 What

- Fix `RelevanceService.computeNextRelevance` returning `1` when null-relevance entities exist in scope
- Fix `RelevanceService.moveRelevance` failing silently when the target entity has `null` relevance
- Fix `moveRelevance` potentially swapping with a `null`-relevance neighbor
- Wrap the relevance swap in a database transaction to prevent partial updates

## 🤷‍♂️ Why

**computeNextRelevance bug**: PostgreSQL sorts `NULL` first with `DESC` ordering. Without filtering, `results[0].relevance` was `null`, and `null + 1 = 1` in JavaScript — colliding with existing items already at `relevance = 1`.

**moveRelevance null-target bug**: If an entity had `relevance = null`, the neighbor query (`WHERE relevance > null`) always returned zero rows in PostgreSQL (comparing with `NULL` yields `UNKNOWN`), causing a misleading 400 "cannot be moved further in that direction".

**moveRelevance null-neighbor bug**: Without filtering null-relevance neighbors, a swap could assign `null` to the target entity, breaking its sort position permanently.

## 🔍 How

- Added `relevance: { '!=': null }` to the `find` query in `computeNextRelevance` so only non-null values contribute to the max.
- Added a guard in `moveRelevance`: if `target.relevance` is `null`, assign `computeNextRelevance()` and return `{ moved, swapped: null }` without performing a swap — the entity is placed at the end of the ranked list.
- Added `relevance: { '!=': null }` to the neighbor query so null-relevance entities are never swap candidates.
- Wrapped the two `updateOne` calls in `sails.getDatastore().transaction()` to make the swap atomic.

## 🧪 Testing

New integration tests in `RelevanceService.test.js`:

- `should return 1 when all existing entities have null relevance`
- `should ignore null-relevance entities and use max of non-null values`
- `should assign relevance and return swapped: null for a null-relevance entity`
- `should skip null-relevance neighbors and swap with the nearest ranked one`

```bash
npm test
```

## 📸 Previews

_N/A — service-layer fix._

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -122,14 +122,8 @@ coverage
 
 docker/esdata/*
 !docker/esdata/for-docker-compose
-.amazonq
-.kiro
 *.iml
 build-info.json
-
-# Kiro IDE
-.kiro
-
 
 # Temporary test output
 *.tmp
@@ -139,3 +133,9 @@ test/.snapshot-stamp
 
 # Error triage agent timestamp
 .last-error-triage
+
+# AI
+CLAUDE.md
+.amazonq
+.kiro
+.claude
--- a/api/services/RelevanceService.js
+++ b/api/services/RelevanceService.js
@@ -141,7 +141,7 @@ module.exports = {
       ...parentScope,
       isDeleted: false,
       id: { '!=': entityId },
-      relevance: { '!=': null },
+      // '>' / '<' comparisons implicitly exclude NULL in PostgreSQL (NULL yields UNKNOWN, not TRUE).
     };
 
     let sort;
--- a/api/services/RelevanceService.js
+++ b/api/services/RelevanceService.js
@@ -2,7 +2,7 @@
  * ENTITY_CONFIG maps each relevance entity type to its Waterline model
  * and the parent fields that define its scope.
  *
- * Models (TLocation, TDescription, etc.) are Sails globals — no imports needed.
+ * Models (TLocation, TDescription, etc.) are Sails globals - no imports needed.
  */
 const ENTITY_CONFIG = {
   location: {
@@ -127,6 +127,9 @@ module.exports = {
 
     // Unranked entity: assign a position first, no swap needed.
     if (target.relevance == null) {
+      sails.log.warn(
+        `moveRelevance: ${label} id=${entityId} has null relevance â€” assigning from computeNextRelevance`
+      );
       const nextRelevance = await this.computeNextRelevance(
         entityType,
         parentScope
@@ -137,11 +140,11 @@ module.exports = {
       return { moved, swapped: null };
     }
 
+    // '>' / '<' comparisons implicitly exclude NULL in PostgreSQL (NULL yields UNKNOWN, not TRUE).
     const where = {
       ...parentScope,
       isDeleted: false,
       id: { '!=': entityId },
-      // '>' / '<' comparisons implicitly exclude NULL in PostgreSQL (NULL yields UNKNOWN, not TRUE).
     };
 
     let sort;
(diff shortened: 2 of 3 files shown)


ClemRz (dismissed):

- **[api/services/RelevanceService.js:144]** The `relevance: { '!=': null }` filter on line 144 is immediately overwritten by `where.relevance = { '>': target.relevance }` (or `'<'`) on lines 149/151. The null-exclusion intent is lost — it's dead code. The behavior is still correct because SQL comparisons (`>`, `<`) naturally exclude NULLs (they evaluate to UNKNOWN, not TRUE), but the code is misleading. A reader would assume the `!= null` guard is active when it isn't.

  Consider removing the dead filter and adding a comment explaining why null exclusion is unnecessary here:
  ```suggestion
  const where = {
    ...parentScope,
    isDeleted: false,
    id: { '!=': entityId },
    // No explicit null filter needed: '>' / '<' comparisons
    // implicitly exclude NULL in PostgreSQL (NULL comparisons yield UNKNOWN).
  };
  ```

### Nitpicks (Optional)

- **[.gitignore]** The `.gitignore` reorganization (consolidating AI tool entries) is unrelated to the relevance fix. Consider splitting it into a separate commit to keep the fix commit focused — makes `git bisect` and changelog generation cleaner.

- **[test/integration/1_services/RelevanceService.test.js:175-176]** The two `sendNativeQuery` INSERT calls could be combined into a single multi-row INSERT for brevity:
  ```suggestion
  await sails.sendNativeQuery(
    `INSERT INTO t_comment (id_author, date_inscription, relevance, title, body, id_entrance, id_language, is_deleted)
     VALUES (1, NOW(), NULL, 'Null relevance test', 'Should be ignored', $1, 'fra', false),
            (1, NOW(), NULL, 'Null relevance test', 'Should be ignored', $1, 'fra', false)`,
    [entranceId]
  );
  ```

ClemRz (dismissed):

- **[api/services/RelevanceService.js:127-128]** The `== null` check on `target.relevance` catches both `null` and `undefined`. Since the model defines `allowNull: false`, null values in the DB are legacy/anomalous data. Consider adding a brief log statement (`sails.log.warn(...)`) when this branch is hit, so the team has visibility into how often unranked entities are encountered in production. This would help inform whether a data migration to backfill null relevance values is warranted.

- **[test/integration/1_services/RelevanceService.test.js:486-598]** The two new `moveRelevance` tests (`null-relevance entity` and `skip null-relevance neighbors`) both use `entranceId` values (6007, 6008) that are cleaned up via `DELETE FROM t_comment WHERE id_entrance = $1`. This cleanup is broader than necessary — it deletes all comments for that entrance, not just the ones created by the test. If fixtures ever seed comments for those entrance IDs, the cleanup would destroy fixture data. Consider tracking created IDs explicitly (as done in other tests in this file) or using entrance IDs that are guaranteed unused by fixtures (e.g., the 9000+ range used by the property tests).

### Nitpicks (Optional)

- **[.gitignore]** The `.gitignore` reorganization (consolidating AI tool entries under a single `# AI` section, adding `.claude` and `CLAUDE.md`) is unrelated to the relevance fix. Keeping unrelated changes in a separate commit makes `git bisect` and changelog generation cleaner. That said, this is minor housekeeping and doesn't block merging.

- **[api/services/RelevanceService.js:143]** The inline comment explaining why `>` / `<` implicitly exclude NULL is helpful. Consider moving it above the `where` object declaration (near line 140) rather than inside the object literal, so it reads as a design note rather than a trailing comment on a property that doesn't exist yet at that point.

ClemRz (approved):

PR #1609, opened 2026-05-23
feat(auth): harden admin account security with TOTP MFA and brute-force protection

### LINKED ISSUE #1608: Harden administrator account security (token TTL, MFA, login protection)

Administrator accounts have elevated privileges (bypass general and DELETE rate limiting, access to critical operations) but do not benefit from stronger security measures compared to regular user accounts. The authentication token has a 90-day lifetime, identical for all roles, which increases the exploitation window in case of a token leak.

For administrator accounts only:

1. **Reduce token validity to 10 days** — Currently `authTokenTTL` is set to 90 days for all roles (`config/custom.js`). Introduce an admin-specific TTL (e.g. `adminAuthTokenTTL: 10 * 24 * 60 * 60`) and apply it when issuing the token in the login controller.

2. **Enable MFA via authenticator app (TOTP)** — Add multi-factor authentication support via TOTP (compatible with Google Authenticator, Aegis, FreeOTP, etc.) for administrator accounts. This involves:
   - Generating and storing a TOTP secret per administrator
   - Enrollment endpoint (QR code + secret)
   - TOTP code verification during login
   - Recovery/backup codes management

3. **Stricter login attempt limiting and/or IP-based protection for admin access** — An `authRateLimit` already exists (10 attempts / 15 min / IP in `config/rateLimit/rateLimiter.js`), applied uniformly to all users. Evaluate whether additional measures are needed for admins:
   - Stricter rate limit on the admin login endpoint (e.g. 5 attempts / 15 min)
   - Temporary account lockout after N consecutive failures
   - Email notification on suspicious login attempts
   - IP whitelist restriction — optional, can be operationally constraining

- Use an external identity provider (Auth0, Keycloak) for admins only — more complex to integrate but delegates MFA management.
- Enforce MFA for all users — too constraining for occasional contributors.
- Reduce the TTL for all roles — would negatively impact UX for regular users.

**Additional context**

- Current token TTL: 90 days (`config/custom.js` → `authTokenTTL`)
- Existing auth rate limiting: 10 req / 15 min / IP (`config/rateLimit/rateLimiter.js` → `authRateLimit`)
- Administrators are currently exempt from general and DELETE rate limiting (but NOT from auth rate limiting, which is correct)
- Suggested TOTP library: `otplib` (well-maintained, Node.js compatible)

ClemRz:
## 🤔 What

Harden the security posture of Administrator accounts with three complementary layers of protection:

- **Mandatory TOTP-based MFA** — admins must enroll in multi-factor authentication and provide a valid TOTP code on every login
- **Stricter brute-force protection** — reduced rate limits (5 req/15min/IP for admin logins), account ban after 5 consecutive failures, email notifications on suspicious activity
- **Reduced token lifetime** — admin tokens expire in 10 days instead of 90 days
- **Token revocation on password change** — all existing admin tokens are invalidated when the password is changed

Closes #1608

## 🤷‍♂️ Why

Administrator accounts have elevated privileges and represent a high-value target. A compromised admin account could modify or delete any data in the system. These measures ensure that stolen credentials alone are insufficient to gain admin access, and that brute-force attacks are detected and blocked quickly.

### New endpoints
- `POST /api/v1/mfa/enroll` — start MFA enrollment, returns TOTP secret + otpauth URI
- `POST /api/v1/mfa/verify` — confirm enrollment with a valid TOTP code, issues full auth token
- `POST /api/v1/mfa/reset` — self-service MFA reset (requires password re-entry)

### Modified endpoints
- `POST /api/v1/login` — new `totpCode` field, new 401 statuses (MfaEnrollmentRequired, MfaRequired, InvalidTotpCode, TotpAlreadyUsed)
- `PATCH /api/v1/account/password` — revokes all admin tokens on password change

### New services
- `MfaService` — TOTP generation, AES-256-GCM encryption, verification, enrollment lifecycle
- `AdminLoginProtectionService` — failure tracking, account banning, email notifications

### Security hardening (from code review)
- Constant-time delay on failed logins to prevent timing-based admin email enumeration
- Atomic DB update for TOTP replay prevention (optimistic locking)
- Version-prefixed encryption format (`v1:iv:ciphertext:authTag`) for future key rotation
- Password re-entry required for MFA reset (prevents stolen-token abuse)
- Production guard against `mfaDevSecret` misconfiguration
- IPv6-safe rate limit key generation via `ipKeyGenerator`
- Isolated otplib instance (`authenticator.clone()`) to avoid global state mutation

### Infrastructure
- New dependency: `otplib@^12.0.1`
- New env var required in production: `MFA_ENCRYPTION_KEY` (32-byte hex string)
- SQL migration: `sql/20250523_admin_mfa.sql` (8 new columns on `t_caver`)
- Email templates: `suspiciousLogin.ejs`, `accountBanned.ejs`
- i18n: all 8 new strings translated in 15 locales

## 🧪 Testing

All tests pass (2405 passing, 0 failing). New test coverage includes:

- MfaService unit tests (encryption round-trip, TOTP verification, replay detection, enrollment lifecycle)
- AdminLoginProtectionService tests (failure tracking, banning, email notifications with cooldown)
- Route integration tests for login MFA flow, enrollment, verification, reset, rate limiting, ban behavior, token revocation
- Policy tests for `mfaEnrollmentAuth`

To test locally:
1. `npm run dev:clean` (rebuilds DB with new columns)
2. `npm run dev` — the dev encryption key is already set in `config/env/development.js`, no env var needed
3. Login as admin — will get `MfaEnrollmentRequired` response with enrollment token
4. Use enrollment token to call `/mfa/enroll`, scan QR code (or use the dev secret `JBSWY3DPEHPK3PXP` directly), then `/mfa/verify`

## 📸 Previews

N/A (API-only changes)


Paul-AUB (changes requested):
This is a solid and well-thought-out security implementation. The crypto choices (AES-256-GCM, version-prefixed ciphertext), the optimistic-lock for TOTP replay prevention, and the constant-time delay for timing-attack mitigation are all excellent. Two security issues need to be resolved before merging.

### Issues (Must Fix)

1. **[api/controllers/v1/account/change-password.js:96–110]** **Admin token revocation missing from the password-reset-token flow.** The PR states "all existing admin tokens are invalidated when the password is changed," but the revocation at lines 42–49 only runs inside the `if (req.token)` branch (authenticated password update). The unauthenticated branch — where an admin follows a forgot-password link — reaches line 96 and updates the password without ever calling `BlacklistService.revoke`. A credential-leak scenario where an admin resets their password via email would leave all existing JWT tokens valid for up to 10 days. The same `isAdmin` + `BlacklistService.revoke` block should be added in `verifyTokenCallback`, using `decodedToken.userId` to identify the account.

2. **[config/http.js:27–32 + config/rateLimit/rateLimiter.js:88–93]** **`/mfa/verify` is effectively unrate-limited for enrollment token holders.** The MFA endpoints are not included in `AUTH_PATHS`, so the `authRateLimit` middleware (10 req/15 min) does not apply to them. More critically, enrollment tokens are issued with the admin's full `groups` array (login.js:101), including `ADMINISTRATOR`. When `parseAuthToken` sets `req.token`, the `generalRateLimit` skip condition fires and the request is exempted entirely. An attacker who obtains a valid 10-minute enrollment token therefore has an uncapped window to brute-force the TOTP code on `/mfa/verify`. The simplest fix is to add `/api/v1/mfa/verify` (and defensively `/api/v1/mfa/enroll`) to `AUTH_PATHS` in `config/http.js` — the existing `authRateLimit` does not skip admins and would cap attempts at 10 per 15 min per IP.

3. **[api/controllers/v1/auth/login.js:156]** **No null guard before `decryptSecret`.** `MfaService.decryptSecret(caver.totpSecret)` is called when `caver.mfaEnabled` is true, but there is no check that `totpSecret` is non-null. A data-integrity issue (e.g., a partially-applied migration or a manual DB edit) would cause `encryptedSecret.split(':')` to throw a `TypeError`, surfacing as an unhandled 500. Adding `if (!caver.totpSecret) { return res.serverError('MFA secret unavailable'); }` before line 156 would make the failure mode explicit and controllable.

4. **[api/services/MfaService.js:86–91]** **Dead "legacy format" branch in `decryptSecret`.** The `else` branch handles ciphertexts without a version prefix ("legacy format"), but since MFA is a brand-new feature there are no pre-existing secrets in the database. This branch can never be reached and risks masking a future regression where secrets are accidentally stored without the version prefix. Consider removing the else branch now; migrating to a new key or algorithm is the right moment to add explicit upgrade logic.

5. **[config/http.js:177–211]** **`adminAuthRateLimit` creates a timing oracle for admin email enumeration.** When the 5-request admin rate limit is exhausted, the middleware returns a 429 before the login controller's `constantDelay()` executes. This makes rate-limited admin-targeted responses measurably faster than non-admin responses, allowing an attacker to distinguish admin email addresses after 5 probes per IP window. The existing `constantDelay` does not help here because the middleware short-circuits before reaching the controller. This is likely an accepted trade-off given the overall threat model, but it should be explicitly documented in the middleware comment so future maintainers understand the boundary.

### Nitpicks (Optional)

6. **[config/http.js:38]** **`AUTH_BODY_LIMIT` (1 KB) does not cover MFA endpoints.** This is a minor inconsistency — login requests now include a `totpCode` field and still fit easily inside 1 KB, and the MFA endpoints accept only a small payload — but adding `/api/v1/mfa/verify` and `/api/v1/mfa/reset` to `AUTH_PATHS` (which would also fix issue 2) would pick up the body-size guard for free.

7. **[config/custom.js:113–120]** **Mixed time units between adjacent config properties.** `mfaEnrollmentTokenTTL` is in seconds (600) while `suspiciousActivityEmailCooldown` is in milliseconds (900000). Both have inline comments, so this is not a bug, but it is a latent footgun for anyone adding a new setting in this block. Consider grouping the TTL-in-seconds and the cooldown-in-milliseconds properties with a brief section comment calling out the unit difference.

ClemRz:
Thanks for the thorough review! All items addressed:

**Issues (Must Fix):**
1. **Admin token revocation in reset-token flow** — Added `BlacklistService.revoke()` in the `verifyTokenCallback` branch, using `decodedToken.userId` to look up groups and revoke if admin.
2. **MFA endpoints unrate-limited** — Added `/api/v1/mfa/enroll` and `/api/v1/mfa/verify` to `AUTH_PATHS`. This applies `authRateLimit` (10 req/15 min/IP) and also picks up the `authBodyLimit` (1 KB) for free (addresses nitpick 6 as well).

**Suggestions (Should Consider):**
3. **Null guard before `decryptSecret`** — Added explicit `if (!caver.totpSecret)` check with `res.serverError('MFA secret unavailable.')`.
4. **Dead legacy branch removed** — Replaced with an explicit `throw` for unknown version prefixes. Updated the test to verify the error.
5. **Timing oracle documented** — Added detailed comment on `adminAuthRateLimit` middleware explaining the accepted trade-off and rationale.

**Nitpicks (Optional):**
6. Covered by fix #2.
7. **Mixed time units** — Grouped config properties by unit type with section comments (`--- TTLs (seconds) ---`, `--- Cooldowns (milliseconds) ---`).

**Bonus:** Added `mfaEnabled` to the `GET /api/v1/account` response so the front-end can determine enrollment status.

Paul-AUB:
# Code Review — PR #1609: `feat(auth): harden admin account security with TOTP MFA and brute-force protection`

## Overview

Substantial PR (~3,700 lines added, 55 files) that hardens administrator account security on three axes:

1. **Mandatory TOTP MFA** for admins (enrollment, verification, reset) with AES-256-GCM-encrypted secret
2. **Brute-force protection**: dedicated rate limit (5 / 15 min), account ban after 5 failures, alert emails
3. **Token hardening**: reduced TTL (10 days vs 90), immediate revocation on password change

The author has already addressed the critical points raised by Paul-AUB (revocation in the reset flow, rate-limit on `/mfa/verify`, etc.).

---

## Strengths

- **Well-thought-out threat model**: MFA + brute-force + timing + revocation, layered as defense in depth.
- **Correct cryptography**: AES-256-GCM, 96-bit random IV, authTag verified, version prefix (`v1:`) for future rotation — [api/services/MfaService.js:53-67](api/services/MfaService.js#L53-L67).
- **TOTP anti-replay protected twice**: `isReplay` check *before* verification, then atomic optimistic lock (`WHERE lastUsedTotp = caver.lastUsedTotp`) to block concurrent races — [api/controllers/v1/auth/login.js:184-200](api/controllers/v1/auth/login.js#L184-L200). Excellent.
- **Separate enrollment tokens**: subject `MfaEnrollment` (10 min) vs `Authentication` (10 days), with the `mfaEnrollmentAuth` policy explicitly rejecting authentication tokens.
- **Very solid test coverage**: unit tests for `MfaService`, integration tests for login/MFA/ban/rate-limit/TTL/reset/revocation, policies. Includes tricky cases (replay, ±1 step codes, email cooldown, fire-and-forget failure).
- **Production safeguards**: `mfaDevSecret` refuses to be used in prod, `mfaEncryptionKey` required via env var.
- **Complete auth configuration**: all thresholds tunable via environment variables (`AUTH_RATE_LIMIT`, `ADMIN_AUTH_RATE_LIMIT`, etc.).

---

## Points of attention

### Medium — `mfaDevSecret` guard is too permissive
[api/services/MfaService.js:30-43](api/services/MfaService.js#L30-L43) — the guard only blocks `mfaDevSecret` in production:
```js
if (process.env.NODE_ENV === 'production' || sails.config.environment === 'production') {
  sails.log.error('... ignoring it for security.');
} else {
  return devSecret;
}
```
Any `NODE_ENV` other than `production` (`staging`, `preview`, `qa`...) will use the hardcoded secret if the config is loaded. Inverting the logic to an allow-list (`development` and `test` only) would be safer.

### Medium — `BlacklistService` is in-memory only (pre-existing)
[api/services/BlacklistService.js:42-46](api/services/BlacklistService.js#L42-L46) — `isRevoked` only consults the local `Map`. In a multi-instance deployment, the revocation triggered by `change-password.js` does not propagate to other instances until a restart. Not introduced by this PR, but it raises its criticality (admin token TTL = 10 days). Worth documenting / tracking.

### Minor — `FAILURE_DELAY_MS` hardcoded
[api/controllers/v1/auth/login.js:11](api/controllers/v1/auth/login.js#L11) defines `FAILURE_DELAY_MS = 200`, while the neighboring `authBruteForceDelay` (500 ms) is configurable. For consistency: move this value into config (with a test override of 0 to keep the test suite fast).

### Minor — Double DB lookup for admins on failed login
- [config/http.js:200-213](config/http.js#L200-L213): `findOne({ mail })` + `populate('groups')` in the `adminAuthRateLimit` middleware
- [api/controllers/v1/auth/login.js:39-41](api/controllers/v1/auth/login.js#L39-L41): same query repeated in the controller for failure tracking

Bounded to 10 req/IP/15 min via `authRateLimit`, so impact is negligible, but the caver could be stashed in `req.adminCaver` by the middleware to save one query.

### Minor — No cleanup of orphaned `totpSecret`s
If an admin starts `startEnrollment` then abandons, the encrypted secret stays in the database until the next `startEnrollment` overwrites it ([api/services/MfaService.js:158-160](api/services/MfaService.js#L158-L160)) or `resetMfa` is called. Not a time bomb (encrypted, unusable without `mfaEnabled = true`), but a periodic cleanup job would be cleaner.

### Minor — `lastUsedTotp` recorded after `confirmEnrollment`
[api/services/MfaService.js:190-196](api/services/MfaService.js#L190-L196) — the code used to finalize enrollment is stored as `lastUsedTotp`. If the admin attempts an immediate login (instead of using the token returned by `/mfa/verify`), they will be rejected as a replay for ~30 s. UX-wise: not a problem because `/mfa/verify` already returns a full auth token, but worth a comment explaining the design.

### Minor — `lastSuspiciousEmailAt` updated only if the email is *scheduled*
[api/services/AdminLoginProtectionService.js:50-58](api/services/AdminLoginProtectionService.js#L50-L58) — the timestamp is written before the email is actually sent. If SES fails (logged but swallowed), the user will receive no alert during the 15-min window. Acceptable since SES retries are rare, but `lastSuspiciousEmailAt` should conceptually represent "last successful send".

---

## Security — Validation of key risks

| Risk | Mitigation |
|---|---|
| Admin enumeration by timing | OK — `constantDelay(200ms)` + existing `authBruteForceDelay(500ms)`. 429 timing-oracle trade-off explicitly documented ([config/http.js:180-189](config/http.js#L180-L189)) |
| TOTP replay | OK — `isReplay` pre-check + atomic optimistic lock |
| Admin login brute-force | OK — rate limit `admin:` namespace 5/15min/IP + DB ban after 5 failures |
| TOTP brute-force | OK — `totpFailedAttempts` counter + ban at 5 |
| Stolen admin JWT | OK — 10d TTL + revocation on `change-password` (both branches) |
| Stolen MFA enrollment token | OK — 10min TTL, restricted scope (subject `MfaEnrollment`) |
| Unauthorized MFA reset | OK — password re-required even with valid token ([api/controllers/v1/mfa/reset.js:19-35](api/controllers/v1/mfa/reset.js#L19-L35)) |
| Body-size bypass via chunked | OK — `authBodyLimit` checks `Buffer.byteLength(JSON.stringify(req.body))` post-parsing |

---

## Recommendation

**Approved subject to the minor follow-ups** suggested above. No blockers. Inverting the `mfaDevSecret` guard (Medium item) is the change I'd prioritize because it's cheap and defensive. The other points can be handled as follow-ups.

ClemRz:
Thanks for the thorough second review! All points addressed:

**Medium — `mfaDevSecret` guard too permissive**
Inverted to an allow-list: only `development` and `test` environments can use the dev secret. Any other env (`staging`, `qa`, `preview`, `production`) is refused with a `warn`-level log.

**Medium — `BlacklistService` in-memory only**
Added a detailed JSDoc comment on `isRevoked()` documenting the multi-instance limitation (revocations don't propagate across instances until restart) and noting it's tracked for future improvement.

**Minor — `FAILURE_DELAY_MS` hardcoded**
Moved to `sails.config.custom.authFailureDelay` (default 200ms) for consistency with `authBruteForceDelay`. Test env overrides it to 0 for speed.

**Minor — Double DB lookup for admins on failed login**
The `adminAuthRateLimit` middleware now stashes the looked-up caver in `req.adminCaver`. The login controller reuses it (with a fallback lookup if the middleware didn't run, e.g. when rate limit hasn't triggered yet).

**Minor — No cleanup of orphaned `totpSecret`s**
Acknowledged — deferring to a follow-up since the secrets are encrypted and unusable without `mfaEnabled = true`.

**Minor — `lastUsedTotp` recorded after `confirmEnrollment`**
Added a comment explaining the design: the enrollment code is stored as a replay guard; acceptable because `/mfa/verify` already returns a full auth token so an immediate login isn't needed.

**Minor — `lastSuspiciousEmailAt` updated before send**
Added a comment explaining the intentional design: prevents duplicate emails from concurrent requests; the trade-off (no alert during cooldown if SES fails) is acceptable.

Paul-AUB (approved):
LGTM !

Issue #1452, opened 2026-02-11
Legislation guidelines


ClemRz:
# Introduction

This feature adds the ability to attach legislation links to countries, regions (ISO 3166-2), and massifs within Grottocenter. Guidelines are directly attached to each geographic entity independently — there is no automatic inheritance between geographic levels, since massifs can span multiple countries or regions. When a user views a cave or entrance page, the API collects guidelines from the directly associated geographic entities (the entrance's country, region, and the cave's massifs) and returns them grouped by source.

All authenticated users can create and update guidelines, following the same collaborative wiki pattern as other Grottocenter features. Soft-delete and restore of deleted content is restricted to Moderators. Full history/versioning allows any user to view changes, and any authenticated user to roll back to a previous version.

# Key design decisions

- **No geographic hierarchy inheritance**: massifs can span multiple countries/regions, so guidelines are attached directly to each entity independently.
- **Open to all authenticated users**: create and update follow the collaborative wiki model — no Leader restriction.
- **Soft-delete restricted to Moderators**: follows the existing permission matrix — only Moderators can soft-delete or restore deleted content.
- **History is public**: anyone (including visitors) can view guideline history.
- **Rollback open to authenticated users**: any authenticated user can roll back a non-deleted guideline to a previous version.
- **Out of scope**: universal conservation guides (e.g., NSS Minimum Impact Guidelines) — those belong on the wiki, not in this legislation feature.

# User stories

- As an authenticated user, I want to create legislation records attached to a country, region, or massif, so that cavers visiting caves in those areas are informed about relevant regulations.
- As a user viewing a country, region, or massif page, I want to see the guidelines attached to that entity, so that I know what legislation applies to that area.
- As an authenticated user, I want to update existing guideline records, so that legislation links and descriptions stay current.
- As a moderator, I want to soft-delete guideline records and restore previously deleted ones, so that outdated or incorrect legislation links can be managed.
- As a user, I want to view the history of changes to a guideline, and as an authenticated user, I want to roll back to a previous version, so that accidental or incorrect modifications can be reverted.
- As a user viewing a cave or entrance page, I want to see all applicable legislation from the directly associated geographic entities, so that I am informed about relevant regulations before visiting.
- As a developer, I want a well-defined data model for guidelines with a history table, so that guideline records are consistent, tracked, and reversible.

⚠️ This ticket is meant to be broken down into several sub-tickets ⚠️

ClemRz:
[...]

PR #1611, opened 2026-05-27
Feat/1452 legislation guidelines

dawoldo:
## 🤔 What

This PR implements the new **Legislation Guidelines** feature, which allows attaching legal/access guidelines to countries, ISO 3166-2 regions, and massifs, and groups them dynamically when retrieving caves and entrances.

### Key Changes:
*   **Database Schema & Triggers:** Added `t_guideline` and `h_guideline` tables, database triggers `histo_update_guideline` and `histo_delete_guideline` to track history snapshots automatically.
*   **Model Layer:** Created Waterline models `TGuideline` (geographic guidelines) and `HGuideline` (history snapshots).
*   **API Services:** Introduced `GuidelineService` for entity validation, retrieval, and hierarchical guideline aggregation.
*   **Endpoint Controllers:** Added CRUD endpoints with soft-delete/restore and version history rollback capabilities.
*   **Entity Integrations:** Integrated guidelines retrieve logic into existing country, region, massif, cave, and entrance find controllers and converters.
*   **Security Policies:** Applied token authentication and moderator group requirements to secure modification, deletion, and restoration endpoints.

---

## 🤷‍♂️ Why

Grottocenter needs a structured way to display geographic-specific guidelines (e.g., local cave access laws, environmental regulations) to cavers. Guidelines are attached directly to their respective levels and aggregated on cave and entrance detail pages so users are aware of all applicable rules from country down to massif levels.

---

### 1. Database & Schema
*   Implemented polymorphic entity mapping using `entityType` (`'country'`, `'region'`, `'massif'`) and `entityId` (`varchar`).
*   Leveraged standard Grottocenter trigger patterns (`histo_delete` and `histo_update`) to manage automated history versioning in `h_guideline`.

### 2. Services & Controllers
*   `GuidelineService` handles programmatic validation of target entities (making sure Spanish guidelines cannot be attached to a non-existent country code, for example).
*   Controllers enforce permissions:
    *   **Create/Update/Rollback:** All authenticated users (ownership checked on update).
    *   **Delete/Restore:** Restricted to Moderators.
*   History snapshots are returned ordered by date descending.

### 3. Converters & Data Aggregation
*   Integrated guideline retrieval into `toCave` and `toEntrance` converters.
*   The API groups output by geographic level (e.g. `{ guidelines: { country: [...], region: [...], massif: [...] } }`) or returns an empty array `[]` when no guidelines apply.

---

## 🧪 Testing

The feature is covered by a dedicated set of integration tests:
1.  **Creation Tests:** Validation of field lengths (title/description), missing fields, invalid language codes, and entity existence.
2.  **Update Tests:** Author and moderator permission boundaries, formatting validations, and trigger-based history creation.
3.  **Delete/Restore Tests:** Soft-delete database behavior and restoration rules.
4.  **Snapshots & Rollback Tests:** Retrieving change history and rolling back fields to previous states.
5.  **Retrieval Tests:** Verifying geographic entity payloads for countries, regions, massifs, caves, and entrances.


ClemRz (changes requested):
### Issues (Must Fix)

- **[api/controllers/v1/auth/verify-email.js:63-70]** The notification check `if (caver.pendingMail)` after the `updateOne` call uses the *original* `caver` object (fetched before the update). Since `updates.pendingMail = null` was set above, the DB row no longer has a `pendingMail`, but the local `caver` variable still does. This means the notification will fire correctly *and* the early return will execute — which is the intended behavior. However, the logic is fragile and confusing: it relies on the stale `caver.pendingMail` being truthy to distinguish "email change verification" from "signup verification". Consider renaming or extracting a boolean like `const isEmailChange = !!caver.pendingMail;` before the update, then using that for the conditional. This makes the intent explicit and avoids future bugs if someone refactors the update logic.

- **[api/controllers/v1/guideline/create.js:69-78]** The initial `HGuideline` snapshot is created manually in the controller with `id: newGuideline.dateInscription.toISOString()`. This sets the `date_reviewed` column (which is the PK in `h_guideline`) to the `dateInscription` value. However, the `histo_update_guideline` trigger also creates snapshots on UPDATE and uses `date_r` (which defaults to `date_inscription` when `date_reviewed` is null). If the first update happens within the same second as creation, there's a potential primary key collision on `(id, date_reviewed)` since both the manual snapshot and the trigger-generated one could have the same timestamp. Consider either:
  1. Not creating the manual snapshot and relying solely on the trigger (the trigger fires on the first UPDATE and captures the pre-update state), or
  2. Adding a small offset or using a different strategy to avoid PK collisions.

  This is the same pattern used by other entities in the project where the trigger handles history — the manual snapshot in the controller is redundant with the trigger's behavior on first update.

- **[sql/0_triggers.sql:1329-1330]** In `change_guideline()`, the first condition `if NEW.is_deleted != OLD.is_deleted` references `OLD` which is NULL on INSERT. In PostgreSQL, `NULL != false` evaluates to NULL (falsy), so the condition falls through to `TG_OP = 'INSERT'` — this works correctly. However, the `change_rigging()` trigger uses a different pattern (`NEW.id_reviewer is null` to detect create vs update) without referencing `OLD` in the INSERT path. The `change_guideline` approach works but is less explicit. Consider adding a guard `if TG_OP = 'UPDATE' then` around the `OLD`-referencing conditions for clarity and safety, matching the pattern:
  ```sql
  if TG_OP = 'INSERT' then
      type_change := 'create';
      id_author := NEW.id_author;
  elsif NEW.is_deleted != OLD.is_deleted AND NEW.is_deleted = true then
      ...
  ```

- **[api/controllers/v1/guideline/rollback.js:20-26]** The snapshot lookup fetches *all* snapshots for the guideline and then filters in JS by comparing timestamps. This is inefficient for guidelines with many history entries. Consider querying directly with a WHERE clause. Since the `id` (PK in Waterline, mapped to `date_reviewed`) is the timestamp string, you could use:
  ```js
  const snapshot = await HGuideline.findOne({ t_id: guidelineId, id: snapshotId });
  ```
  This leverages the database index instead of loading all rows into memory.

- **[api/services/GuidelineService.js:44-46]** `getGuidelinesForEntrance` fetches the entrance without populating `country`, so `entrance.country` will be the raw FK value (a string ISO code). The code handles this with `entrance.country?.id || entrance.country`, which works. However, `entrance.iso_3166_2` is accessed directly — consider adding a comment explaining that this field stores the region code directly (not a FK reference) to avoid confusion for future readers.

- **[api/services/mapping/converters.js:50-56]** In `toCave`, the guidelines conversion uses `toList('massif', { massif: source.guidelines || [] }, ...)`. The `toList` utility expects `data[key]` to be an array, but `source.guidelines` could be an array of guideline objects (from `GuidelineService.getGuidelinesForCave`). The wrapping `{ massif: source.guidelines || [] }` is correct but the key name `'massif'` is a bit confusing since it's being used as a data accessor key rather than describing the entity type. This works but could be clearer with a comment.

- **[api/controllers/v1/account/update.js]** The email change feature (pendingMail flow) is a significant addition bundled into a PR primarily about legislation guidelines. Consider splitting this into a separate PR for cleaner review history and easier rollback if issues arise with either feature independently.

- **[api/services/DocumentService.js:402]** Good catch fixing `TCountry.find` → `TMassif.find` for the massifs field. This was clearly a bug. Consider adding a brief note in the PR description about this unrelated bugfix so reviewers don't miss it.

- **[sql/0_tables.sql:1371]** The `entity_id` column is `varchar(10)`. This accommodates ISO country codes (2 chars), ISO 3166-2 region codes (up to ~6 chars), and massif IDs cast to text. However, if massif IDs ever exceed 9,999,999,999 (10 digits), this would truncate. While unlikely in practice, consider using `varchar(20)` for future-proofing, or document the assumption.

- **[api/models/TGuideline.js:95-101]** The `customToJSON()` method mutates `this.entityId` in place. This is a side effect that could cause issues if the same object is serialized multiple times (the second call to `parseEntityId` would receive an already-parsed value). Consider returning a new object instead:
  ```js
  customToJSON() {
    const obj = { ...this };
    if (obj.entityType && obj.entityId) {
      obj.entityId = GuidelineService.parseEntityId(obj.entityType, obj.entityId);
    }
    return obj;
  }
  ```

### Nitpicks (Optional)

- **[api/controllers/v1/guideline/delete.js:5]** `RecentChangeService` is imported but the `setDeleteRestoreAuthor` call happens after `destroyOne`. If `destroyOne` triggers the `histo_delete` function (which sets `is_deleted = true` via UPDATE), the `change_guideline` trigger already inserts a `t_last_change` row with `type_change = 'delete'`. The subsequent `setDeleteRestoreAuthor` then updates that row's `id_author`. This works but the two-step dance (trigger inserts row → controller updates author) could be documented with a brief comment for maintainability.

- **[test/integration/4_routes/Guidelines/find-for-entity.test.js:18-19]** The test "should return 200 with empty list if no guidelines exist" uses `country/ES` but fixture data includes a guideline for ES (id=2). This test might fail depending on test execution order if other tests haven't deleted it. Consider using a country code that definitely has no guidelines in fixtures (e.g., `IT` or `DE`).

- **[api/controllers/v1/guideline/create.js:37-38]** The `getEntityTypes()` method queries `pg_enum` on every create request. Consider caching this value since enum types don't change at runtime, or simply validating against the hardcoded `isIn` values already defined in the model.

ClemRz (dismissed):

- **[api/controllers/v1/account/update.js]** The email change feature (pendingMail flow) is a significant addition bundled into a PR primarily about legislation guidelines. This includes a new migration (`sql/9_03_2026_05_08_add_caver_pending_mail.sql`), model changes to `TCaver`, new `AccountNotificationService`, and substantial controller logic. Consider splitting this into a separate PR for cleaner review history and easier rollback if issues arise with either feature independently. This was flagged in the previous review and still applies.

- **[api/controllers/v1/guideline/create.js:37-38]** `TGuideline.getEntityTypes()` queries `pg_enum` on every create/find-for-entity request. Since enum types don't change at runtime, consider caching the result (e.g., a module-level variable populated on first call). This avoids a round-trip to the DB on every guideline mutation. The same applies to `find-for-entity.js:10`.

- **[api/services/GuidelineService.js:44-90]** `getGuidelinesForEntrance` makes 4 sequential/parallel DB calls (entrance lookup + 3 guideline queries). If the entrance doesn't exist, it returns `null`, but the `entrance/find.js` controller doesn't guard against this — it passes `guidelines: null` into the converter. The `toEntrance` converter handles `source.guidelines || {}` gracefully, so this won't crash, but it means a non-existent entrance ID passed to `getGuidelinesForEntrance` silently returns `null` while the entrance lookup in the controller returns `null` and triggers a 404. The parallel `Promise.all` in `entrance/find.js` means the guideline queries run even when the entrance doesn't exist — a minor waste of DB resources. Consider short-circuiting: if `entrance` is null after the parallel call, the guidelines result is simply discarded, which is fine for correctness but slightly wasteful.

- **[api/services/mapping/converters.js:50-56]** In `toCave`, the guidelines conversion uses `toList('massif', { massif: source.guidelines || [] }, c.toSimpleGuideline)`. The key `'massif'` is used as a data accessor key for `toList` but reads confusingly since it's not describing the entity type being converted — it's just the property name on the wrapper object. A brief inline comment would help future readers understand this pattern.

### Nitpicks (Optional)

- **[sql/0_triggers.sql:1288]** The `histo_insert_guideline` function is unique to guidelines — other entities (cave, entrance, massif, etc.) don't have an AFTER INSERT trigger for history. This is a deliberate design choice (the initial snapshot is created by the trigger rather than manually in the controller), which is cleaner. Consider adding a brief comment at the top of the function explaining why guidelines differ from other entities in this regard, so future contributors don't try to "fix" it by removing the trigger and adding manual snapshot creation.

- **[test/integration/1_services/GuidelineService.test.js:155-175]** The `getGuidelinesForCave` tests iterate over all caves to find one with/without massifs. This is fragile if fixture data changes. Consider using a known fixture cave ID directly (e.g., cave 1 which has entrance 1 with known coordinates inside massif 1's polygon, or a cave that's known to have no spatial overlap).

- **[api/controllers/v1/guideline/get-snapshots.js:12-15]** Returning 404 when a guideline has no history snapshots is slightly misleading — the guideline exists but has no history. A 200 with an empty list might be more semantically correct (matching how `find-for-entity` returns 200 with `[]` for entities with no guidelines). However, this is a minor API design choice and the current behavior is documented in the Swagger spec.

ClemRz (dismissed):

- **[api/controllers/v1/account/update.js]** The email change feature (pendingMail flow) is a significant addition bundled into a PR primarily about legislation guidelines. It includes a new migration (`sql/9_03_2026_05_08_add_caver_pending_mail.sql`), model changes to `TCaver`, a new `AccountNotificationService`, the `verify-email.js` rewrite, and substantial controller logic with its own test file (`change-email.test.js`). This was flagged in both previous reviews and still applies — consider splitting it into its own PR for cleaner review/rollback boundaries.

- **[api/controllers/v1/guideline/create.js:37-38 / find-for-entity.js:10]** `TGuideline.getEntityTypes()` queries `pg_enum` on every create/find-for-entity request. Since PostgreSQL enum values are immutable at runtime (they only change via DDL), this round-trip can be avoided. Consider a module-level cache populated on first call:
  ```js
  let cachedEntityTypes = null;
  // In getEntityTypes():
  if (cachedEntityTypes) return cachedEntityTypes;
  // ... query ...
  cachedEntityTypes = result.rows.map((r) => r.enumlabel);
  return cachedEntityTypes;
  ```

- **[api/controllers/v1/guideline/rollback.js:20-26]** The snapshot lookup fetches *all* snapshots for the guideline via `HGuideline.find({ t_id: guidelineId })` and then filters in JS by timestamp comparison. For guidelines with many history entries, this loads all rows into memory. Consider querying directly:
  ```js
  const snapshot = await HGuideline.findOne({ t_id: guidelineId, id: snapshotId });
  ```
  The `id` field in the Waterline model maps to `date_reviewed` (the string PK from the fixture), so a direct lookup should work and leverages the DB index.

- **[api/controllers/v1/entrance/find.js:28-32]** The `Promise.all` fires `getGuidelinesForEntrance(entranceId)` in parallel with the entrance lookup. If the entrance doesn't exist, the guideline service also fetches the entrance internally (`TEntrance.findOne(entranceId)`) resulting in a duplicate query. The result is simply discarded if the entrance is null, so it's correct but slightly wasteful. One option: pass the already-fetched entrance object to `getGuidelinesForEntrance` or skip guidelines when entrance is null.

- **[test/integration/4_routes/Guidelines/find-for-entity.test.js]** The test data in `tguideline.json` references massif IDs `1` and `2`, while `3_14_mock_guidelines.sql` (production seed data) references massif `4` and `490`. This divergence means the test fixtures and production seed scripts tell different stories. This isn't a bug (test fixtures are independent of production seeds) but worth being aware of for future debugging.

- **[sql/0_triggers.sql:1329-1393]** The `change_guideline()` trigger function uses `NEW.is_deleted != OLD.is_deleted` in the INSERT path. In PostgreSQL, `OLD` is NULL during INSERT, so `NULL != false` evaluates to NULL (falsy) and falls through to the `TG_OP = 'INSERT'` check — which works correctly. However, this was restructured since the last review to check `TG_OP = 'INSERT'` first, which makes the logic clear and safe. Good improvement.

### Nitpicks (Optional)

- **[api/services/GuidelineService.js:44]** Minor readability note: `getGuidelinesForEntrance` fetches the entrance without populating `country`, so `entrance.country` is the raw FK value (ISO code string). The code handles this with `entrance.country?.id || entrance.country`, which works. A brief inline comment (like the one for `iso_3166_2` on the next line) would help future readers understand why the optional chaining is needed here.

- **[api/services/mapping/converters.js:60-66]** In `toCave`, the pattern `toList('massif', { massif: source.guidelines || [] }, c.toSimpleGuideline)` uses `'massif'` as a property accessor key on a synthetic wrapper object. This works but reads oddly at first glance since `'massif'` looks like it describes the entity type rather than a data key. A brief comment would improve clarity, e.g. `// 'massif' is the property key on the wrapper object for toList`.

- **[test/integration/1_services/GuidelineService.test.js:155-175]** The `getGuidelinesForCave` tests iterate over all caves with `await TCave.find()` to dynamically discover one with/without massifs. This is robust to fixture changes but makes the tests harder to reason about in isolation. Using a known fixture cave ID with an inline comment explaining the expectation would be more readable.

ClemRz:
@dawoldo I rebased your branch because I forced-push a squash of your 16 commits for #1452  on `develop`. Make sure you always squash your commits before merging to `develop`.

You had 14 tests failing before my rebase, they are still failing.

ClemRz (changes requested):
### Issues (Must Fix)

- **[test/integration/4_routes/Guidelines/delete.test.js:31-32, 58-59] and [test/integration/4_routes/Guidelines/restore.test.js:38-39]** The tests create guidelines with `entityType: 'country'` and `entityId: 'ES'`, but the `TGuideline` model no longer has these attributes — it uses junction tables (`countries`, `regions`, `massifs`) via `addToCollection`. Waterline silently ignores unknown attributes, so these guidelines are created without any entity association. The tests still pass because they only verify soft-delete/restore behavior (not entity links), but the fixture setup is incorrect and misleading. Consider replacing with the correct junction table pattern or simply removing those fields since they're not needed for the test assertions.

- **[test/customSQL.js]** The junction table indexes defined in `sql/0_tables.sql` (`idx_j_guideline_country_guideline`, `idx_j_guideline_country_country`, `idx_j_guideline_region_guideline`, `idx_j_guideline_region_region`, `idx_j_guideline_massif_guideline`, `idx_j_guideline_massif_massif`) are NOT mirrored in `test/customSQL.js`. Per project convention, all indexes from SQL files must also appear in `test/customSQL.js` to ensure the test database has the same performance characteristics. Add these `CREATE INDEX IF NOT EXISTS` statements to the `CREATE_GUIDELINE_TRIGGERS` block.

- **[assets/swaggerV1.yaml]** The `GET /api/v1/guidelines` (find-all) endpoint is defined in `config/routes.js` but is missing from the Swagger specification. The spec only documents `POST /guidelines` under the `/guidelines` path. Add a `get:` definition documenting the endpoint (public access, returns array of guidelines).

- **[api/controllers/v1/entrance/find.js:28-32]** The `Promise.all` fires `getGuidelinesForEntrance(entranceId)` in parallel with the entrance lookup. Inside `getGuidelinesForEntrance`, there's a duplicate `TEntrance.findOne(entranceId)` call. If the entrance doesn't exist, both the main lookup and the guideline service query the entrance separately, wasting a DB round-trip. The guideline result is discarded when entrance is null, so correctness is fine — but consider passing the already-fetched entrance to avoid the duplicate query, or at least short-circuiting guidelines when entrance is null (e.g., sequential: fetch entrance first, then fetch guidelines only if it exists).

- **[api/services/GuidelineService.js:73-132]** `getGuidelinesForEntrance` performs multiple rounds of DB queries (entrance → country/region/massif records → guideline lookups). For entrances with many massifs, this generates significant DB load. The design is correct but consider whether a single raw SQL query with JOINs across the junction tables would be more efficient for this aggregation use case. This is a performance consideration for later, not a blocker.

- **[api/controllers/v1/guideline/find-all.js]** This endpoint fetches ALL non-deleted guidelines with full population (`author`, `reviewer`, `countries`, `regions`, `massifs`). There's no pagination. If the guidelines table grows large, this will become a performance problem. Consider adding pagination parameters (`limit`, `skip`) consistent with other find-all endpoints in the project. Not a blocker for initial implementation, but worth planning for.

- **[sql/0_triggers.sql:1342-1350]** The `histo_update_guideline` trigger condition `if NEW.is_deleted = OLD.is_deleted OR (OLD.is_deleted = false AND NEW.is_deleted = true)` skips snapshot creation on restore (when `OLD.is_deleted = true` and `NEW.is_deleted = false`). This is intentional per the comment, but the logic is a bit hard to follow because the condition means "take snapshot on regular updates AND on soft-deletes, but NOT on restores." Consider adding a more descriptive comment explaining the full truth table for future maintainers.

- **[api/controllers/v1/guideline/delete.js:33]** After `destroyOne` (which triggers soft-delete via the `histo_delete` trigger), the code manually sets `guideline.isDeleted = true` on the in-memory object. This works but is fragile — if the trigger logic ever changes (e.g., to do a hard delete for certain cases), this manual mutation would be incorrect. Consider re-fetching the guideline after the operation, consistent with what `restore.js` does (it calls `getGuideline` after the update).

### Nitpicks (Optional)

- **[api/controllers/v1/guideline/find-for-entity.js:9-13]** The `validEntityTypes` array is hardcoded. This is fine and actually preferable (as noted in prior reviews — querying `pg_enum` at runtime adds unnecessary overhead). Good choice keeping it as a constant.

- **[test/fixtures/tguideline.json vs sql/3_14_mock_guidelines.sql]** The test fixtures reference massif IDs `1` and `2` while the production mock SQL references massifs `4` and `490`. Different titles too ("Vercors Massif" vs "Bois de Monnié"). This divergence is expected (test fixtures are independent of production seeds) but worth documenting in a comment if it ever causes confusion during debugging.

- **[api/controllers/v1/auth/verify-email.js]** This file was modified as part of this PR but appears unrelated to the legislation guidelines feature. The change itself looks correct (proper email change verification with race condition handling). Consider noting this as an unrelated fix in the PR description so reviewers don't conflate the two changes.

config/bootstrap.js
```diff
@@ -28,6 +28,11 @@ module.exports.bootstrap = async function (done) {
     ) {
       const modelMatch = message.match(/for model `([^`]+)`/);
       const model = modelMatch ? modelMatch[1] : 'unknown';
+      // Suppress validation warnings for history models (starting with 'h') where Waterline expects
+      // a string/number primary key, but the database returned a Date object due to the legacy composite PK.
+      if (model.startsWith('h')) {
+        return;
+      }
```
ClemRz, line 51: This might have a negative effect if the warning is not caused by the composite PK.

ClemRz (changes requested):
### Issues (Must Fix)

- **[api/controllers/v1/guideline/rollback.js:1]** `const moment = require('moment')` will throw `MODULE_NOT_FOUND` at runtime. The project migrated from `moment` to `dayjs` (commit `2a17d964`), and `moment` is no longer listed in `package.json` or installed in `node_modules`. This endpoint will crash on any rollback request. Replace with:
  ```js
  const dayjs = require('../../../utils/dayjs');
  ```
  and update line 22:
  ```js
  const formattedSnapshotId = dayjs(snapshotId).format('YYYY-MM-DD HH:mm:ss.SSS');
  ```

- **[api/controllers/v1/guideline/find-all.js]** This endpoint loads ALL non-deleted guidelines with full population and no pagination. As the guidelines table grows, this will become a performance concern. Consider adding `limit`/`skip` parameters for consistency with other find-all patterns in the project. Not a blocker for merging, but worth planning as a follow-up.

- **[api/controllers/v1/guideline/update.js:45]** `getGuideline(guidelineId)` is called a second time (after the initial `TGuideline.findOne` on line 10) just to get the populated current entity associations. Since you already have `rawGuideline` from line 10, consider populating it directly or using `TGuideline.findOne(guidelineId).populate('countries').populate('regions').populate('massifs')` for the initial fetch, avoiding the double query.

- **[config/bootstrap.js:34-36]** Suppressing all warnings for models starting with `'h'` is quite broad — it silences legitimate PK issues in any future history model. Consider checking for the specific model name (`hguideline`) or at least adding a comment listing the known affected models so future developers understand the scope.

### Nitpicks (Optional)

- **[api/services/GuidelineService.js:83]** The `entrance.country?.id || entrance.country` pattern works but benefits from a short comment explaining that `country` may be a populated object or a raw FK string depending on whether it was populated upstream. A one-liner would save future readers time.

- **[api/services/mapping/converters.js:60-66]** In `toCave`, the pattern `toList('massif', { massif: source.guidelines || [] }, c.toSimpleGuideline)` uses `'massif'` as a property accessor key on a synthetic wrapper object. This reads oddly at first glance since `'massif'` looks like it describes the entity type rather than a data key. The inline comment `// 'massif' is the property key on the wrapper object for toList` added elsewhere addresses this partially, but adding one here as well would help.

ClemRz (changes requested):
### Issues (Must Fix)

- **[api/controllers/v1/guideline/find-all.js:34-37]** The `catch` block swallows the error without logging it. If this endpoint fails in production, there will be no trace of what went wrong — only the generic message reaches the client. Add `sails.log.error(err)` before returning:
  ```suggestion
  } catch (err) {
    sails.log.error(err);
    return res.serverError(
      'There was a problem while retrieving the guidelines.'
    );
  }
  ```

- **[assets/swaggerV1.yaml:3254]** The `POST /guidelines` description states "Authenticated action (moderator or author)" but the policy is just `tokenAuth` — any authenticated user can create a guideline, with no role restriction in the controller. This mismatch between the docs and the implementation could confuse API consumers. Update to:
  ```suggestion
      description: Create a new guideline. Authenticated action (any logged-in user).
  ```

- **[api/controllers/v1/guideline/find-all.js:6-7]** `req.param()` returns a string from query parameters. `Math.max('abc', 0)` yields `NaN`, which then propagates to `.skip(NaN)` and `.limit(NaN)`. Consider parsing to integer first:
  ```js
  const skip = Math.max(parseInt(req.param('skip', 0), 10) || 0, 0);
  const limit = Math.max(Math.min(parseInt(req.param('limit', 50), 10) || 50, 100), 1);
  ```

- **[assets/swaggerV1.yaml:3253-3295]** The `POST /guidelines` schema lists `countries`, `regions`, and `massifs` as individually optional, but the controller requires at least one of them to be non-empty. Consider adding a note in the description or as a `description` on the schema itself to clarify this "at least one" constraint, since OpenAPI doesn't natively express `anyOf` for required request body properties.

- **[api/services/GuidelineService.js:83]** The comment on line 83 explains the `entrance.country?.id || entrance.country` pattern, which is helpful. Consider a similar brief note for `entrance.cave?.id || entrance.cave` on line 85, since the same populated-vs-raw FK ambiguity applies.

### Nitpicks (Optional)

- **[api/controllers/v1/guideline/create.js:44 / update.js:42]** The `ensureArray` utility function is defined inline in both `create.js` and `update.js`. If this pattern continues to spread, consider extracting it to a shared utility (e.g., in `api/utils/` or as a helper). Not urgent given only two occurrences.

- **[config/routes.js:325-327]** The catch-all route `GET /api/v1/guidelines/:entityType/:entityId` sits after `GET /api/v1/guidelines/:id/snapshots` which relies on route ordering for correctness. A request like `GET /api/v1/guidelines/country/snapshots` would be incorrectly matched by the snapshots route (`:id = 'country'`), returning a 404 rather than looking up guidelines for a hypothetical entity. This is an unlikely edge case but worth documenting with a comment explaining the route ordering dependency.

ClemRz (changes requested):
### Issues (Must Fix)

- **[config/routes.js:326-328]** The route ordering comment acknowledges that `GET /api/v1/guidelines/:id/snapshots` must precede `GET /api/v1/guidelines/:entityType/:entityId`. However, the route `POST /api/v1/guidelines/:id/rollback/:snapshotId` also sits between these, meaning a request like `GET /api/v1/guidelines/123/snapshots` works correctly. But `GET /api/v1/guidelines/country/snapshots` (asking for guidelines of country code "snapshots", which presumably doesn't exist) will be caught by the snapshots route — returning a 404 ("no history snapshots") rather than a 200 with empty array from `find-for-entity`. While unlikely in practice (no country has ISO code "snapshots"), consider adding a numeric check in `get-snapshots.js` for the `:id` parameter to explicitly fail fast when it's not a number:
  ```js
  const guidelineId = Number(req.param('id'));
  if (Number.isNaN(guidelineId)) {
    return res.badRequest({ message: 'Guideline ID must be a number.' });
  }
  ```
  This hardens all `:id` routes against misrouted requests and documents the intent explicitly.

- **[api/controllers/v1/guideline/update.js:21]** The `isAuthor` check compares `rawGuideline.author` (which is the raw FK integer from Waterline since `author` is not populated here) with `req.token.id`. However, `rawGuideline` is fetched with `.populate('countries').populate('regions').populate('massifs')` but NOT `.populate('author')`. Since `author` is a `model: 'TCaver'` association that is not populated, `rawGuideline.author` will be the integer FK value — so the comparison `rawGuideline.author === req.token.id` works correctly as long as both are the same type. This is fine in practice, but if `req.token.id` is ever a string (e.g., from JSON parsing), the strict equality will fail silently. Consider using `Number(rawGuideline.author) === Number(req.token.id)` for robustness.

- **[api/controllers/v1/guideline/rollback.js:22-24]** The `dayjs(snapshotId).format('YYYY-MM-DD HH:mm:ss.SSS')` conversion assumes the incoming `snapshotId` is a valid date string. If a user passes garbage (e.g., `POST /guidelines/1/rollback/not-a-date`), dayjs will return "Invalid Date" and the `HGuideline.findOne` will simply return null (resulting in a 404). This works but the error message "History snapshot... not found" is misleading when the real problem is an invalid date format. Consider adding a guard:
  ```js
  const parsed = dayjs(snapshotId);
  if (!parsed.isValid()) {
    return res.badRequest({ message: 'snapshotId must be a valid ISO date string.' });
  }
  ```

- **[api/services/GuidelineService.js:133-158]** `getGuidelinesForCave` performs two sequential DB rounds: first fetching all massifs for the cave, then querying `TMassif.find({ id: massifIds }).populate('guidelines')` to get junction links, and finally a third query to get the full guideline objects. Since `CaveService.getMassifs` already returns massif records, consider populating guidelines directly via the junction (like `getGuidelinesForEntrance` does with its native query approach). This would reduce from 3 queries to 2. Not a blocker but worth noting for future optimization.

- **[config/policies.js:197-204]** The guideline routes that accept `:id` (`update`, `delete`, `restore`, `rollback`, `get-snapshots`) don't use the `validateId` policy middleware. Some other entity routes (e.g., `device/update`, `device/delete`) use `['validateId', 'tokenAuth']`. Consider whether adding `validateId` would be appropriate here for early rejection of non-numeric IDs, though the controllers do handle invalid IDs gracefully (Waterline returns null for non-numeric lookups, leading to a 404).

- **[api/services/mapping/converters.js:60-67]** In `toCave`, the `source.guidelines` passed from `cave/find.js` is a flat array of guideline objects (from `getGuidelinesForCave` which only returns massif-level guidelines). Wrapping it as `{ massif: source.guidelines || [] }` and using `'massif'` as the key is correct but semantically confusing. The comment helps, but the pattern diverges from `toEntrance` which receives an already-grouped `{ country: [], region: [], massif: [] }` object. Consider making the API consistent: either always pass grouped objects to converters, or always pass flat arrays and let the converter handle grouping. This inconsistency could trip up future contributors.

- **[api/controllers/v1/auth/verify-email.js]** This file's changes (the `isEmailChange` boolean extraction) are unrelated to the legislation guidelines feature. While the change itself is correct and improves readability, bundling unrelated changes makes the PR harder to review and rollback independently. Consider noting this as an unrelated improvement in the PR description.

### Nitpicks (Optional)

- **[test/integration/4_routes/Guidelines/rollback.test.js:34-60]** The rollback test modifies guideline ID 1's title to "French National Caving Guideline Draft". Since tests may run in parallel across shards, and the `update.test.js` also modifies guideline 1's title, there's a potential for test ordering sensitivity within the same shard (the `4_routes/Guidelines/` folder will be kept together). The rollback test dynamically fetches the snapshot ID which helps, but the final assertions depend on knowing the pre-rollback state. Consider using a dedicated guideline created in a `before()` hook to isolate this test from mutations by other test files in the same folder.

- **[sql/0_triggers.sql:1305]** The `histo_insert_guideline()` trigger function declares `date_r` but uses it as a local variable — this is fine in PostgreSQL. Minor style note: the other trigger functions in this file use consistent formatting for `if/else` blocks, but the guideline triggers use slightly different indentation. Not functional, just a consistency observation.

- **[api/models/HGuideline.js:15-20]** The `id` attribute has `unique: true` but this is a composite PK table (`PRIMARY KEY (id, date_reviewed)`). Since `id` maps to `date_reviewed` in Waterline and the real PK is `(id, date_reviewed)` in the DB, the `unique: true` declaration is slightly misleading at the model level. Waterline doesn't enforce composite PKs, so this is cosmetic — but a comment clarifying the composite PK relationship with `t_id` would help future readers.

ClemRz (changes requested):
### Issues (Must Fix)

- **[assets/swaggerV1.yaml:7707-7710]** The `t_id` field in the `Guideline` schema is documented as `type: string, format: date-time` with description "Only present in history snapshots as the snapshot ID." This is incorrect on both counts. In the code (`HGuideline.t_id`), this field is `type: 'number'` — it's the **integer ID of the parent guideline** the snapshot belongs to, not the snapshot ID itself (the snapshot ID is the `id` field, which is the `date_reviewed` timestamp). This will mislead API consumers trying to use the snapshots endpoint. Fix:
  ```yaml
  t_id:
    type: integer
    description: "Only present in history snapshots. The ID of the parent guideline this snapshot belongs to."
  ```

- **[api/controllers/v1/guideline/update.js:56-63]** When validating newly provided entities, the code passes empty arrays for unchanged entity types to `resolveEntitiesExist`. This is correct but means the function's early return guards (`if (countryIds && countryIds.length > 0)`) handle the short-circuit. Consider adding a brief comment above the call explaining why only changed entities are validated, since the `resolveEntitiesExist` function's signature doesn't make this conditional validation obvious.

- **[api/services/GuidelineService.js:34-65]** `getGuidelinesForEntity` performs a two-step query pattern: first fetching the entity with `.populate('guidelines')` to get IDs, then fetching the full guidelines separately. Since `.populate('guidelines')` already returns the junction records (which include the full guideline objects in Waterline), you could filter `isDeleted: false` in-memory and then populate author/reviewer/collections only for the surviving records. This would save one DB round-trip for entities with few guidelines. Not a blocker — the current approach is clearer and correct.

- **[config/bootstrap.js:34-47]** The `legacyHistoryModels` array explicitly lists all history models. If a new history model is added in the future (beyond `hguideline`), a developer would need to remember to update this list. Consider using a pattern match (e.g., `model.startsWith('h') && model !== 'health'`) or a comment noting that new H-prefixed models should be added here.

### Nitpicks (Optional)

- **[api/controllers/v1/guideline/create.js:44-46]** The `massifs` array is mapped through `Number(m)` to coerce IDs to integers. If a client sends a non-numeric string (e.g., `massifs: ['abc']`), `Number('abc')` produces `NaN`, which would then be passed to `resolveEntitiesExist` → `TMassif.find({ id: [NaN] })`. Waterline handles this gracefully (returns empty results, causing the 404 response), so it's not a bug — but an explicit `filter(Number.isFinite)` or early validation would make the intent clearer.

- **[api/controllers/v1/guideline/find-for-entity.js:9]** The `validEntityTypes` array is hardcoded as `['country', 'region', 'massif']`. This is the right approach (avoiding the runtime `pg_enum` query from earlier iterations). Consider extracting this to a shared constant (e.g., in `GuidelineService` or `CommonService`) since `create.js` uses the same concept implicitly through its entity validation logic.

ClemRz (changes requested):
### Issues (Must Fix)

- **[config/routes.js:325-328]** Route ordering conflict: `GET /api/v1/guidelines/:id/snapshots` is placed *before* `GET /api/v1/guidelines/:entityType/:entityId`, with an inline comment explaining the priority. However, the route `GET /api/v1/guidelines/:id/snapshots` uses `validateId` as its policy (from `config/policies.js`), which means a request like `GET /api/v1/guidelines/country/snapshots` will be routed to `get-snapshots` and *rejected* by `validateId` (because `"country"` is not a valid integer ID). This means it's impossible to retrieve guidelines for an entity whose ID is literally `"snapshots"` — which is unlikely in practice. More critically though, if a user navigates to `GET /api/v1/guidelines/1/FR` expecting it to be the entity route, it would instead hit `get-snapshots` (where `:id = 1` and `"FR"` becomes an unmatched path segment ignored by Sails). The comment acknowledges this but the implications for `entityId = "snapshots"` specifically (hitting the wrong controller) aren't handled. Consider adding a guard in `find-for-entity.js` that returns 400 if `entityId === 'snapshots'` or `entityId === 'restore'`, or restructure the routes to use a dedicated prefix (e.g., `/api/v1/guidelines/by-entity/:entityType/:entityId`).

- **[api/controllers/v1/guideline/find-for-entity.js:8-13]** The `find-for-entity` controller does not validate that the referenced entity actually exists. If `entityType` is `"country"` and `entityId` is `"XX"` (non-existent), `getGuidelinesForEntity` returns `[]` and the endpoint returns `200 []`. This is semantically correct (no guidelines exist for that entity) but means the consumer cannot distinguish between "entity has no guidelines" and "entity doesn't exist." This matches the test expectations (`find-for-entity.test.js:23` tests for `200 []` on `XX`), so the design is intentional — but the Swagger doc for this endpoint doesn't have a `404` response for non-existent entities. **Either add documentation noting that non-existent entities return 200 with an empty array, or add entity existence validation returning 404 when the entity itself doesn't exist.** The current spec implies only `400` for invalid entityType but doesn't clarify the non-existent entity behavior.

- **[api/services/mapping/converters.js:60-68]** In `toCave`, the guidelines conversion wraps the flat `source.guidelines` array into `{ massif: source.guidelines || [] }` and extracts it back via `toList('massif', ...)`. The naming is confusing — `'massif'` is just a synthetic key for `toList()` but reads as if guidelines are being treated as massifs. The inline comment helps, but the approach creates an inconsistency with the entrance converter (which expects a grouped object `{ country: [], region: [], massif: [] }`). For caves, `getGuidelinesForCave` returns a flat array (only massif-level guidelines), so the wrapper returns `{ massif: [...] }` or `[]`. Consider renaming the wrapper key to something more neutral (e.g., `'items'`) or restructuring `getGuidelinesForCave` to return `{ massif: [...] }` directly, matching the entrance pattern and eliminating the synthetic wrapper.

- **[api/controllers/v1/guideline/update.js:44-48]** The shadow variable `c` in `.map((c) => c.id || c)` conflicts with the module-level constant `c` pattern used elsewhere in the codebase (converters). While this file doesn't have that conflict, the parameter name `c` for the country mapping function shadows the outer scope conceptually and is inconsistent with the `r` and `m` names used in the very next lines for regions and massifs. This appears to be a leftover from an earlier review iteration where `c` was the converters object. The ESLint `no-shadow` rule may or may not catch this depending on configuration, but renaming to something like `(co) => co.id || co` would be clearer.

- **[api/services/GuidelineService.js:88-97]** In `getGuidelinesForEntrance`, the raw SQL uses `$${params.length}` for parameterized queries, which is correct. However, the `params.push(massifIds)` call pushes an array directly. With `sails.sendNativeQuery`, passing an array as a parameter value for use with `ANY($N::int[])` depends on the adapter correctly serializing it to a PostgreSQL array literal. This works with the `sails-postgresql` adapter, but it's worth a brief inline comment noting this reliance for future maintainers unfamiliar with how the adapter handles array params.

- **[api/controllers/v1/guideline/create.js:44-45]** The `massifs` array filtering uses `.filter(Number.isFinite)` after `.map(Number)`, which correctly removes `NaN` values from non-numeric strings. However, it also silently drops invalid IDs without any feedback to the caller. If a client sends `massifs: ['abc', 1]`, the result is `[1]` — the invalid entry is ignored. Consider returning a 400 error if any massif ID fails to parse, rather than silently dropping it. This would be consistent with the `resolveEntitiesExist` check that validates all passed IDs exist.

- **[api/controllers/v1/cave/find.js:27]** `cave.guidelines = await GuidelineService.getGuidelinesForCave(caveId)` mutates the object returned by `CaveService.getPopulatedCave`. This works because Waterline records are mutable POJOs after population, but it's a mild code smell — the mutation happens between the population and the response. The entrance controller avoids this by spreading (`{ ...entrance, guidelines }`). Consider using the spread pattern here too for consistency: `{ ...cave, guidelines }` passed to `treatAndConvert`.

- **[test/integration/4_routes/Entrances/find.test.js:53-58]** The test `should return empty guidelines array when entrance has no guidelines` uses entrance 3 and asserts `guidelines` is `[]`. This depends on entrance 3 having no country, region, or massif with guidelines attached. If future fixture changes link entrance 3's country/region to a guideline, this test will break without obvious cause. Consider adding a brief comment documenting the assumption (e.g., `// Entrance 3 has country/region with no associated guidelines in fixtures`).

### Nitpicks (Optional)

- **[api/controllers/v1/auth/verify-email.js]** The `isEmailChange` boolean extraction is a good improvement addressing a prior review comment. The change itself is correct. However, this is unrelated to the legislation guidelines feature. Consider noting it in the PR description as a separate cleanup to help reviewers track independent changes.

- **[sql/0_triggers.sql:1405]** The `histo_delete_guideline` trigger uses the shared `histo_delete()` function. This is clean and consistent with other entities. No issue here, just noting for completeness that the shared function handles the soft-delete pattern correctly for guidelines.

- **[test/integration/4_routes/Guidelines/rollback.test.js:25-62]** The `before()` hook creates a dedicated guideline and manually inserts a history snapshot via raw SQL to isolate this test from other files in the same shard. This is well-designed and addresses the test isolation concern raised in prior reviews. Good improvement over relying on shared fixture data.

ClemRz (changes requested):
### Issues (Must Fix)

- **[api/controllers/v1/guideline/rollback.js]** The rollback endpoint has no authorization check — any authenticated user can roll back any guideline to any historical snapshot. Since rollback mutates the guideline's title, description, and language, it is functionally equivalent to an update. The `update` controller enforces that the caller must be the author, a moderator, or an admin. The same check should apply here. Without it, any logged-in user can revert other users' edits.

- **[api/services/GuidelineService.js:192-196]** `resolveEntitiesExist` compares `countryIds.length` (or `regionIds.length` / `massifIds.length`) against the DB count, but does not deduplicate the input array. If a client sends `countries: ['FR', 'FR']`, the SQL `WHERE iso IN ('FR', 'FR')` returns count `1`, while `countryIds.length` is `2` — causing a spurious "entity does not exist" 404. The same issue applies to regions and massifs. Additionally, passing duplicate IDs to `addToCollection` (in the create controller) will hit the junction table's `PRIMARY KEY (id_guideline, id_country)` unique constraint and throw an unhandled error. Consider deduplicating early:
  ```js
  const uniqueCountryIds = [...new Set(countryIds)];
  const count = await TCountry.count({ id: uniqueCountryIds });
  if (count !== uniqueCountryIds.length) return false;
  ```
  And in `create.js`/`update.js`, deduplicate before calling `addToCollection`/`replaceCollection`.

- **[api/controllers/v1/guideline/rollback.js:30-33]** The `formattedSnapshotId` uses `dayjs(...).format('YYYY-MM-DD HH:mm:ss.SSS')` to match against the `HGuideline` model's `id` field (mapped to `date_reviewed` as a timestamp column). Whether this format matches what Waterline/PostgreSQL serializes depends on the adapter's timestamp → string conversion. In the test fixtures, `id` is stored as ISO 8601 (`"2026-01-01T10:00:00.000Z"`), which Waterline may return as-is or reformat. If there's a mismatch, `findOne` silently returns `null` and the endpoint returns 404. The rollback test passes because it manually inserts with a controlled format, but production data created by triggers (which use `now()`) may serialize differently. Consider querying with a small range or normalizing both sides to the same format to be safe.

- **[api/controllers/v1/massif/find.js:25-28]** This controller mutates the `massif` object directly (`massif.guidelines = ...`). The cave and entrance controllers use the spread pattern (`{ ...cave, guidelines }`, `{ ...entrance, guidelines }`). Consider aligning for consistency:
  ```js
  const guidelines = await GuidelineService.getGuidelinesForEntity('massif', String(massifId));
  // pass { ...massif, guidelines } to treatAndConvert
  ```

- **[config/policies.js:200]** `'v1/guideline/get-snapshots': ['validateId']` is functionally public access with ID format validation. This works, but the pattern is inconsistent with the analogous `'v1/rigging/get-snapshots': true` (which is explicitly public). Consider using `true` for clarity if snapshots are intentionally public, or adding `'tokenAuth'` if authentication is desired.

- **[api/controllers/v1/guideline/create.js:44]** `Number('')` returns `0` and `Number.isFinite(0)` is `true`, so an empty string in the massifs array (`massifs: ['']`) would pass the non-numeric validation and attempt to look up massif ID 0. It would be caught by `resolveEntitiesExist` (massif 0 doesn't exist), but the error message would say "entity does not exist" rather than "invalid number." Consider adding a check for `m <= 0` or filtering empty strings before `Number()` conversion.

- **[test/integration/4_routes/Caves/update.test.js:106-110]** Good fix — `populate()` order is not guaranteed. This is an unrelated bug fix; consider noting it in the PR description so reviewers can track independent changes.

### Nitpicks (Optional)

- **[api/controllers/v1/auth/verify-email.js]** The `isEmailChange` boolean extraction is a clean improvement from a prior review. Unrelated to legislation guidelines — consider noting it separately in the PR description.

- **[sql/0_triggers.sql:1305-1335]** The `histo_insert_guideline` AFTER INSERT trigger is a nice design choice that other entities don't use (they often create the initial snapshot manually in the controller). Consider adding a brief SQL comment explaining this deliberate divergence to prevent future contributors from "fixing" it by adding manual snapshot creation.

- **[api/services/GuidelineService.js:81]** The comment about `country` being a raw FK or populated object is at the top of the function block. A brief inline comment next to `entrance.country?.id || entrance.country` would make the intent immediately visible without scrolling up.

ClemRz (changes requested):
### Issues (Must Fix)

- **[api/controllers/v1/guideline/update.js:87-92]** The validation order creates a logic gap. If the caller sends `title: undefined` (not specified) but the combined entity lists (`countries`, `regions`, `massifs`) all end up empty because the *persisted* record's collections happened to be empty (e.g., after a previous `replaceCollection` with `[]`), the endpoint returns 400 "At least one country, region, or massif must be specified." However, the entity existence validation on lines 93-101 only validates caller-supplied values (not persisted ones), meaning a caller can never fix a guideline that somehow lost all its entity associations without also sending at least one entity. This is an unlikely but reachable state if another concurrency issue removes all junction rows. More critically: if `newTitle` is `undefined` and all three `new*` params are also `undefined`, the `updatedFields` object contains only `{ reviewer: req.token.id }`, and the endpoint silently updates `reviewer` without changing anything else — effectively a no-op that still triggers the history snapshot trigger. Consider returning 400 early if no updatable field is supplied.

- **[api/controllers/v1/guideline/get-snapshots.js:7-9]** The `validateId` policy is intentionally omitted (per the comment in `config/policies.js`), and the controller manually converts the `id` param with `Number(req.param('id'))`. However, if `id` is a non-numeric string like `"abc"`, `Number("abc")` returns `NaN`, and the code correctly returns 400. But if `id` is something like `"1abc"`, `Number("1abc")` also returns `NaN`, while `parseInt("1abc", 10)` would return `1`. The current behavior is safe (rejects ambiguous inputs), but differs from `rollback.js` line 8 where `guidelineId = req.param('id')` is passed directly to `TGuideline.findOne(guidelineId)` without numeric conversion — Waterline will coerce strings to numbers there. This inconsistency means `rollback` silently accepts `"1abc"` as guideline 1 while `get-snapshots` rejects it. Consider normalizing: either both validate explicitly or both delegate to Waterline. The `rollback` endpoint should ideally reject non-numeric IDs too.

- **[api/controllers/v1/guideline/rollback.js:8]** `const guidelineId = req.param('id')` stays as a string throughout. It's passed to `TGuideline.findOne(guidelineId)` (which coerces internally) and later to `HGuideline.find({ t_id: guidelineId })`. Since `HGuideline.t_id` is defined as `type: 'number'`, Waterline may silently coerce the string, but it would be cleaner and more defensive to parse it explicitly:
  ```js
  const guidelineId = Number(req.param('id'));
  if (Number.isNaN(guidelineId)) {
    return res.badRequest({ message: 'Guideline ID must be a number.' });
  }
  ```
  This aligns with how `get-snapshots.js` handles the same parameter.

- **[api/services/GuidelineService.js:92-142]** `getGuidelinesForEntrance` uses `sails.sendNativeQuery` with parameterized placeholders — good for SQL injection prevention. However, the `massifIds` array is passed as a single parameter and cast with `ANY($N::int[])`. This relies on the `sails-postgresql` adapter correctly serializing JS arrays to PostgreSQL array literals. This works in practice with the current adapter version, but the behavior isn't documented in Waterline. Consider adding a brief comment noting this adapter-specific assumption, so future maintainers know to verify compatibility if the adapter is upgraded.

- **[api/controllers/v1/auth/verify-email.js]** This file's change (extracting `isEmailChange` boolean) is unrelated to the legislation guidelines feature. The change itself is clean and addresses a prior review comment. Consider noting it in the PR description as an unrelated improvement so reviewers can track independent changes.

- **[api/services/mapping/converters.js:597]** In `toMassif`, guidelines are rendered via `toList('guidelines', source, c.toSimpleGuideline)`. When `getGuidelinesForEntity` returns an array (the massif/find controller spreads `{ ...massif, guidelines }` where guidelines is the array), this works because `toList` will look for `source['guidelines']`. However, the cave and entrance converters receive guidelines as a *grouped object* (`{ massif: [...] }` or `{ country: [...], region: [...], massif: [...] }`), which requires different handling (reading sub-properties). This divergence is correct but could confuse future contributors. A brief inline comment on the `toMassif` guidelines line noting "flat array, unlike cave/entrance which receive grouped objects" would help.

- **[sql/0_tables.sql:1424-1430]** The `h_guideline.id` column is defined as `serial NOT NULL` but also has a foreign key to `t_guideline.id`. The `serial` type implies a separate auto-increment sequence for `h_guideline.id`, but the trigger inserts `NEW.id` (from `t_guideline`) — it never uses the sequence. The sequence is harmless (it's created but never used), but using `int4 NOT NULL` instead of `serial` would be more accurate and avoid creating an unused sequence object in the database. Not a blocker, but a cleanup opportunity.

- **[test/integration/4_routes/Caves/update.test.js:106-110]** The fix for non-deterministic `populate()` ordering is good and unrelated to guidelines. Consider noting it in the PR description.

### Nitpicks (Optional)

- **[api/controllers/v1/guideline/delete.js:37]** After `destroyOne` triggers soft-delete, the controller re-fetches via `GuidelineService.getGuideline(guidelineId)`. This is the correct pattern and matches `restore.js`. Good consistency.

- **[sql/0_triggers.sql:1305-1335]** The `histo_insert_guideline` AFTER INSERT trigger is a deliberate design choice unique to guidelines. The comment in `test/customSQL.js` already covers this. In production SQL, a brief one-liner like `-- Creates initial snapshot on insert (unlike other entities which rely on controllers)` above the function definition would make the intent immediately visible.

- **[config/policies.js:205]** `'v1/guideline/get-snapshots': true` — the comment explaining why `validateId` is omitted is helpful. Consider adding a similar one-liner for `find-for-entity` explaining that `:entityType` and `:entityId` are non-numeric, so `validateId` doesn't apply.

ClemRz (changes requested):
### Issues (Must Fix)

- **[assets/swaggerV1.yaml: rollback endpoint]** The `POST /guidelines/{id}/rollback/{snapshotId}` spec documents `200`, `400`, `401`, and `404` responses, but the controller (`rollback.js:39-49`) returns `403 Forbidden` when the caller is neither the author, a moderator, nor an admin. This mismatch means API consumers won't know that `403` is a possible response code. Add a `'403'` response entry:
  ```yaml
  '403':
    description: Forbidden (user is not the author, a moderator, or an administrator).
  ```

- **[api/controllers/v1/guideline/update.js:87-92]** There's a reachable edge case where a no-op mutation still triggers a history snapshot. If a caller sends `title: undefined` (not in the payload) but sends `countries: []` explicitly, and `regions`/`massifs` are also undefined, the effective entity lists are taken from the persisted record. If the persisted record happens to already have empty collections (e.g., from a prior buggy replaceCollection), the endpoint correctly returns 400 "At least one country…". However, if only one of (`countries`, `regions`, `massifs`) is sent as `[]` while the others are left undefined, and the persisted record's remaining collections are non-empty, the combined list passes. The real issue: if **only** entity arrays are sent (no `title`, `description`, `language`), the `updateOne` call still runs with just `{ reviewer: req.token.id }`, which bumps `date_reviewed` via the trigger and creates a snapshot capturing the previous state — even though nothing meaningful changed. The "no updatable field" check on lines 46-54 correctly guards against a completely empty body, but doesn't catch the case where only entities were changed (the `replaceCollection` calls below don't flow through `updatedFields`). Consider adding `dateReviewed: new Date()` to `updatedFields` only when entities actually changed, or treating entity-only updates as a valid mutation (which is arguably already the case). This is more of a semantic clarity issue than a crash, but worth documenting the intent.

- **[api/controllers/v1/guideline/rollback.js:57-62]** The snapshot lookup loads all snapshots for the guideline into memory and resolves via JS timestamp comparison. The comment explains why (adapter timestamp serialization varies), which is fair. However, consider at least a `.limit()` safety net if a guideline ever accumulates many snapshots. In practice history rows are few, so this is non-blocking — just a future-proofing thought.

- **[api/services/GuidelineService.js:109-113]** The `sails.sendNativeQuery` call passes a JS array as a parameter for `ANY($N::int[])`. The inline comment already notes this is adapter-specific behavior. Good documentation — no action needed, just confirming the comment adds clarity for future maintainers.

- **[api/controllers/v1/guideline/create.js:55-59 / update.js:79-83]** The massif ID validation logic (`!Number.isFinite(m) || m <= 0`) is duplicated across both controllers. Consider extracting this to a shared helper (e.g., `GuidelineService.validateMassifIds(massifs)`) to DRY up the two files. Not urgent with only two occurrences, but would reduce the surface for divergence if the validation logic evolves.

- **[sql/0_tables.sql:1424]** The `h_guideline.id` column is defined as `serial NOT NULL`. Since the trigger inserts `NEW.id` from `t_guideline` (never using the auto-increment sequence), this creates an unused sequence object in the database. Using `int4 NOT NULL` would be more precise. Not a blocker — the unused sequence is harmless — but worth noting as a cleanup opportunity.

- **[api/services/mapping/converters.js:60-67]** In `toCave`, the pattern `toList('massif', source.guidelines || {}, c.toSimpleGuideline)` uses `'massif'` as a property key on the grouped object from `getGuidelinesForCave`. The inline comment explains this, which helps. Since `getGuidelinesForCave` now returns `{ massif: [...] }` (matching the entrance pattern), the wrapper is consistent. Good.

### Nitpicks (Optional)

- **[api/controllers/v1/auth/verify-email.js]** The `isEmailChange` boolean extraction is a clean improvement addressing a prior review comment. This change is unrelated to the legislation guidelines feature — consider noting it in the PR description so reviewers (and git blame) can track independent improvements.

- **[test/integration/4_routes/Caves/update.test.js:106-110]** The fix for non-deterministic `populate()` ordering is correct and unrelated to guidelines. Worth a brief mention in the PR description.

- **[sql/0_triggers.sql:1308]** Consider adding a one-liner above `histo_insert_guideline()` explaining that the AFTER INSERT trigger is a deliberate design choice unique to guidelines (other entities create the initial snapshot manually in the controller). This prevents future contributors from "fixing" it by adding manual snapshot creation in `create.js`.

ClemRz (approved):

ClemRz:
## 🚀 Release Procedure

This PR introduces new database tables, indexes, and triggers. The SQL must be executed **manually** against the production PostgreSQL database **before** the deployment reaches production (or immediately after merge, before traffic hits the new endpoints).

### Pre-deployment: Run SQL migration

Connect to the production PostgreSQL instance and execute the following in order:

#### 1. Create tables

```sql
-- t_guideline
CREATE TABLE IF NOT EXISTS t_guideline (
  id serial NOT NULL,
  title varchar(150) NOT NULL,
  description varchar(500) NULL,
  id_author int4 NOT NULL,
  id_reviewer int4 NULL,
  id_language bpchar(3) NOT NULL,
  date_inscription timestamp NOT NULL DEFAULT now(),
  date_reviewed timestamp NULL,
  is_deleted bool NOT NULL DEFAULT false,
  CONSTRAINT t_guideline_pk PRIMARY KEY (id),
  CONSTRAINT t_guideline_t_caver_author_fk FOREIGN KEY (id_author) REFERENCES t_caver(id),
  CONSTRAINT t_guideline_t_caver_reviewer_fk FOREIGN KEY (id_reviewer) REFERENCES t_caver(id),
  CONSTRAINT t_guideline_t_language_fk FOREIGN KEY (id_language) REFERENCES t_language(id)
);

CREATE INDEX IF NOT EXISTS idx_t_guideline_is_deleted ON t_guideline(is_deleted);

-- h_guideline
CREATE TABLE IF NOT EXISTS h_guideline (
  id int4 NOT NULL,
  title varchar(150) NOT NULL,
  description varchar(500) NULL,
  id_author int4 NOT NULL,
  id_reviewer int4 NULL,
  id_language bpchar(3) NOT NULL,
  date_inscription timestamp NOT NULL DEFAULT now(),
  date_reviewed timestamp NOT NULL,
  is_deleted bool NOT NULL DEFAULT false,
  CONSTRAINT h_guideline_pk PRIMARY KEY (id, date_reviewed),
  CONSTRAINT h_guideline_t_guideline_fk FOREIGN KEY (id) REFERENCES t_guideline(id),
  CONSTRAINT h_guideline_t_caver_author_fk FOREIGN KEY (id_author) REFERENCES t_caver(id),
  CONSTRAINT h_guideline_t_caver_reviewer_fk FOREIGN KEY (id_reviewer) REFERENCES t_caver(id),
  CONSTRAINT h_guideline_t_language_fk FOREIGN KEY (id_language) REFERENCES t_language(id)
);

-- j_guideline_country
CREATE TABLE IF NOT EXISTS j_guideline_country (
  id_guideline int4 NOT NULL,
  id_country bpchar(2) NOT NULL,
  CONSTRAINT j_guideline_country_pk PRIMARY KEY (id_guideline, id_country),
  CONSTRAINT j_guideline_country_t_guideline_fk FOREIGN KEY (id_guideline) REFERENCES t_guideline(id),
  CONSTRAINT j_guideline_country_t_country_fk FOREIGN KEY (id_country) REFERENCES t_country(iso)
);
CREATE INDEX IF NOT EXISTS idx_j_guideline_country_guideline ON j_guideline_country(id_guideline);
CREATE INDEX IF NOT EXISTS idx_j_guideline_country_country ON j_guideline_country(id_country);

-- j_guideline_region
CREATE TABLE IF NOT EXISTS j_guideline_region (
  id_guideline int4 NOT NULL,
  id_region varchar(10) NOT NULL,
  CONSTRAINT j_guideline_region_pk PRIMARY KEY (id_guideline, id_region),
  CONSTRAINT j_guideline_region_t_guideline_fk FOREIGN KEY (id_guideline) REFERENCES t_guideline(id),
  CONSTRAINT j_guideline_region_t_iso3166_2_fk FOREIGN KEY (id_region) REFERENCES t_iso3166_2(iso)
);
CREATE INDEX IF NOT EXISTS idx_j_guideline_region_guideline ON j_guideline_region(id_guideline);
CREATE INDEX IF NOT EXISTS idx_j_guideline_region_region ON j_guideline_region(id_region);

-- j_guideline_massif
CREATE TABLE IF NOT EXISTS j_guideline_massif (
  id_guideline int4 NOT NULL,
  id_massif int4 NOT NULL,
  CONSTRAINT j_guideline_massif_pk PRIMARY KEY (id_guideline, id_massif),
  CONSTRAINT j_guideline_massif_t_guideline_fk FOREIGN KEY (id_guideline) REFERENCES t_guideline(id),
  CONSTRAINT j_guideline_massif_t_massif_fk FOREIGN KEY (id_massif) REFERENCES t_massif(id)
);
CREATE INDEX IF NOT EXISTS idx_j_guideline_massif_guideline ON j_guideline_massif(id_guideline);
CREATE INDEX IF NOT EXISTS idx_j_guideline_massif_massif ON j_guideline_massif(id_massif);
```

#### 2. Create trigger functions and triggers

```sql
-- histo_insert_guideline (AFTER INSERT — creates initial history snapshot)
CREATE OR REPLACE FUNCTION histo_insert_guideline() RETURNS trigger AS $$
DECLARE date_r timestamp;
BEGIN
  IF NEW.date_reviewed IS NULL THEN date_r := NEW.date_inscription;
  ELSE date_r := NEW.date_reviewed;
  END IF;
  INSERT INTO h_guideline (id, title, description, id_author, id_reviewer, id_language, date_inscription, date_reviewed, is_deleted)
  VALUES (NEW.id, NEW.title, NEW.description, NEW.id_author, NEW.id_reviewer, NEW.id_language, NEW.date_inscription, date_r, NEW.is_deleted);
  RETURN NEW;
END;
$$ LANGUAGE plpgsql;

-- histo_update_guideline (BEFORE UPDATE — snapshots old state)
CREATE OR REPLACE FUNCTION histo_update_guideline() RETURNS trigger AS $$
DECLARE date_r timestamp;
BEGIN
  IF NEW.date_reviewed IS NULL THEN date_r := NEW.date_inscription;
  ELSE date_r := NEW.date_reviewed;
  END IF;
  IF NEW.is_deleted = OLD.is_deleted OR (OLD.is_deleted = false AND NEW.is_deleted = true) THEN
    IF NOT EXISTS (SELECT 1 FROM h_guideline WHERE id = OLD.id AND date_reviewed = date_r) THEN
      INSERT INTO h_guideline (id, title, description, id_author, id_reviewer, id_language, date_inscription, date_reviewed, is_deleted)
      VALUES (OLD.id, OLD.title, OLD.description, OLD.id_author, OLD.id_reviewer, OLD.id_language, OLD.date_inscription, date_r, OLD.is_deleted);
    END IF;
  END IF;
  NEW.date_reviewed := now();
  RETURN NEW;
END;
$$ LANGUAGE plpgsql;

-- change_guideline (tracks in t_last_change)
CREATE OR REPLACE FUNCTION change_guideline() RETURNS trigger AS $$
DECLARE type_change varchar(20);
DECLARE id_author int4;
BEGIN
  type_change := '';
  IF TG_OP = 'INSERT' THEN
    type_change := 'create';
    id_author := NEW.id_author;
  ELSIF NEW.is_deleted != OLD.is_deleted AND NEW.is_deleted = true THEN
    type_change := 'delete';
    id_author := COALESCE(NEW.id_reviewer, NEW.id_author);
  ELSIF NEW.is_deleted != OLD.is_deleted AND NEW.is_deleted = false THEN
    type_change := 'restore';
    id_author := COALESCE(NEW.id_reviewer, NEW.id_author);
  ELSIF NEW.is_deleted = false THEN
    type_change := 'update';
    id_author := COALESCE(NEW.id_reviewer, NEW.id_author);
  END IF;
  IF type_change != '' THEN
    INSERT INTO t_last_change (type_entity, type_change, date_change, id_entity, id_author, name)
    VALUES ('guideline', type_change, now(), NEW.id, id_author, NEW.title);
  END IF;
  RETURN NEW;
END;
$$ LANGUAGE plpgsql;

-- Attach triggers
CREATE OR REPLACE TRIGGER histo_insert_guideline AFTER INSERT ON t_guideline FOR EACH ROW EXECUTE PROCEDURE histo_insert_guideline();
CREATE OR REPLACE TRIGGER histo_update_guideline BEFORE UPDATE ON t_guideline FOR EACH ROW EXECUTE PROCEDURE histo_update_guideline();
CREATE OR REPLACE TRIGGER histo_delete_guideline BEFORE DELETE ON t_guideline FOR EACH ROW EXECUTE PROCEDURE histo_delete();
CREATE OR REPLACE TRIGGER last_change_guideline BEFORE INSERT OR UPDATE ON t_guideline FOR EACH ROW EXECUTE PROCEDURE change_guideline();
```

#### 3. (Optional) Seed mock/demo data

Only run this if you want to pre-populate guidelines for testing or demo purposes. **Skip for a clean production deploy.**

See `sql/3_14_mock_guidelines.sql`.

---

### Deployment

After the SQL above is applied:

1. Merge this PR into `develop`
2. The GitHub Actions workflow (`deploy.yml`) will automatically build and deploy to Azure App Service
3. Sails starts with `migrate: safe` in production — no auto-schema changes, which is why the manual SQL above is required

### Post-deployment verification

1. `GET /api/v1/guidelines` — should return `200` with an empty array (or seeded data)
2. `GET /api/v1/caves/:id` — response should now include a `guidelines` property
3. `GET /api/v1/entrances/:id` — response should now include a `guidelines` property
4. `GET /api/v1/countries/:id` — response should now include a `guidelines` property
5. `GET /api/v1/massifs/:id` — response should now include a `guidelines` property

### Rollback

If issues arise, the feature can be rolled back by:

1. Reverting the merge commit on `develop` (triggers re-deploy without guideline code)
2. The tables can remain in the database harmlessly — or be dropped if a clean rollback is preferred:

```sql
DROP TRIGGER IF EXISTS last_change_guideline ON t_guideline;
DROP TRIGGER IF EXISTS histo_delete_guideline ON t_guideline;
DROP TRIGGER IF EXISTS histo_update_guideline ON t_guideline;
DROP TRIGGER IF EXISTS histo_insert_guideline ON t_guideline;
DROP FUNCTION IF EXISTS change_guideline();
DROP FUNCTION IF EXISTS histo_update_guideline();
DROP FUNCTION IF EXISTS histo_insert_guideline();
DROP TABLE IF EXISTS j_guideline_massif;
DROP TABLE IF EXISTS j_guideline_region;
DROP TABLE IF EXISTS j_guideline_country;
DROP TABLE IF EXISTS h_guideline;
DROP TABLE IF EXISTS t_guideline;
```

PR #1614, opened 2026-06-02
fix(rateLimit): suppress ERR_ERL_KEY_GEN_IPV6 validation warning

ClemRz:
## 🤔 What

- Suppress `ERR_ERL_KEY_GEN_IPV6` validation warnings from `express-rate-limit` during boot
- Add `validate: { ipKeyGenerator: false }` to all four `rateLimit()` calls in `config/rateLimit/rateLimiter.js`

## 🤷‍♂️ Why

After the latest deploy, production logs show repeated `ValidationError` warnings at boot:

> Custom keyGenerator appears to use request IP without calling the ipKeyGenerator helper function for IPv6 addresses.

This happens because our custom `keyGenerator` accesses `req.ip` directly (to handle Azure App Service's `ip:port` format) without using the library's built-in `ipKeyGenerator` helper. The warning is harmless but noisy — it fires for each limiter instance on every boot.

## 🔍 How

The `express-rate-limit` library exposes granular validation toggles. Setting `validate: { ipKeyGenerator: false }` disables only the IPv6 key-generator check while keeping all other validations active.

Our `keyGenerator` treats the IP as an opaque string key for bucket identification — it doesn't need IPv6 normalization. This is safe because rate limiting only requires a *consistent* key per client, not a canonicalized IP address.

## 🧪 Testing

- Full test suite passes (2456 tests, 0 failures)
- ESLint clean
- Verified the warning no longer appears on local boot

## 📸 Previews

N/A — no UI changes.

### DIFF
--- a/config/rateLimit/rateLimiter.js
+++ b/config/rateLimit/rateLimiter.js
@@ -8,7 +8,13 @@ const RightService = require('../../api/services/RightService');
  * with ERR_ERL_INVALID_IP_ADDRESS. Since the key only needs to be a consistent
  * string per client, req.ip works directly without normalization.
  *
+ * We also set `validate: { ipKeyGenerator: false }` on each limiter to suppress
+ * the ERR_ERL_KEY_GEN_IPV6 warning. The library detects that we reference
+ * req.ip without calling its ipKeyGenerator helper, but our use-case is safe:
+ * we treat the IP as an opaque string key rather than parsing it.
+ *
  * @see https://express-rate-limit.github.io/ERR_ERL_INVALID_IP_ADDRESS/
+ * @see https://express-rate-limit.github.io/ERR_ERL_KEY_GEN_IPV6/
  */
 const keyGenerator = (req) => req.ip || 'unknown';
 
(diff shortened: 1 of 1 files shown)


Paul-AUB (dismissed):
### Nitpicks (Optional)

1. **[config/rateLimit/rateLimiter.js:189]** Minor property ordering inconsistency: in `adminAuthRateLimit`, `validate` appears before `keyGenerator`, whereas in the other three limiters (`generalRateLimit`, `deleteRateLimit`, `authRateLimit`) `keyGenerator` is listed before `validate`. Consider reordering to match the convention used elsewhere in the file (`keyGenerator` → `validate` → `skip`). No functional impact.

---

The fix is minimal, targeted, and correctly scoped. Using `validate: { ipKeyGenerator: false }` as a partial object disables only the IPv6 key-generator check while leaving all other `express-rate-limit` validations active — the PR description confirms this. The updated JSDoc block in the `keyGenerator` comment is accurate and helpful. All four limiters are covered consistently.

PR #1616, opened 2026-06-02
fix(document): return 400 when issue field is set on unsupported type

### LINKED ISSUE #1615: bug(api): [23514] tdocument - check constraint violation on issue field for Article type
**Describe the bug**
Creating a document of type "Article" (id_type=18) with an `issue` field (e.g., `"102"`) fails because the PostgreSQL check constraint `t_document_check` only allows the `issue` column to be non-NULL when `id_type` is 16 or 17 (periodical types like Issue/Collection). Articles logically belong to a periodical issue, so the API should either strip the `issue` field before insert or the constraint should be relaxed to allow articles to carry issue numbers.

**To Reproduce**
Based on the trace logs:
1. POST /api/v1/documents
2. Include `issue: "102"` in the payload with a document type of "Article" (id_type=18)
3. The request returns a 500 Server Error

**Input Data**

- **Endpoint:** `POST /api/v1/documents`
- **Request Body:**
```json
{
  "identifier": "http://www.speleomag.com",
  "identifierType": {"id": "url"},
  "datePublication": "2018-07",
  "authors": [{"id": "37330"}],
  "editor": {"id": "794", "name": "Spéléo Magazine"},
  "library": {"id": "3705", "name": "CNDS"},
  "type": "Article",
  "title": "Grotte du Coutal. Les-Vignes, Lozère (48)",
  "description": "Une grande cavité et une classique du Causse de Sauveterre...",
  "subjects": [{"code": "2"}],
  "pages": "22-27",
  "issue": "102",
  "license": {"id": "1"},
  "mainLanguage": "fra",
  "iso3166": [{"type": "region", "iso": "FR-48", "name": "Lozère"}],
  "parent": {"id": "22597"}
}
```
- **User:** id=32211, nickname="XF", role=User

**Expected behavior**
The API should handle creating an Article with an issue number gracefully without returning a 500 error. Either:
1. The `t_document_check` constraint should be updated to also allow `issue` when `id_type = 18` (Article), or
2. The API should validate and return a 400 error with a clear message explaining that articles cannot have an issue number.

**Error Details**
- Error Type: OperationalError [AdapterError]
- Occurrences: 1
- Time Range: 2026-06-02T14:21:29Z to 2026-06-02T14:21:29Z
- Representative Trace ID: 019e88b5-ff02-73d9-ba6f-0b688438ebaf
- Stack Trace:
```
OperationalError [AdapterError]: Unexpected error from database adapter: new row for relation "t_document" violates check constraint "t_document_check"
    at Object.during (/home/site/wwwroot/api/services/DocumentService.js:295:47)

raw error:
  severity: 'ERROR'
  code: '23514'
  detail: 'Failing row contains (233971, 32211, ..., 22-27, http://www.speleomag.com, 102, url, ..., 794, 3705, 18, 22597, 1, ...)'
  constraint: 't_document_check'
  table: 't_document'
```

**Additional context**
- Service: `api/services/DocumentService.js:295`
- Model: `tdocument`
- DB Error Code: `23514` (check_violation)
- Constraint: `t_document_check` requires `issue IS NULL` unless `id_type IN (16, 17)`
- The user was creating an Article (id_type=18) referencing parent collection "Spéléo Magazine" (id=22597) with issue number "102"

ClemRz:
## 🤔 What

- Add API-level validation that returns a **400 Bad Request** when the `issue` field is set on a document type that doesn't support it (anything other than Book or Issue)
- Closes #1615

## 🤷‍♂️ Why

A user submitted an Article (type 18) with `issue: "102"`, which violated the `t_document_check` PostgreSQL constraint and caused an unhandled **500 Internal Server Error**. The DB constraint is correct — articles don't carry the `issue` field directly; they reference a parent document of type Issue instead. The API should reject this upfront with a clear error message.

## 🔍 How

- Created `config/constants/document.js` with named document type IDs (`BOOK`, `ISSUE`, `ARTICLE`) and a derived `TYPES_ALLOWING_ISSUE` array
- Added validation in `api/controllers/v1/document/create.js` before calling `createDocument`
- Added the same validation in `api/controllers/v1/document/update.js` (using the existing document's type as fallback when type isn't being changed)
- Both return `res.badRequest(...)` with a message explaining the business rule

## 🧪 Testing

- Added integration test in `create.test.js`: Article with `issue` → expects 400
- Added integration test in `update.test.js`: setting `issue` on an existing Article → expects 400
- Full test suite passes (2458 tests, 0 failures)

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/api/controllers/v1/document/create.js
+++ b/api/controllers/v1/document/create.js
@@ -2,6 +2,9 @@ const ControllerService = require('../../../services/ControllerService');
 const DocumentService = require('../../../services/DocumentService');
 const { toDocument } = require('../../../services/mapping/converters');
 const FileService = require('../../../services/FileService');
+const {
+  TYPES_ALLOWING_ISSUE,
+} = require('../../../../config/constants/document');
 
 module.exports = async (req, res) => {
   const documentData = await DocumentService.getConvertedDataFromClient(
@@ -11,6 +14,17 @@ module.exports = async (req, res) => {
     req.body,
     req.token.id
   );
+
+  // Validate that the issue field is only set for types that support it
+  if (
+    documentData.issue != null &&
+    !TYPES_ALLOWING_ISSUE.includes(documentData.type)
+  ) {
+    return res.badRequest(
+      'The "issue" field is only allowed for documents of type Book or Issue. ' +
+        'Articles should be linked to a parent document of type Issue instead.'
+    );
+  }
 
   const cleanedData = {
     ...documentData,
--- a/api/controllers/v1/document/update.js
+++ b/api/controllers/v1/document/update.js
@@ -4,6 +4,9 @@ const NotificationService = require('../../../services/NotificationService');
 const FileService = require('../../../services/FileService');
 const RightService = require('../../../services/RightService');
 const { toDocument } = require('../../../services/mapping/converters');
+const {
+  TYPES_ALLOWING_ISSUE,
+} = require('../../../../config/constants/document');
 
 const { INVALID_FORMAT, INVALID_NAME, ERROR_DURING_UPLOAD_TO_AZURE } =
   FileService;
@@ -24,6 +27,21 @@ module.exports = async (req, res) => {
         `You are not authorized to update a document with modifications waiting a moderator approval.`
       );
     }
+  }
+
+  // Validate that the issue field is only set for types that support it
+  // (must run before file upload to avoid Azure side effects on invalid requests)
+  const documentDataForValidation =
+    await DocumentService.getConvertedDataFromClient(req.body);
+  const effectiveType = documentDataForValidation.type ?? document.type;
+  if (
+    documentDataForValidation.issue != null &&
+    !TYPES_ALLOWING_ISSUE.includes(effectiveType)
+  ) {
+    return res.badRequest(
+      'The "issue" field is only allowed for documents of type Book or Issue. ' +
+        'Articles should be linked to a parent document of type Issue instead.'
+    );
   }
 
   // Add new files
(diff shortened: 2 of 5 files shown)


Paul-AUB (dismissed):

1. **[api/controllers/v1/document/update.js:75-85]** The issue validation runs *after* the file upload block (lines 32–66). If a user POSTs an Article payload with both `issue: '102'` and attached files, those files are uploaded to Azure before the 400 is returned — wasting storage and incurring cost. Consider moving the `effectiveType` / `TYPES_ALLOWING_ISSUE` check to before the file-upload block (i.e., after the `modifiedDocJson` guard on line 30), consistent with how the document existence and permission checks are done before any side effects.

2. **[test/integration/4_routes/Documents/update.test.js:117-133]** The `TDocument.destroy` cleanup call is not guarded by `try/finally`. If the `expect(400)` assertion throws (e.g., the endpoint unexpectedly returns 200), the created `articleDoc` document leaks into subsequent tests. Wrapping the supertest call in a `try/finally` would ensure cleanup regardless of assertion outcome.

### Nitpicks (Optional)

3. **[test/integration/4_routes/Documents/update.test.js:125-130]** The update test omits the `.set('Accept', 'application/json')` header that all other supertest calls in both test files include. Minor inconsistency worth aligning for uniformity.

4. **[test/integration/4_routes/Documents/create.test.js:92]** The create test wraps the assertion in a named `describe('Article with issue field', ...)` block, but the update test adds a bare `it(...)` directly inside the existing `describe`. Keeping the same nesting pattern in both test files would improve readability.

---

Overall, this is a clean, focused fix. The constants file is well-structured (IDs match the seed file and the DB constraint), the `effectiveType` fallback in the update controller correctly handles the case where the type is not being changed, and both happy-path validations are covered by integration tests. The only non-trivial concern is item 1 (Azure file upload before early-exit validation).

ClemRz:
Thanks for the review! All 4 items were legit — addressed in the force-pushed commit:

1. **Issue validation moved before file upload** — `getConvertedDataFromClient` now runs before the Azure upload block so invalid requests get a 400 with no side effects. The result is reused downstream (no double call).
2. **Cleanup wrapped in `try/finally`** — `TDocument.destroy` runs regardless of assertion outcome.
3. **Added `.set('Accept', 'application/json')`** — aligned with the rest of the test suite.
4. **Wrapped in `describe('Article with issue field', ...)`** — matches the nesting pattern in the create test.

PR #1618, opened 2026-06-02
fix(document): map titleAndDescriptionLanguage for description creation

### LINKED ISSUE #1617: bug(api): [23502] tdescription - null id_language on document create
**Describe the bug**
When creating a document of type "Event" via `POST /api/v1/documents`, if no `language` field is provided in the request body, the `DocumentService.createDocument` method passes `undefined` as the `language` value to `TDescription.create()`. This results in a PostgreSQL NOT NULL constraint violation on the `id_language` column of `t_description`.

The API should either require the `language` field (and return 400 if missing) or default to a sensible language (e.g., inferred from the user's profile or the first preferred language).

**To Reproduce**
Based on the trace logs:
1. Authenticate as a user
2. `POST /api/v1/documents` with a body that includes `type: "Event"` but no `language` field

**Input Data**

- **Endpoint:** `POST /api/v1/documents`
- **Request Body:**
```json
{
  "datePublication": "2026-05",
  "authors": [{"id": "60181", "nickname": "Paul AUBERTIN"}],
  "type": "Event",
  "title": "Evènement pour tester [à supprimer]",
  "description": "Description de l'évènement.",
  "iso3166": [{"type": "region", "iso": "FR-01", "name": "Ain"}]
}
```
- **User:** id 60181 (Paul AUBERTIN)

**Expected behavior**
The API should handle this case gracefully without returning a 500 error. Either:
1. Validate that `language` is required and return a 400 Bad Request if missing, or
2. Default to a sensible language value when not provided.

**Error Details**
- Occurrences: 3
- Time Range: 2026-06-02T18:56:22Z to 2026-06-02T19:16:17Z
- Representative Trace ID: `019e89b1-a9ff-77ca-bf6c-828878632d25`
- Stack Trace:
```
OperationalError [AdapterError]: Unexpected error from database adapter: null value in column "id_language" of relation "t_description" violates not-null constraint
    at Object.during (/home/site/wwwroot/api/services/DocumentService.js:299:26)
  adapterMethodName: 'create',
  modelIdentity: 'tdescription',
  raw: error: null value in column "id_language" of relation "t_description" violates not-null constraint
    code: '23502',
    table: 't_description',
    column: 'id_language'
```

**Additional context**
- Service: `api/services/DocumentService.js` (line 299)
- Model: `tdescription`
- DB Error Code: `23502` (not_null_violation)
- The `TDescription.create()` call at line 299 sets `language: descriptionData.language`, which is `undefined` when the request body doesn't include it.

ClemRz:
## 🤔 What

- Fix `POST /api/v1/documents` crashing with a 500 (PostgreSQL NOT NULL violation on `t_description.id_language`) when `titleAndDescriptionLanguage` is missing from the request body
- Correctly map front-end field names to service-layer field names for document language fields
- Closes #1617

## 🤷‍♂️ Why

The front-end sends `titleAndDescriptionLanguage` and `documentMainLanguage` but the service was reading `body.mainLanguage` — a field that doesn't exist in the actual request payload. This caused `undefined` to be passed to `TDescription.create()`, which PostgreSQL rejects due to its NOT NULL constraint on `id_language`. The bug was hidden in tests because the test database (created by Waterline `migrate: drop`) doesn't enforce the same constraint.

## 🔍 How

- `getDescriptionDataFromClient()`: reads `body.titleAndDescriptionLanguage?.id` with fallback to `body.mainLanguage` for backward compatibility
- `getConvertedDataFromClient()`: reads `body.documentMainLanguage?.id` with fallback to `body.mainLanguage`, extracted into a `getDocumentLanguages()` helper
- Controller validation: returns 400 early if neither `titleAndDescriptionLanguage` nor `mainLanguage` is present
- Tests updated to use correct field names and cover the validation path

## 🧪 Testing

```bash
npm run test -- --grep "Document create|DocumentService"
```

All 2460 tests pass. Lint clean.

## 📸 Previews

N/A (API-only change)

### DIFF
--- a/api/controllers/v1/document/create.js
+++ b/api/controllers/v1/document/create.js
@@ -7,6 +7,16 @@ const {
 } = require('../../../../config/constants/document');
 
 module.exports = async (req, res) => {
+  // Validate required language fields
+  const hasDescriptionLanguage = Boolean(
+    req.body.titleAndDescriptionLanguage?.id ?? req.body.mainLanguage
+  );
+  if (!hasDescriptionLanguage) {
+    return res.badRequest(
+      'Missing required field: titleAndDescriptionLanguage.'
+    );
+  }
+
   const documentData = await DocumentService.getConvertedDataFromClient(
     req.body
   );
--- a/api/controllers/v1/document/update-with-new-entities.js
+++ b/api/controllers/v1/document/update-with-new-entities.js
@@ -36,6 +36,13 @@ module.exports = async (req, res) => {
   }
 
   if (isArrNotEmpty(newDescriptions)) {
+    const missingLanguage = newDescriptions.some((desc) => !desc.language);
+    if (missingLanguage) {
+      return res.badRequest(
+        'Each new description must include a language field.'
+      );
+    }
+
     const descParams = newDescriptions.map((desc) => ({
       ...desc,
       document: documentId,
--- a/api/services/DocumentService.js
+++ b/api/services/DocumentService.js
@@ -13,6 +13,14 @@ const {
 const normalizeToId = (item) =>
   item != null && typeof item === 'object' ? (item.id ?? item) : item;
 
+// Extract the document's main language from the request body.
+// Prefers `documentMainLanguage.id` (current front-end field) with
+// fallback to `mainLanguage` (legacy/backward-compat).
+const getDocumentLanguages = (body) => {
+  const lang = body.documentMainLanguage?.id ?? body.mainLanguage;
+  return lang ? [lang] : [];
+};
+
 module.exports = {
   async deleteInSearch(documentId) {
     await SearchService.deleteDocument('documents', documentId);
@@ -83,7 +91,7 @@ module.exports = {
     author: authorId,
     body: body.description,
     title: body.title,
-    language: body.mainLanguage,
+    language: body.titleAndDescriptionLanguage?.id ?? body.mainLanguage,
   }),
 
   getChangedFileFromClient: (fileObjectArray) =>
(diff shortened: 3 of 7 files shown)


Paul-AUB (dismissed):
The fix is well-targeted and correctly addresses the reported 500. The fallback chain (`titleAndDescriptionLanguage?.id ?? mainLanguage`) is consistently applied in both `getDescriptionDataFromClient` and `getDocumentLanguages`, and the service-layer unit tests cover the priority/fallback cases properly.

1. **[test/integration/4_routes/Documents/create.test.js]** There is no route-level integration test for the `mainLanguage` backward-compatibility path. The service unit tests cover it, but an integration test confirming that `POST /api/v1/documents` with `mainLanguage` (and no `titleAndDescriptionLanguage`) still returns 200 would make the backward-compatibility guarantee explicit and guard against future regressions at the controller level.

2. **[api/controllers/v1/document/update-with-new-entities.js:38-44]** The same NOT NULL hazard exists here: `TDescription.createEach(descParams)` is called with items directly from `req.body.newDescriptions`, and no language validation is performed before the DB call. Out of scope for this PR, but it is the same root pattern — worth a follow-up issue or a quick validation guard.

### Nitpicks (Optional)

3. **[api/controllers/v1/document/create.js:8-13]** The extracted `descriptionLanguage` value is immediately discarded after the truthy check — only its boolean value is used, while the actual language is re-read independently inside `getDescriptionDataFromClient`. Consider renaming the variable to `hasDescriptionLanguage` (boolean) or passing the already-computed value into the service to make the intent clearer and avoid the silent duplication.

ClemRz:
Thanks for the review @Paul-AUB! All three items addressed:

1. **Integration test for `mainLanguage` backward-compat** — Added a route-level test that sends `POST /api/v1/documents` with `mainLanguage` (no `titleAndDescriptionLanguage`) and asserts 200. Also fixed the existing tests that were using `res.body.id` instead of the correct `res.body.document.id` for cleanup.

2. **NOT NULL guard in `update-with-new-entities.js`** — Added validation that returns 400 if any item in `newDescriptions` is missing the `language` field, plus a corresponding integration test.

3. **Renamed `descriptionLanguage` to `hasDescriptionLanguage`** — Makes it explicit that the variable is only used as a boolean check.

PR #1629, opened 2026-06-03
chore(ci): group all dependabot updates into a single PR

ClemRz:
## 🤔 What

- Group all Dependabot npm updates into a single PR instead of one per package
- Ignore ESLint >=9 major bumps (airbnb-base doesn't support flat config)
- Remove `open-pull-requests-limit` (unnecessary with grouping)

## 🤷‍♂️ Why

Dependabot was opening up to 10 separate PRs per month, creating review fatigue. With grouping, all semver-compatible updates land in one PR that can be reviewed and merged in a single pass.

The ESLint ignore rule codifies the constraint documented in our dependency upgrade steering file — `eslint-config-airbnb-base` does not support ESLint v9's flat config yet.

## 🔍 How

Uses Dependabot's native `groups` configuration with a wildcard pattern (`*`) to batch all npm ecosystem updates into one PR.

## 🧪 Testing

No code changes — CI config only. Next monthly Dependabot run will produce a single grouped PR.

## 📸 Previews

N/A

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,8 +5,18 @@
 
 version: 2
 updates:
-  - package-ecosystem: 'npm' # See documentation for possible values
-    directory: '/' # Location of package manifests
+  - package-ecosystem: 'npm'
+    directory: '/'
     schedule:
       interval: 'monthly'
-    open-pull-requests-limit: 10
+    groups:
+      all-dependencies:
+        patterns:
+          - '*'
+    ignore:
+      # ESLint must stay at v8 — eslint-config-airbnb-base does not support flat config (v9+)
+      - dependency-name: 'eslint'
+        versions: ['>=9.0.0']
+      # eslint-config-airbnb-base is coupled to ESLint v8; a major bump may require flat config (v9+)
+      - dependency-name: 'eslint-config-airbnb-base'
+        update-types: ['version-update:semver-major']


Paul-AUB:
I think this would need a proper rebase before review !

ClemRz:
Thank you @Paul-AUB, done ✅

Paul-AUB:
Thanks, I have look !

Paul-AUB (dismissed):
Clean, minimal change. The grouping config is syntactically correct, the `*` wildcard correctly batches all npm dependencies, and the ESLint ignore rule is well-justified by the comment.

1. **[.github/dependabot.yml:12-15]** The `groups` block currently has no `update-types` filter, so Dependabot will bundle **major** version bumps together with minor/patch updates in the same PR. The PR description says "semver-compatible updates", which typically means minor + patch only. If the intent is to keep major bumps separate (usually a good idea — they carry breaking-change risk and may need individual review), consider adding:
   ```yaml
   groups:
     all-dependencies:
       patterns:
         - '*'
       update-types:
         - 'minor'
         - 'patch'
   ```
   With this, major bumps would still land in separate PRs while still dramatically reducing monthly noise.

2. **[.github/dependabot.yml:16-19]** The `eslint` pin keeps the package on v8, but if Dependabot later proposes a `eslint-config-airbnb-base` update that internally requires ESLint v9+, it would land in the grouped PR without the matching ESLint upgrade. That grouped PR would then fail CI. Consider also ignoring major bumps of `eslint-config-airbnb-base` (or at least tracking that both packages need to be upgraded together when airbnb-base adds flat-config support).

PR #1631, opened 2026-06-04
fix(search): guard against non-string sort parameter

### LINKED ISSUE #1630: bug(api): [500] search - sort.split TypeError on non-string sort param
**Describe the bug**
The `POST /api/v1/advanced-search` endpoint crashes with `TypeError: sort.split is not a function` when the `sort` parameter is a non-string truthy value (e.g., an empty object `{}`). The code at line 23 of `advanced-search.js` checks `if (sort)` before calling `sort.split(',')`, but objects are truthy in JavaScript so the guard doesn't prevent the call on non-string values.

**To Reproduce**
1. POST `/api/v1/advanced-search`
2. Send a JSON body with `"sort": {}` (empty object instead of a string)

**Input Data**

- **Endpoint:** `POST /api/v1/advanced-search`
- **Request Body:**
```json
{"query":"gouffre","entity":"entrances","sort":{},"filter":{},"matchAllFields":true,"page":0,"size":20}
```
- **User:** [not logged] (unauthenticated — no auth policy on this route)

**Expected behavior**
The API should validate that `sort` is a string before calling `.split()` on it. If `sort` is not a string (or is empty/falsy), it should either be ignored or return a 400 Bad Request with a descriptive message.

**Error Details**
- Error Type: `TypeError`
- Error Message: `sort.split is not a function`
- Occurrences: 4
- Time Range: 2026-06-03T09:00:25Z to 2026-06-04T03:16:25Z
- Representative Trace ID: `019e90a1-d55b-7796-b27d-cca9c7172596`
- Stack Trace:
```
TypeError: sort.split is not a function
    at module.exports [as v1/search/advanced-search] (/home/site/wwwroot/api/controllers/v1/search/advanced-search.js:23:29)
    at /home/site/wwwroot/node_modules/sails/lib/router/bind.js:236:53
    at routeTargetFnWrapper (/home/site/wwwroot/node_modules/sails/lib/router/bind.js:395:9)
```

**Additional context**
- Service: `api/controllers/v1/search/advanced-search.js`
- Model: `search` (controller domain)
- DB Error Code: N/A (TypeError, not a database error)
- Fix suggestion: Change the guard on line 22 to check that `sort` is a string: `if (sort && typeof sort === 'string')`

ClemRz:
## 🤔 What

- Fix `TypeError: sort.split is not a function` in the advanced search endpoint
- Closes #1630

## 🤷‍♂️ Why

The front-end occasionally sends the `sort` parameter as an empty object `{}` instead of a string. The existing guard `if (sort)` passes because `{}` is truthy, then `.split(',')` throws a TypeError since it's a string method.

This was causing intermittent 500 errors in production (4 occurrences over the last 40 hours).

## 🔍 How

- Coerce the raw `sort` parameter to `undefined` if it's not a string, before any validation or downstream use
- This ensures non-string values are silently ignored rather than crashing the endpoint

## 🧪 Testing

- Added 5 integration tests covering: empty object, number, array, null, and valid string sort values
- All 2477 existing tests continue to pass

### DIFF
--- a/api/controllers/v1/search/advanced-search.js
+++ b/api/controllers/v1/search/advanced-search.js
@@ -16,7 +16,8 @@ module.exports = async (req, res) => {
   if (!matchAllFields || matchAllFields === 'false') matchAllFields = false;
 
   const entity = req.param('entity') ?? '';
-  const sort = req.param('sort');
+  const rawSort = req.param('sort');
+  const sort = typeof rawSort === 'string' ? rawSort : undefined;
 
   // Validate entity-specific sort fields
   if (sort) {
--- a/test/integration/4_routes/Search/advanced-search-sort-type.test.js
+++ b/test/integration/4_routes/Search/advanced-search-sort-type.test.js
@@ -1,0 +1,122 @@
+/* eslint-disable global-require */
+const supertest = require('supertest');
+const sinon = require('sinon');
+const should = require('should');
+
+describe('Advanced Search - sort parameter type handling', () => {
+  let SearchService;
+
+  before(() => {
+    SearchService = require('../../../../api/services/SearchService');
+  });
+
+  afterEach(() => {
+    sinon.restore();
+  });
+
+  it('should not crash when sort is an empty object', async () => {
+    sinon.stub(SearchService, 'collectionSearch').resolves({
+      hits: [{ document: { id: '1', name: 'Cave A' } }],
+      found: 1,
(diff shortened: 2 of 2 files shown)


Paul-AUB (approved):
Surgical fix for the reported crash. The two-line change is correct, readable, and the five integration tests cover the exact combinations that matter (empty object, number, array, null, and a valid string golden path).

PR #1632, opened 2026-06-04
fix(rateLimit): use correct validate option name for express-rate-limit v8

ClemRz:
## 🤔 What

- Fix the `validate` option key in all four `rateLimit()` calls in `config/rateLimit/rateLimiter.js`
- Replace `ipKeyGenerator` (non-existent) with `keyGeneratorIpFallback` (correct for v8.x)

## 🤷‍♂️ Why

PR #1614 intended to suppress the `ERR_ERL_KEY_GEN_IPV6` warning by setting `validate: { ipKeyGenerator: false }`. However, `ipKeyGenerator` is not a recognized validation option in `express-rate-limit` v8.4.1. This caused two warnings per limiter on every boot:

1. `ERR_ERL_UNKNOWN_VALIDATION` — the library doesn't recognize `ipKeyGenerator`
2. `ERR_ERL_KEY_GEN_IPV6` — the original warning, still unsuppressed

## 🔍 How

The correct option name in v8.x is `keyGeneratorIpFallback`. The library's error message lists all supported keys, and `keyGeneratorIpFallback` is the one that controls the IPv6 key-generator check. Swapping the key name is the only change needed.

## 🧪 Testing

- Server boots without any `ValidationError` warnings in console
- Full test suite passes (no behavioral change — this only affects boot-time validation)

## 📸 Previews

N/A — no UI changes.

### DIFF
--- a/config/rateLimit/rateLimiter.js
+++ b/config/rateLimit/rateLimiter.js
@@ -8,10 +8,10 @@ const RightService = require('../../api/services/RightService');
  * with ERR_ERL_INVALID_IP_ADDRESS. Since the key only needs to be a consistent
  * string per client, req.ip works directly without normalization.
  *
- * We also set `validate: { ipKeyGenerator: false }` on each limiter to suppress
- * the ERR_ERL_KEY_GEN_IPV6 warning. The library detects that we reference
- * req.ip without calling its ipKeyGenerator helper, but our use-case is safe:
- * we treat the IP as an opaque string key rather than parsing it.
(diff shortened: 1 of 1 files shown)


Paul-AUB (approved):
Clean one-liner fix applied consistently across all four limiters. The rename from `ipKeyGenerator` to `keyGeneratorIpFallback` is the right correction — the old key was unrecognized by express-rate-limit v8.x, causing a double warning storm on every boot without suppressing the original one. No behavioral change to rate limiting logic.

PR #1635, opened 2026-06-04
refactor(auth): upgrade otplib from v12 to v13

ClemRz:
## 🤔 What

Upgrade `otplib` from v12 to v13 (pinned at `13.4.1`).

- Migrate from deprecated `authenticator` API to the new `TOTP` class with explicit plugins
- Convert `MfaService.verifyCode()` from synchronous to async
- Extract a shared `test/helpers/totp.js` to eliminate TOTP boilerplate duplication across 6 test files

## 🤷‍♂️ Why

otplib v12 is end-of-life. v13 brings a cleaner plugin architecture (`NobleCryptoPlugin`, `ScureBase32Plugin`), better tree-shaking, and removes the deprecated `authenticator` singleton. Doing it now while the MFA surface is small keeps the migration contained.

## 🔍 How

- **`MfaService.js`**: Replace `authenticator.clone()` with `new TOTP({ period, epochTolerance, digits, algorithm, crypto, base32, guardrails })`. Use `epochTolerance: [30, 30]` (seconds-based) instead of v12's `window: 1` (step-based). `verifyCode()` is now `async` because `TOTP.verify()` returns a Promise in v13.
- **`login.js`**: Add `await` before `MfaService.verifyCode()`.
- **`test/helpers/totp.js`**: New shared helper exporting a configured `totp` instance and `generateCode()` utility, used by all MFA test files.
- **Test files (9)**: Replace `authenticator.generate(secret)` with `await generateCode()`. Fake-timer tests adapted to work with async generate/verify.
- **`package.json`**: Pin `otplib` to exact `13.4.1` (major bump with breaking API changes).

Key discovery: v13 uses `epochTolerance` (in seconds) for TOTP time-window tolerance, not `window` (which is now HOTP-only counter tolerance). The migration guide in `OTPLIB_UPGRADE.md` was inaccurate on this point.

## 🧪 Testing

All 2474 tests pass (`npm run test`). No additional setup needed — the dev secret and test infrastructure remain unchanged.

## 📸 Previews

N/A — backend-only refactor with no user-visible changes.

### DIFF
--- a/api/controllers/v1/auth/login.js
+++ b/api/controllers/v1/auth/login.js
@@ -168,7 +168,7 @@ module.exports = async (req, res) => {
       return res.serverError('MFA secret unavailable.');
     }
     const secret = MfaService.decryptSecret(caver.totpSecret);
-    const isValid = MfaService.verifyCode(totpCode, secret);
+    const isValid = await MfaService.verifyCode(totpCode, secret);
 
     if (!isValid) {
       const totpResult =
--- a/api/services/MfaService.js
+++ b/api/services/MfaService.js
@@ -6,16 +6,34 @@
  */
 
 const crypto = require('crypto');
-const { authenticator } = require('otplib');
+const {
+  TOTP,
+  NobleCryptoPlugin,
+  ScureBase32Plugin,
+  createGuardrails,
+} = require('otplib');
 
-// Use a cloned instance to avoid mutating global otplib state.
-const totpAuth = authenticator.clone();
-totpAuth.options = {
-  step: 30,
-  window: 1,
+// Plugins must be instantiated and passed explicitly in otplib v13.
+const otpCrypto = new NobleCryptoPlugin();
+const base32 = new ScureBase32Plugin();
+
+// Allow 10-byte secrets (the dev secret JBSWY3DPEHPK3PXP is 10 bytes).
+// Production secrets are 20 bytes and always satisfy the default minimum.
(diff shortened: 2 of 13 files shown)


Paul-AUB (dismissed):
Well-executed migration. The API choices are correct: explicit plugin instantiation, `epochTolerance: [30, 30]` properly replacing `window: 1`, per-call TOTP instance creation for verification, and the `toURI()` rename. All three call sites of `verifyCode` are properly awaited (`login.js`, `confirmEnrollment`, internal). The test helper extraction is clean — re-using `TOTP_OPTIONS` from the service ensures tests exercise the exact production configuration.

The fake-timer tests for ±1 step tolerance are particularly well-written: aligning to step boundaries before introducing the clock offset is the right approach to avoid flakiness at step edges.

1. **[api/services/MfaService.js:140-141]** `totp.verify(code)` is called with the assumption that it always returns an object with a `valid` field. In authentication paths, defensive coding is especially worthwhile. Consider:
   ```js
   const result = await totp.verify(code);
   return result?.valid ?? false;
   ```
   This costs nothing and makes the function resilient to any future library change that returns `null` on an internal error, rather than surfacing an unhandled `TypeError` to the caller.

ClemRz (commented):
Applied the defensive null-safe access suggestion:

```js
const result = await totp.verify(code);
return result?.valid ?? false;
```

All 2474 tests pass. Thanks for the review!

PR #1638, opened 2026-06-04
feat(superset): add Docker setup and CI workflow for BI dashboards

### LINKED ISSUE #1637: Deploy and configure Apache Superset for scientific data dashboards
## Summary

Deploy and configure Apache Superset as the BI layer for scientific observation data, embedded seamlessly into the Grottocenter front-end via the `@superset-ui/embedded-sdk`.

## Context

Scientific observation data (time series, contamination records, human activities) is stored in the transactional database. Superset provides interactive dashboards for researchers to explore this data — time series charts, geo maps, cross-filtering — without building custom visualization code.

Architecture is documented in `BI_INFRASTRUCTURE.md`, `BI_DASHBOARD_STRATEGY.md`, `BI_STAR_SCHEMA.md`, and `BI_COST.md` on branch `wip/scientific-data-bi-poc`.

### Infrastructure

- Deploy Superset as a Docker container on Azure App Service (`bi.grottocenter.org`)
- Provision Azure Cache for Redis (Basic C0) for Superset caching
- Configure Superset to connect to the existing Azure PostgreSQL Flexible Server (read-only)
- Set up DNS: `bi.grottocenter.org` CNAME to the App Service
- Configure CORS to allow embedding from `grottocenter.org`

### Configuration

- Register `v_measurement_wide` as the primary Superset dataset
- Define standard metrics: AVG(value), COUNT(measurement_id), MAX-MIN range
- Create the Tier 1 system template dashboard (automatic per-observation)
- Configure guest token embedding (no Superset login required for viewers)
- Set up row-level security for observation/cave/point scoping

### API Integration

- Add `GET /api/v1/superset/guest-token` endpoint on the Grottocenter API
- The endpoint mints a Superset guest token scoped via RLS to the requested observation/cave/point
- Requires `tokenAuth` policy

### Estimated Cost

~€18/month (App Service B1 + Redis Basic C0)

## Related

- #1562 — Parent: backend scientific observations infrastructure
- #1636 — Bulk measurement import endpoint (provides the data Superset visualizes)
- GrottoCenter/grottocenter-front#1365 — Front-end ETL wizard
- Reference docs on branch `wip/scientific-data-bi-poc`: `BI_INFRASTRUCTURE.md`, `BI_DASHBOARD_STRATEGY.md`, `BI_STAR_SCHEMA.md`, `BI_COST.md`, `BI_COMPARISON.md`

ClemRz:
## 🤔 What

Add Docker setup and CI workflow for Apache Superset, enabling embedded BI dashboards for scientific observation data.

- Superset Dockerfile with custom config and branding
- `docker-compose.superset.yml` for local development
- GitHub Actions workflow for Superset deployment
- Initial dashboard and chart exports (temperature comparison)
- Guest token embedding configuration (CORS, RLS)

Closes #1637

## 🤷‍♂️ Why

Researchers need interactive dashboards to explore scientific observation data (time series, geo maps, cross-filtering). Superset provides this without building custom visualization code. It embeds seamlessly into the Grottocenter front-end via `@superset-ui/embedded-sdk`.

Estimated cost: ~€18/month on Azure (App Service B1 + Redis Basic C0).

## 🔍 How

- **Docker**: Custom image based on `apache/superset` with pre-configured `superset_config.py` (CORS, guest token, Redis cache, PostgreSQL metadata)
- **Branding**: Custom favicon and logo replacing Superset's default
- **Entrypoint**: Auto-initializes admin user, imports dashboard exports on first boot
- **CI**: GitHub Actions workflow builds and deploys the container to Azure App Service
- **Exports**: YAML-based dashboard/chart/dataset exports for version-controlled dashboard definitions

Architecture: Superset connects read-only to the existing PostgreSQL instance and queries `v_measurement_wide` directly. The API mints guest tokens for embedded viewing.

## 🧪 Testing

```bash
# Start Superset locally
docker compose -f docker/docker-compose.superset.yml up -d

# Access at http://localhost:8088
# Default credentials: admin / admin
```

Verify:
- Dashboard loads with sample temperature chart
- Guest token embedding works (no Superset chrome visible)
- `v_measurement_wide` dataset is registered and queryable

## 📸 Previews

N/A — infrastructure PR, dashboards visible after scientific seed data is loaded.


Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[docker/superset/docker-entrypoint.sh:15]** `--password admin` hardcodes the admin password. When Docker Compose is used, it overrides the Dockerfile `CMD` with this entrypoint, so `SUPERSET_ADMIN_PASSWORD` is completely ignored and the Superset admin always gets `admin/admin`. `startup.sh` already handles this correctly with `${SUPERSET_ADMIN_PASSWORD:-changeme}` — apply the same pattern here, or consolidate into a single script.

2. **[docker/superset/docker-entrypoint.sh:7]** The PostgreSQL readiness probe hardcodes `root:root@dbserver:5432/grottoce`. If any of those values differ from the local defaults, the health loop never exits and the container hangs on startup. The `environment:` block in `docker-compose.superset.yml` already passes `SQLALCHEMY_DATABASE_URI` — use that (or the individual `POSTGRES_*` vars) for the probe connection string.

3. **[docker/superset/export/databases/Grottocenter_PostgreSQL.yaml:2]** `sqlalchemy_uri: postgresql+psycopg2://root:root@dbserver:5432/grottoce` commits plaintext credentials. These are dev credentials today, but Superset database exports are often re-run against production; future re-exports will carry real credentials. Strip the user/password before committing (use `postgresql+psycopg2://@dbserver:5432/grottoce` or a placeholder) and document how to restore them on import.

4. **[docker/superset/superset_config.py:144]** `SESSION_COOKIE_SECURE = True` breaks login over HTTP. Since this file is also volume-mounted in local dev (plain HTTP on localhost:8088), users will silently fail to authenticate. Gate it on an env var: `SESSION_COOKIE_SECURE = os.environ.get('SESSION_COOKIE_SECURE', 'false').lower() == 'true'`.

5. **[docker/superset/docker-entrypoint.sh vs startup.sh]** Two initialization scripts diverge in credential handling, readiness checks, and server startup method (`gunicorn` direct vs `run-server.sh`). The Dockerfile uses `startup.sh`; Docker Compose overrides it with `docker-entrypoint.sh`. Consider merging them into a single entrypoint so the dev and production paths can't silently drift.

6. **[.github/workflows/superset.yml:49]** `|| true` on the Azure webhook call hides deployment failures — CI stays green while the container in production is stale. Consider dropping `|| true` so a webhook failure fails the job, or moving the webhook trigger to a separate non-blocking step that still surfaces the failure status.

7. **[docker/superset/README.md:171]** Manual deployment instructions hardcode `--docker-registry-server-user ClemRz`. If this person's token expires or they leave the project, ops teams have no guidance on what to update. Document using a service account or the organization's bot instead.

### Nitpicks (Optional)

8. **[package.json]** The diff adds `"otplib": "^12.0.1"`, which is unrelated to this Superset PR and duplicates the same entry already on `develop`. This appears to be a rebase artifact — worth squashing or reverting to keep the git history clean.

9. **[docker/superset/README.md:176]** The example webhook URL uses `$grottocenter-superset` as what looks like a shell variable, but `-` in the name causes it to expand to empty string in bash. This is a copy-paste bug in the documentation.

10. **[docker/superset/superset_config.py:122]** `'allow_headers': ['*']` is maximally permissive for CORS. Narrowing to headers actually required (e.g., `Content-Type`, `Authorization`, `X-GuestToken`) reduces the attack surface with no functional cost.

11. **[docker/superset/Dockerfile:3]** `uv pip install psycopg2-binary flask-cors` installs unpinned versions. Consider pinning to specific versions for reproducible image builds.

ClemRz:
Thanks for the thorough review @Paul-AUB! All legitimate items have been addressed in the amended commit:

### Issues (Must Fix)

1. ✅ **Hardcoded admin password** — `docker-entrypoint.sh` now uses `${SUPERSET_ADMIN_PASSWORD:-changeme}` (same pattern as `startup.sh`). Docker-compose sets it to `admin` for local dev.

2. ✅ **Hardcoded DB connection in readiness probe** — Now uses `${SQLALCHEMY_DATABASE_URI:-postgresql://root:root@dbserver:5432/grottoce}` so it picks up the compose env var.

3. ✅ **Plaintext credentials in database export YAML** — Replaced with `USERNAME:PASSWORD` placeholders. Superset re-applies actual credentials on import via its configured database connection.

4. ✅ **`SESSION_COOKIE_SECURE = True` breaking HTTP locally** — Gated on env var: `os.environ.get('SESSION_COOKIE_SECURE', 'false').lower() == 'true'`. Docker-compose explicitly sets it to `false`; production should set `true` in Azure App Settings.

5. ✅ **Two diverging init scripts** — Aligned credential handling (both now use env vars with defaults). Kept them separate because they serve genuinely different purposes: `startup.sh` is the production Dockerfile CMD (uses `/usr/bin/run-server.sh`), while `docker-entrypoint.sh` adds local-dev-specific logic (postgres readiness wait, dashboard import, direct gunicorn).

6. ✅ **`|| true` hiding webhook failures** — Replaced with `continue-on-error: true` on the step. The failure is now visible in the workflow UI but doesn't block the job.

7. ✅ **Hardcoded `ClemRz` user in docs** — Replaced with `GrottoCenter` (org name) and added a comment recommending a machine/bot account with a `read:packages` PAT.

### Nitpicks (Optional)

8. ⏭️ **`otplib` rebase artifact** — Not present in the current diff; likely already resolved during rebase.

9. ✅ **`$grottocenter-superset` invalid shell variable** — Fixed to `grottocenter-superset:<webhook-password>` (removed the `$`).

10. ✅ **`allow_headers: ['*']`** — Narrowed to `['Content-Type', 'Authorization', 'X-GuestToken', 'X-CSRFToken']`.

11. ✅ **Unpinned pip install** — Pinned to `psycopg2-binary==2.9.12` and `flask-cors==6.0.2`.

Paul-AUB (changes requested):
All 11 items from the prior review have been addressed — thank you for the thorough response. After re-reading the updated code I found the following remaining concerns.

### Issues (Must Fix)

1. **[docker/superset/README.md:139]** `npm run dev:up` is documented as "Starts Superset + PostgreSQL + Redis", but `scripts/developmentSetup.js` (the script behind `dev:up`) contains no reference to the Superset compose file. The correct command is `npm run superset:up` (added in this PR's `package.json`). A developer following the README will start the main app stack, get no Superset container, and wonder why `localhost:8088` is unreachable.

2. **[docker/superset/init-superset.sh:20–24]** `init-superset.sh` still hardcodes `--password admin`, diverging from the env-var pattern now in place in both `docker-entrypoint.sh` and `startup.sh`. Since this is a host-side helper script, the risk is low — but using `${SUPERSET_ADMIN_PASSWORD:-admin}` here would make a credential change a one-variable update instead of a multi-file hunt.

3. **[docker/docker-compose.superset.yml:96–98]** `grottocenter_default` is declared as an `external` network with no existence guard. A developer who starts fresh with only this compose file will get `network grottocenter_default declared as external, but could not be found` and no guidance on how to fix it. Consider adding a prerequisite note to the README's "Local Development" section (e.g., "The main grottocenter stack must be running first, or create the network manually: `docker network create grottocenter_default`").

### Nitpicks (Optional)

4. **[docker/superset/docker-entrypoint.sh:257]** `superset import-directory --overwrite "$EXPORT_DIR/" || true` silently swallows import failures. A failed import leaves Superset running but dashboardless — a confusing state. Consider `|| echo "==> Warning: dashboard import failed, continuing anyway"` so at least the failure appears in container logs.

5. **[docker/superset/superset_config.py:188]** `'unsafe-eval'` in `script-src` alongside `'unsafe-inline'` is required by Superset's React build, but without a comment a future security reviewer will likely try to remove it and break the UI. A one-line note (`# Required by Superset's React/webpack bundle`) would prevent repeated churn.

ClemRz:
Thanks for the second pass @Paul-AUB! All items addressed:

### Issues (Must Fix)

1. ✅ **README documents wrong command** — Fixed: now shows `npm run dev:up` for the main stack and `npm run superset:up` for Superset separately.

2. ✅ **`init-superset.sh` hardcodes `--password admin`** — Now uses `${SUPERSET_ADMIN_PASSWORD:-admin}` (and other env vars for username/firstname/etc.).

3. ✅ **`grottocenter_default` external network with no guidance** — Added prerequisite note in the README explaining the network dependency and a manual `docker network create` workaround.

### Nitpicks (Optional)

4. ✅ **`|| true` hiding import failures** — Replaced with `|| echo "==> Warning: dashboard import failed, continuing anyway"` so the failure is visible in logs.

5. ✅ **`unsafe-eval` in CSP without comment** — Added `# 'unsafe-eval' is required by Superset's React/webpack bundle`.

### Additional fix

- **`USERNAME:PASSWORD` placeholder in database export YAML** — `docker-entrypoint.sh` now copies the export to `/tmp`, patches the placeholder with `${POSTGRES_USER:-root}:${POSTGRES_PASSWORD:-root}` via sed, then imports from the copy (so the volume-mounted source stays clean). Docker-compose passes `POSTGRES_USER`/`POSTGRES_PASSWORD` to the container. Same pattern applied to `init-superset.sh`.

Paul-AUB (dismissed):
All 16 items from the two prior review rounds are confirmed addressed in the current diff. Final pass below.

1. **[docker/superset/superset_config.py:136]** `GUEST_ROLE_NAME = 'Public'` — In a fresh Superset install, the `Public` role has no permissions on any database or dataset. Guest tokens run under this role, so an embedded dashboard will return 403 errors on the first chart query unless the `Public` role is explicitly granted access to the Grottocenter PostgreSQL connection and the `v_measurement_wide` dataset. Neither the init scripts nor the README document this step. Consider adding a `superset fab add-permission-role` call to `init-superset.sh` / `docker-entrypoint.sh`, or at minimum a "Required setup" section in the README explaining how to configure `Public` role permissions.

2. **[docker/superset/export/dashboards/Cave_id_5_1.yaml:7]** `published: false` — The imported dashboard will not be visible to non-admin users browsing Superset directly. The embedding path (guest token + dashboard UUID) bypasses this flag, so it is not a blocker for the front-end integration, but researchers who are given direct Superset access will see an empty dashboard list. Consider setting it to `true` or documenting the expectation.

### Nitpicks (Optional)

3. **[docker/superset/init-superset.sh:44]** The script now uses `${SUPERSET_ADMIN_PASSWORD:-admin}` for the created account, but the final status message still hardcodes `admin / admin`. If someone overrides `SUPERSET_ADMIN_PASSWORD`, the printed credentials will be wrong. A one-word fix: `echo "    Login: ${SUPERSET_ADMIN_USERNAME:-admin} / ${SUPERSET_ADMIN_PASSWORD:-admin}"`.

Issue #1636, opened 2026-06-04
Add POST /api/v1/observations/import bulk measurement import endpoint


ClemRz:
## Summary

Add a `POST /api/v1/observations/import` endpoint that accepts a raw CSV/TSV/TXT file and a profile JSON via multipart/form-data, and atomically creates the full observation data chain (point → observation → document → time series → measurements → quality log).

## Context

Researchers deploy data loggers in caves that produce CSV exports with timestamps and multiple measurement columns. This endpoint bridges the gap between raw heterogeneous spreadsheets and the structured scientific observation schema.

The front-end ETL wizard (GrottoCenter/grottocenter-front#1365) handles the user-facing column mapping and profile creation. This endpoint receives the result and performs the server-side import.

## What the endpoint does

1. Validates the profile JSON (required fields, IANA timezone, column roles, entity references)
2. Parses the CSV using profile instructions (encoding, header row, skip rows, number locale, column mappings)
3. Converts timestamps to UTC and measurement values to SI units
4. Creates entities atomically within a single transaction:
   - `t_point` (if new, optional)
   - `t_observation` (type = physical_measurements)
   - `t_document` (raw file archive, linked to cave or observation)
   - `t_time_series` (one per measurement column)
   - `t_measurement` (bulk insert)
   - `t_time_series_quality_log` (initial audit entry)
5. Returns a summary (created IDs, measurement count, observation date)

## Related

- #1562 — Parent: backend scientific observations infrastructure
- GrottoCenter/grottocenter-front#1365 — Front-end ETL wizard
- Spec: `.kiro/specs/scientific-data-import-endpoint/`

ClemRz:
# Requirements Document

## Introduction

This feature adds a single API endpoint (`POST /api/v1/observations/import`) that accepts a raw CSV/TSV/TXT file and a profile JSON describing how to interpret the file. The endpoint performs an atomic ETL operation: it parses measurement data from the file using the profile instructions, archives the raw file as a document, and creates the full observation data chain in the database (point → observation → document → time series → measurements → quality log).

The endpoint serves researchers and cavers who deploy data loggers in caves and need to bulk-import measurement datasets (temperature, humidity, pressure, CO₂, etc.) into Grottocenter's scientific observation system.

## Glossary

- **Import_Endpoint**: The `POST /api/v1/observations/import` controller action that receives multipart/form-data requests containing a data file and a profile JSON
[...]

PR #1639, opened 2026-06-04
feat(observation): add scientific observations schema and models

ClemRz:
## 🤔 What

Add the database schema and Waterline models for the scientific observations system.

- DDL for all observation tables (partitioned measurements, time series, devices, sensor configs, contamination, human activities)
- Lookup enumerations: quantity kinds, units, media, methods, observation types, contaminant types, human activity types
- Demo seed data with generated measurements for development/testing
- Waterline models for all new entities
- Cosmetic: rename SQL files to follow naming convention

Related:
- #1636
- #1562

## 🤷‍♂️ Why

This is the foundational schema that enables scientific data collection in Grottocenter. Researchers record observations at geolocated points, attach time series from data loggers, and track contamination and human activities in caves.

This schema is a prerequisite for:
- The bulk measurement import endpoint (#1636)
- The Superset BI dashboards (#1637)
- The front-end ETL wizard (grottocenter-front#1365)

## 🔍 How

- `sql/9_04_*_ddl.sql`: Creates tables, constraints, partitions, indexes, triggers, and the `v_measurement_wide` BI view
- `sql/9_05_*_enumerations.sql`: Seeds lookup tables with ontology-linked values
- `sql/9_06_*_seed.sql`: Demo data (devices, points, observations, time series, generated measurements)
- `api/models/T*.js`: Waterline model definitions matching the new tables
- SQL renames: `20250523_admin_mfa.sql` → `1_2025_05_23_admin_mfa.sql`, `3_mock_data.sql` → `3_15_mock_data.sql` (prefix consistency)

## 🧪 Testing

```bash
npm run dev:clean   # Rebuild containers from scratch
docker logs grotto-postgres  # Verify no SQL errors
npm run test        # Ensure existing tests still pass
```

Verify demo data:
```bash
docker exec grotto-postgres psql -U root -d grottoce \
  -c "SELECT COUNT(*) FROM t_measurement;"
# Should return ~44k rows (generated demo data)
```

## 📸 Previews

N/A — schema-only PR.


Paul-AUB (changes requested):
This PR adds ~1 700 lines, so the analysis prioritized the DDL schema, Waterline model consistency, and data integrity concerns.

### Issues (Must Fix)

1. **[sql/9_04…ddl.sql:130]** `t_observation.observation_date` is declared as `timestamp` (no time zone) while `t_measurement.timestamp` is `timestamptz`. Researchers may import historical data from sensors in different time zones; storing observation dates without a timezone loses that information and makes it impossible to reliably compare the two columns. Consider changing `observation_date timestamp NOT NULL` → `observation_date timestamptz NOT NULL`, and applying the same fix to `t_time_series.start_date` / `end_date` (line ~190) for the same reason.

2. **[sql/9_04…ddl.sql:437–455]** There is a 3-year partition gap in `t_measurement`: partitions exist for 2020–2022 Q1 and then jump to 2025 Q1. Any measurement with a `timestamp` between `2022-04-01` and `2025-01-01` will fail with `no partition of relation found for row`. There are also no partitions beyond 2026 Q4. Historical sensor data from 2022–2024 is likely — this is a silent data-loss bug that won't surface until the import endpoint ships. Add the missing quarterly partitions (2022 Q2 → 2024 Q4 at minimum) and a `DEFAULT` partition as a catch-all safety net.

3. **[sql/9_04…ddl.sql:122–135]** `t_method` has no `UNIQUE` constraint on `url`, unlike every other lookup table in this PR (`t_observation_type`, `t_quantity_kind`, `t_medium`, `t_human_activity_type`, `t_contaminant_type` all have `UNIQUE (code)`). Nothing prevents two methods with the same URL from being inserted, which would silently corrupt the ontology data. Add `CONSTRAINT t_method_url_key UNIQUE (url)`.

4. **[sql/9_04…ddl.sql:317–375]** Denormalized fields (`cave_name`, `observation_type_code`, `point_label`, `latitude`, `longitude` in `t_observation`; `quantity_kind_code`, `unit_symbol`, `medium_code` in `t_time_series`; `medium_code` in `t_contamination`) are populated at insertion time but have no triggers to stay in sync when source data changes. A cave rename, a lookup code update, or a point relocation will leave stale BI data. Consider adding `AFTER UPDATE` triggers on the source tables, or documenting that these fields are write-once and add a `CHECK` constraint to enforce that (e.g., `medium_code = (SELECT code FROM t_medium WHERE id = id_medium)` enforced via a trigger rather than a static constraint).

5. **[sql/9_04…ddl.sql:342–350]** The `v_measurement_wide` GeoJSON is constructed via string concatenation: `'{"type":"Point","coordinates":[' || o.longitude || ',' || o.latitude || ']}'`. While `numeric` values are safe from injection, the approach is fragile (no null guard for the individual components) and non-idiomatic. Consider `ST_AsGeoJSON(ST_MakePoint(o.longitude, o.latitude))::json` which handles nulls, produces properly typed JSON, and stays correct if the geometry column is used later.

6. **[sql/9_06…seed.sql]** The seed file is labeled "local dev only" in a comment, but it lives in the same `sql/` directory alongside production migrations and follows the same `9_06_` naming scheme. Nothing prevents it from running in production environments. Consider moving dev-only seeds to a separate directory (e.g., `sql/dev/`) or adopting a naming convention that an automated check can enforce (e.g., `99_dev_…`).

### Nitpicks (Optional)

7. **[sql/2_2022_09_07_notification_types.sql + sql/10_01 deletion]** The consolidation of the `REJECT` notification type into the older migration file is unrelated to scientific observations. Bundling it here makes the git history harder to follow. Consider moving it to a separate commit (or it could stay as-is since it is genuinely a cleanup).

8. **[api/models/TQuantityKind.js:22]** Missing blank line between the `url` and `symbolSi` attribute blocks, inconsistent with every other attribute definition in this file and across the other models in this PR.

ClemRz:

**Fixed:**

1. ✅ `observation_date` → `timestamptz`, same for `t_time_series.start_date`/`end_date`
2. ✅ Added missing quarterly partitions (2022 Q2 → 2024 Q4) + a `DEFAULT` partition as safety net
3. ✅ Added `UNIQUE (url)` constraint on `t_method`
5. ✅ Replaced string-concatenation GeoJSON with `ST_AsGeoJSON(ST_MakePoint(...))::json`
8. ✅ Added missing blank line in `TQuantityKind.js`

**Not addressed (intentional):**

4. Denormalized BI fields are write-once by design — they're populated at insertion time for Superset dashboards. Adding sync triggers would introduce complexity for a scenario (renaming a cave or updating a lookup code) that we handle via data reprocessing scripts when needed. Happy to add a code comment clarifying this intent if you think it helps.
6. Seed file stays in `sql/` — it follows the Docker init ordering convention. The "local dev only" comment is the guard; a separate directory would require changes to the Docker entrypoint. Can revisit if we add more dev-only seeds.
7. The REJECT notification consolidation is a one-line cosmetic cleanup that avoids a standalone migration file. Kept it bundled for simplicity.

All 2485 tests passing.

Paul-AUB (changes requested):
All three Must Fix items from the previous review have been addressed — `observation_date` / `start_date` / `end_date` are now `timestamptz` in the DDL, the partition gap is filled (2022 Q2 → 2026 Q4 + DEFAULT), and `t_method` has `UNIQUE (url)`. Thanks for the thorough response and rationale.

One new Must Fix surfaced during this pass, plus two minor suggestions.

### Issues (Must Fix)

1. **[api/models/TObservation.js:46, api/models/TTimeSeries.js:74,79]** The DDL fix (✅ DDL now uses `timestamptz`) was not carried through to the Waterline models: `observationDate`, `startDate`, and `endDate` all still declare `columnType: 'timestamp'`. In any environment where Sails.js runs with `migrate: 'drop'` or `migrate: 'alter'` (e.g., the test suite), Waterline will (re)create these columns as `timestamp`, silently undoing the timezone fix and reintroducing the data-loss risk from the previous review. Fix: change `columnType: 'timestamp'` → `columnType: 'timestamptz'` for those three attributes (use `TTimeSeriesQualityLog.js:49` — `changedAt` — as the pattern already in this PR).

2. **[sql/9_04_2026_06_03_scientific_observations_ddl.sql:231–238]** `t_measurement` has no FK constraint on `id_time_series`. PostgreSQL 12+ supports FK constraints on partitioned tables, so the engine version is not a blocker. Without it, a bug in the import endpoint (or a direct INSERT) could silently create orphaned measurements with a non-existent `id_time_series` — there is no database-level safeguard. Consider adding `CONSTRAINT t_measurement_t_time_series_fk FOREIGN KEY (id_time_series) REFERENCES t_time_series(id)` to the parent table definition.

3. **[sql/9_05_2026_06_03_scientific_observations_enumerations.sql:101]** The `t_method` INSERT uses bare `ON CONFLICT DO NOTHING`, while every other lookup table in the same file uses `ON CONFLICT (code) DO NOTHING`. The bare form silently swallows conflicts on *any* unique constraint including the PK — if a future developer adds a row with `id = 1` by mistake, the conflict is swallowed instead of surfacing. Prefer `ON CONFLICT (url) DO NOTHING` to be explicit and consistent with the rest of the file.

### Nitpicks (Optional)

4. **[api/models/TTimeSeries.js:83–97]** `measurementCount`, `minValue`, `maxValue`, `startDate`, `endDate` are denormalized aggregates that require application-level maintenance (no triggers keep them in sync). The project already knows how to annotate this intent — `TContamination.js` uses `// Denormalized field for BI` comments. A one-line comment near these fields (e.g., `// Denormalized aggregates — maintained by application at import time`) would make the maintenance contract explicit for future contributors.

ClemRz:
Thanks @Paul-AUB! All four items addressed:

**Fixed:**

1. ✅ `columnType: 'timestamp'` → `'timestamptz'` for `observationDate` in `TObservation.js`, and `startDate`/`endDate` in `TTimeSeries.js` — models now match the DDL.
2. ✅ Added `CONSTRAINT t_measurement_t_time_series_fk FOREIGN KEY (id_time_series) REFERENCES t_time_series(id)` to the `t_measurement` partitioned table.
3. ✅ `ON CONFLICT DO NOTHING` → `ON CONFLICT (url) DO NOTHING` for the `t_method` INSERT — consistent with all other lookup tables.
4. ✅ Added `// Denormalized aggregates — maintained by application at import time` comments near `measurementCount`, `minValue`, `maxValue`, `startDate`, `endDate` in `TTimeSeries.js`.

All 2485 tests passing.

Paul-AUB (changes requested):
All Must Fix items from the two previous rounds have been resolved. Good iteration. Three new issues surfaced on this pass — two at Must Fix level.

### Issues (Must Fix)

1. **[sql/9_05_2026_06_03_scientific_observations_enumerations.sql:25]** `symbol_si = '%'` for `RelativeHumidity` is inconsistent with how `value_si` is actually stored. The seed file (9_06, line 116) stores `value_si = (96.0 + ...)/100`, i.e., a dimensionless fraction in the range 0–1. The `si_to_display_factor = 100` correctly converts that fraction to a percent for `value_display`. But the `v_measurement_wide` view exposes `qk.symbol_si AS unit_si` — so Superset would label the `value_si` column as `%` while the values are ~0.96, not 96. A researcher reading the BI dashboard would see `RelativeHumidity (%)` with values near zero and interpret them as near-zero humidity, not ~96%. Fix: change `symbol_si` from `'%'` to `'1'` (dimensionless ratio, per SI/QUDT convention for a mass fraction or pure number). The `display_symbol = '%'` and factor = 100 already handle the presentation layer correctly. This fix also makes the row consistent with CO₂ (`symbol_si = 'mol/mol'`, dimensionless fraction) and the Kelvin-based temperature rows.

2. **[api/models/TObservation.js]** `cave_name` is absent from the Waterline model, but every other denormalized BI field in the same DDL table (`observation_type_code`, `point_label`, `latitude`, `longitude`) is present as an attribute. This is almost certainly an omission: `cave_name` is populated at import time (the seed backfills it via `UPDATE`) and is used in `v_measurement_wide` (`o.cave_name`). Without it in the model, any ORM-level query that needs cave name would have to fall back to raw SQL, making the import endpoint (#1636) harder to build consistently. Consider adding:
   ```js
   // Denormalized field for BI
   caveName: {
     type: 'string',
     allowNull: true,
     columnName: 'cave_name',
   },
   ```

3. **[api/models/TMeasurement.js:8, sql/9_04…ddl.sql:237]** The model declares `primaryKey: 'id'` and `unique: true` on `id`, but the DDL defines a composite PK `(id, timestamp)` — required by PostgreSQL for partitioned tables (the partition key must be part of every unique constraint). In any `migrate: 'alter'` or `migrate: 'drop'` environment, Waterline would either try to create `t_measurement` as a regular unpartitioned table, or attempt to add a UNIQUE constraint on `id` alone, both of which would break the partitioned structure. Since the tests apparently pass (implying `migrate: 'safe'` is the effective mode), the risk is latent — but it would cause a hard startup failure if someone changes migrate mode during development. Consider either setting `migrate: 'safe'` explicitly at the model level, or adding a comment documenting that this model must never be created/altered by Waterline:
   ```js
   // Partitioned table — schema is managed exclusively by SQL migrations.
   // Do NOT run Waterline in migrate: 'alter' or 'drop' mode with this model.
   ```

4. **[sql/9_04…ddl.sql:393–395]** `v_measurement_wide` joins `t_quantity_kind` via `JOIN t_quantity_kind qk ON qk.code = ts.quantity_kind_code` (INNER JOIN). Any `t_time_series` row with a `quantity_kind_code` that doesn't match a `t_quantity_kind.code` — e.g., if a future code rename or bulk-import typo gets through — is silently excluded from the view: measurements are invisible in Superset with no error. A `LEFT JOIN` would surface the anomaly as NULL values in the `unit_si` / `quantity_unit` columns, which are easier to detect in a dashboard than missing rows. Alternatively, joining through the FK chain (`ts.id_sensor_configuration → t_sensor_configuration.id_quantity_kind → t_quantity_kind`) avoids the denormalized string lookup entirely.

5. **[sql/9_04…ddl.sql:97]** The migration drops and recreates `t_point` on the assumption it is empty in all environments. The comment documents the intent, but there is no runtime guard. If that assumption fails silently in any environment, the data loss is immediate and unrecoverable. Adding a pre-condition takes one line:
   ```sql
   DO $$ BEGIN
     ASSERT (SELECT COUNT(*) FROM t_point) = 0,
       't_point must be empty — abort migration';
   END $$;
   ```

### Nitpicks (Optional)

6. **[sql/9_06…seed.sql:63]** The four `observation_date` literals (`'2025-06-01 10:00:00'`, etc.) are inserted without an explicit timezone offset. The column type is `timestamptz`, so PostgreSQL interprets them in the session timezone (UTC in Docker — correct outcome, fragile assumption). The `generate_series` calls in the same file use `::timestamptz` consistently. Appending `+00:00` to the four values would make the intent unambiguous and consistent with the rest of the file.

ClemRz (commented):
Thanks @Paul-AUB! All items from round 3 addressed:

**Fixed (Must Fix):**

1. ✅ `symbol_si = '%'` → `'1'` for `RelativeHumidity` — the SI column now correctly represents the dimensionless ratio stored in `value_si`. The `display_symbol = '%'` and `si_to_display_factor = 100` handle presentation.
2. ✅ Added missing `caveName` attribute to `TObservation.js` — consistent with all other denormalized BI fields in the model.

**Fixed (Suggestions):**

3. ✅ Added partition documentation comment to `TMeasurement.js` explaining the composite PK constraint and warning against `migrate: 'alter'/'drop'`.
4. ✅ `v_measurement_wide`: `JOIN t_quantity_kind` → `LEFT JOIN` — mismatched codes now surface as NULLs instead of silently hiding rows.
5. ✅ Added `ASSERT (SELECT COUNT(*) FROM t_point) = 0` guard before the `DROP TABLE` — migration now aborts explicitly if the assumption is violated.

**Fixed (Nitpick):**

6. ✅ Appended `+00:00` to the four `observation_date` literals in the seed file — consistent with the `::timestamptz` casts used elsewhere.

All 2485 tests passing.

Paul-AUB (changes requested):
All three Must Fix items from round 3 have been resolved — `symbol_si` for `RelativeHumidity` is correctly `'1'`, `caveName` is now in `TObservation.js`, the `t_measurement` partition guard exists, the `v_measurement_wide` LEFT JOIN is in place, and the `observation_date` seed literals now carry `+00:00`. One new Must Fix and three minor items surfaced on this pass.

### Issues (Must Fix)

1. **[docker/superset/export/datasets/Grottocenter_PostgreSQL/v_measurement_wide_1.yaml — `observation_date` column entry]** The Superset dataset export was generated before the DDL was patched to use `timestamptz` for `observation_date`. The YAML still describes the column as `type: TIMESTAMP WITHOUT TIME ZONE`. When `docker-entrypoint.sh` calls `superset import-directory` on a fresh instance, Superset's column metadata will classify `observation_date` as a naive timestamp: TZ-aware datetime pickers are disabled for that column in dashboard filters, and time-range queries may produce incorrect results for data recorded outside UTC. Since this YAML is the bootstrap source for the production Superset instance at `bi.grottocenter.org`, it needs to match the current DDL. Regenerate it with `superset export-zip -t dataset` and replace the stale file.

2. **[sql/9_06_2026_06_03_scientific_observations_seed.sql:87–91]** The five `t_time_series` inserts use bare timestamp literals for `start_date` and `end_date` (e.g., `'2025-06-01 00:00:00'`) without an explicit timezone offset. Round 3 fixed exactly this pattern for the four `observation_date` literals (appending `+00:00`). Applying the same fix to these five rows would make all timestamp literals in the seed file consistently timezone-explicit, and avoids relying on the session timezone being UTC.

3. **[.github/workflows/superset.yml:52,59,62,65]** The workflow pins actions by mutable major-version tags (`actions/checkout@v6`, `docker/login-action@v4`, `docker/setup-buildx-action@v4`, `docker/build-push-action@v7`) rather than commit SHAs. A force-pushed tag at any of those upstream repos could silently inject malicious code into the CI/CD pipeline. This is a higher-than-usual risk here because the workflow writes to GHCR and triggers an Azure deployment webhook that restarts the production Superset instance. GitHub's security-hardening guide recommends pinning to full SHAs (e.g., `actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683`). Consider switching to SHA pinning together with a tool like Dependabot or `pin-github-action` to keep them updated.

### Nitpicks (Optional)

4. **[docker/docker-compose.superset.yml:14]** `SUPERSET_ADMIN_PASSWORD=admin` is hardcoded in the local dev compose file, and `docker-entrypoint.sh` falls back to `changeme` if the variable is absent. A short inline comment (e.g., `# override in production`) on that line would make the risk explicit for anyone who runs this file outside a strictly local context, consistent with the guidance already in the README.

ClemRz:
Thanks @Paul-AUB! Here's what I addressed from round 4:

**Fixed (Must Fix):**

1. ✅ `observation_date` type in Superset dataset YAML → `TIMESTAMP WITH TIME ZONE` — now matches the DDL.

**Fixed (Suggestions):**

2. ✅ Added `+00:00` to all 10 bare `start_date`/`end_date` literals in the `t_time_series` seed INSERT — consistent with `observation_date` and the `::timestamptz` casts elsewhere in the file.
4. ✅ Added `# override in production` inline comment on `SUPERSET_ADMIN_PASSWORD=admin` in the compose file.

**Not addressed (intentional):**

3. SHA-pinning GitHub Actions is a valid security hardening concern, but it's orthogonal to this schema PR. I'll track it as a separate chore task to avoid scope creep here.

All 2490 tests passing.

Paul-AUB (changes requested):
All Must Fix items from round 4 have been resolved — `observation_date` in the Superset YAML is now `TIMESTAMP WITH TIME ZONE`, the `t_time_series` seed timestamps carry `+00:00`, and the docker-compose password carries the `# override in production` comment. One new Must Fix surfaced during this pass.

### Issues (Must Fix)

1. **[docker/superset/export/datasets/Grottocenter_PostgreSQL/v_measurement_wide_1.yaml:239]** The YAML lists `column_name: id_sensor`, but the current DDL view exposes the column as `ts.id_sensor_configuration` with no alias (line 396 of `9_04_…ddl.sql`), so the actual PostgreSQL column name is `id_sensor_configuration`. When `superset import-directory` bootstraps a fresh instance from this YAML, the `id_sensor_configuration` column will have no configured metadata (not marked filterable, no type hint), while the stale `id_sensor` entry will silently reference a column that does not exist in the view. Fix: either add `AS id_sensor` to the view's last SELECT column, or rename the YAML entry to `id_sensor_configuration`. The fastest path is to add the alias in the DDL (`ts.id_sensor_configuration AS id_sensor`) since the YAML was clearly generated from a version of the view that used that alias — but either way the two artefacts must agree.

2. **[.github/workflows/superset.yml:52,59,62,65]** SHA-pinning of CI actions was deferred in round 4 as out of scope. The concern stands: this workflow pushes to GHCR and triggers an Azure deployment webhook for the production Superset instance, so a compromised mutable tag at any of the four upstream repos (`actions/checkout@v6`, `docker/login-action@v4`, `docker/setup-buildx-action@v4`, `docker/build-push-action@v7`) becomes a supply-chain risk with production impact. Tracking as a separate chore is reasonable, but worth flagging again so it doesn't slip.

Paul-AUB (approved):
All Must Fix items from rounds 1–5 have been resolved. This is a clean pass. Two new suggestions surfaced on this pass.

1. **[api/models/TPoint.js]** `TPoint.js` doesn't declare a `pointGeom` attribute, and — unlike `TMeasurement.js` — it has no disclaimer warning against Waterline `migrate: 'alter'` or `migrate: 'drop'` mode. The DDL (line 115 of `9_04_…ddl.sql`) adds a `point_geom geometry(Point, 4326)` PostGIS column and a GiST index on it. In `migrate: 'alter'` mode, Waterline would attempt to drop that column; in `migrate: 'drop'` mode it would recreate the table without it, silently removing the spatial index. The pattern established for `TMeasurement.js` — a comment block explaining the constraint — would make the intent explicit:
   ```js
   // Contains a PostGIS column (point_geom) managed exclusively by SQL migrations.
   // Do NOT run Waterline in migrate: 'alter' or 'drop' mode with this model.
   ```

2. **[docker/superset/export/datasets/Grottocenter_PostgreSQL/v_measurement_wide_1.yaml]** The YAML covers 17 of the 26 columns produced by `v_measurement_wide`. Nine columns have no entry: `value_si`, `value_display`, `cave_name`, `geom`, `quantity_unit`, `unit_si`, `unit_display`, `quantity_unit_si`, `quantity_unit_display`. Superset auto-discovers these on dataset sync, so they remain accessible — but `cave_name` (primary cave filter dimension) and `geom` (GeoJSON for map charts) would benefit from explicit `filterable: true` / `groupby: true` / type metadata. A full `superset export-zip -t dataset` regeneration from the final DDL state would capture all 26 columns.

### Nitpicks (Optional)

3. **[.github/workflows/superset.yml:52,59,62,65]** SHA-pinning of `actions/checkout@v6`, `docker/login-action@v4`, `docker/setup-buildx-action@v4`, and `docker/build-push-action@v7` was deferred in round 4 as out of scope. The risk remains: this workflow pushes to GHCR and triggers an Azure deployment webhook for production, so a compromised mutable tag is a supply-chain vector. Noting again for tracking purposes.

PR #1643, opened 2026-06-09
refactor(dates): migrate moment.js to dayjs

ClemRz:
## 🤔 What

- Replace `moment` (^2.30.1) and `moment-duration-format` (^2.3.2) with `dayjs` (1.11.13)
- Add centralized dayjs config module at `api/utils/dayjs.js` with `customParseFormat` and `duration` plugins
- Migrate all date/duration logic across the codebase

## 🤷‍♂️ Why

`moment.js` is in maintenance mode and weighs ~300 KB. `dayjs` provides an almost identical API at ~7 KB, with better tree-shaking and active maintenance. This reduces bundle size and removes a deprecated dependency.

## 🔍 How

- Created `api/utils/dayjs.js` as the single import point (pre-loads plugins)
- Migrated date parsing in `karstlink.js` and `csvHelper.js` to dayjs strict mode
- Migrated date diff in `computeEntranceDataQuality.js`
- Migrated duration formatting in `CommentService.js` (accumulates days into hours as the original did)
- Added property-based tests (`dayjs.property.test.js`) covering parsing equivalence, invalid date rejection, and duration formatting
- Removed `moment` and `moment-duration-format` from `package.json`

## 🧪 Testing

```bash
npm run test
```

All existing tests pass. New property-based tests validate behavioral equivalence with the former moment-based logic.

## 📸 Previews

N/A — backend refactor, no UI changes.


Paul-AUB (dismissed):
Clean, well-scoped migration. The new `api/utils/dayjs.js` central module is the right pattern, the property-based tests are thorough, and removing two deprecated packages in exchange for one actively-maintained one is a clear win. A few things worth looking at before merge:

1. **[api/services/CommentService.js:73–84]** The implementation accumulates `days` into hours and then builds a duration with only `{hours, minutes, seconds}`, silently discarding any `months` or `years` that a PostgreSQL interval object might contain. The original `moment.duration(Object.assign(emptyDuration, pgInterval))` accepted the full merged object and handled all fields automatically. For caving trip durations this is almost certainly irrelevant in practice, but a brief comment (e.g. `// PG interval for trip durations never includes months/years`) or an explicit assertion would make the assumption explicit and prevent surprise if the schema ever changes.

2. **[package.json:14]** `"dayjs": "1.11.13"` is pinned to an exact version rather than a semver range. Every other direct dependency in the file uses `^`. Consider using `"^1.11.13"` for consistency so that automated patch/minor updates (Dependabot, etc.) can pick it up without a manual version bump.

3. **[test/integration/2_utils/dayjs.property.test.js:164]** The Property 3 duration tests live in a top-level `describe('dayjs duration - Property 3: …')` block rather than inside the main `describe('dayjs - Property Tests', …)` that wraps Properties 1 and 2. Test reporters will show them as a separate root-level suite. Consider nesting all three under the same outer describe for consistent grouping.

### Nitpicks (Optional)

4. **[api/services/CommentService.js:73]** `Object.assign(emptyDuration, pgInterval)` mutates `emptyDuration` in place. Since `emptyDuration` is a fresh local object each invocation the result is correct, but `const merged = { ...emptyDuration, ...pgInterval }` reads as an immutable merge and is marginally clearer.

ClemRz:
Thanks for the thorough review @Paul-AUB! All four items addressed in the latest push:

1. **CommentService.js** — Added comment documenting the assumption that PG intervals for trip durations never include months/years.
2. **package.json** — Changed to `^1.11.13` for consistency with other deps and Dependabot compatibility.
3. **dayjs.property.test.js** — Nested Property 3 inside the main `describe('dayjs - Property Tests', ...)` block.
4. **CommentService.js (nitpick)** — Replaced `Object.assign` with spread syntax for clearer immutable-style merge.

All tests passing (2490 pass, 0 fail), lint clean.

PR #1644, opened 2026-06-10
Feat/1598 organization responsible geo entity

### LINKED ISSUE #1598: Organization responsible for geographic entity
# Requirements Document

## Introduction

A system to associate organizations with geographic entities (countries, regions, massifs) in Grottocenter. This enables users to identify which organization**s** is responsible for or manages a given geographic area. The relationship is constrained to **many** organization per geographic entity **to take into account the need to identify organizations with different roles (e.g., speleological organization manager, speleological rescue manager)**. Any authenticated user can create, modify, or delete these associations from a geographic entity page. If the target organization does not yet exist in Grottocenter, the user can create a minimal organization record (name only) inline during the association process.

## Glossary

- **Association_System**: The backend subsystem responsible for creating, reading, updating, and deleting links between organizations and geographic entities.
- **Organization**: A record in the t_grotto table representing a caving club, federation, or managing body. Identified by a numeric ID.
- **Geographic_Entity**: A country (TCountry), region (TISO31662), or massif (TMassif) record in the Grottocenter database.
- **Country**: A record in t_country identified by its ISO 3166-1 alpha-2 code (e.g., "FR").
- **Region**: A record in t_iso3166_2 identified by its ISO 3166-2 code (e.g., "FR-75").
- **Massif**: A record in t_massif identified by a numeric auto-increment ID.
- **Association**: A link between one Organization and one Geographic_Entity, stored in a dedicated junction table. Each Geographic_Entity can have **many** Association.
- **Registered_User**: An authenticated caver with a valid JWT token who is not banned and not soft-deleted.

## API Endpoints

| Method | Path | Description |
|--------|------|-------------|
| PUT | `/api/v1/countries/:id/organization` | Set or update the associated organization for a country |
| DELETE | `/api/v1/countries/:id/organization` | Remove the organization association from a country |
| PUT | `/api/v1/countries/:countryId/regions/:regionId/organization` | Set or update the associated organization for a region |
| DELETE | `/api/v1/countries/:countryId/regions/:regionId/organization` | Remove the organization association from a region |
| PUT | `/api/v1/massifs/:id/organization` | Set or update the associated organization for a massif |
| DELETE | `/api/v1/massifs/:id/organization` | Remove the organization association from a massif |

### Existing Endpoints Modified

| Method | Path | Change |
|--------|------|--------|
| GET | `/api/v1/countries/:id` | Include associated organizations as child entities in the response |
| GET | `/api/v1/countries/:countryId/regions/:regionId` | Include associated organizations as child entities in the response |
| GET | `/api/v1/massifs/:id` | Include associated organizations as child entities in the response |
| GET | `/api/v1/organizations/:id` | Include managed geographic entities (countries, regions, massifs) as child entities in the response |

### Requirement 1: Associate an Organization with a Geographic Entity

**User Story:** As a Registered_User, I want to associate an existing organization with a geographic entity, so that other users can identify which organization is responsible for that area.

#### Acceptance Criteria

1. WHEN a Registered_User submits a PUT request with a valid organization ID for a Geographic_ THE Association_System SHALL create the Association and return a response containing the Geographic_Entity identifier and the associated organization ID.
2. IF a Registered_User submits a PUT request with an organization ID that does not exist or references a soft-deleted organization, THEN THE Association_System SHALL reject the request and return a not-found error indicating the organization was not found.
3. IF a Registered_User submits a PUT request with a Geographic_Entity ID that does not exist, THEN THE Association_System SHALL reject the request and return a not-found error indicating the geographic entity was not found.
4. IF a Registered_User submits a PUT request with a request body that does not contain a valid numeric organizationId, THEN THE Association_System SHALL reject the request and return a bad-request error indicating the required field is missing.

### Requirement 2: Remove an Association

**User Story:** As a Registered_User, I want to remove the association between an organization and a geographic entity, so that outdated or incorrect links can be corrected.

#### Acceptance Criteria

1. WHEN a Registered_User submits a DELETE request for a Geographic_Entity that has an existing Association, THE Association_System SHALL remove all Association records from the junction table and return a 204 No Content response with an empty body.
2. IF a Registered_User submits a DELETE request for a Geographic_Entity that has no Association to any organization, THEN THE Association_System SHALL return a not-found error indicating that no association exists for the specified Geographic_Entity.
3. IF a Registered_User submits a DELETE request with a Geographic_Entity ID that does not correspond to any existing record in the respective entity table (TCountry, TISO31662, or TMassif), THEN THE Association_System SHALL return a not-found error indicating that the Geographic_Entity does not exist.

### Requirement 3: Include the Associated Organization in Geographic Entity Responses

**User Story:** As a user, I want to see which organizations is associated with a geographic entity when I view that entity's page, so that I can contact the responsible organization.

#### Acceptance Criteria

1. WHEN any user requests a Geographic_Entity (country, region, or massif) via its existing GET endpoint, THE Association_System SHALL include all associated Organizations as a child entity in the response if an or many Association exists.
2. WHEN the associated Organization is not soft-deleted, THE response SHALL include the Organizations' ID, main names (the TName record marked as isMain), and the names' language.
3. WHEN the associated Organization is soft-deleted (isDeleted is true), THE response SHALL include the Organization's ID, isDeleted flag, and redirectTo value (if set), without exposing the full organization details.
4. WHEN a Geographic_Entity has no Association, THE response SHALL include a null value for the organization field.
5. IF the associated Organization has a non-null redirectTo value, THEN THE response SHALL include the redirectTo Organization ID so the client can resolve the merged entity.

### Requirement 4: Include Managed Geographic Entities in Organization Response

**User Story:** As a user, I want to see which geographic entities an organization manages when I view that organization's page, so that I can understand the scope of that organization.

#### Acceptance Criteria

1. WHEN any user requests an Organization via its existing GET endpoint, THE Association_System SHALL include the associated Geographic_Entities as child entities in the response, grouped by type (countries, regions, massifs).
2. FOR each associated country, THE response SHALL include its ISO 3166-1 alpha-2 code and its localized names (nativeName, enName, frName, esName, deName, bgName, itName, caName, nlName, rsName).
3. FOR each associated region, THE response SHALL include its ISO 3166-2 code and its localized names (name, nameEn, nameFr, nameEs, nameDe, nameBg, nameIt, nameCa, nameNl).
4. FOR each associated massif, THE response SHALL include its numeric ID and its names from the TName collection.
5. WHEN an Organization has no Associations, THE response SHALL include an empty list for each of the three groups (countries, regions, massifs).

### Requirement 5: Authorization and Access Control

**User Story:** As a platform operator, I want association management endpoints to enforce proper authorization, so that only authenticated users can modify associations.

#### Acceptance Criteria

1. THE Association_System SHALL require a valid JWT token (tokenAuth policy) for PUT and DELETE endpoints.
2. IF a request to a PUT or DELETE endpoint does not include a valid JWT token (missing, expired, or malformed), THEN THE Association_System SHALL return an HTTP 401 Unauthorized response.
3. THE Association_System SHALL allow any Registered_User to create, update, or delete Associations regardless of role (User, Leader, Moderator, Administrator), requiring only a valid authenticated session with no additional role-based restriction.
4. THE Association_System SHALL allow unauthenticated access to all GET endpoints without requiring a JWT token.

### Requirement 6: Data Integrity

**User Story:** As a platform operator, I want the association data to remain consistent, so that orphaned or invalid links do not accumulate.

#### Acceptance Criteria

1. THE Association_System SHALL store the ID of the caver who created each Association (author) and the ID of the caver who last modified it (reviewer), as separate fields.
2. THE Association_System SHALL store the timestamp of creation (dateInscription) and the timestamp of last modification (dateReviewed) for each Association.
3. IF an Organization referenced by an Association is soft-deleted, THEN THE Association_System SHALL retain the Association record and the GET endpoints SHALL include the organization's isDeleted flag set to true and, when applicable, the redirectTo identifier of the merged entity.
4. THE Association_System SHALL perform Association creation within a database transaction so that if any step fails, all changes are rolled back and no partial data is persisted.

dawoldo:
## 🤔 What

- Adds junction tables and their corresponding Waterline models (`JOrganizationCountry`, `JOrganizationRegion`, `JOrganizationMassif`) to the database schema.
- Exposes new API endpoints to set and remove organizations associated with each type of geographic entity.
- Updates existing `find` routes for geographic entities to return arrays of associated organizations.
- Updates the Swagger API documentation with the newly created endpoints and schemas.
- Adds SQL database fixtures for organization management in the development environment.
- Resolves #1598

## 🔍 How

- **Database & Models:** Created `JOrganizationCountry`, `JOrganizationRegion`, `JOrganizationMassif` Waterline models. Added the corresponding SQL migration scripts to create the junction tables, explicitly ensuring the removal of unique constraints on the geographic entity foreign keys.
- **Service Layer:** Introduced `GeoAssociationService` to cleanly centralize the business logic for adding and removing geographic associations, preventing redundant code across controllers.
- **Controllers & Routes:** Added `set-organization` and `remove-organization` controllers for countries, regions, and massifs. Wired them up in `config/routes.js` and secured them via `config/policies.js`.
- **Find Endpoints:** Updated `api/controllers/v1/{country,region,massif}/find.js` to ensure GET responses correctly query and populate the array of associated organizations.
- **Documentation & Testing:** Synchronized OpenAPI specs (`assets/swaggerV1.yaml`) with the changes and introduced integration tests for the new association endpoints to ensure stability.


ClemRz (changes requested):
### Issues (Must Fix)

- **[test/customSQL.js]** The SQL migration creates 6 indexes (`idx_j_organization_country_grotto`, `idx_j_organization_country_country`, `idx_j_organization_region_grotto`, `idx_j_organization_region_region`, `idx_j_organization_massif_grotto`, `idx_j_organization_massif_massif`) but none of them are added to `test/customSQL.js` in the `INDEX_OPTIMIZATION_MIGRATION` block. Per project conventions, indexes defined in SQL files must also be added there so the test database (created by Waterline `migrate:drop`) matches production. Without this, tests may pass by accident with small datasets but the test DB schema diverges from production.

- **[test/fixtures + test/fixtureOrder.js]** No JSON fixture files are provided for the new junction models (`jorganizationcountry.json`, `jorganizationregion.json`, `jorganizationmassif.json`), and they aren't registered in `test/fixtureOrder.js`. The SQL mock file (`9_05_mock_organization_geo_associations.sql`) only seeds the dev database — the test database is seeded via Fixted (JSON fixtures). This means the tests that assert "should add the organization without replacing existing ones" and check for pre-existing associations are relying on associations created earlier in the same test file rather than on known fixture state. While the tests may still pass because of test ordering, there's no stable fixture data for these junction tables. Consider adding JSON fixture files and registering them in `fixtureOrder.js` (after `tgrotto` and `tmassif`/`tiso31662`/`tcountry`) to ensure a predictable test state.

- **[config/policies.js:236-237]** The `massif/set-organization` and `massif/remove-organization` routes only apply `'tokenAuth'` but the massif route parameters include a numeric `:id` and `:organizationId`. Other massif write operations (like `massif/find`) use `['validateId']`. Without it, malformed or out-of-range IDs (e.g., `99999999999`) will bypass the early check and hit the database adapter directly, potentially causing unhandled errors. Consider using `['tokenAuth', 'validateId']` for these endpoints:
  ```suggestion
  'v1/massif/set-organization': ['tokenAuth', 'validateId'],
  'v1/massif/remove-organization': ['tokenAuth', 'validateId'],
  ```

- **[api/services/GeoAssociationService.js:33]** The JSDoc comment says "Enforces that a geographic entity can only have one managing organization" but the actual behavior (per the issue update and the code) supports many organizations per entity. Consider updating the comment to match reality.

- **[api/controllers/v1/country/find.js:12-15]** The `getFormattedOrganizations` call is `await`ed without a try/catch. If it throws (e.g., DB connection issue), the request will result in an unhandled error. The `region/find-by-country.js` controller wraps similar logic in a try/catch. Consider adding consistent error handling here, or wrapping the whole controller in a try/catch pattern similar to the region controller.

- **[api/services/GeoAssociationService.js:51]** The `organizationId` validation uses `!organizationId || Number.isNaN(parseInt(organizationId, 10))`. The `!organizationId` check will consider `"0"` and `0` as invalid (since `!0` is true). While organization IDs naturally start at 1 in PostgreSQL sequences, an explicit `parseInt(organizationId, 10) <= 0` check would be clearer about intent. Also consider validating against PostgreSQL's max integer (2147483647) like the `validateId` policy does.

- **[api/services/GeoAssociationService.js:186-200]** The `getFormattedOrganizations` function executes 3 sequential queries (find junctions → find grottos → find names). For entities with many organizations, this is fine, but the pattern would benefit from using Waterline's `.populate('grotto')` on the junction model query to reduce round-trips. Not urgent, but worth considering for performance.

- **[sql/9_04_2026_06_05_organization_geo_associations.sql]** The migration creates tables without `IF NOT EXISTS`, which is fine for initial creation but makes re-running the script non-idempotent. Per project conventions, consider using `CREATE TABLE IF NOT EXISTS` for resilience during local development.

- **[assets/swaggerV1.yaml – Organization schema]** The `managedCountries`, `managedRegions`, and `managedMassifs` arrays in the Organization schema use bare `type: object` without defining properties. Consider adding the property schemas (matching what the `toOrganization` converter actually returns) for better API documentation and client code generation.

### Nitpicks (Optional)

- **[api/controllers/v1/country/set-organization.js:19-22 / massif/set-organization.js:19-22]** The success response uses `res.ok({ ... })` directly. Per project conventions, success responses should go through `ControllerService.treat` or `ControllerService.treatAndConvert`. While `res.ok()` works, using `ControllerService.treat` would be more consistent with the rest of the codebase.

- **[test/integration/4_routes/Countries/set-organization.test.js]** The test sends `organizationId` in both the URL path and the request body (`.send({ organizationId: ... })`). Since the controller reads from `req.param('organizationId')` which resolves from the URL path, the body payload is unused. This is slightly misleading — consider removing the `.send()` calls from PUT requests since the ID is in the URL.

- **[api/services/GeoAssociationService.js:89-91]** When updating an existing junction, `grotto: orgId` is set again even though it's part of the query criteria and doesn't change. This is harmless but redundant.

ClemRz (dismissed):
Good work addressing the feedback from the previous review — fixtures, indexes in `test/customSQL.js`, `validateId` on massif routes, `ControllerService.treat` for success paths, `.populate('grotto')`, `IF NOT EXISTS`, and full Swagger schemas are all in place now.

- **[api/controllers/v1/massif/find.js:22-25]** The `getFormattedOrganizations` call is not wrapped in a try/catch. If it throws, Sails will return a generic 500 without the structured error logging present in the country and region controllers. Consider wrapping the whole controller body in a try/catch (like `country/find.js` and `region/find-by-country.js` now do) for consistent error handling.

- **[config/policies.js:217-218]** The country and region `set-organization` / `remove-organization` routes only apply `'tokenAuth'` without `validateId`. This is correct since country/region IDs are strings, but it means `organizationId` in the URL path (e.g., `/countries/FR/organizations/abc`) is validated at the service layer (returning 400) rather than at the policy layer (returning 404 like the massif routes do). This is a minor behavioral inconsistency — an invalid `organizationId` on country/region routes returns 400, while on massif routes it returns 404. Not blocking, but worth noting for API consistency.

### Nitpicks (Optional)

- **[api/controllers/v1/country/set-organization.js:9-11 / remove-organization.js:5-11]** The `if (!countryId)` and `if (!organizationId)` guards are unreachable since both come from required URL path parameters (`:id` and `:organizationId`) — if they're missing, the route won't match and Sails returns 404 before the controller runs. Same for massif and region controllers. Harmless defense-in-depth, but could be removed for cleanliness.

- **[test/integration/4_routes/Countries/set-organization.test.js:90]** `should(junction.author).equal(3)` relies on knowing which caver ID the test token maps to. If that fixture changes, the test breaks silently. Consider extracting the expected user ID into a constant or deriving it from the token for clarity.

ClemRz (dismissed):
Good iteration on this PR — the previous review feedback has been addressed well. Fixtures, indexes in `test/customSQL.js`, `validateId` on all organization routes, `ControllerService.treat` for success paths, `.populate('grotto')`, `IF NOT EXISTS` for idempotent SQL, and full Swagger property schemas are all in place.

- **[api/services/GeoAssociationService.js:52]** The `E_BAD_REQUEST` for invalid `organizationId` (NaN, <= 0, > max int) will never be reached on routes where `validateId` is in the policy chain, since the policy rejects invalid numeric IDs before the controller runs. This is fine as defense-in-depth (the service could be called from elsewhere), but worth noting that for the current routes, an invalid organizationId will return 404 (from `validateId`) rather than 400 (from the service). If the 400 vs 404 distinction matters for API clients, consider aligning the policy response with the service behavior — or accept the inconsistency since it only affects clearly invalid input.

- **[api/controllers/v1/country/find.js:14]** The `getFormattedOrganizations` call happens outside the "entity not found" check — if `TCountry.findOne` returns a result but `getFormattedOrganizations` fails with a DB error, it will be caught by the outer try/catch and return a 500. This is correct behavior. However, there's a subtle edge case: if `countryId` is `undefined` (which can't actually happen since it's a path param), the `findOne` would match the first record. Not a real issue, just confirming the pattern is sound.

- **[test/integration/4_routes/Countries/set-organization.test.js:84-100]** The "should add the organization without replacing existing ones" test asserts `should(junctions.length).equal(3)`. This relies on: 2 from fixtures (grotto 1 + grotto 2 for FR) + the upsert of grotto 1 from the previous test (still 2) + grotto 3 from this test = 3. The test order dependency is acceptable since Mocha runs `it()` blocks sequentially within a `describe`, but a comment explaining why 3 is the expected count would help future maintainers.

- **[api/services/GeoAssociationService.js:190]** `const organizationIds = grottos.map((g) => g.id)` is declared but only used in the `TName.find` query below. If `.populate('grotto')` returns a grotto object where the organization was soft-deleted (which is possible since the junction record is retained per Requirement 6.3), the `names` query will still fetch names for it, but `getFormattedOrganizations` won't use them — it returns the `isDeleted` shape instead. This is harmless (one extra name fetched but unused), though you could optimize by only fetching names for non-deleted grottos: `const activeGrottoIds = grottos.filter(g => !g.isDeleted).map(g => g.id)`.

### Nitpicks (Optional)

- **[api/controllers/v1/massif/find.js:34]** The catch block calls `ControllerService.treatAndConvert(req, err, null, params, res, null)` with `null` as the converter. This works because when `err` is truthy, the converter is never invoked. However, it's slightly inconsistent with the country and region catch blocks which call `ControllerService.treat(req, err, null, params, res)`. Both produce the same 500 behavior, but using `treat` in all three would be more uniform.

- **[test/integration/4_routes/Massifs/set-organization.test.js:63]** The "should add without replacing" test asserts `should(junctions.length).equal(2)` — this works because: 1 from fixtures (grotto 1 → massif 1) + the upsert from previous test (still 1) + grotto 2 from this test = 2. Clear and correct.

ClemRz (dismissed):
Good work on this feature. The architecture is clean — `GeoAssociationService` centralizing the logic, proper use of transactions, and solid test coverage. Previous review feedback has been well addressed (fixtures, indexes, `validateId`, `ControllerService.treat`, `.populate('grotto')`, `IF NOT EXISTS`, Swagger schemas).

- **[api/services/GeoAssociationService.js:57-62]** The validation for invalid `organizationId` (NaN, <= 0, > max int) throws with `error.code = 'E_NOT_FOUND'`, which causes the controller to return 404. Semantically, an invalid format (e.g., `"abc"` or `-5`) is a client input error, not "resource not found". Consider using `'E_BAD_REQUEST'` for this case to return 400, and reserving 404 for the actual "organization does not exist in DB" check below. The same applies to `removeAssociation` at line ~140. That said, since `validateId` catches these before the service runs on massif routes, this mostly affects country/region routes where `organizationId` isn't validated at the policy layer.

- **[api/controllers/v1/massif/find.js:34]** The catch block calls `ControllerService.treat(req, err, null, params, res)` while the country and region catch blocks use the same pattern. However, the success path uses `ControllerService.treatAndConvert`. This is fine since the converter isn't invoked when `err` is truthy — just noting the slight asymmetry for consistency. No action needed.

- **[api/services/GeoAssociationService.js:193-197]** `getFormattedOrganizations` returns `[]` when there are no associations. The issue's Requirement 3.4 states "include a null value for the organization field" when no association exists. The empty array is arguably better for a many-to-many relationship (avoids null checks on the client), but worth confirming this divergence from the spec is intentional.

- **[api/controllers/v1/country/find.js:14]** The `getFormattedOrganizations` call runs after verifying the country exists — good. However, if `countryId` is a string like `"FR"`, the `!countryId` guard at line 5 would not catch an empty string `""` since `req.param('id')` for a matched route is always truthy (the route wouldn't match otherwise). The guard is harmless defense-in-depth but unreachable.

### Nitpicks (Optional)

- **[api/services/GeoAssociationService.js:100]** When updating an existing junction, only `reviewer` and `dateReviewed` are set. The `grotto` field (which equals `orgId`) is no longer redundantly written — good cleanup from the earlier revision.

- **[test/integration/4_routes/Countries/set-organization.test.js:37-38]** The test for "organizationId is missing" sends a PUT to `/api/v1/countries/FR/organizations/` with an empty body. This actually tests route matching (no route matches a trailing slash without the param), so it returns 404 from Sails routing, not from the controller. The test passes, but the description is slightly misleading — it's really testing "unmatched route" rather than "missing organizationId validation."

- **[test/integration/4_routes/Caves/update.test.js]** The refactoring from callback-style (`done`) to async/await is a nice cleanup. Unrelated to the organization feature but improves readability.

ClemRz (dismissed):
Good work on this feature. The architecture is solid — `GeoAssociationService` cleanly centralizes the logic, the `validateId` policy correctly skips string-based IDs while still validating `organizationId`, fixtures and test indexes are in place, and the test coverage is comprehensive. Prior review feedback (4 rounds) has been well addressed.

- **[api/services/GeoAssociationService.js:57-62]** The validation for invalid `organizationId` format (NaN, <= 0, > max int) throws with `error.code = 'E_NOT_FOUND'`, which maps to 404 in the controllers. Semantically, an invalid format (e.g., `"abc"` or `-5`) is a client input error (400), not "resource not found" (404). Since `validateId` now catches these on all routes before the service runs, this is effectively unreachable for the current API — but if the service is ever called from another code path, it would return 404 for what is logically a bad request. Consider using `'E_BAD_REQUEST'` here for correctness:
  ```js
  error.code = 'E_BAD_REQUEST';
  ```
  Same applies to `removeAssociation` at line ~140.

- **[api/controllers/v1/country/remove-organization.js:13 / region/remove-organization.js:16 / massif/remove-organization.js:13]** The `res.status(204).send()` is a raw Express response. The project conventions prefer Sails custom responses for error paths and `ControllerService.treat` for success paths. For 204 No Content there's no practical alternative since Sails doesn't provide a built-in `res.noContent()`, so this is acceptable — just flagging it for awareness in case a custom Sails response (e.g., `api/responses/noContent.js`) is ever introduced.

### Nitpicks (Optional)

- **[api/services/GeoAssociationService.js:186-189]** `getFormattedOrganizations` returns `[]` when no associations exist. The issue's Requirement 3.4 says "include a null value for the organization field." The empty array is arguably better for a many-to-many relationship (avoids null checks on the client), but worth confirming this intentional divergence from the spec is communicated to the front-end team.

- **[test/integration/4_routes/Countries/set-organization.test.js:37-38]** The "should return 404 if organizationId is missing" test sends a PUT to `/api/v1/countries/FR/organizations/` — this returns 404 because no route matches the trailing slash without the param, not because the controller validates the missing param. The test passes and protects against regressions, but the description is slightly misleading (it tests route matching, not controller validation).

ClemRz (dismissed):
Good work on this feature. The architecture is clean — `GeoAssociationService` centralizes the business logic, the `validateId` policy correctly handles string-based vs numeric IDs, junction models and SQL migration are well-structured with proper FK constraints and idempotent `IF NOT EXISTS`, fixtures are registered in `test/fixtureOrder.js`, indexes are mirrored in `test/customSQL.js`, and the test coverage is comprehensive. Prior rounds of review feedback have been addressed well.

- **[api/services/GeoAssociationService.js:57-62]** The validation for invalid `organizationId` format (NaN, <= 0, > max int) throws with `error.code = 'E_BAD_REQUEST'`, which maps to `res.badRequest()` (400) in the controllers. However, since `validateId` now runs on all six organization routes and catches invalid numeric IDs before the service runs (returning 404), this service-level validation is effectively unreachable for the current API. It's fine as defense-in-depth (the service could be called from elsewhere), but worth noting the behavioral discrepancy: `validateId` returns 404 for invalid IDs while the service returns 400. If API clients rely on the distinction, consider aligning `validateId` to return 400 for malformed IDs vs 404 for well-formed-but-nonexistent ones — or document that the 404 from `validateId` covers both cases.

- **[api/services/GeoAssociationService.js:186-189]** `getFormattedOrganizations` returns `[]` when no associations exist. The issue's Requirement 3.4 states "include a null value for the organization field." The empty array is arguably better for a many-to-many relationship (avoids null checks on the client), but worth confirming this intentional divergence from the spec is communicated to the front-end team so they handle both shapes or agree on the array contract.

- **[api/services/GeoAssociationService.js:193-197]** When a junction's `.populate('grotto')` resolves to a soft-deleted organization, `getFormattedOrganizations` correctly includes the `isDeleted` shape per Requirement 6.3. However, the `TName.find` query still runs for `activeGrottoIds` only — which is correct. One edge case: if `.populate('grotto')` returns `null` (e.g., the grotto record was hard-deleted bypassing the junction FK), `.filter(Boolean)` handles it gracefully. Just confirming this is solid.

- **[api/controllers/v1/massif/find.js:25]** The `organizations` field is assigned directly to the `massif` object (`massif.organizations = ...`) before passing it to `toMassif`. The `toMassif` converter passes it through verbatim (`organizations: source.organizations`). This works because `getFormattedOrganizations` already returns the formatted shape. However, other collections in `toMassif` go through `toList()` with a sub-converter — the inconsistency is minor but means there's no schema enforcement on the organizations array at the converter level. If the `getFormattedOrganizations` shape ever changes, the converter won't catch the mismatch. Not blocking, just a design note.

- **[api/controllers/v1/country/remove-organization.js:13 / region/remove-organization.js:16 / massif/remove-organization.js:13]** The `res.status(204).send()` is a raw Express response. The project conventions prefer Sails custom responses for error paths and `ControllerService.treat` for success paths. For 204 No Content there's no built-in Sails alternative, so this is acceptable — just noting it for awareness in case a custom `api/responses/noContent.js` is ever introduced.

### Nitpicks (Optional)

- **[test/integration/4_routes/Countries/set-organization.test.js:37-38]** The "should return 404 if organizationId is missing" test sends a PUT to `/api/v1/countries/FR/organizations/` — this returns 404 because no route matches the trailing slash without the param, not because the controller validates the missing param. The test passes and protects against regressions, but the description is slightly misleading. Consider renaming to something like "should return 404 when organizationId path segment is absent" to clarify it's testing route matching.

- **[test/integration/4_routes/Caves/update.test.js]** The refactoring from callback-style (`done`) to async/await is a nice cleanup. The change to non-deterministic entrance order assertion (`containEql` instead of index-based) is a good improvement. Unrelated to the organization feature but improves test stability.

- **[api/services/GeoAssociationService.js:1-5]** The file header comment is clean, but `mapEntityType` could benefit from a brief JSDoc explaining that it maps entity type strings to their Waterline model names and junction config — helpful for new contributors navigating the service.

ClemRz (dismissed):

- **[api/services/GeoAssociationService.js:57-62]** The `organizationId` format validation (NaN, <= 0, > max int) throws with `error.code = 'E_BAD_REQUEST'` (maps to 400). However, since `validateId` now runs on all six organization routes and rejects invalid numeric IDs before the service runs (returning 404), this service-level validation is effectively unreachable for the current API. The behavioral discrepancy (404 from `validateId` vs 400 from the service) is fine as defense-in-depth, but worth documenting with a short comment so future maintainers don't wonder about the inconsistency. Same applies to `removeAssociation` around line 140.

- **[api/services/mapping/converters.js:684-720]** `managedCountries`, `managedRegions`, and `managedMassifs` use `source.managedCountries?.map(...)` which returns `undefined` when the field is absent. This differs from the `toList` pattern used elsewhere, which returns `[]` for missing arrays. In practice `getPopulatedOrganization` always populates them (so they'll be empty arrays, not undefined), but for robustness — and to match the Swagger schema (which declares them as `type: array`) — consider defaulting: `managedCountries: source.managedCountries?.map(...) || []`. This prevents `undefined` from leaking to the response if the converter is ever called without prior population.

- **[api/services/GeoAssociationService.js:186-189]** `getFormattedOrganizations` returns `[]` when no associations exist. The issue's Requirement 3.4 states "include a null value for the organization field." The empty array is arguably better for a many-to-many relationship (avoids null checks on the client), and the Swagger schemas declare it as an array — so this seems intentional. Consider adding a brief note in the issue or PR description to confirm this intentional divergence from the spec wording.

- **[api/controllers/v1/country/remove-organization.js:13 / region/remove-organization.js:16 / massif/remove-organization.js:13]** `res.status(204).send()` is a raw Express response. The project conventions prefer Sails custom responses or `ControllerService.treat` for the success path. Since Sails doesn't provide a built-in `res.noContent()`, this is pragmatic and acceptable — just flagging for awareness in case a custom `api/responses/noContent.js` is ever introduced.

### Nitpicks (Optional)

- **[test/integration/4_routes/Countries/set-organization.test.js:37-38]** The "should return 404 when organizationId path segment is absent" test sends a PUT to `/api/v1/countries/FR/organizations/` — this returns 404 because no route matches the trailing slash without the param, not because the controller validates anything. The description "when organizationId path segment is absent" is accurate (good improvement from a previous iteration), but a reader might still assume controller validation is being tested. Minor — no action needed.

- **[test/integration/4_routes/Caves/update.test.js]** The refactoring from callback-style (`done`) to async/await and the switch to non-deterministic entrance order assertion (`containEql` instead of index-based) is a nice cleanup. Unrelated to the organization feature but improves test stability.

ClemRz (dismissed):

ClemRz:
## 🚀 Deployment Procedure

This PR includes a SQL migration that must be run **before** deploying the new application code.

### Steps

1. **Put the API in maintenance mode** (or schedule during low-traffic window)
2. **Run the migration script** against the production database:
   ```sql
   \i sql/9_04_2026_06_05_organization_geo_associations.sql
   ```
   This script uses `IF NOT EXISTS` / `IF EXISTS` — it is idempotent and safe to re-run.
3. **Deploy the application code**
4. **Verify**:
   - `GET /api/v1/countries/FR` returns an `organizations` array
   - `GET /api/v1/massifs/1` returns an `organizations` array
   - `POST /api/v1/countries/:iso/organizations` with a valid body returns 200 (requires Moderator+ role)
   - `SELECT COUNT(*) FROM j_organization_country` returns expected rows
   - `SELECT COUNT(*) FROM j_organization_region` returns expected rows
   - `SELECT COUNT(*) FROM j_organization_massif` returns expected rows

### What the migration does

| Step | Action |
|------|--------|
| 1 | Creates `j_organization_country` junction table (grotto ↔ country, with author/reviewer tracking) |
| 2 | Creates `j_organization_region` junction table (grotto ↔ region) |
| 3 | Creates `j_organization_massif` junction table (grotto ↔ massif) |
| 4 | Adds indexes on foreign key columns for all three tables |
| 5 | Adds UNIQUE constraints on `(id_grotto, id_country)`, `(id_grotto, id_region)`, `(id_grotto, id_massif)` to prevent duplicate associations |

### ⚠️ Important notes

- **Order matters**: migration must run before code deployment. The new code references the junction tables (`JOrganizationCountry`, `JOrganizationRegion`, `JOrganizationMassif`) and will fail with "relation does not exist" errors if the tables are missing.
- **No data loss risk**: this migration only creates new tables — it does not modify or drop existing structures.
- **Mock data**: `sql/9_05_mock_organization_geo_associations.sql` is for development only. Do **not** run it in production.
- **Rollback**: drop the three junction tables if needed:
  ```sql
  DROP TABLE IF EXISTS j_organization_massif;
  DROP TABLE IF EXISTS j_organization_region;
  DROP TABLE IF EXISTS j_organization_country;
  ```

### New endpoints

| Method | Route | Auth | Description |
|--------|-------|------|-------------|
| POST | `/api/v1/countries/:iso/organizations` | Moderator+ | Associate an organization with a country |
| DELETE | `/api/v1/countries/:iso/organizations/:organizationId` | Moderator+ | Remove association |
| POST | `/api/v1/regions/:id/organizations` | Moderator+ | Associate an organization with a region |
| DELETE | `/api/v1/regions/:id/organizations/:organizationId` | Moderator+ | Remove association |
| POST | `/api/v1/massifs/:id/organizations` | Moderator+ | Associate an organization with a massif |
| DELETE | `/api/v1/massifs/:id/organizations/:organizationId` | Moderator+ | Remove association |

ClemRz (approved):
.

PR #1648, opened 2026-06-10
fix(search): validate columns and columnsName in export endpoint

### LINKED ISSUE #1645: bug(api): [500] search - undefined columnsName crashes advanced-search-export
**Describe the bug**
The `POST /api/v1/advanced-search/export` endpoint crashes with a `TypeError` when `columnsName` is undefined or not provided in the request body. At line 143 of `advanced-search-export.js`, the code calls `.map()` on `columnsName` without checking if it's defined first.

**To Reproduce**
Based on the trace logs:
1. `POST /api/v1/advanced-search/export` from the production front-end
2. The request either omits `columnsName` or sends it as `null`/`undefined`
3. The controller proceeds past the Typesense search, then crashes when writing the CSV header

**Input Data**

- **Endpoint:** `POST /api/v1/advanced-search/export`
- **Request Body:** [not logged — only client metadata visible in trace]
- **User:** [not logged]

**Expected behavior**
The API should validate that `columnsName` is present and is an array before proceeding. If missing, it should return a 400 Bad Request response instead of crashing with a 500 error.

**Error Details**
- Error Type: `TypeError: Cannot read properties of undefined (reading 'map')`
- Occurrences: 2
- Time Range: 2026-06-08T07:10:38Z to 2026-06-08T07:11:59Z
- Representative Trace ID: `019ea612-f1e3-7162-a1b2-cf5d1ac73891`
- Stack Trace:
```
TypeError: Cannot read properties of undefined (reading 'map')
    at module.exports [as v1/search/advanced-search-export] (/home/site/wwwroot/api/controllers/v1/search/advanced-search-export.js:143:32)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
```

**Additional context**
- Service: `api/controllers/v1/search/advanced-search-export.js`
- Model: `search` (no DB model — Typesense search)
- DB Error Code: N/A (runtime TypeError, not a database error)
- Note: A secondary `ERR_HTTP_HEADERS_SENT` error occurs because the streaming response has already sent headers before the crash happens on subsequent iterations, but the root cause is the missing `columnsName` validation.
- The same undefined check should likely be applied to `columns` as well (line 142 uses it in `documentsToCSV`).

ClemRz:
## 🤔 What

- Validate `columns` and `columnsName` request body parameters in `POST /api/v1/advanced-search/export`
- Return 400 when either parameter is missing or is an empty array
- Add 4 integration tests covering the validation scenarios
- Closes #1645

## 🤷‍♂️ Why

Production 500 errors (`TypeError: Cannot read properties of undefined (reading 'map')`) when clients call the export endpoint without providing `columnsName` in the request body. The code called `.map()` on `undefined` without any guard.

## 🔍 How

Added early validation at the top of the controller, checking both `columns` and `columnsName` are non-empty arrays before proceeding. Follows the existing pattern of `res.badRequest(...); return;` used elsewhere in the file.

## 🧪 Testing

```bash
npm run test -- --grep "Advanced Search Export"
```

All 18 tests pass (14 existing + 4 new validation tests).

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/api/controllers/v1/search/advanced-search-export.js
+++ b/api/controllers/v1/search/advanced-search-export.js
@@ -94,6 +94,18 @@ module.exports = async (req, res) => {
     res.badRequest('columnsName must be a non-empty array');
     return;
   }
+  if (columns.length !== columnsName.length) {
+    res.badRequest('columns and columnsName must have the same length');
+    return;
+  }
+  if (!columns.every((c) => typeof c === 'string' && c.length > 0)) {
+    res.badRequest('each element in columns must be a non-empty string');
+    return;
+  }
+  if (!columnsName.every((c) => typeof c === 'string' && c.length > 0)) {
+    res.badRequest('each element in columnsName must be a non-empty string');
+    return;
+  }
 
   const params = {
     query: req.param('query'),
--- a/api/controllers/v1/search/advanced-search-export.js
+++ b/api/controllers/v1/search/advanced-search-export.js
@@ -85,6 +85,16 @@ module.exports = async (req, res) => {
   if (!matchAllFields || matchAllFields === 'false') matchAllFields = false;
   const columns = req.param('columns');
   const columnsName = req.param('columnsName');
+
+  if (!Array.isArray(columns) || columns.length === 0) {
+    res.badRequest('columns must be a non-empty array');
+    return;
+  }
+  if (!Array.isArray(columnsName) || columnsName.length === 0) {
+    res.badRequest('columnsName must be a non-empty array');
+    return;
+  }
+
   const params = {
     query: req.param('query'),
     entity: req.param('entity') ?? '',
--- a/test/integration/4_routes/AdvancedSearchExport.test.js
+++ b/test/integration/4_routes/AdvancedSearchExport.test.js
@@ -75,6 +75,63 @@ describe('Advanced Search Export features', () => {
         });
     });
 
+    it('should return 400 when columns and columnsName have different lengths', (done) => {
+      supertest(sails.hooks.http.app)
+        .post('/api/v1/advanced-search/export')
+        .send({
+          query: 'test',
+          entity: 'organizations',
+          columns: ['id', 'name'],
+          columnsName: ['ID'],
+        })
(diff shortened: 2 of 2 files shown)


Paul-AUB (dismissed):

1. **[api/controllers/v1/search/advanced-search-export.js:89-96]** The validation correctly guards against missing/empty arrays but does not check that `columns.length === columnsName.length`. If a caller sends 3 column keys but only 2 header names (or vice versa), the exported CSV will have a mismatched header row vs. data rows — a silent data quality issue that could confuse consumers of the file. Consider adding:
   ```js
   if (columns.length !== columnsName.length) {
     res.badRequest('columns and columnsName must have the same length');
     return;
   }
   ```

### Nitpicks (Optional)

2. **[api/controllers/v1/search/advanced-search-export.js:89-96]** The array checks do not validate that each element is a non-empty string. For example, `columns: [null]` or `columns: [42]` passes the current guard and would later cause `key.includes('.')` to throw in `resolveField`. This is a low-probability edge case since the callers are likely trusted front-end clients, but a `columns.every(c => typeof c === 'string' && c.length > 0)` check would make the contract explicit.

3. **[test/integration/4_routes/AdvancedSearchExport.test.js:12-76]** The mismatched-length scenario (e.g. `columns: ['id', 'name'], columnsName: ['ID']`) is not covered. If suggestion #1 is implemented, a corresponding test would be worth adding.

---

The fix is minimal, correct, and follows the existing `res.badRequest(...); return;` pattern used elsewhere in the file. The four new integration tests cleanly cover both missing and empty-array cases for each parameter. Good catch on guarding `columns` as well as `columnsName`.

ClemRz:
Thanks for the review @Paul-AUB! All three suggestions implemented in cc7b9197:

1. ✅ Added `columns.length !== columnsName.length` check
2. ✅ Added element-type validation (`typeof c === 'string' && c.length > 0`) for both arrays
3. ✅ Added 3 integration tests covering the new validation cases (mismatched length, non-string element, empty string)

PR #1649, opened 2026-06-10
fix(name): remove populate point from name controllers

### LINKED ISSUE #1647: Définir la langue sur les organisation ou sur un massif génère une erreur à l'édition
Quand on associe une langue a une organisation ou a un massif à l'édition on a un message d'erreur. A priori les modifications sont bien enregistrées sauf le champ langue

Image

Voila l'erreur qui apparait dans la console

Image

ClemRz:
## 🤔 What

- Remove `.populate('point')` from the name `update` and `set-as-main` controllers
- Closes #1647

## 🤷‍♂️ Why

The production `t_point` table was restructured (the `type` column was removed). When updating or setting a name as main, the controllers called `.populate('point')` which triggered a Waterline JOIN generating `t_point__point.type` — a column that no longer exists, causing `"Unexpected error from database adapter: column t_point__point.type does not exist"`.

## 🔍 How

The `toName` converter never uses the `point` association in its response — it only maps `id`, `name`, `isMain`, `language`, dates, `isDeleted`, `author`, and `reviewer`. The `.populate('point')` call was unnecessary and is now removed from both controllers.

## 🧪 Testing

- PATCH `/api/v1/names/:id` with a valid name body should succeed without database errors
- POST `/api/v1/names/:id/setAsMain` should succeed without database errors

## 📸 Previews

N/A

### DIFF
--- a/api/controllers/v1/name/set-as-main.js
+++ b/api/controllers/v1/name/set-as-main.js
@@ -55,7 +55,6 @@ module.exports = async (req, res) => {
       .populate('grotto')
       .populate('language')
       .populate('massif')
-      .populate('point')
       .populate('reviewer');
 
     const params = {};
--- a/api/controllers/v1/name/update.js
+++ b/api/controllers/v1/name/update.js
@@ -35,7 +35,6 @@ module.exports = async (req, res) => {
       .populate('grotto')
       .populate('language')
       .populate('massif')
-      .populate('point')
       .populate('reviewer');
 
     const params = {};


Paul-AUB (approved):

1. **[api/models/TPoint.js:21-26]** The root cause of #1647 is that the `TPoint` model still declares the `type` attribute (`columnName: 'type'`) even though the column was removed from the production `t_point` table. This PR removes the only remaining `.populate('point')` calls, so the symptom is fixed, but the model remains out of sync with the schema: any future query or populate touching `TPoint` will reintroduce the same `column ... does not exist` error. Consider removing the stale `type` attribute from the model in this PR (or a follow-up) so the trap is fully defused.

2. **[api/controllers/v1/name/update.js:31-38, api/controllers/v1/name/set-as-main.js:51-58]** Since `toName` only converts `author` and `reviewer` to objects (and passes `language` through), the `.populate('cave')`, `.populate('entrance')`, `.populate('grotto')`, and `.populate('massif')` calls are also unused by the response. They each generate an extra JOIN/query per request. Not blocking, but they could be removed in the same spirit as this fix for a small performance win and to avoid similar surprises if those tables are ever restructured.

### Nitpicks (Optional)

3. **[api/controllers/v1/name/set-as-main.js:16-27]** Pre-existing, but worth noting while in this file: the early-return path for an already-main name passes the unpopulated `currentName` to `toName`, so `author`/`reviewer` come back as raw ids instead of objects, unlike the success path. The response shape differs depending on whether the name was already main.

---

Verified: `toName` (api/services/mapping/converters.js:566) never reads `source.point`, and no other `.populate('point')` call remains in the codebase. The fix is correct and minimal. LGTM.

PR #1651, opened 2026-06-10
fix(search): use facet search in field-search to fix wrong massif results

### LINKED ISSUE #1646: Les massifs affichent les entrées d'un autre massif
Quand vous allez sur la page d'un massif qui n'a pas de périmètre défini la liste des cavités proposées est la liste associée à un autre massif

Le test peut être réalisé sur le massif suivant

- causse d'Aumelas : https://grottocenter.org/ui/massifs/128 donne les entrées du causse de Viols-le-Fort
Il ne devra

Pour les massifs qui n'ont pas de périmètre il ne devrait pas y avoir de liste de cavités

Pour les massifs qui ont un périmètre la liste des cavités correspond a un autre massif

https://grottocenter.org/ui/massifs/1000/
https://grottocenter.org/ui/massifs/1000/entrances

ClemRz:
## Description

Fixes #1646 — Massif pages display entrances from unrelated massifs.

## Why

The `/api/v1/field-search` endpoint returned wrong massif names when searching entrances by `massifs.name`. The old implementation used Typesense's `query_by` + `group_by`, which does token-based text search: a query like "Aumelas (causse d')" was tokenized into ["aumelas", "causse", "d"], matched documents containing *any* massif with those tokens (e.g., "Viols-le-Fort (causse de)"), then grouped by *all* distinct massif names in matched documents — surfacing unrelated massifs.

## What

Switched `fieldSearch` from `query_by` + `group_by` to Typesense's **facet search** (`facet_by` + `facet_query`). Facet search matches tokens within individual facet values rather than across documents, so only massif names whose text actually matches the query are returned. No changes to the HTTP response contract.

## Testing

- All existing tests pass (2492 passing, 1 unrelated flaky failure in rate limiting)
- Updated `SearchService.test.js` with tests covering the new facet-based implementation
- Linter passes

## Checklist

- [x] Tests pass
- [x] Linter passes
- [x] No breaking changes to the API contract

### DIFF
--- a/api/services/SearchService.js
+++ b/api/services/SearchService.js
@@ -178,11 +178,18 @@ async function fieldSearch({
   // documents containing other massif names that share tokens (e.g. "causse"),
   // then group_by would surface those unrelated massif names.
   // Facet search matches against the actual facet values, not document text.
+
+  // Cap max_facet_values at 100 to avoid surfacing raw Typesense validation
+  // errors for out-of-range values (mirrors collectionSearch's per_page cap).
+  const maxFacetValues = size ? Math.min(size, 100) : size;
+
   const params = {
     q: '*',
+    // query_by is required by the Typesense API even for wildcard queries;
+    // it does not participate in matching when q is '*'.
     query_by: field,
     facet_by: field,
-    max_facet_values: size,
+    max_facet_values: maxFacetValues,
     per_page: 0, // We only need facet counts, not document hits
     ...(q !== '*' && { facet_query: `${field}:${q}` }),
     ...(filterBy && { filter_by: filterBy }),
@@ -195,10 +202,19 @@ async function fieldSearch({
   // the controller. The controller reads: found, found_docs, grouped_hits, page.
   const facetInfo = result.facet_counts?.find((f) => f.field_name === field);
   const counts = facetInfo?.counts ?? [];
+
+  // found_docs: sum of document counts for the returned facet values.
+  // Note: on array fields (e.g. massifs.name), a document with multiple
+  // matching values is counted once per value, so this may exceed the true
+  // distinct document count. This is acceptable for the UI's display purpose.
   const totalDocuments = counts.reduce((sum, c) => sum + c.count, 0);
 
+  // Use stats.total_values for the true count of distinct facet values
+  // matching the query, which is not capped by max_facet_values.
+  const totalDistinct = facetInfo?.stats?.total_values ?? counts.length;
+
   return {
-    found: counts.length,
+    found: totalDistinct,
     found_docs: totalDocuments,
     grouped_hits: counts.map((c) => ({
       group_key: [c.value],
--- a/api/services/SearchService.js
+++ b/api/services/SearchService.js
@@ -172,18 +172,40 @@ async function fieldSearch({
     isLogicalCompareAnd
   );
 
+  // Use faceted search to get distinct values for the field.
+  // The previous approach (query_by + group_by) caused wrong results on
+  // array fields like massifs.name: a token-based text search would match
+  // documents containing other massif names that share tokens (e.g. "causse"),
+  // then group_by would surface those unrelated massif names.
+  // Facet search matches against the actual facet values, not document text.
   const params = {
-    q,
+    q: '*',
     query_by: field,
-    group_by: field,
-    group_limit: 1,
-    sort_by: '_group_found:desc',
-    per_page: size,
+    facet_by: field,
+    max_facet_values: size,
+    per_page: 0, // We only need facet counts, not document hits
+    ...(q !== '*' && { facet_query: `${field}:${q}` }),
     ...(filterBy && { filter_by: filterBy }),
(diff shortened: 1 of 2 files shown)


Paul-AUB (changes requested):
Nice find on the root cause — the token-based `query_by` + `group_by` explanation matches the behavior reported in #1646, and facet search is the right tool here. The eligibility constraint is unchanged too: `group_by` already required `facet: true` fields, so any field that worked before still works. A few points below, one of which I think needs a fix before merging.

### Issues (Must Fix)

1. **[api/services/SearchService.js:201]** `found: counts.length` changes the meaning of `totalDistinct` in the HTTP response. The old grouped search returned `found` = the **total** number of distinct groups matching the query, independent of `per_page`. The new code returns at most `size` (default 10), because `counts` is truncated by `max_facet_values`. Any client that displays or paginates on `totalDistinct` will now see it silently capped at `size` (e.g., 50 distinct massifs → `totalDistinct: 10`). Typesense's facet response includes `facet_counts[].stats.total_values` (the total number of distinct values for the facet); consider using `facetInfo?.stats?.total_values ?? counts.length` instead. Worth verifying against a live instance that `total_values` reflects the `facet_query`-filtered set — if it doesn't, this semantics change should at least be called out in the PR description rather than stated as "no changes to the HTTP response contract".

2. **[api/services/SearchService.js:198]** `found_docs` (`totalDocuments` in the response) also changes meaning, in two ways: (a) it only sums counts for the top `size` facet values, so it under-counts when there are more distinct values than `size`; and (b) on array fields like `massifs.name`, a document with several matching massif names is counted once per value, so the sum can exceed the real document count. The old value was the number of documents matching the text query. This is probably acceptable for the UI use case, but consider documenting the new semantics in the code comment (and the PR description) so future readers don't assume it's a distinct-document count.

3. **[test/integration/1_services/SearchService.test.js]** All `fieldSearch` tests stub `typesense.search`, so they verify the parameters sent and the response transform, but not the actual facet-matching behavior that fixes #1646 (the "Aumelas (causse d')" vs "Viols-le-Fort (causse de)" scenario). That's consistent with how the rest of this suite works, so not blocking — but if there's any harness for tests against a real Typesense instance, a regression test reproducing the issue's query would be valuable, since this bug shipped precisely because the stub-level tests couldn't catch it.

### Nitpicks (Optional)

4. **[api/services/SearchService.js:183]** `query_by: field` is ignored by Typesense when `q: '*'`, so it now serves no purpose beyond satisfying the API's required-parameter check (if it's even still required for wildcard queries). A short comment saying it's intentionally kept for API validity (or removing it if Typesense accepts wildcard searches without it) would prevent readers from thinking text search still participates in the matching.

5. **[api/services/SearchService.js:185]** `max_facet_values` receives `size` unvalidated (the controller default is 10, but clients control it). `collectionSearch` caps `size` at 1000 for the equivalent reason; a similar cap here would avoid surfacing raw Typesense validation errors for out-of-range values. Pre-existing behavior (the old `per_page: size` had the same issue), so optional.

ClemRz:
Thanks for the thorough review @Paul-AUB! All points addressed in 058d9c03:

**1. `found: counts.length` capping `totalDistinct` (Must Fix)** — Fixed. Now using `facetInfo?.stats?.total_values ?? counts.length` so the response reflects the true number of distinct facet values matching the query, not just the truncated set returned by `max_facet_values`.

**2. `found_docs` semantics change (Should Consider)** — Added code comments documenting that on array fields the sum may exceed the true distinct document count, and that it only sums counts for the returned facet values. Acceptable for the UI use case.

**3. Integration test against real Typesense** — Acknowledged. Consistent with the existing stub-based test approach. No change.

**4. `query_by: field` purpose (Nitpick)** — Added a comment clarifying it's required by the Typesense API even for wildcard queries.

**5. `max_facet_values` unbounded (Nitpick)** — Capped at 100 (mirrors `collectionSearch`'s `per_page` cap at 1000). Added test coverage for the cap.

Paul-AUB (approved):
All five points from the previous review have been properly addressed in 058d9c03. Here's a quick verification of each:

1. **`totalDistinct` capping (was Must Fix)** — Fixed. `facetInfo?.stats?.total_values ?? counts.length` correctly uses the uncapped total from Typesense's facet stats, with `counts.length` as a safe fallback. New test coverage confirms both the happy path and the fallback.

2. **`found_docs` semantics (was Should Consider)** — The in-code comments now clearly document both caveats (partial sum for large result sets, over-count on array fields). Good enough for a UI use case.

3. **Integration test against real Typesense (was Should Consider)** — Acknowledged as consistent with the existing stub-based test strategy. No change needed.

4. **`query_by: field` purpose (was Nitpick)** — Comment added explaining it's required by the Typesense API even for wildcard queries. Clear and accurate.

5. **`max_facet_values` unbounded (was Nitpick)** — Capped at 100 with `Math.min(size, 100)`, plus two tests verifying both the normal case (size 50 → 50) and the cap (size 500 → 100).

The implementation is clean, the response contract is preserved, and the test suite now provides good coverage of the new transform logic. No further issues.

PR #1656, opened 2026-06-11
feat(observation): add scientific observations import pipeline with device & sensor CRUD

### LINKED ISSUE #1636: Add POST /api/v1/observations/import bulk measurement import endpoint

ClemRz:
## 🤔 What

Implements the full scientific observations import pipeline along with Device and Sensor Configuration CRUD endpoints.

- Device CRUD: `GET/POST/PATCH/DELETE /api/v1/devices`, search via Typesense, soft-delete/restore
- Sensor Configuration CRUD: nested under devices (`/api/v1/devices/:deviceId/configurations`), soft-delete/restore
- Observation import endpoint: `POST /api/v1/observations/import` — accepts a CSV/TSV file + JSON profile, performs atomic multi-stage ETL into the database
- Removes the unused `TMethod` model and associated references
- Refactors `TName` and `TDescription` controllers to remove unnecessary populates
- Adds `id_observation` FK columns to `t_name` and `t_description` for naming/describing observations
- Updates Superset dashboard and materialized view definitions
- Full OpenAPI/Swagger documentation for all new endpoints

Closes #1636

## 🤷‍♂️ Why

Grottocenter needs to support ingestion of scientific measurement data (temperature, humidity, pressure, CO₂, water level, etc.) collected by cave monitoring instruments. This requires:

1. A way to register physical devices and their sensor configurations (units, precision, detection limits)
2. A pipeline that parses raw CSV/TSV data files, converts timestamps to UTC, converts measurement values to SI units, and persists everything atomically (observation → time series → measurements)

## 🔍 How

The import pipeline is structured as six sequential stages, each with a single responsibility:

```
ProfileValidator → ReferenceValidator → Parser → TimestampConverter → SIConverter → EntityBuilder
```

- **ProfileValidator** — structural validation of the JSON profile (no DB)
- **ReferenceValidator** — verifies all referenced entity IDs exist (parallel DB lookups)
- **Parser** — decodes the file buffer, detects delimiter, splits into rows, handles RFC 4180 quoting
- **TimestampConverter** — parses timestamps in 4 modes (datetime, split date+time, components, Unix epoch) with IANA timezone conversion
- **SIConverter** — locale-aware number parsing + unit conversion via `(value - offset) / factor`
- **EntityBuilder** — creates all DB entities in a single transaction (point, observation, document, time series, measurements in 1000-row batches, quality log, file uploads to Azure)

The `t_measurement` table uses PostgreSQL range partitioning by quarter. The `PartitionManager` ensures partitions exist before bulk inserts.

Devices are indexed in Typesense for full-text search, following the same pattern as documents and organizations.

## 🧪 Testing

- Property-based tests (fast-check) for all pure pipeline stages: ProfileValidator, Parser, TimestampConverter, SIConverter, PartitionManager, EntityBuilder, ReferenceValidator
- Conventional integration tests for all Device and SensorConfiguration routes (create, find, update, delete, restore, search)
- Integration tests for the import controller (missing file, missing profile, malformed JSON, auth, validation errors, reference errors, success with stubbed EntityBuilder)
- Run with: `npm run test`

## 📸 Previews

N/A — backend-only changes.


ClemRz:
## Known functional gaps (tracked)

The following follow-up issues have been identified and added to the [Gestion des observations scientifiques](https://github.com/orgs/GrottoCenter/projects/13) project:

| # | Issue | Priority |
|---|-------|----------|
| #1657 | **Azure blob rollback on transaction failure** — uploaded blobs are orphaned if the DB transaction rolls back | Medium |
| #1658 | **Add `label` to sensor configurations and `serialNumber` to devices** — better identification of channels and physical instruments | Medium |
| #1659 | **Expose scientific observation enumerations via public API** — avoid hardcoding quantity kinds, units, media, etc. in the front-end | Medium |
| #1660 | **Support attaching to an existing point by ID** — alternative to label-based point resolution | Low |
| #1661 | **Support attaching time series to an existing observation** — avoid duplicate observations for multi-file campaigns | Low |
| #1662 | **Support appending measurements to an existing time series** — incremental data ingestion for long-running monitoring | Low |

None of these block this PR — the current implementation is complete and functional. They represent enhancements for future iterations.

Paul-AUB (changes requested):
Impressive piece of work — the six-stage pipeline is cleanly separated, the property-based test coverage is excellent, and the controllers follow the existing project patterns closely (validateId, soft-delete two-phase pattern, treatAndConvert, Typesense indexing via dbSync entities). The known functional gaps already tracked in #1657–#1662 are not repeated here.

The must-fix items below all share one theme: the import endpoint promises structured 400 responses for invalid profiles, but several malformed-input shapes slip past `ProfileValidator` and crash deeper in the pipeline as unhandled errors → HTTP 500. Since the profile is client-supplied JSON, the validator should pin down types before anything else consumes them.

> Note: given the size of the PR (~12.7k additions), this review prioritized business logic, security-sensitive code, and the public API surface (pipeline services, controllers, policies/routes, SQL DDL); test files were skimmed rather than line-reviewed.

### Issues (Must Fix)

1. **[api/services/observation-import/ProfileValidator.js:216-224]** `profile.documentTitle.trim()` throws a `TypeError` when `documentTitle` is a non-string JSON value (e.g. `123` or `{}`). The error escapes `validate()` (it is not wrapped in the orchestrator), so the controller returns HTTP 500 instead of a 400 `IMPORT_VALIDATION_ERROR`. Consider checking `typeof profile.documentTitle === 'string'` first (same for `pointLabel`, which is later used in `TPoint.findOne`).

2. **[api/services/observation-import/ProfileValidator.js:75]** There is no error when `columnMappings` is present but not an array. A truthy non-array (e.g. a string) passes the required-field check, silently skips all role/structure checks behind the `Array.isArray` guard, and later crashes in `Parser.parse` with `columnMappings.filter is not a function` — surfaced to the client as a misleading `IMPORT_PARSE_ERROR`. Consider adding an explicit `columnMappings must be an array` validation error.

3. **[api/services/observation-import/ProfileValidator.js:45-59]** Referenced IDs (`caveId`, `licenseId`, `authorId`, `mediumId`, `sensorConfigurationId`) are never validated as positive integers. A non-numeric value (e.g. `"abc"`) reaches `TCave.findOne({ id: 'abc' })` in `ReferenceValidator`, which rejects with a Waterline `UsageError` → unhandled → HTTP 500. An object value would even be interpreted as a Waterline criteria modifier (e.g. `{">": 0}`). Consider validating ID shape in `ProfileValidator` (the same rule as `api/policies/validateId.js`: positive integer ≤ 2147483647) so these fail as structured 400s.

4. **[api/services/observation-import/TimestampConverter.js:215-224]** In epoch mode the result of `dayjs.unix(epochSeconds)` is never checked with `isValid()` (unlike the other three modes). An out-of-range epoch such as `1e15` passes `Number.isInteger` but produces an Invalid Date, which then propagates NaN into `EntityBuilder` (`observationDate` becomes Invalid Date) and `PartitionManager` (`t_measurement_NaN_qNaN`, and `computeBoundaries` throws a `RangeError` mid-transaction) → HTTP 500. Adding the same `!m.isValid()` guard used by the other modes would turn this into a proper 400.

5. **[api/services/observation-import/PartitionManager.js:68-81]** `Date.UTC(year, …)` maps years 0–99 to 1900–1999, so a component-mode profile with a two-digit year column (e.g. `year=24`) would create partition `t_measurement_24_q1` with 1924 boundaries — and the subsequent insert of year-24 rows fails with a cryptic partition error. Either use `setUTCFullYear` in `computeBoundaries`, or (probably better) validate a sane year range (e.g. 1900–2100) in `TimestampConverter` so absurd dates are rejected up front with a row number.

6. **[api/services/observation-import/SIConverter.js:97-103]** A measurement column whose `columnIndex` is not found in `columnIndices` (typo, beyond the file's width, or accidentally also marked excluded) is silently skipped — the import "succeeds" with missing time series (`processColumn` returns `null` for it). `TimestampConverter.findPosition` throws on the same condition. Consider failing (or at least reporting in the summary) for consistency, since silent data loss in a scientific import is hard to spot.

7. **[api/services/observation-import/SIConverter.js:54-77]** An empty measurement cell aborts the entire import (`parseFloat('') → NaN → IMPORT_CONVERSION_ERROR`). Real data-logger CSVs frequently contain gaps (sensor offline, multi-sensor files with different sampling rates). `EntityBuilder.processColumn` already supports sparse columns (the `find`/`filter(Boolean)` per-row logic). If fail-fast is the intended spec behavior this is fine — otherwise consider skipping empty cells per column. This might also be worth a follow-up issue alongside #1657–#1662.

8. **[sql/9_04_2026_06_03_scientific_observations_ddl.sql]** This PR edits the existing migration file in place (new audit columns inside `CREATE TABLE IF NOT EXISTS t_device` / `t_sensor_configuration`, and `id_author int4 NOT NULL DEFAULT 1`). If 9_04 has already been applied to any environment, `IF NOT EXISTS` will skip the table creation and those columns will never be added there. Worth confirming no environment ran the original version; otherwise the additions need `ALTER TABLE … ADD COLUMN IF NOT EXISTS` statements like the ones used for `t_name.id_observation`.

9. **[api/services/observation-import/PartitionManager.js:101-121]** Production DDL declares `t_measurement_default`. If any row ever lands in the default partition for a not-yet-created quarter (any insert path bypassing `PartitionManager`), PostgreSQL will reject `CREATE TABLE … PARTITION OF` for that range and all imports targeting that quarter will 500 until the default partition is manually repartitioned. Probably acceptable for now, but worth a comment/monitoring note. Relatedly, the test topology (`CONVERT_MEASUREMENT_TO_PARTITIONED` in test/customSQL.js) creates no default partition and uses `DOUBLE PRECISION` where production uses `numeric`, so tests don't exercise the production layout exactly.

10. **[api/services/observation-import/Parser.js:160-172]** `headerRow` is only checked against the upper bound (`> lines.length`). `headerRow: 0` (or a negative value) passes, then `lines[-1]` is `undefined` and `splitLine` throws a `TypeError` surfaced as a confusing `IMPORT_PARSE_ERROR`. A `headerRow must be an integer ≥ 1` check in `ProfileValidator` would give a clearer message.

### Nitpicks (Optional)

11. **[api/controllers/v1/observation/import.js:17]** `MAX_FILE_SIZE = 100 * 1024 * 1024` (104,857,600) vs. multer's `fileSize: 100000000` in config/http.js:306. Files between the two limits are rejected by multer (`LIMIT_FILE_SIZE`) before the controller runs, so the friendly `IMPORT_FILE_TOO_LARGE` path is mostly unreachable. Aligning the two constants (or lowering the controller's) would make the structured error the one users actually see.

12. **[api/services/observation-import/EntityBuilder.js:494]** `timeSeriesMap` is keyed by `sensorConfigurationId`, so if two measurement columns legitimately share one sensor configuration, both time series are created but the returned summary only reports the last one. Keying by `columnIndex` would keep the summary lossless.

ClemRz:
## Review addressed — 078e34ad

Thanks for the thorough review @Paul-AUB. All must-fix items and most suggestions are addressed in the latest commit.

### Fixed

| # | Item | Fix |
|---|------|-----|
| 1 | `documentTitle.trim()` on non-string | Added `typeof === 'string'` guard; same for `pointLabel` |
| 2 | `columnMappings` not validated as array | Explicit "must be an array" error when truthy non-array |
| 3 | Referenced IDs not validated as integers | All ID fields now checked: positive integer ≤ 2147483647 |
| 4 | Epoch mode missing `isValid()` | Added `isValid()` + year range check (1900–2100) |
| 5 | `Date.UTC` 2-digit year bug | Switched to `setUTCFullYear` in `computeBoundaries` |
| 6 | Missing measurement column silently skipped | Now throws `IMPORT_CONVERSION_ERROR` with clear message |
| 10 | `headerRow: 0` or negative passes | Validated as integer >= 1 in ProfileValidator |
| 11 | multer `fileSize` misconfigured | Moved into `limits: { fileSize: ... }` object |
| 12 | `timeSeriesMap` keyed by sensorConfigurationId | Now keyed by `columnIndex` for lossless summary |

### Follow-up issues created

| # | Item | Issue |
|---|------|-------|
| 7 | Empty measurement cells abort import | #1663 |
| 9 | Default partition overflow monitoring | #1664 |

### Not changed (by design)

| # | Item | Reason |
|---|------|--------|
| 8 | Editing existing migration in place | DDL is already in prod; `IF NOT EXISTS` makes it idempotent |

Swagger spec updated with new constraints. All 2764 tests pass, lint clean.

urien (commented):
Comme discuté sur Slack il y a un sujet autour de la possibilité de modifier/supprimer les devices et les Sensor Configuration
Peut être qu'il faudrait le rendre possible uniquement par les auteurs et les administrateurs (ou les modérateurs)
Les auteurs peuvent se rendre compte qu'ils ont fait des erreurs et souhaiter les corriger.
Une erreur peut être repérée longtemps après et nécessiter une correction
Une correction par tous les personnes authentifiées risque de mettre en difficulté les analyses qui sont réalisées et exploitées par l'auteur sur Super set

Paul-AUB (changes requested):
Re-review of `078e34ad`. All four previous must-fix items are properly fixed and the test updates match: `documentTitle`/`pointLabel` type guards, `columnMappings` array check, top-level ID shape validation, the epoch `isValid()` + year-range guard, the `setUTCFullYear` boundary fix, the multer `limits` correction, the `columnIndex`-keyed `timeSeriesMap`, and the SIConverter fail-fast on missing measurement columns. The Swagger spec was updated consistently. Thanks for the quick turnaround and for filing #1663/#1664.

Regarding @urien's comment on edit/delete permissions: the requested model is already implemented in the controllers — device and sensor-configuration **update** require the original author or a moderator ([api/controllers/v1/device/update.js:35-43](https://github.com/GrottoCenter/grottocenter-api/blob/078e34adeb31f9f20c2550c07ed0955771b94bca/api/controllers/v1/device/update.js#L35-L43), [api/controllers/v1/sensor-configuration/update.js:59-69](https://github.com/GrottoCenter/grottocenter-api/blob/078e34adeb31f9f20c2550c07ed0955771b94bca/api/controllers/v1/sensor-configuration/update.js#L59-L69)), **soft delete/restore** require moderator, and **permanent delete** requires administrator. Authenticated users without rights get a 403, so analyses built on top of this data are protected from arbitrary edits.

One residual gap remains from previous item 3 (non-integer IDs reaching Waterline → HTTP 500), hence the request-changes:

### Issues (Must Fix)

1. **[api/services/observation-import/ProfileValidator.js:72 / ReferenceValidator.js:123-139]** The new `idFields` check validates a **top-level** `profile.mediumId` — but the pipeline never reads that field. The `mediumId` that is actually consumed lives **per column mapping** (`col.mediumId`, see ReferenceValidator.js:43-50 and the Swagger `columnMappings` schema), and it is still not shape-validated. A profile with `columnMappings: [{ role: 'measurement', sensorConfigurationId: 1, mediumId: "abc" }]` passes ProfileValidator, then `TMedium.findOne({ id: "abc" })` rejects with a Waterline `UsageError` → unhandled → HTTP 500 (an object value would again act as a criteria modifier). Consider moving/duplicating the integer check into the `columnMappings.forEach` loop (same rule as `sensorConfigurationId`), and removing the dead top-level entry unless it is meant to be supported.

2. **[api/services/observation-import/ProfileValidator.js:75 / ReferenceValidator.js:82]** The `val !== ''` escape in the ID checks lets `caveId: ""` through: it skips integer validation, satisfies `hasCaveId` (line 219 only excludes `undefined`/`null`), and ReferenceValidator's presence check (`!== undefined && !== null`) then runs `TCave.findOne({ id: '' })` → Waterline `UsageError` → HTTP 500. The same applies to `col.mediumId: ""` and `col.sensorConfigurationId: ""` on a non-measurement column (ReferenceValidator collects `sensorConfigurationId` from **all** columns, while ProfileValidator only validates it on `role === 'measurement'` ones). Consider treating `''` as invalid in the ID checks (rather than skipping it), or aligning ReferenceValidator's presence predicates with the validator's.

3. **[assets/swaggerV1.yaml:7609]** The `ImportProfile` schema documents a top-level `deviceId` ("ID of the device that produced the data"), but nothing in the pipeline reads `profile.deviceId` — the device association comes implicitly through each column's sensor configuration. Consider removing it from the spec (or wiring it up) so front-end integrators don't send a field that is silently ignored.

### Nitpicks (Optional)

4. **[api/services/observation-import/TimestampConverter.js:228-236]** The new 1900–2100 sanity range applies only to epoch mode. A component-mode profile with a two-digit `year` column (e.g. `24`) now imports "successfully" into a year-0024 partition thanks to the `setUTCFullYear` fix — correct mechanically, but probably never what the user meant. Applying the same year-range guard to all four modes would catch this class of data-entry error uniformly. Fine as a follow-up.

ClemRz:
Pushed `b6e4ff81` addressing all items from the re-review:

**Must Fix (1 & 2):**
- Per-column `mediumId` is now shape-validated on measurement columns (the top-level `mediumId` entry was dead code and has been removed from `idFields`)
- Empty strings `""` no longer bypass ID validation — `isBlank` now catches them before they reach Waterline
- `sensorConfigurationId` and `mediumId` on non-measurement columns are silently ignored (both in ProfileValidator and ReferenceValidator)

**Should Consider (3):**
- Removed `deviceId` from the Swagger `ImportProfile` schema and from the inline description (nothing in the pipeline reads it)

**Refactoring:**
- Extracted `api/utils/isBlank.js` (covers `undefined`, `null`, whitespace-only strings)
- Extracted `api/utils/isNonBlankString.js` (true iff string with at least one visible character)
- Extracted `api/utils/isValidId.js` (positive integer ≤ 2147483647, strict type — no string coercion)
- All three utilities have property-based tests (`test/integration/2_utils/`)

All 2764 tests pass.

Paul-AUB (approved):
Re-review of `b6e4ff81`. All items from the second review are correctly resolved:

1. **Per-column `mediumId` shape-validation** — `ProfileValidator.js:124` now checks `!isBlank(col.mediumId) && !isValidId(col.mediumId)` for each measurement column; the dead top-level `mediumId` entry has been removed from `idFields`. `ReferenceValidator` collects `mediumId` only from `measurementColumns` (`col.role === 'measurement'`), so non-measurement columns are silently ignored in both validators. ✓

2. **Empty-string bypass closed** — The old `val !== ''` escape is gone. The `idFields` loop now uses `val !== undefined && val !== null`, so `caveId: ""` reaches `isValidId("")` → false → structured 400. The new `isBlank` utility returns true for empty strings, so `""` mediumIds/sensorConfigIds are filtered out in `ReferenceValidator` before any DB call. ✓

3. **`deviceId` removed from Swagger** — The `ImportProfile` schema no longer documents the unused top-level `deviceId` field. ✓

The three extracted utilities (`isBlank`, `isNonBlankString`, `isValidId`) are clean, correctly scoped to parsed-JSON values (no string coercion), and have thorough property-based test coverage. The orchestrator's fail-fast on ProfileValidator errors means the corrected validators collectively close the HTTP-500-on-bad-input class of issues flagged in both prior reviews.

No new concerns found in `b6e4ff81`.

PR #1665, opened 2026-06-13
fix(observation): correct row numbers and skip empty measurements

### LINKED ISSUE #1663: Observation import: support empty measurement cells (data gaps)
## Summary

The observation import pipeline currently aborts the entire import when a measurement cell is empty (`parseFloat('') → NaN → IMPORT_CONVERSION_ERROR`). Real data-logger CSVs frequently contain gaps (sensor offline, multi-sensor files with different sampling rates).

## Current Behaviour

In `SIConverter.parseLocaleNumber`, an empty string is parsed via `parseFloat('')` which returns `NaN`, triggering an error that aborts the entire import.

## Proposed Fix

Skip empty measurement cells gracefully instead of failing the import. The `EntityBuilder.processColumn` logic already supports sparse columns (the `filter(Boolean)` per-row logic). The change would be:

1. In `SIConverter.convertAll`, return `null` for empty cells instead of throwing
2. Document that empty cells produce gaps in the time series (no measurement stored for that timestamp)
3. Include a count of skipped cells in the import summary for transparency

This aligns with the scientific data model where gaps are normal and expected.

## Complexity

Low. A single `if` guard in `parseLocaleNumber` or `convertAll` + summary enrichment.

## Priority

Medium — blocks real-world data ingestion for any file with sensor gaps.

ClemRz:
## 🤔 What

- Fix row numbers in observation import validation errors to reference the original file line numbers
- Skip empty measurement cells (`""`) silently instead of throwing a parse error

Fixes #1663

## 🤷‍♂️ Why

When the import pipeline reported validation errors (timestamp parse failures, measurement conversion failures), the row numbers referenced the internal parsed-data array index — not the actual line in the user's file. Users had to mentally add the header and skip-first-rows offset to locate the problem, which was confusing.

Additionally, empty measurement cells are normal in scientific time series data (sensor gaps, logger power loss) and should not be treated as errors.

## 🔍 How

**Row number fix** (`ObservationImportService.js`, `TimestampConverter.js`, `SIConverter.js`):
- Compute `fileRowOffset = headerRow + skipFirstRows` once after parsing in the orchestrator
- Pass it as a parameter to `TimestampConverter.convert()` and `SIConverter.convertAll()`
- Each error throw site uses `rowIndex + 1 + rowOffset` directly — plain arithmetic, no post-hoc string manipulation

**Empty cell handling** (`SIConverter.js`):
- Switch from `.map()` to `.reduce()` so that when `rawString === ''`, no entry is added to the row's measurement array
- EntityBuilder already handles missing entries via `.find()` → `undefined` → filtered by `.filter(Boolean)` — no downstream changes needed

## 🧪 Testing

All 2778 existing tests pass. To manually verify:
1. Import a CSV with `headerRow: 1` and `skipFirstRows: 1` containing a bad timestamp on line 5 → error should report "row 5" (not "row 3")
2. Import a CSV with empty measurement cells → import succeeds, those cells produce no `TMeasurement` record

### DIFF
--- a/api/services/ObservationImportService.js
+++ b/api/services/ObservationImportService.js
@@ -105,6 +105,17 @@ const execute = async (file, profile, requestAuthorId) => {
   const { rows, columnIndices } = parsedResult;
 
   // -------------------------------------------------------------------------
+  // Compute the file-line offset so that row numbers in error messages
+  // reference the original file line rather than the parsed-data index.
+  // Offset = (headerRow lines consumed) + (skipFirstRows lines consumed).
+  // The Parser already removed these lines from the data, so any 1-based
+  // row number from downstream stages must be shifted by this amount.
+  // -------------------------------------------------------------------------
+  const fileRowOffset =
+    (profile.headerRow != null ? profile.headerRow : 0) +
+    (profile.skipFirstRows || 0);
+
+  // -------------------------------------------------------------------------
   // Stage 4: TimestampConverter — parse timestamps → UTC Date objects
   // -------------------------------------------------------------------------
   let convertedTimestamps;
@@ -112,7 +123,8 @@ const execute = async (file, profile, requestAuthorId) => {
     convertedTimestamps = TimestampConverter.convert(
       rows,
       profile,
-      columnIndices
+      columnIndices,
+      fileRowOffset
     );
   } catch (err) {
     throw new ImportError('IMPORT_TIMESTAMP_ERROR', err.message, []);
@@ -143,7 +155,8 @@ const execute = async (file, profile, requestAuthorId) => {
       rows,
       sensorConfigMap,
       columnIndices,
-      profile
+      profile,
+      fileRowOffset
     );
   } catch (err) {
     throw new ImportError('IMPORT_CONVERSION_ERROR', err.message, []);
--- a/api/services/ObservationImportService.js
+++ b/api/services/ObservationImportService.js
@@ -60,6 +60,7 @@ class ImportError extends Error {
  *   documentId: number,
  *   timeSeriesMap: Object<string, number>,
  *   measurementCount: number,
+ *   skippedMeasurements: number,
  *   observationDate: Date,
  *   importedAt: Date,
  *   importedBy: number,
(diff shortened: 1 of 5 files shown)


Paul-AUB (dismissed):

1. **[ObservationImportService.js:114–116 / TimestampConverter.js:47 / SIConverter.js:95–100]** No tests were added for the `rowOffset` parameter — the core fix of this PR. The PR states "2778 existing tests pass," but those tests don't cover the new behavior. If someone changes the offset formula in the future there is no regression protection. Consider adding unit tests for `TimestampConverter.convert` and `SIConverter.convertAll` that pass a non-zero `rowOffset` and assert that error messages report the correct file-line number (e.g., `rowOffset=2`, bad row at index 0 → message contains "row 3").

2. **[SIConverter.js:123]** Issue #1663 proposed "Include a count of skipped cells in the import summary for transparency." The PR silently drops empty cells without surfacing how many gaps were skipped. The import result returned by `EntityBuilder.build()` could include a `skippedMeasurements` counter. This is purely a UX improvement and not a blocker, but it was part of the original specification and may surprise users who expect import feedback.

### Nitpicks (Optional)

3. **[SIConverter.js:123]** The empty-cell guard checks `rawString === ''` (strict equality). Whitespace-only cells like `' '` or `'\t'` — which can appear in files exported from Excel or LibreOffice — would fall through to `parseLocaleNumber`, fail with a NaN error, and abort the import rather than being treated as gaps. If real-world files are expected to have padded empty fields, consider `rawString.trim() === ''`. Low priority given the current issue scope, but worth noting for future imports.

ClemRz:
Thanks for the review @Paul-AUB — all three items were legitimate. Addressed in 6229d02f:

**1. rowOffset regression tests** — Added 6 unit tests (3 for `TimestampConverter`, 3 for `SIConverter`) that pass a non-zero `rowOffset` and assert error messages contain the correct file-line number.

**2. skippedMeasurements counter** — `SIConverter.convertAll` now returns `{ measurements, skippedMeasurements }`. The count is surfaced in the import response and documented in the Swagger spec.

**3. Whitespace-only cell guard** — Changed `rawString === ''` to `rawString.trim() === ''` so padded empty cells (`' '`, `'\t'`) from Excel/LibreOffice exports are treated as sensor gaps instead of aborting the import. Added dedicated tests.

All 2788 tests passing, lint clean.

PR #1667, opened 2026-06-13
feat(device): add serialNumber to devices and label to sensor configurations

### LINKED ISSUE #1658: Add label field to sensor configurations and serial number to devices
## Summary

Add missing identification fields to the scientific instrumentation models:

1. **Sensor Configuration** — add a `label` field (free-text, optional) so users can give a human-readable name to each configuration (e.g. "Channel 1 - Temperature", "Outdoor probe")
2. **Device** — add a `serial_number` field (optional, varchar) to uniquely identify physical instruments by their manufacturer serial number

## Motivation

- Multiple sensor configurations on the same device are currently distinguished only by their quantity kind and unit, which can be ambiguous when a device has multiple channels measuring the same physical quantity
- Devices of the same model are indistinguishable without serial numbers — important when multiple loggers are deployed simultaneously at different points

### Database

```sql
ALTER TABLE t_sensor_configuration ADD COLUMN IF NOT EXISTS label varchar(300) NULL;
ALTER TABLE t_device ADD COLUMN IF NOT EXISTS serial_number varchar(200) NULL;
```

### Model

- `TSensorConfiguration.js`: add `label` attribute (`type: 'string', allowNull: true, maxLength: 300`)
- `TDevice.js`: add `serialNumber` attribute (`type: 'string', allowNull: true, columnName: 'serial_number', maxLength: 200`)

### API

- Accept `label` in sensor configuration create/update endpoints
- Accept `serialNumber` in device create/update endpoints
- Include both in the response converters (`toDevice`, `toSensorConfiguration`)
- Index `serialNumber` in the Typesense device schema for searchability

### Swagger

- Update `assets/swaggerV1.yaml` to document both new fields

ClemRz:
## 🤔 What

Add two optional identification fields to the scientific instrumentation models:

- **`serialNumber`** (varchar 200) on `TDevice` — manufacturer serial number for physical instruments
- **`label`** (varchar 300) on `TSensorConfiguration` — human-readable name for configurations (e.g. "Channel 1 - Temperature")

Closes #1658

## 🤷‍♂️ Why

- Multiple sensor configurations on the same device are distinguished only by quantity kind and unit, which is ambiguous when a device has multiple channels measuring the same physical quantity
- Devices of the same model are indistinguishable without serial numbers — important when multiple loggers are deployed simultaneously at different points

## 🔍 How

- SQL migration adds both columns with `IF NOT EXISTS`
- Waterline models updated with appropriate `allowNull`, `maxLength`, and `columnName`
- Create/update controllers accept and validate the new fields (length checks)
- Response converters (`toDevice`, `toSimpleDevice`, `toSensorConfiguration`) include the new fields
- Typesense device schema adds `serialNumber` as an optional searchable field in `query_by`
- `DeviceService.updateInSearch` passes `serialNumber` to the search index
- Swagger spec updated for all relevant component schemas

## 🧪 Testing

All 2783 existing + new tests pass. New test coverage:

- Device create/update: validation (400 on >200 chars), success with `serialNumber`
- SensorConfiguration create/update: validation (400 on >300 chars), success with `label`
- Property tests updated to include both fields in arbitraries and shape assertions
- Device find test verifies `serialNumber` in response

Run `npm run test` — no additional setup needed (Waterline `migrate: drop` handles the column in tests).

For local dev DB, apply the migration:
```
docker exec grotto-postgres psql -U root -d grottoce -f /docker-entrypoint-initdb.d/9_07_2026_06_13_device_serial_number_sensor_label.sql
```

## 📸 Previews

N/A (backend-only change)


Paul-AUB (approved):
Thorough, well-tested PR that cleanly follows the patterns already established for the device/sensor-configuration controllers. The changes are consistent across the whole stack (model → migration → controllers → converters → search schema → swagger → fixtures → tests), and the validation, `|| null` vs `?? null` handling, and swagger placement all match the existing `brandName`/`manufacturerUrl` conventions. No blocking issues found.

1. **[api/dbSync/entities/device.js:52]** Adding `serialNumber` to `query_by` only takes effect on the live `devices` collection after a full `dbSync` (which rebuilds the collection via `createTimestampedCollection` + `switchCollectionAlias`). Until that sync runs post-deploy, `DeviceService.updateInSearch` will push the `serialNumber` field into the old collection where it is stored but not indexed/searchable. Worth calling out in the deployment notes that a `dbSync` is required for serial-number search to work.

2. **[sql/9_07_2026_06_13_device_serial_number_sensor_label.sql:4]** Issue #1658 motivates `serial_number` as a way to "uniquely identify physical instruments," but no uniqueness constraint is added (DB or app level). This is likely intentional — manufacturer serials can collide across brands, and the field is optional — but confirming the intent here avoids a surprise later if the frontend assumes uniqueness.

### Nitpicks (Optional)

3. **[api/controllers/v1/sensor-configuration/create.js:48]** The `label` length check uses `if (label && label.length > 300)` while the matching device create check is identical in spirit — good. Just note that, as with the existing string fields, a non-string `label` (e.g. a numeric JSON value) would bypass the length guard since `.length` is `undefined`. This mirrors the pre-existing `brandName`/`productUrl` behavior, so no change is needed for consistency; flagging only for awareness if input typing is ever tightened project-wide.

PR #1669, opened 2026-06-14
feat(observation): support multiple authors on import

### LINKED ISSUE #1666: Observation import: support multiple authors
## Context

The observation import endpoint (`POST /api/v1/observations/import`) currently accepts a single `authorId` in the profile payload. This author is stored on the resulting `TObservation` and `TDocument` records.

## Problem

Scientific observation campaigns are typically collaborative — multiple people contribute to data collection, logger installation, and analysis. The current single-author model doesn't reflect this reality and forces users to pick one person as "the" author.

## Expected behavior

The import profile should accept multiple author IDs (e.g., `authorIds: [1, 5, 12]`). All referenced authors should be linked to the observation and its associated document, preserving attribution for all contributors.

## Considerations

- ~Backward compatibility: the existing `authorId` field should continue working (single author as a shorthand)~
- The `TDocument` model already supports multiple authors via the `j_document_caver_author` junction table
- `TObservation` currently has a single `id_author` column — this may need a junction table or a different attribution model
- The UI will need corresponding changes to allow selecting multiple authors

ClemRz:
## 🤔 What

- Replace the single `authorId` field with `authorIds` (array) in the observation import profile
- All listed authors are linked to the created document via `j_document_caver_author`
- Closes #1666

## 🤷‍♂️ Why

Scientific observation campaigns are collaborative — multiple people contribute to data collection. The single-author model forced users to pick one person as "the" author. The `TDocument` model already supports multiple authors via the junction table, so the pipeline just needed to leverage it.

## 🔍 How

- **ProfileValidator**: `authorIds` is now a required field (non-empty array of positive integers). `authorId` is removed.
- **ReferenceValidator**: Resolves each unique ID from `authorIds` against `TCaver`. Returns `resolved.authors` (array).
- **EntityBuilder**: Deduplicates `authorIds`, sets `TDocument.author` FK to the first entry, and creates a `JDocumentCaverAuthor` row for each author sequentially within the transaction.
- `TObservation.author` remains unchanged — always the authenticated user (`requestAuthorId`).
- No schema migration needed.

## 🧪 Testing

All existing tests updated. New test added for mixed author existence (some IDs valid, some not). Run:

```
npm run test -- --grep "EntityBuilder|ProfileValidator|ReferenceValidator|observations/import"
```

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/PERMISSION_SYSTEM.md
+++ b/PERMISSION_SYSTEM.md
@@ -12,7 +12,7 @@ Roles are **not hierarchical**; for instance, an Administrator does not automati
 - **Description**: Non-authenticated users
 - **Permissions**:
   - View public cave/entrance/document/organisation/massif/person data
-  - Search through cave/entrance/document/organisation/massif/person data
+  - Search through cave/entrance/document/organisation/massif/person/device data
   - View complete entity history
   - View statistics
   - View all organization details
@@ -23,12 +23,24 @@ Roles are **not hierarchical**; for instance, an Administrator does not automati
 - **Description**: Default authenticated users role
 - **Permissions**: All Visitor permissions plus:
   - **Content Management**:
-    - Create caves, entrances, documents, Massifs, Organisations, comments, descriptions, locations, riggings, histories
-    - Edit caves, entrances, documents, Massifs, Organisations, descriptions, locations, riggings, histories
+    - Create caves, entrances, documents, massifs, organisations, comments, descriptions, locations, riggings, histories
+    - Edit caves, entrances, documents, massifs, organisations, descriptions, locations, riggings, histories
     - Edit own comments
     - Unlink documents from entities
     - Rollback to a previous version of any content
     - Add/remove own explored entrances
+    - Set main name for entities
+  - **Scientific Data Management**:
+    - Create devices and sensor configurations
+    - Update own devices and sensor configurations
+    - Import observation data (CSV/TSV)
+  - **Messaging**:
+    - Send and receive private messages
+    - View, archive, and unarchive own conversations only
+    - Read messages only in conversations where the user is a participant
+  - **Notifications**:
+    - View own notifications
+    - Mark notifications as read
   - **Personal Management**:
     - Manage personal profile and settings
     - Join/leave organizations (own membership only)
@@ -50,7 +62,9 @@ Roles are **not hierarchical**; for instance, an Administrator does not automati
 - **Description**: Content reviewers and validators
 - **Permissions**: All User permissions plus:
   - **Content Moderation**:
-    - Delete/restore any caves, entrances, documents, comments, descriptions, locations, riggings, histories, organisations, massifs, documents
+    - Delete/restore any caves, entrances, documents, comments, descriptions, locations, riggings, histories, organisations, massifs
+    - Delete/restore devices and sensor configurations
+    - Update any device or sensor configuration (regardless of ownership)
     - Update any user's comments
     - Validate documents
     - Manage duplicates (documents and entrances)
@@ -67,16 +81,25 @@ Roles are **not hierarchical**; for instance, an Administrator does not automati
     - Assign/remove user roles
     - Manage any user accounts
     - Delete user accounts
-    - View user lists (authors, contributors, users)
+    - Ban/unban users (revokes all active tokens)
+    - View user lists (authors, contributors, users, banned, invalid-email)
     - View complete user information
     - Cancel any user's subscriptions
+  - **MFA Management**:
+    - Enroll MFA (generate TOTP secret)
+    - Verify MFA enrollment (activate TOTP)
+    - Reset MFA (requires password re-entry)
   - **System Operations**:
     - Import data from CSV files (documents and entrances)
     - System configuration and maintenance
   - **Sensitive Data Management**:
+    - View coordinates of sensitive entrances
     - Remove sensitive flag from entrances
     - Modify coordinates of sensitive entrances
-    - Access all sensitive entrance data
+    - Mark/unmark massifs as sensitive (cascades to contained entrances)
+    - Preview massif sensitivity impact
+  - **Permanent Deletion**:
+    - Permanently delete any content (caves, entrances, documents, devices, sensor configurations, etc.)
   - **Advanced Organization Management**:
     - Add/remove explored caves for any organization
     - Manage organization memberships for any user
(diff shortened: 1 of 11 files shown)


Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[assets/swaggerV1.yaml:7536]** The `ObservationImportProfile` schema component still lists `authorId` (singular) in its `required` array. This needs to be replaced with `authorIds` so the published OpenAPI contract matches the actual API behaviour.

2. **[assets/swaggerV1.yaml:7632-7636]** Same schema component: the property `authorId` is defined as `type: integer` with a single-caver description. It must be replaced with `authorIds` defined as `type: array, items: { type: integer, minimum: 1 }` (non-empty). As-is, any client code-generated from the spec will send the wrong shape and receive a validation error with no clear explanation.

3. **[api/services/observation-import/ProfileValidator.js, ReferenceValidator.js]** The original issue (#1666) explicitly listed backward compatibility as a consideration: *"the existing `authorId` field should continue working (single author as a shorthand)"*. This PR removes `authorId` entirely — existing callers will receive a `Missing required field: authorIds` error with no migration guidance. Consider either (a) accepting both shapes during a transition period (coerce `authorId` to `authorIds: [authorId]` before validation), or (b) explicitly documenting the break in the PR description so consumers are aware.

4. **[test/integration/1_services/observation-import/ReferenceValidator.property.test.js:104]** The property-based test only ever generates a single-element `authorIds` array (`authorIds: [base.authorId]`). The new multi-author code path — multiple parallel `TCaver.findOne` promises, partial resolution, correct error counting — is therefore never exercised by property testing. Consider switching to `fc.uniqueArray(idArb, { minLength: 1, maxLength: 3 })` so the property test covers the multi-author scenario.

### Nitpicks (Optional)

5. **[api/services/observation-import/EntityBuilder.js:290-293]** The fallback `profile.authorIds || []` in the `authorIds` derivation would yield an empty array if the resolver returned no authors and `profile.authorIds` is somehow absent, causing `authorIds[0]` to be `undefined` on line 419. This path is unreachable in normal pipeline flow (ProfileValidator guards it), but a short guard comment — or removing the fallback in favour of a hard assertion — would make the intent clearer.

ClemRz:
Thanks for the review @Paul-AUB! All addressed in 7c2fb0e0:

**Item 1 & 2 (Swagger):** Fixed — `authorId` replaced with `authorIds` (array, minItems: 1) in both the `required` array and the property definition of `ObservationImportProfile`.

**Item 3 (Backward compatibility):** Intentionally not implemented. There are no existing consumers of the `authorId` field — this endpoint hasn't been released to any client yet. Adding a deprecation/coercion layer for a field nobody uses would be dead code. The breaking change is documented in the PR description.

**Item 4 (Property test):** Fixed — the arbitrary now generates `authorIds` with 1–3 unique entries via `fc.uniqueArray`, each with independent existence flags. This exercises the multi-author resolution, partial failure, and deduplication paths.

**Item 5 (EntityBuilder fallback):** Added a guard comment explaining the `profile.authorIds || []` fallback is defensive-only since ProfileValidator guarantees `authorIds` is non-empty and ReferenceValidator resolves them before EntityBuilder runs.

Paul-AUB (changes requested):
All five items from the previous review have been addressed. This follow-up focuses on a new finding in the updated code and one bundling note.

### Issues (Must Fix)

1. **[api/services/observation-import/ReferenceValidator.js:115–127 / EntityBuilder.js:293–297]** **Non-deterministic primary-author ordering.** All `TCaver.findOne` calls are pushed into `checks` and awaited via `Promise.all` — they run in parallel. Each `.then()` appends to `resolved.authors` in *resolution order*, not in the order of `profile.authorIds`. In `EntityBuilder.js`, `authorIds[0]` (derived from `authors.map(a => a.id)`) is then used as the `TDocument.author` FK. With two or more authors, the entity stored as the primary author could be any of the provided IDs, not necessarily `profile.authorIds[0]` as the caller intended.

   Simplest fix — use the already-validated profile field directly in `EntityBuilder.js`:
   ```js
   // profile.authorIds already validated by ProfileValidator and resolved by ReferenceValidator
   const authorIds = [...new Set(profile.authorIds || [])];
   ```
   This removes the dependency on the resolution-order-dependent `authors` array and preserves the user-specified order. The `authors` array from `resolvedEntities` is not used for anything beyond extracting IDs, so nothing else is affected.

2. **[PERMISSION_SYSTEM.md]** This file accounts for +141/−20 lines in the diff, covering MFA enrollment, ban/unban, token blacklist, massif sensitivity, messaging, notifications, and device/sensor configurations — none of which are part of the multiple-author import feature described in the PR. These changes would ideally live in a dedicated PR (or the PR that introduced each feature) to keep the history clean and make future bisects easier. If they were pulled in during a rebase on `develop`, consider splitting them out before merge.

ClemRz:
### Review items addressed

**Second review, Item 1 (Non-deterministic primary-author ordering):** Fixed in 40aa9322. `EntityBuilder` now uses `profile.authorIds` directly (deduplicated with `Set` preserving insertion order) instead of deriving IDs from the resolution-order-dependent `resolved.authors` array. This guarantees `authorIds[0]` — used as the `TDocument.author` FK — always matches the caller-specified first author, regardless of `Promise.all` resolution timing.

**Second review, Item 2 (PERMISSION_SYSTEM.md):** Intentionally included in this PR — those docs reflect features already merged on `develop` and were added here during rebase.

Paul-AUB (dismissed):
All items from both previous reviews have been addressed. The fix for non-deterministic primary-author ordering (using `profile.authorIds` directly in EntityBuilder rather than deriving IDs from the resolution-order-dependent `resolved.authors` array) is correct and clean.

1. **[api/services/observation-import/ReferenceValidator.js:122–124 / ObservationImportService.js:86]** `resolved.authors` is still populated in `ReferenceValidator` (each found `TCaver` is pushed to it) and returned in `resolvedEntities`, but `EntityBuilder.js:286` now only destructures `{ cave, license }` — `authors` is never accessed downstream. The array is built, passed through, and silently discarded. Since the existence check itself (the `errors.push(...)` path) is the only part of the author resolution that matters now, consider either:
   - Removing the `resolved.authors = []` init and the `.push(author)` line from `ReferenceValidator`, keeping only the error-push path; or
   - Keeping the array but documenting in a comment that it is not used by `EntityBuilder` (which reads `profile.authorIds` directly for determinism).

   This is not a bug — the pipeline is correct — but the unused array may mislead a future maintainer into thinking `resolved.authors` is the authoritative source of author IDs downstream.

### Nitpicks (Optional)

2. **[api/services/observation-import/ProfileValidator.js:84]** The comment on line 83 notes that `isBlank([])` returns `false` for arrays, so an empty array slips past the required-field check and is caught by the explicit `length === 0` guard. This is an existing subtlety in `isBlank`; the comment is appreciated. No change needed — just confirming it is correct.

ClemRz:
Addressed suggestion #1 from the latest review: removed the unused `resolved.authors` array from `ReferenceValidator`. The existence check (error path) is preserved — only the array population and downstream references were removed since `EntityBuilder` reads `profile.authorIds` directly for deterministic ordering.

Regarding suggestion #2 (PERMISSION_SYSTEM.md in a separate PR): leaving as-is since it's already isolated in its own commit and the review was approved.

PR #1672, opened 2026-06-15
fix(organization): trim string inputs and widen postalCode to varchar(10)

### LINKED ISSUE #1670: bug(api): [E_INVALID_VALUES_TO_SET] tgrotto - postalCode with trailing whitespace causes 500
**Describe the bug**
When updating an organization via `PUT /api/v1/organizations/:id`, the API returns a 500 error if the `postalCode` field contains trailing whitespace that makes it exceed the 5-character maximum length defined in the `TGrotto` model. The error should be caught and returned as a 400 Bad Request, but instead it propagates as an unhandled server error.

**To Reproduce**
Based on the trace logs:
1. `PUT /api/v1/organizations/3699`
2. Include a `postalCode` value with trailing whitespace (e.g., `"56220 "` — 6 chars)
3. The Waterline validation rejects the value but the error is not handled gracefully

**Input Data**

- **Endpoint:** `PUT /api/v1/organizations/3699`
- **Request Body:**
```json
{
  "id": 3699,
  "address": "Rue de l'Étang ",
  "city": "Rochefort-en-Terre",
  "country": {"id": "FR"},
  "customMessage": "La paroisse a pour mission d'accompagner...",
  "latitude": null,
  "longitude": null,
  "mail": null,
  "postalCode": "56220 ",
  "url": "https://www.paroisses-questembert-rochefort.com/rochefort-en-terre"
}
```
- **URL/Query Parameters:** `id = 3699`
- **User:** id=63690, nickname="Aissatou", groups=[]

**Expected behavior**
The API should either:
1. Trim whitespace from `postalCode` before validation, or
2. Return a 400 Bad Request with a clear validation message instead of 500

**Error Details**
- Error Type: OperationalError [UsageError]
- Error Code: E_INVALID_VALUES_TO_SET
- Occurrences: 3
- Time Range: 2026-06-15T10:49:56Z to 2026-06-15T11:02:28Z
- Representative Trace ID: 019ecaf2-78b0-77ba-abfa-b0f4f15099c9
- Stack Trace:
```
Details:
  Could not use specified `postalCode`.  Violated one or more validation rules:
  • Value was -1 characters longer than the configured maximum length (5)

    at flaverr (node_modules/flaverr/index.js:94:15)
    at _.extend._WLModel [as _handleExec] (node_modules/waterline/lib/waterline/methods/update-one.js:138:15)
    at Deferred.exec (node_modules/parley/lib/private/Deferred.js:286:10)
```

**Additional context**
- Service: `api/controllers/v1/organization/update.js`
- Model: `tgrotto`
- Waterline Error Code: `E_INVALID_VALUES_TO_SET`
- The `postalCode` field has `maxLength: 5` in the TGrotto model. The input `"56220 "` has a trailing space making it 6 characters. The controller should trim string inputs or catch `E_INVALID_VALUES_TO_SET` errors and return a 400.

ClemRz:
## 🤔 What

- Trim all string fields in `GrottoService.getConvertedDataFromClientRequest()` before persistence
- Widen `postal_code` column from `varchar(5)` to `varchar(10)` in both `t_grotto` and `h_grotto`
- Update `TGrotto` model `maxLength` from 5 to 10
- Add SQL migration `1_2026_06_15_widen_grotto_postal_code.sql`
- Closes #1670

## 🤷‍♂️ Why

- Users submitting `postalCode` with trailing whitespace (e.g., `"56220 "`) trigger an unhandled Waterline `E_INVALID_VALUES_TO_SET` error → HTTP 500
- `varchar(5)` is too restrictive for international postal codes (UK: `"SW1A 2AA"`, Brazil: `"12345-678"`, Iran: 10 digits)

## 🔍 How

- Added a `trimIfString` helper in `GrottoService` that safely trims string values (passes through `null`/`undefined`)
- Applied it to all string fields extracted from the client request
- Widened the DB column to `varchar(10)` to accommodate the longest postal codes worldwide

## 🧪 Testing

- New integration test: PUT organization with `postalCode: "56220 "` → 200, returns trimmed `"56220"`
- New unit tests: `getConvertedDataFromClientRequest` trims whitespace, handles null/undefined gracefully
- All existing tests pass


ClemRz:
Migrations already applied in prod ✅

Paul-AUB (dismissed):
Clean fix — the `trimIfString` approach is the right call over catching `E_INVALID_VALUES_TO_SET` (which would silently discard meaningful data rather than sanitize it). The migration, model, and tests are all aligned.

1. **[api/services/GrottoService.js:14]** `trimIfString` is a general-purpose utility that other services touching string fields (CaveService, DocumentService, etc.) might need independently. The project already has a `api/utils/` directory with analogous helpers (`coerceToNumeric`, `coerceBool`, `isBlank`). Consider moving `trimIfString` there so it can be imported wherever needed without duplicating the logic.

2. **[test/integration/1_services/GrottoService.test.js:103]** The "null and undefined" test only exercises the `null` path (via `req.param.returns(null)`). Since `trimIfString` uses a strict `typeof === 'string'` check, it correctly passes `undefined` through too — but that branch isn't explicitly covered. A second stub returning `undefined` for a few fields would make the contract explicit and guard against future changes.

3. **[sql/1_2026_06_15_widen_grotto_postal_code.sql]** Existing rows in production with trailing whitespace in `address`, `city`, `region`, etc. won't be cleaned up by this migration — only new writes will be sanitized. If data quality is a concern, a one-time `UPDATE t_grotto SET field = TRIM(field) WHERE field <> TRIM(field)` per affected column could be worth running separately (not necessarily in this PR).

### Nitpicks (Optional)

4. **[api/services/GrottoService.js:31]** `yearBirth` is the only field in `getConvertedDataFromClientRequest` not wrapped in `trimIfString` — intentionally, since it's `type: 'number'` in the model and `trimIfString` would be a no-op anyway. A brief inline comment (`// numeric field — no trim needed`) would make the asymmetry obviously intentional rather than an accidental omission for the next reader.

5. **[test/integration/4_routes/Organization/update.test.js:62]** The integration test only verifies `postalCode` is trimmed in the response, but the PR also trims `address`, `city`, `region`, etc. An additional assertion on a second field (e.g., send `address: "Rue de l'Étang "` and assert the trimmed value) would give end-to-end coverage of the general-case trimming, not just the specific bug that motivated the PR.

ClemRz:
Thanks for the review @Paul-AUB — addressed in 026adc90:

| # | Item | Action |
|---|------|--------|
| 1 | Move `trimIfString` to `api/utils/` | ✅ Created `api/utils/trimIfString.js`, service now imports from there |
| 2 | Test the `undefined` branch | ✅ Split into two explicit tests (null + undefined) |
| 3 | Data cleanup migration | Skipped per your note — will handle separately if needed |
| 4 | Comment on `yearBirth` | ✅ Added inline comment |
| 5 | Assert second trimmed field in integration test | ✅ Added `address` assertion |

ClemRz:
I cleaned up the t_grotto fields:
- `postal_code`: 1 row affected
- `address`: 41 rows affected
- `city`: 16 rows affected
- `custom_message`: 24 rows affected
- `mail`: 5 rows affected
- `region`: 1 row affected
- `url`: 1 row affected

PR #1673, opened 2026-06-15
fix(account): prevent 500 when duplicate nicknames exist in DB

### LINKED ISSUE #1671: bug(api): [500] tcaver - findOne crash on duplicate nickname during account update
**Describe the bug**
When updating account details via `PATCH /api/v1/account`, the nickname uniqueness check uses `TCaver.findOne({ nickname })` which throws an unhandled `OperationalError` when multiple cavers already share the same nickname in the database. The `findOne()` method in Waterline throws if more than one record matches. This means users with pre-existing duplicate nicknames (IDs 427 and 32181 both have nickname "Frédéric Urien") cannot update any of their account fields without triggering a 500.

**To Reproduce**
Based on the trace logs:
1. `PATCH /api/v1/account`
2. Authenticate as user 32181 (nickname "Frédéric Urien")
3. Submit any update that includes `nickname` matching the current value
4. The controller calls `TCaver.findOne({ nickname: 'Frédéric Urien' })` at line 150
5. Waterline finds two matching records (IDs 427 and 32181) and throws

**Input Data**

- **Endpoint:** `PATCH /api/v1/account`
- **Request Body:**
```json
{
  "nickname": "Frédéric Urien",
  "name": "Frédéricc",
  "surname": "Urien"
}
```
- **User:** id=32181, nickname="Frédéric Urien", groups=["Administrator", "Moderator", "User", "Leader"]

**Expected behavior**
The API should handle pre-existing duplicate nicknames gracefully. The uniqueness check should either:
1. Use `TCaver.find({ nickname }).limit(2)` instead of `findOne()` and check manually, or
2. Exclude the current user from the query: `TCaver.findOne({ nickname, id: { '!=': req.token.id } })`
3. Additionally, a UNIQUE constraint should be added to the `nickname` column to prevent future duplicates, and existing duplicates should be resolved via data migration.

**Error Details**
- Error Type: OperationalError
- Occurrences: 4
- Time Range: 2026-06-14T13:26:57Z to 2026-06-15T06:21:54Z
- Representative Trace ID: 019ec9f1-9784-7648-b673-ab0ea631275c
- Stack Trace:
```
OperationalError: More than one matching record found for `Tcaver.findOne()`:
···
427,32181
···

Criteria used:
···
{
  where: { nickname: 'Frédéric Urien' },
  limit: 2,
  skip: 0,
  sort: []
}
···
    at module.exports [as v1/account/update] (/home/site/wwwroot/api/controllers/v1/account/update.js:150:40)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
```

**Additional context**
- Service: `api/controllers/v1/account/update.js` (line 150)
- Model: `tcaver`
- Root cause: The database has duplicate nicknames (caver IDs 427 and 32181 both have nickname "Frédéric Urien"). The `findOne()` call at line 150 does not exclude the current user, so when user 32181 submits their unchanged nickname, it finds both records and throws.
- Fix: Change the findOne query to exclude the requesting user: `TCaver.findOne({ nickname: req.body.nickname, id: { '!=': req.token.id } })` — which is actually what's already done on the current `develop` branch at line 150. The deployed code may be slightly behind or the logic path differs. Additionally, a data cleanup is needed for existing duplicates (caver IDs 427 and 32181).

ClemRz:
## 🤔 What

- Fix `PATCH /api/v1/account` crashing with 500 when duplicate nicknames exist in the database
- Closes #1671

## 🤷‍♂️ Why

- `TCaver.findOne({ nickname })` throws when multiple records match (Waterline expects 0 or 1 results from `findOne`)
- Production has legacy duplicate nicknames (e.g., caver IDs 427 and 32181 share the same nickname)
- User 32181 repeatedly fails to update their account because of this crash

## 🔍 How

- Replaced `findOne({ nickname })` + post-hoc ID check with `find({ nickname, id: { '!=': req.token.id } }).limit(1)`
- This excludes the current user upfront and uses `find` which handles multiple matches gracefully
- If any other caver already has the nickname, returns 409 Conflict

## 🧪 Testing

- Added regression test that creates two cavers with the same nickname and verifies the endpoint returns 409 (not 500)
- All existing account tests pass (41 total)

### DIFF
--- a/api/controllers/v1/account/update.js
+++ b/api/controllers/v1/account/update.js
@@ -147,10 +147,11 @@ module.exports = async (req, res) => {
     if (!req.body.nickname) {
       return res.badRequest('You must provide a nickname.');
     }
-    const existingCaver = await TCaver.findOne({
+    const existingCavers = await TCaver.find({
       nickname: req.body.nickname,
-    });
-    if (existingCaver && existingCaver.id !== req.token.id) {
+      id: { '!=': req.token.id },
+    }).limit(1);
+    if (existingCavers.length > 0) {
       return res.conflict('This nickname is already used.');
     }
     updates.nickname = req.body.nickname;
--- a/api/controllers/v1/account/update.js
+++ b/api/controllers/v1/account/update.js
@@ -147,11 +147,14 @@ module.exports = async (req, res) => {
     if (!req.body.nickname) {
       return res.badRequest('You must provide a nickname.');
     }
-    const existingCavers = await TCaver.find({
-      nickname: req.body.nickname,
-      id: { '!=': req.token.id },
-    }).limit(1);
-    if (existingCavers.length > 0) {
+    const nicknameConflict =
+      (
+        await TCaver.find({
+          nickname: req.body.nickname,
+          id: { '!=': req.token.id },
+        }).limit(1)
+      ).length > 0;
+    if (nicknameConflict) {
       return res.conflict('This nickname is already used.');
     }
     updates.nickname = req.body.nickname;
--- a/test/integration/4_routes/Account.test.js
+++ b/test/integration/4_routes/Account.test.js
@@ -245,6 +245,36 @@ describe('Account features', () => {
         .expect(409);
     });
 
+    it('should return 409 even when duplicate nicknames exist in DB', async () => {
+      // Simulate legacy data: create a caver with a nickname that already exists
+      const duplicateNickname = 'DuplicatedNick';
+      const caver1 = await TCaver.create({
+        nickname: duplicateNickname,
+        mail: 'dup1@test.com',
+        password: 'hashed',
+        language: '000',
+      }).fetch();
+      const caver2 = await TCaver.create({
+        nickname: duplicateNickname,
+        mail: 'dup2@test.com',
+        password: 'hashed',
+        language: '000',
+      }).fetch();
+
+      // The authenticated user (id 3) trying to claim this duplicated nickname
+      // should get 409, not 500
+      await supertest(sails.hooks.http.app)
+        .patch('/api/v1/account')
+        .send({ nickname: duplicateNickname })
+        .set('Authorization', userToken)
+        .set('Content-type', 'application/json')
+        .set('Accept', 'application/json')
+        .expect(409);
+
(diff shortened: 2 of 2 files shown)


Paul-AUB (dismissed):

1. **[test/integration/4_routes/Account.test.js:265–275]** The cleanup `await TCaver.destroy(...)` is placed inline at the end of the test body. If the `supertest` assertion on line 272 fails (e.g., the server returns 500 instead of 409), it will throw before reaching the `destroy` call, leaving `caver1` and `caver2` in the test database. Depending on how the test DB is reset between runs this could pollute subsequent tests. Consider wrapping in a `try/finally`:

   ```js
   try {
     await supertest(sails.hooks.http.app)
       .patch('/api/v1/account')
       .send({ nickname: duplicateNickname })
       ...
       .expect(409);
   } finally {
     await TCaver.destroy({ id: [caver1.id, caver2.id] });
   }
   ```

2. **[test/integration/4_routes/Account.test.js]** Consider adding a companion test that verifies a user can still submit their own current nickname without hitting a conflict (e.g., updating `name` and `nickname` simultaneously where `nickname` is unchanged). The fix handles this correctly via the `id: { '!=': req.token.id }` exclusion, but an explicit regression test would guard against future regressions in that path.

### Nitpicks (Optional)

3. **[api/controllers/v1/account/update.js:150–153]** The variable name `existingCavers` is accurate, but since only 0 or 1 results are ever returned (due to `.limit(1)`), `existingCaver` (singular) with an array check, or a boolean idiom like `const nicknameConflict = (await TCaver.find({ ... }).limit(1)).length > 0` would read slightly more naturally. Minor preference only.

ClemRz:
All three review items addressed in 75404ecf:

1. **try/finally cleanup** — Duplicate-nickname test now wraps the assertion in `try/finally` so test cavers are always destroyed, even on assertion failure.
2. **Companion test** — Added a test verifying that submitting your own current nickname (alongside other fields) does not trigger a 409.
3. **Variable rename** — `existingCavers` replaced with a `nicknameConflict` boolean idiom for clarity given the `.limit(1)`.

Lint clean, all tests passing (88 passing, 0 failing).

PR #1674, opened 2026-06-15
chore(db): add missing index and drop dead indexes

ClemRz:
## 🤔 What

- Add missing index on `h_document.id_massif` to eliminate sequential scans during FK constraint checks
- Drop 8 dead indexes that have had 0 scans over 11 months (stats window since 2024-07-21)
- Update test infrastructure (`customSQL.js` and `IndexOptimization.property.test.js`)

## 🤷‍♂️ Why

A production database health audit revealed:
- `h_document.id_massif` FK constraint checks were causing 3.3s sequential scans on a 684k-row table because the composite PK `(id, date_reviewed)` doesn't help queries filtering only on `id_massif`
- 8 indexes consuming write overhead on every INSERT/UPDATE/DELETE without ever being read (confirmed by `pg_stat_user_indexes` over 11 months)
- `t_entrance` had accumulated 22,979 dead tuples (16% bloat) degrading scan performance

## 🔍 How

Migration file:
- `sql/9_05_2026_06_15_h_document_id_massif_index.sql` — `CREATE INDEX CONCURRENTLY` (non-blocking)

Dropped indexes:
| Index | Table | Size | Reason |
|-------|-------|------|--------|
| `idx_t_entrance_of_interest` | t_entrance | 16 kB | 0 scans |
| `idx_t_entrance_country_active` | t_entrance | 1.3 MB | 0 scans |
| `idx_h_name_id` | h_name | 2.5 MB | 0 scans |
| `idx_time_series_date_range` | t_time_series | 16 kB | 0 scans |
| `idx_quantity_kind_code` | t_quantity_kind | 16 kB | 0 scans |
| `idx_t_message_sender` | t_message | 16 kB | 0 scans |
| `idx_point_geom` | t_point | 8 kB | 0 scans |
| `idx_tsql_time_series` | t_time_series_quality_log | 8 kB | 0 scans |

## 🧪 Testing

- `npm run test` passes (IndexOptimization property tests updated)
- Production impact: run migrations, then verify with `SELECT idx_scan FROM pg_stat_user_indexes WHERE indexrelname = 'idx_h_document_id_massif'` after a few hours

### DIFF
--- a/sql/9_05_2026_06_15_h_document_id_massif_index.sql
+++ b/sql/9_05_2026_06_15_h_document_id_massif_index.sql
@@ -1,0 +1,6 @@
+\c grottoce;
+
+-- Add index on h_document.id_massif to avoid seq scans during FK constraint checks.
+-- The composite PK (id, date_reviewed) does not help queries filtering only on id_massif.
+CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_h_document_id_massif
+    ON h_document (id_massif);
--- a/sql/9_06_2026_06_15_drop_dead_indexes.sql
+++ b/sql/9_06_2026_06_15_drop_dead_indexes.sql
@@ -1,0 +1,25 @@
+\c grottoce;
+
+-- Drop indexes with 0 scans over 11 months (since 2024-07-21 stats_reset).
+-- They add write overhead on t_entrance without any read benefit.
+DROP INDEX IF EXISTS idx_t_entrance_of_interest;
+DROP INDEX IF EXISTS idx_t_entrance_country_active;
+
+-- Drop unused indexes on other tables (0 scans in 11 months).
+-- idx_h_name_id: h_name has 0 idx_scans total, this index is never used.
+DROP INDEX IF EXISTS idx_h_name_id;
+
+-- idx_time_series_date_range: never used since creation.
+DROP INDEX IF EXISTS idx_time_series_date_range;
+
+-- idx_quantity_kind_code: never used since creation.
+DROP INDEX IF EXISTS idx_quantity_kind_code;
+
+-- idx_t_message_sender: never used, t_message has only 205 rows.
+DROP INDEX IF EXISTS idx_t_message_sender;
+
+-- idx_point_geom: never used, t_point has very few rows.
+DROP INDEX IF EXISTS idx_point_geom;
+
+-- idx_tsql_time_series: never used since creation.
+DROP INDEX IF EXISTS idx_tsql_time_series;
--- a/sql/9_06_2026_06_15_drop_dead_indexes.sql
+++ b/sql/9_06_2026_06_15_drop_dead_indexes.sql
@@ -1,25 +1,27 @@
 \c grottoce;
 
 -- Drop indexes with 0 scans over 11 months (since 2024-07-21 stats_reset).
--- They add write overhead on t_entrance without any read benefit.
-DROP INDEX IF EXISTS idx_t_entrance_of_interest;
-DROP INDEX IF EXISTS idx_t_entrance_country_active;
+-- Using CONCURRENTLY to avoid blocking writes on busy tables.
+-- Note: CONCURRENTLY cannot run inside a transaction block; these statements
+-- are executed by psql which auto-commits each one individually.
+DROP INDEX CONCURRENTLY IF EXISTS idx_t_entrance_of_interest;
+DROP INDEX CONCURRENTLY IF EXISTS idx_t_entrance_country_active;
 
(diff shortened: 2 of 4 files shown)


Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[PR description vs implementation]** `VACUUM ANALYZE t_entrance` is listed as one of the four deliverables in the "What" section of the PR body, but it does not appear in either migration file (`9_05_…` or `9_06_…`). This creates a mismatch between stated scope and actual implementation. Either add it to `9_06_2026_06_15_drop_dead_indexes.sql` (note: `VACUUM` cannot run inside a transaction, so the migration runner must support it outside one), or update the PR description to explicitly call it a manual DBA step and document how/when it should be run.

2. **[sql/9_06_2026_06_15_drop_dead_indexes.sql:5-25]** `DROP INDEX IF EXISTS` acquires a `ShareLock` on the index's table, which can briefly block `INSERT`/`UPDATE`/`DELETE` on busy production tables. `t_entrance` in particular is a high-traffic table. Consider using `DROP INDEX CONCURRENTLY IF EXISTS` for each statement to make these drops non-blocking — consistent with the approach taken for the new index in `9_05_…`. Note: `DROP INDEX CONCURRENTLY` cannot run inside a transaction block, so this may require verifying the migration runner's transaction wrapping behaviour.

3. **[test/integration/0_models/IndexOptimization.property.test.js:87-88]** The JSDoc comment reads "Covers: all 7 dead indexes with 0 scans over the diagnostic window", but `DEAD_INDEXES` now contains 10 entries after this PR. The count should be updated to keep the documentation accurate.

4. **[test/integration/0_models/IndexOptimization.property.test.js / test/customSQL.js]** `idx_t_message_sender` and `idx_quantity_kind_code` are both removed from `customSQL.js` (test DB setup) and dropped by the migration, yet neither is added to `DEAD_INDEXES` in the property test — unlike `idx_t_entrance_of_interest`, `idx_t_entrance_country_active`, and `idx_h_name_id`, which are. Consider adding them to `DEAD_INDEXES` for consistency so all indexes removed from `customSQL.js` in this batch have explicit test coverage for their absence.

ClemRz:
I removed mentions to the VACUUM.

Paul-AUB (dismissed):
Thank you for the update. The only Must Fix from the previous review — the VACUUM/ANALYZE mismatch — has been resolved by removing it from the PR description.

Two open suggestions from the prior review remain unaddressed (items 3 and 4). Listing them below for tracking; neither is blocking.

---

1. **[sql/9_06_2026_06_15_drop_dead_indexes.sql:5-25]** The eight `DROP INDEX IF EXISTS` statements are still using the blocking form. `DROP INDEX CONCURRENTLY IF EXISTS` would avoid brief write-locks on `t_entrance` and other tables during the migration. Note: `CONCURRENTLY` cannot run inside a transaction block, so it depends on whether the migration runner wraps each file in a transaction. The asymmetry with the `CREATE INDEX CONCURRENTLY` in `9_05_…` is worth noting either way.

2. **[test/integration/0_models/IndexOptimization.property.test.js:86]** The JSDoc comment still reads "Covers: all 7 dead indexes with 0 scans over the diagnostic window", but `DEAD_INDEXES` now contains **10** entries (7 pre-existing + 3 added by this PR). Consider updating the count to `10`.

3. **[test/integration/0_models/IndexOptimization.property.test.js:42-60 / test/customSQL.js]** `idx_t_message_sender` and `idx_quantity_kind_code` are both removed from `customSQL.js` and dropped by `9_06_…`, but neither is added to `DEAD_INDEXES`. Unlike the three other indexes dropped in this batch (`idx_t_entrance_of_interest`, `idx_t_entrance_country_active`, `idx_h_name_id`), they have no explicit Property 9 coverage. Consider adding them (with their respective tables `t_message` and `t_quantity_kind`) to maintain consistency.

ClemRz:
All three suggestions addressed in 07a203d4:

1. **`DROP INDEX CONCURRENTLY`** — switched all 8 drops to `DROP INDEX CONCURRENTLY IF EXISTS` for non-blocking behavior, consistent with the `CREATE INDEX CONCURRENTLY` in `9_05_…`.
2. **JSDoc count** — updated from "all 7 dead indexes" to "all 15 dead indexes".
3. **Missing `DEAD_INDEXES` entries** — added all 5 indexes dropped in `9_06_…` that were missing from the test array (`idx_time_series_date_range`, `idx_quantity_kind_code`, `idx_t_message_sender`, `idx_point_geom`, `idx_tsql_time_series`).

All 2799 tests pass, lint clean.

PR #1675, opened 2026-06-15
fix(name): support language update on names, organizations and massifs

### LINKED ISSUE #1652: `PATCH /api/v1/names/:id` does not support updating the language
# `PATCH /api/v1/names/:id` does not support updating the language

## Summary

The `PATCH /api/v1/names/:id` endpoint currently only accepts the `name` (text) field. There is no way to update the `language` of an existing name through the API.

## Current behavior

`api/controllers/v1/name/update.js` builds the update payload as:

```js
const cleanedData = {
  name: nameText,   // only the text is updated
};
await TName.updateOne({ id: nameId }).set(cleanedData);
```

Any `language` field sent in the request body is silently ignored.

## Impact

- Once a name is created with a wrong language, there is no way to correct it — neither through the UI nor through the API.
- On the frontend, the language field is explicitly **disabled** in edit mode for organizations (and other entities) precisely because this endpoint does not support it. This is a workaround for a missing backend feature, not an intentional design choice.

## Expected behavior

`PATCH /api/v1/names/:id` should accept an optional `language` field (ISO 639-3 code, matching `TLanguage.id`) and update it when provided:

```js
const cleanedData = {
  name: nameText,
};
if (req.param('language')) {
  cleanedData.language = req.param('language');
}
await TName.updateOne({ id: nameId }).set(cleanedData);
```

## Related

- `api/services/GrottoService.js` — `createGrotto` already handles language correctly on creation (accepts ISO 639-3 codes directly as `TLanguage` PKs).
- Frontend: `packages/web-app/src/components/appli/EntitiesForm/Organization/OrganizationFields.jsx` — `disabled={!isNewOrganization}` on the language input.
- Frontend: `packages/web-app/src/actions/Name.js` — `updateName` action that calls this endpoint.

ClemRz:
## 🤔 What

- `PATCH /api/v1/names/:id` now accepts an optional `language` field (ISO 639-3 code) to update the language of an existing name
- `PUT /api/v1/organizations/:id` now accepts an optional `name: { text, language }` field to update the main name atomically in the same request
- `PUT /api/v1/massifs/:id` now accepts the same `name: { text, language }` field
- Closes #1652

## 🤷‍♂️ Why

Once a name was created with a wrong language, there was no way to correct it through the API. The `PATCH /names/:id` endpoint only accepted the `name` text field and silently ignored `language`.

Additionally, organizations and massifs required two separate API calls to update entity data + name (one PUT for the entity, one PATCH for the name), which is non-transactional and inconsistent with how entrances and caves already handle it inline.

## 🔍 How

- **`PATCH /names/:id`**: Added `language` as an optional body parameter. If provided, it's validated against `TLanguage` (must exist, cannot be null). Both `name` and `language` are independently optional (PATCH semantics).

- **`PUT /organizations/:id`** and **`PUT /massifs/:id`**: Added support for an optional `name` object in the request body with `text` and `language` sub-fields. When present, the main name (`isMain: true`) is updated inline — same pattern already used by `PUT /entrances/:id` and `PUT /caves/:id`.

All changes are fully backward compatible. Omitting the new fields preserves existing behavior.

## 🧪 Testing

- Integration tests added for all three endpoints covering: successful updates, both fields together, invalid language, null language, name too long
- Verified manually via curl against local dev server with GET roundtrips confirming persistence

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/api/controllers/v1/cave/update.js
+++ b/api/controllers/v1/cave/update.js
@@ -43,16 +43,39 @@ module.exports = async (req, res) => {
     return res.badRequest(nameError);
   }
 
+  // Validate language if provided
+  const nameLanguage = req.body.name?.language;
+  if (nameLanguage !== undefined) {
+    if (nameLanguage === null) {
+      return res.badRequest('Language cannot be null.');
+    }
+    const foundLanguage = await TLanguage.findOne({ id: nameLanguage });
+    if (!foundLanguage) {
+      return res.badRequest('The provided language does not exist.');
+    }
+  }
+
   // Handle name manually
   // Currently, use only one name per cave (even if the model can handle multiple names)
   // Done before the TCave update so the last_change_cave DB trigger will fetch the last updated name
-  await TName.updateOne({
-    cave: caveId,
-    isMain: true,
-  }).set({
-    name: nameText,
-    language: req.body.name?.language,
-  });
+  if (req.body.name) {
+    const nameUpdate = {};
+    if (nameText !== undefined) {
+      nameUpdate.name = nameText;
+    }
+    if (nameLanguage !== undefined) {
+      nameUpdate.language = nameLanguage;
+    }
+    if (Object.keys(nameUpdate).length > 0) {
+      const updatedName = await TName.updateOne({
+        cave: caveId,
+        isMain: true,
+      }).set(nameUpdate);
+      if (!updatedName) {
+        sails.log.warn(`Cave ${caveId} has no main name record to update.`);
+      }
+    }
+  }
 
   await TCave.updateOne({ id: caveId }).set(updatedFields);
 
--- a/api/controllers/v1/massif/update.js
+++ b/api/controllers/v1/massif/update.js
@@ -2,6 +2,7 @@ const ControllerService = require('../../../services/ControllerService');
 const MassifService = require('../../../services/MassifService');
 const NotificationService = require('../../../services/NotificationService');
 const { toMassif } = require('../../../services/mapping/converters');
+const { validateNameLength } = require('../../../utils/nameValidation');
 
 module.exports = async (req, res) => {
   const massifId = req.param('id');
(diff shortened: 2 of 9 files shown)


Paul-AUB (dismissed):
Good implementation overall — the inline name update pattern is consistent with the existing cave/entrance controllers, and the test coverage is solid. A few things worth considering before merge.

1. **[api/controllers/v1/cave/update.js]** The existing cave controller silently passes `language: req.body.name?.language` (which can be `undefined` or `null`) without the null-guard this PR adds for massif/org/names. Sending `{ name: { language: null } }` to `PUT /caves/:id` would pass validation and attempt to set `language: null` in the DB. Consider fixing `cave/update.js` for consistency, since this PR is already touching the same feature area. If intentionally left out of scope, a TODO comment would help.

2. **[api/controllers/v1/name/update.js:35-46]** If neither `name` nor `language` is provided in the PATCH body, `cleanedData` ends up as `{}` and `TName.updateOne({ id: nameId }).set({})` is still called — a no-op that wastes a DB round-trip and obscures intent. Consider short-circuiting with a 400 ("At least one of `name` or `language` must be provided") when `Object.keys(cleanedData).length === 0`, or at minimum skipping the `updateOne` call.

3. **[api/controllers/v1/massif/update.js:61-64 / organization/update.js:68-71]** `TName.updateOne({ massif/grotto: id, isMain: true })` is silently a no-op if the entity has no main name (Waterline's `updateOne` returns `undefined` without throwing when nothing matches). Consider asserting that the name exists before attempting the update, or at least logging a warning, so data integrity issues don't go undetected.

### Nitpicks (Optional)

4. **[assets/swaggerV1.yaml]** The polygon area limit description change from `8 000 km²` to `35 000 km²` is bundled into the same line as the new `name` field description. The new value is correct (matches `MAX_AREA_KM2 = 35000` in `MassifService.js`), but it's an unrelated fix not mentioned in the PR body — worth calling out in the commit message or splitting to a separate commit for a clean git history.

5. **[api/controllers/v1/name/update.js:24 vs api/controllers/v1/massif/update.js:40]** `name/update.js` reads `language` via `req.param('language')` (body + query string), while the massif/org controllers read via `req.body.name?.language` (body only). Both work correctly for JSON clients, but the inconsistency could be surprising if someone tries to pass `language` as a query param to the massif/org endpoints and wonders why it's silently ignored.

ClemRz (commented):
Thanks for the thorough review! Addressed items 1–3 in fc8dfae2:

1. **Cave language null-guard** — Added the same language validation (null check + `TLanguage` existence check) to `cave/update.js`, plus wrapped the name update in `if (req.body.name)` so it only fires when a name object is present.

2. **Name empty-body short-circuit** — `PATCH /names/:id` now returns 400 with a clear message when neither `name` nor `language` is provided, avoiding the useless DB round-trip.

3. **Warning on missing main name** — All three controllers (cave, massif, organization) now log a warning when `TName.updateOne(...)` returns `undefined` (no main name found).

---

Items 4 and 5 left as-is:

4. The swagger area limit change is correct (`35000` matches `MAX_AREA_KM2` in `MassifService.js`) — acknowledged it's bundled but not worth a separate commit at this point.

5. `req.param('language')` vs `req.body.name?.language` is intentional — `PATCH /names/:id` accepts flat body fields while `PUT /massifs/:id` and `PUT /organizations/:id` use a nested `name` object. Different API contracts by design.

PR #1676, opened 2026-06-15
refactor(search): remove unused arrays from Typesense index and add nbEntrances to caves

### LINKED ISSUE #1580: perf(search): remove unused arrays from Typesense index and search responses
## Problem

The advanced search endpoint (`POST /api/v1/advanced-search`) currently stores and returns large, indefinitely-growing arrays in Typesense documents and API responses that are **never consumed by the frontend** in search results.

For example, every entrance document in Typesense stores its full `descriptions`, `comments`, `histories`, `riggings`, `locations`, `documents`, and `names` arrays. These grow unboundedly as users contribute content, yet the frontend only uses scalar fields (`name`, `latitude`, `cave.depth`, etc.) from search results.

### Typesense index bloat
- These arrays are stored in every document in the Typesense index, consuming RAM and disk
- As contributions grow, document sizes increase without bound — a well-documented entrance could have dozens of descriptions, comments, riggings, etc.
- Full re-syncs become slower as more data is serialized per document
- This affects all 5 indexed entities (entrances, caves, massifs, documents, organizations)

### API response payload
- Search responses are significantly larger than necessary
- A single entrance result can carry kilobytes of unused nested arrays
- Paginated results (10–50 items) multiply this waste
- Increased bandwidth, slower parsing on the client, higher memory pressure

### Maintenance cost
- The `processRows` generator in dbSync performs expensive JOINs to populate these arrays during sync
- `updateInSearch` re-fetches all related entities on every entrance update
- More code to maintain for data that is never displayed

## Analysis

A full audit of the frontend codebase confirms the following arrays are **not used** in search results:

| Entity | Unused arrays |
|--------|--------------|
| **Entrances** | `cave.exploringOrganizations`, `comments`, `descriptions`, `documents`, `histories`, `locations`, `names`, `riggings` |
| **Caves** | `entrances`, `exploringOrganizations` |
| **Massifs** | `descriptions`, `documents`, `networks` |
| **Documents** | `authorsOrganization`, `deletedFiles`, `files`, `modifiedFiles`, `newFiles`, `parent.files` |
| **Organizations** | `cavers`, `documents`, `exploredEntrances`, `exploredNetworks`, `partnerEntrances`, `partnerNetworks` |

For **documents**, 4 arrays are used but can be slimmed to only the fields needed:
- `authors` → only `id` and `nickname`
- `entrances` → only `name`
- `iso3166` → only `iso`
- `massifs` → only `name`

## Proposed solution

1. **Stop indexing unused arrays in Typesense** — remove them from the schema and `processRows`/`importFormater`
2. **Stop populating them in `updateInSearch`** — skip the expensive JOINs for data that won't be searched or returned
3. **Strip them in the `toSearchResult` converter** — for arrays that must stay in Typesense for filtering/faceting but aren't needed in the response
4. **Slim down document arrays** — only index/return the specific sub-fields the frontend needs

## Benefits

- Smaller Typesense index (less RAM, faster syncs)
- Smaller API responses (less bandwidth, faster client rendering)
- Simpler sync code (fewer JOINs, faster `updateInSearch`)
- Reduced risk of hitting Typesense document size limits as content grows

ClemRz:
## 🤔 What

- Add `nbEntrances` scalar field to the Caves Typesense index (same pattern as massifs)
- Remove 21 unused arrays from Typesense indexing and search API responses across all 5 entities
- Keep the `entrances` array on caves for now (frontend still uses `.length` in quicksearch)

Closes #1580

## 🤷‍♂️ Why

The Typesense index stores large, unbounded arrays that are never used in search results. This wastes RAM, inflates response payloads, and slows down syncs with expensive JOINs for data that is never displayed.

## 🔍 How

**Index (dbSync processRows):**
- Entrances: removed 6 joins (descriptions, locations, riggings, histories, documents, comments full fields). Kept comments join slimmed to only rating fields (`aestheticism`, `caving`, `approach`) for `commentsRating` computation.
- Caves: removed descriptions and documents joins. Added `COUNT(e.id) AS "nbEntrances"` via `LEFT JOIN t_entrance`.
- Massifs: removed descriptions and documents joins.
- Documents: removed files join.
- Organizations: already clean (no array joins).

**Real-time updates (updateInSearch):**
- Entrances: replaced individual array mappings with `commentsRating` aggregate. Removed `locations = []` for sensitive entrances (no longer relevant).
- Caves: added `nbEntrances: c.entrances?.length ?? 0`.

**Search responses (toSearchResult converter):**
- Strips unused arrays per entity type after conversion, so non-search API endpoints remain unaffected.
- Removed `exploringOrganizations` from `toSimpleCave` (used only in search context).

## 🧪 Testing

All 2792 existing tests pass. Verified each entity via local curl against quicksearch:
- Caves: `nbEntrances` present, `exploringOrganizations` absent
- Entrances: no comments/descriptions/histories/riggings/locations/documents/names arrays
- Massifs: no descriptions/documents/networks
- Documents: no authorsOrganization/files/newFiles/modifiedFiles/deletedFiles
- Organizations: no cavers/documents/exploredEntrances/exploredNetworks/partnerEntrances/partnerNetworks

After deploy, a full Typesense re-sync is needed to remove the arrays from existing documents.

### DIFF
--- a/api/dbSync/entities/cave.js
+++ b/api/dbSync/entities/cave.js
@@ -31,26 +31,10 @@ async function* processRows(source) {
   for await (const rows of source) {
     const joins = [
       {
-        table: 't_description d',
-        foreignField: 'id_cave',
-        rows,
-        localField: 'descriptions',
-        fields: ['title', 'body', exportUtils.dateAndAuthorFields('d')],
-        join: exportUtils.dateAndAuthorJoins('d'),
-      },
-      {
         table: 't_entrance',
         foreignField: 'id_cave',
         rows,
         localField: 'entrances',
-        fields: ['id'],
-        transform: (e) => e.id,
-      },
-      {
-        table: 't_document',
-        foreignField: 'id_cave',
-        rows,
-        localField: 'documents',
         fields: ['id'],
         transform: (e) => e.id,
       },
--- a/api/dbSync/entities/cave.js
+++ b/api/dbSync/entities/cave.js
@@ -29,6 +29,7 @@ const query = `
 
 async function* processRows(source) {
   for await (const rows of source) {
+    // Keep until frontend migrates from entrances.length to nbEntrances
     const joins = [
       {
         table: 't_entrance',
@@ -50,7 +51,7 @@ function importFormater(d) {
   d.id = `${d.id}`;
   d.dateInscription = new Date(d.dateInscription).getTime();
   if (d.dateReviewed) d.dateReviewed = new Date(d.dateReviewed).getTime();
-  if (d.nbEntrances) d.nbEntrances = parseInt(d.nbEntrances, 10);
+  if (d.nbEntrances != null) d.nbEntrances = parseInt(d.nbEntrances, 10);
   return d;
 }
 /* eslint-enable no-param-reassign */
--- a/api/dbSync/entities/cave.js
+++ b/api/dbSync/entities/cave.js
@@ -14,11 +14,13 @@ const query = `
       c.depth,
       c.length,
       c.temperature,
-      c.is_diving AS "isDiving"
+      c.is_diving AS "isDiving",
+      COUNT(e.id) AS "nbEntrances"
     FROM t_cave AS c
     LEFT JOIN t_name n ON n.id_cave = c.id AND n.is_main = true
     LEFT JOIN t_caver a ON a.id = c.id_author
     LEFT JOIN t_caver r ON r.id = c.id_reviewer
+    LEFT JOIN t_entrance e ON e.id_cave = c.id AND e.is_deleted = false
     WHERE c.is_deleted = false
     GROUP BY c.id, n.name, n.id_language, r.nickname, a.nickname
     ORDER BY c.id ASC
@@ -64,6 +66,7 @@ function importFormater(d) {
   d.id = `${d.id}`;
   d.dateInscription = new Date(d.dateInscription).getTime();
   if (d.dateReviewed) d.dateReviewed = new Date(d.dateReviewed).getTime();
+  if (d.nbEntrances) d.nbEntrances = parseInt(d.nbEntrances, 10);
   return d;
 }
 /* eslint-enable no-param-reassign */
@@ -90,6 +93,7 @@ module.exports = {
         { name: 'length', type: 'int32', optional: true, sort: true },
         { name: 'temperature', type: 'float', optional: true, sort: true },
         { name: 'isDiving', type: 'bool', optional: true, sort: true },
+        { name: 'nbEntrances', type: 'int32', optional: true },
       ],
       default_sorting_field: 'dateInscription',
     },
--- a/api/dbSync/entities/document.js
+++ b/api/dbSync/entities/document.js
@@ -43,8 +43,6 @@ const query = `
     ${exportUtils.PAGGING_PLACEHOLDER}
   `;
 
-const baseUrl = 'https://grottocenter.blob.core.windows.net/documents/';
-
 async function* processRows(source) {
   for await (const rows of source) {
     const joins = [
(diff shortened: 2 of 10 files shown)


Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[api/services/CaveService.js:242]** `nbEntrances: c.entrances?.length ?? 0` counts soft-deleted entrances. `getPopulatedCave` calls `.populate('entrances')` without a `where` clause, so deleted entrances (`isDeleted: true`) are included. The dbSync SQL query correctly uses `AND e.is_deleted = false`, so the two sync paths will diverge whenever a cave is updated (via REST) after an entrance has been soft-deleted — `nbEntrances` will be inflated until the next full re-sync. Since the whole point of this field is to eventually replace the frontend's `entrances.length` check, an incorrect count would be a visible user-facing bug. Fix: `(c.entrances ?? []).filter(e => !e.isDeleted).length`.

2. **[api/dbSync/entities/cave.js:30–45]** The `entrances` join in `processRows` still runs a full JOIN to populate the `entrances` array during every dbSync, even though `nbEntrances` is now computed via a `COUNT` in the main SQL query (line 18). The join is kept intentionally pending the frontend migration, but there's no comment explaining this. Consider adding a brief note (e.g. `// Keep until frontend migrates from entrances.length to nbEntrances`) so the redundancy doesn't look accidental to the next reader.

3. **[api/dbSync/entities/entrance.js:9–13 vs api/services/EntranceService.js:32–36]** `average` and `filterFn` are defined identically in both files. Now that the same computation appears in two places, extracting them to a shared utility (e.g. `api/utils/commentsRating.js`) would prevent future drift.

### Nitpicks (Optional)

4. **[api/dbSync/entities/cave.js:53]** `if (d.nbEntrances) d.nbEntrances = parseInt(d.nbEntrances, 10)` — the guard passes when `d.nbEntrances` is `"0"` (a string from PostgreSQL's `COUNT`, which is truthy), so the behaviour is correct today. However, `if (d.nbEntrances != null)` would make the intent clearer and be robust if the value ever arrives as the number `0`.

ClemRz:
All four review items addressed in aba9e07f:

1. **Fixed** — `CaveService.updateInSearch` now filters soft-deleted entrances: `(c.entrances ?? []).filter((e) => !e.isDeleted).length`
2. **Done** — Added comment explaining the retained `entrances` join in `cave.js` processRows
3. **Done** — Extracted `average`/`filterPositive`/`computeCommentsRating` to `api/utils/commentsRating.js`, updated both consumers
4. **Done** — Changed `if (d.nbEntrances)` to `if (d.nbEntrances != null)` in cave `importFormater`

All 2792 tests pass, lint clean.

Paul-AUB (dismissed):
All four items from the previous review have been addressed. The soft-deleted entrance filter, the comment on the retained `entrances` join, the extracted `commentsRating` utility, and the `!= null` guard are all in place. This is clean, well-scoped work.

1. **[api/services/EntranceService.js:31, ~338–348]** `EntranceService` imports `average` and `filterPositive` from `commentsRating.js` but re-implements the three-field averaging pattern inline rather than calling `computeCommentsRating`. The point of the extraction was to share exactly this logic. Consider replacing the inline block with:
   ```js
   commentsRating: computeCommentsRating(comments ?? []),
   ```
   `computeCommentsRating` doesn't guard against `null`, so passing `comments ?? []` is the right call site fix. This also aligns the real-time update path with the dbSync path, where `computeCommentsRating` is already used correctly.

### Nitpicks (Optional)

2. **[api/utils/commentsRating.js:13–19]** The JSDoc block above `computeCommentsRating` is accurate but adds noise for a function whose name and signature are already self-documenting. A single `// Shared by dbSync and EntranceService` would carry the same load.

3. **[api/services/mapping/converters.js:~295]** `delete data.comments` in the `entrances` branch of `toSearchResult` will be a no-op for documents indexed after this deploy (since `importFormater` already strips `comments`). The intent is transitional cleanup for old indexed documents. Worth noting in a brief inline comment, or remove after the mandatory re-sync is confirmed complete.

ClemRz:
Thanks for the review! All items addressed:

**Item 1 (Suggestion — fixed):** Replaced the inline `commentsRating` averaging in `EntranceService.updateInSearch` with `computeCommentsRating(comments ?? [])`, aligning the real-time path with dbSync.

**Item 2 (Nitpick — skipped):** Keeping the JSDoc — it provides type info for IDE consumers.

**Item 3 (Nitpick — fixed):** Added a transitional comment above the `delete` block in `toSearchResult` explaining these target old indexed documents until the post-deploy full re-sync.

ClemRz:
Follow-up issues created:

1. **Frontend**: GrottoCenter/grottocenter-front#1402 — migrate quicksearch to use `nbEntrances` scalar instead of `entrances.length`
2. **API**: #1682 — remove the `entrances` array from the caves Typesense index once the frontend migration is deployed

PR #1678, opened 2026-06-16
feat(search): add GeoJSON, KML, and GPX export formats

ClemRz:
## 🤔 What

Add GeoJSON, KML, and GPX export formats to the advanced search export endpoint.

- Extends `POST /api/v1/advanced-search/export` with a `format` query parameter (`csv`, `geojson`, `kml`, `gpx`)
- Geographic formats are restricted to entrance searches only
- Sensitive entrances and entries without coordinates are filtered out
- Optional `columns`/`columnsName` support for field selection and aliasing in geo formats
- Dot-notation field resolution works across all formats (`cave.depth`, `commentsRating.approach`, etc.)
- Closes #1677

## 🤷‍♂️ Why

Cavers need to export search results directly into formats compatible with GIS tools (QGIS, Google Earth) and GPS devices. Until now, only CSV was available which requires manual conversion.

## 🔍 How

- **Serializer registry pattern** — each format module (`geojson.js`, `kml.js`, `gpx.js`) exposes `prologue()`, `serializeBatch()`, `epilogue()` for streaming
- **Shared `filterDocuments` utility** — removes sensitive and coordinate-less entries before any serializer sees them
- **Format dispatch in controller** — validates format, checks entity restriction, then delegates to the appropriate streaming path
- **No new npm dependencies** — serializers use template literals, `JSON.stringify`, and a hand-written `escapeXml` function
- **Streaming in batches of 1,000** — same pagination model as the existing CSV path
- **Pre-resolution of dot-notation fields** — the controller uses `resolveField` to traverse nested Typesense documents before passing to serializers
- **GPX `<desc>` element** — includes field data as plain-text lines since GPX 1.1 has no structured metadata equivalent
- **KML nested objects** — JSON-serialized in `<Data>` values instead of `[object Object]`

## 🧪 Testing

- 19 property-based tests (fast-check, 100 iterations each) covering structural validity, coordinate ordering, exclusion logic, metadata invariants, and round-trip stability
- 22 route integration tests covering Content-Type headers, Content-Disposition, field filtering/aliasing, sensitive exclusion, empty results, and backward compatibility
- Unit tests for `filterDocuments` (15 tests)

```bash
# Run all geo-serializer tests
npx mocha --timeout 30000 test/integration/1_services/geo-serializers/*.test.js

# Run route tests (requires sails lifted)
npm run test -- --grep "Geo Export"
```

Manual testing:
```bash
# GeoJSON with field selection
curl -X POST http://localhost:1337/api/v1/advanced-search/export?format=geojson \
  -d '{"query":"*","entity":"entrances","columns":["name","city","cave.depth"],"columnsName":["Name","City","Depth"]}'

# KML (all fields)
curl -X POST http://localhost:1337/api/v1/advanced-search/export?format=kml \
  -d '{"query":"*","entity":"entrances"}'

# GPX with aliases
curl -X POST http://localhost:1337/api/v1/advanced-search/export?format=gpx \
  -d '{"query":"*","entity":"entrances","columns":["name","cave.depth"],"columnsName":["Nom","Profondeur"]}'
```

## 📸 Previews

N/A (API-only change, no UI)

### DIFF
--- a/api/controllers/v1/search/advanced-search-export.js
+++ b/api/controllers/v1/search/advanced-search-export.js
@@ -89,6 +89,40 @@ const GEO_FORMATS = ['geojson', 'kml', 'gpx'];
 const BATCH_SIZE = 1000;
 const MAX_NB_ROW_EXPORT = 10000;
 
+/**
+ * Validate columns and columnsName arrays.
+ * Returns an error message string if invalid, or null if valid.
+ *
+ * @param {Array} columns - Column keys
+ * @param {Array|null} columnsName - Column display names (optional for geo formats)
+ * @param {object} options
+ * @param {boolean} options.requireColumnsName - Whether columnsName is mandatory
+ * @returns {string|null} Error message or null
+ */
+function validateColumns(columns, columnsName, { requireColumnsName }) {
+  const hasColumns = Array.isArray(columns) && columns.length > 0;
+  const hasColumnsName = Array.isArray(columnsName) && columnsName.length > 0;
+
+  if (!hasColumns) {
+    return 'columns must be a non-empty array';
+  }
+  if (!columns.every((c) => typeof c === 'string' && c.length > 0)) {
+    return 'each element in columns must be a non-empty string';
+  }
+  if (requireColumnsName && !hasColumnsName) {
+    return 'columnsName must be a non-empty array';
+  }
+  if (hasColumnsName) {
+    if (columns.length !== columnsName.length) {
+      return 'columns and columnsName must have the same length';
+    }
+    if (!columnsName.every((c) => typeof c === 'string' && c.length > 0)) {
+      return 'each element in columnsName must be a non-empty string';
+    }
+  }
+  return null;
+}
+
 module.exports = async (req, res) => {
   const format = req.param('format') || 'csv';
   const entity = req.param('entity') ?? '';
@@ -115,30 +149,15 @@ module.exports = async (req, res) => {
   const columns = req.param('columns');
   const columnsName = req.param('columnsName');
 
-  // For geo formats, columns/columnsName are optional (when absent, all fields are included).
-  // When provided, they must pass the same validation as CSV.
   const hasColumns = Array.isArray(columns) && columns.length > 0;
   const hasColumnsName = Array.isArray(columnsName) && columnsName.length > 0;
 
   if (!isGeoFormat) {
-    if (!hasColumns) {
-      res.badRequest('columns must be a non-empty array');
-      return;
-    }
-    if (!hasColumnsName) {
-      res.badRequest('columnsName must be a non-empty array');
-      return;
-    }
-    if (columns.length !== columnsName.length) {
-      res.badRequest('columns and columnsName must have the same length');
-      return;
-    }
-    if (!columns.every((c) => typeof c === 'string' && c.length > 0)) {
-      res.badRequest('each element in columns must be a non-empty string');
-      return;
-    }
-    if (!columnsName.every((c) => typeof c === 'string' && c.length > 0)) {
-      res.badRequest('each element in columnsName must be a non-empty string');
+    const error = validateColumns(columns, columnsName, {
+      requireColumnsName: true,
+    });
+    if (error) {
+      res.badRequest(error);
       return;
     }
   }
@@ -146,25 +165,16 @@ module.exports = async (req, res) => {
   // For geo formats with columns provided, validate and build field mapping
   let fieldMapping = null;
   if (isGeoFormat && hasColumns) {
-    if (!columns.every((c) => typeof c === 'string' && c.length > 0)) {
-      res.badRequest('each element in columns must be a non-empty string');
+    const error = validateColumns(columns, columnsName, {
+      requireColumnsName: false,
+    });
+    if (error) {
+      res.badRequest(error);
       return;
     }
     if (hasColumnsName) {
-      if (columns.length !== columnsName.length) {
-        res.badRequest('columns and columnsName must have the same length');
-        return;
-      }
-      if (!columnsName.every((c) => typeof c === 'string' && c.length > 0)) {
-        res.badRequest(
-          'each element in columnsName must be a non-empty string'
-        );
-        return;
-      }
-      // Map each column key to its display alias
       fieldMapping = columns.map((key, i) => ({ key, alias: columnsName[i] }));
     } else {
-      // columns provided without aliases — use key names as-is
       fieldMapping = columns.map((key) => ({ key, alias: key }));
     }
   }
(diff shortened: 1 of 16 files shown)


Paul-AUB (changes requested):
This is a well-scoped feature with solid test coverage (property-based + integration). The streaming approach and serializer-registry pattern are clean. The diff exceeds 500 lines so this review prioritises business logic, security, and API correctness. Swagger and pure test files are not examined in depth.

### Issues (Must Fix)

1. **[advanced-search-export.js:241-255 / geojson.js:39-41 / kml.js:84-92]** When `fieldMapping` is applied, the controller builds pre-mapped documents as `{id, latitude, longitude, ...userAliasedFields}` and then passes `null` as the `fieldMapping` argument to all serializers. In `geojson.js`, the `null` path spreads the whole document (`properties = { ...doc }`), so `id`, `latitude`, and `longitude` always appear in `properties` — even when the user requested only `columns: ['name']`. In `kml.js`, the same `null` path emits all keys as `<Data>` elements in `<ExtendedData>`, again leaking the coordinate fields. This breaks the field-selection contract. `gpx.js` already handles this correctly by explicitly excluding `['latitude', 'longitude', 'altitude', 'name', 'id']` from `<desc>`. Apply the same exclusion in the GeoJSON and KML serializers (or strip the hidden geometry fields from the pre-mapped doc before serialising).

2. **[kml.js:72]** Altitude defaults to `0` when missing: `const alt = doc.altitude != null ? doc.altitude : 0`. In 3D-capable KML viewers (Google Earth, QGIS 3D) a hardcoded `0` places placemarks at sea level rather than at an unknown elevation, which is visually misleading for mountain or underground sites. `gpx.js` correctly omits `<ele>` when altitude is `null`. KML should do the same — the `<coordinates>` element accepts `lon,lat` without an altitude component:
   ```
   <Point><coordinates>${lon},${lat}${doc.altitude != null ? `,${alt}` : ''}</coordinates></Point>
   ```

3. **[kml.js:99]** The `<coordinates>` content `${lon},${lat},${alt}` is inserted verbatim without going through `escapeXml`. While these values are expected to be numeric, the rest of the file routes everything through `escapeXml`, and `filterDocuments` only guards against `null`/`undefined`, not malformed strings. Consider wrapping: `${escapeXml(String(lon))},${escapeXml(String(lat))},${escapeXml(String(alt))}`.

4. **[advanced-search-export.js:255 / geojson.js:30-53 / kml.js:66-103 / gpx.js:69-85]** The `fieldMapping` parameter accepted by all three serializers' `serializeBatch` functions is never used — the controller always passes `null` after pre-mapping. This makes the parameter misleading for future maintainers. Consider either removing it from the serializer signatures (and documenting that pre-mapping is the controller's responsibility) or removing the pre-mapping from the controller and delegating to the serializers.

5. **[advanced-search-export.js:215-222 and 309-316]** In both the geo and CSV streaming loops, the `hasSentHeader` guard inside `results.found > MAX_NB_ROW_EXPORT` is unreachable. `results.found` is the total result count (constant across pages), and headers are written only after this check passes, so `hasSentHeader` is always `false` when the limit is first exceeded. The `break` branch inside `if (!hasSentHeader)` can never execute. Consider simplifying to an early return on the first page only.

6. **[gpx.js:19-27 / kml.js:35-41]** `escapeXml` is defined independently in both files with a subtle difference: `gpx.js` calls `String(str)` directly (producing the string `"null"` when `str` is `null`), while `kml.js` routes through `toStringValue` which returns `''` for null inputs. This inconsistency can cause literal `"null"` to appear in GPX `<desc>` for missing fields. Extracting a shared utility (e.g., `api/services/geo-serializers/xml-utils.js`) would resolve both the duplication and the inconsistency.

### Nitpicks (Optional)

7. **[advanced-search-export.js:123-169]** The column validation block for `isGeoFormat && hasColumns` (lines 148-169) mirrors the CSV validation block (lines 123-143) almost exactly. Extracting a `validateColumns(columns, columnsName, res)` helper would reduce duplication and make both paths easier to read.

8. **[geojson.js:21]** `name`, `description`, and `timestamp` are non-standard top-level properties on a `FeatureCollection` (RFC 7946 defines only `type`, `features`, `bbox`, and `id`). Strict GIS parsers may ignore or warn on them. Consider moving them into a custom `metadata` sub-object, or accepting that they are informational-only extras (low risk, but worth acknowledging in the PR description or a code comment).

ClemRz:
All review items addressed in ab307af0:

**Issues (Must Fix)**
1. ✅ Geometry fields (`id`, `latitude`, `longitude`, `altitude`) now excluded from GeoJSON `properties` and KML `<ExtendedData>`. Field-selection contract is honoured across all serializers.
2. ✅ KML omits altitude from `<coordinates>` when null (produces `lon,lat` instead of `lon,lat,0`). Matches GPX behavior.

**Suggestions (Should Consider)**
3. ✅ KML coordinates now routed through `escapeXml` for consistency.
4. ✅ Removed unused `fieldMapping` parameter from all serializer `serializeBatch` signatures. Pre-mapping is the controller's responsibility (documented in module headers).
5. ✅ Simplified `MAX_NB_ROW_EXPORT` check to first-page-only guard — removed unreachable `break` branch.
6. ✅ Extracted shared `api/services/geo-serializers/xml-utils.js` with `escapeXml` and `toStringValue`. Both KML and GPX now use consistent null → `''` handling.

**Nitpicks (Optional)**
7. ✅ Extracted `validateColumns()` helper in the controller to deduplicate CSV/geo validation blocks.
8. ✅ Added code comment acknowledging non-standard RFC 7946 top-level properties (low risk, kept as informational extras).

Tests and linter pass.

Paul-AUB (dismissed):
All 8 items from the previous review are cleanly resolved in ab307af0. The streaming architecture, serializer-registry pattern, and test coverage remain solid. Two minor items worth noting on the updated code:

1. **[kml.js:25]** KML's `GEOMETRY_FIELDS` does not include `'name'`, while GPX's does (`['latitude', 'longitude', 'altitude', 'name', 'id']`). When no field mapping is active, `name` appears both in the `<name>` element (via `doc.name`) and as a `<Data>` entry in `<ExtendedData>`, producing a duplicate visible in KML viewers. Adding `'name'` to KML's `GEOMETRY_FIELDS` would match GPX's behavior and eliminate the redundancy.

2. **[geojson.js:50 / kml.js:73]** Both serializers unconditionally append a serializer-generated `url` entry after processing the document fields. In GeoJSON `properties.url` is set after the field spread, so if a user aliases a column as `"url"` (e.g. `columnsName: ["url"]`), the serializer overwrites it silently. The same collision applies to KML's `urlData` entry. Using a fixed, prefixed key like `"grottocentreUrl"` would be collision-proof without breaking existing consumers.

### Nitpicks (Optional)

3. **[advanced-search-export.js:343 vs 270]** Minor inconsistency: the CSV exit condition is `documents.length !== BATCH_SIZE` while the geo exit condition uses `documents.length < BATCH_SIZE`. Both are safe (Typesense never returns more than the requested page size), but aligning them to `<` would be slightly more defensive and easier to reason about.

ClemRz:
All three items from the second review addressed in 0478f232:

**Suggestions (Should Consider)**
1. ✅ Added `'name'` to KML's `GEOMETRY_FIELDS` — `name` no longer appears as both a `<name>` element and a `<Data>` entry in `<ExtendedData>`. Matches GPX behavior.
2. ✅ Renamed serializer-generated URL key from `url` to `grottocenterUrl` in GeoJSON (`properties.grottocenterUrl`) and KML (`<Data name="grottocenterUrl">`). Prevents silent overwriting if a user aliases a column as `"url"`. GPX is unaffected (URL lives in a `<link href>` attribute, not a named data key).

**Nitpicks (Optional)**
3. ✅ Aligned CSV exit condition from `!== BATCH_SIZE` to `< BATCH_SIZE`, matching the geo streaming loop.

Tests and linter pass.

PR #1679, opened 2026-06-16
feat(observations): add isotope delta quantity kind and per mil unit

ClemRz:
## 🤔 What

- Add `IsotopeDelta` quantity kind (per mil relative to a reference standard)
- Add `per_mil` unit (‰)
- Update test fixtures accordingly
- Relates to GrottoCenter/grottocenter-front#1400

## 🤷‍♂️ Why

Isotope delta values (δ¹⁸O, δ²H, δ¹³C) expressed in ‰ relative to VSMOW or VPDB are standard measurements in cave hydrology and paleoclimate research. These cannot currently be represented in sensor configurations because the unit and quantity kind are missing.

## 🔍 How

- New quantity kind `IsotopeDelta` with `symbol_si = '1'` (dimensionless), `display_symbol = '‰'`, `factor = 1000`, `offset = 0` — aligned with how `RelativeHumidity` handles dimensionless quantities (store true ratio in SI, display in ‰).
- New unit `per_mil` with symbol `‰`.
- The reference standard (VSMOW, VPDB) is not encoded here — it will be derivable from the substance/analyte field once GrottoCenter/grottocenter-front#1399 lands.

## 🧪 Testing

Run `npm run dev:clean` to rebuild the local database with the new seed data, then verify:
```sql
docker exec grotto-postgres psql -U root -d grottoce -c "SELECT * FROM t_quantity_kind WHERE code = 'IsotopeDelta';"
docker exec grotto-postgres psql -U root -d grottoce -c "SELECT * FROM t_unit WHERE code = 'per_mil';"
```

## 📸 Previews

N/A — data-only change, no UI impact on the API side.

### DIFF
--- a/sql/9_05_2026_06_03_scientific_observations_enumerations.sql
+++ b/sql/9_05_2026_06_03_scientific_observations_enumerations.sql
@@ -45,7 +45,8 @@ INSERT INTO t_quantity_kind (id, code, url, symbol_si, display_symbol, si_to_dis
   (22, 'LightIntensity', 'http://qudt.org/vocab/quantitykind/Illuminance', 'lx', 'lx', 1, 0),
   (23, 'AirVelocity', 'http://qudt.org/vocab/quantitykind/Speed', 'm/s', 'm/s', 1, 0),
   (24, 'WaterVelocity', 'http://qudt.org/vocab/quantitykind/Speed', 'm/s', 'm/s', 1, 0),
-  (25, 'RadonConcentration', 'http://qudt.org/vocab/quantitykind/ActivityConcentration', 'Bq/m³', 'Bq/m³', 1, 0)
+  (25, 'RadonConcentration', 'http://qudt.org/vocab/quantitykind/ActivityConcentration', 'Bq/m³', 'Bq/m³', 1, 0),
+  (26, 'IsotopeDelta', 'http://qudt.org/vocab/quantitykind/DimensionlessRatio', '1', '‰', 1000, 0)
 ON CONFLICT (code) DO NOTHING;
 
 -- ============================================================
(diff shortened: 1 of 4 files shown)


Paul-AUB (dismissed):

1. **[test/fixtures/tunit.json]** The new fixture entry uses `per_mil` as its `code` value, which matches the production SQL code exactly. However, the two existing fixture entries use abbreviated uppercase codes (`DEG_C`, `HPA`) that deliberately differ from their production counterparts (`degree_celsius`, `hectopascal`). This creates a mixed convention: two entries are test-only codes, one is the real production code. If any test ever checks `unit.code` by value, this inconsistency could produce surprising results. Consider either aligning the new entry to the abbreviated style (`PER_MIL`) or, if this PR is the opportunity to fix the pattern, updating the existing entries to use their production codes as well.

### Nitpicks (Optional)

2. **[test/fixtures/tsensorconfiguration.json]** There is no sensor configuration fixture that references the new `IsotopeDelta` quantity kind (id 3) or the new `per_mil` unit (id 3). Adding one would confirm end-to-end that the enum values are valid foreign keys in context, though this is optional for a data-only seed change.

3. **[sql/9_05_2026_06_03_scientific_observations_enumerations.sql:49]** The QUDT URL `http://qudt.org/vocab/quantitykind/DimensionlessRatio` is a reasonable fit. Worth noting for future readers: QUDT does not yet have a dedicated `IsotopeDelta` quantity kind, so `DimensionlessRatio` is the closest available term. If QUDT adds one later, this URL would need updating.

ClemRz:
Thanks for the review @Paul-AUB! Addressed in bd5fe680:

1. **Unit code convention** — Renamed `per_mil` → `PER_MIL` in `tunit.json` to align with the existing abbreviated uppercase convention (`DEG_C`, `HPA`).

2. **IsotopeDelta sensor config fixture** — Added a new fixture (id 4) referencing `quantityKind: 3` (IsotopeDelta) and `unit: 3` (PER_MIL) to confirm end-to-end FK validity.

3. **QUDT URL note** — Acknowledged, no action needed. Left as-is since `DimensionlessRatio` is the closest available QUDT term.

PR #1680, opened 2026-06-16
feat(observations): add substance field to sensor configurations

ClemRz:
## 🤔 What

Add a `substance` field to sensor configurations and time series, enabling tracking of what specific substance is being measured for Concentration and IsotopeDelta quantity kinds.

- Add `substance` column to `TSensorConfiguration` and `TTimeSeries` models
- Add validation: substance is required for Concentration/IsotopeDelta QKs, forbidden for others
- Propagate substance from sensor config to time series during observation import
- Auto-clear substance when quantity kind changes to a non-substance type on update

## 🤷‍♂️ Why

Concentration and isotope delta measurements are meaningless without knowing what substance is being measured (e.g., NO₃⁻, δ¹⁸O). This field allows distinguishing between multiple sensor configurations that measure different substances with the same quantity kind.

## 🔍 How

- Model: `substance` attribute (string, max 100 chars, nullable) on `TSensorConfiguration` and `TTimeSeries`
- Create controller: validates substance presence/absence based on quantity kind code
- Update controller: validates substance rules, auto-clears when QK changes to non-substance type
- `EntityBuilder`: copies substance from sensor config to time series at import time
- SQL migration: adds `substance` column to both tables, adds Concentration quantity kind and mg/L unit
- Swagger: documents the new field on relevant endpoints

## 🧪 Testing

- `npm run test -- --grep "substance"` runs all substance-related tests
- Property-based tests cover create/update validation logic and entity builder propagation
- Integration tests cover CRUD operations and find responses

## 📸 Previews

N/A — backend only.

### DIFF
--- a/api/controllers/v1/sensor-configuration/create.js
+++ b/api/controllers/v1/sensor-configuration/create.js
@@ -77,7 +77,7 @@ module.exports = async (req, res) => {
     detectionLimitMin: detectionLimitMin ?? null,
     detectionLimitMax: detectionLimitMax ?? null,
     label: label || null,
-    substance: substance || null,
+    substance: SensorConfigurationService.normalizeSubstance(substance),
     author: req.token.id,
     dateInscription: new Date(),
   };
--- a/api/controllers/v1/sensor-configuration/create.js
+++ b/api/controllers/v1/sensor-configuration/create.js
@@ -48,12 +48,22 @@ module.exports = async (req, res) => {
   const detectionLimitMin = req.param('detectionLimitMin');
   const detectionLimitMax = req.param('detectionLimitMax');
   const label = req.param('label');
+  const substance = req.param('substance');
 
   // Validate label length
   if (label && label.length > 300) {
     return res.badRequest(
       'The sensor configuration label must not exceed 300 characters.'
     );
+  }
+
+  // Validate substance
+  const substanceError = SensorConfigurationService.validateSubstance(
+    substance,
+    existingQuantityKind.code
+  );
+  if (substanceError) {
+    return res.badRequest(substanceError);
   }
 
   // 7. Build data object
@@ -67,6 +77,7 @@ module.exports = async (req, res) => {
     detectionLimitMin: detectionLimitMin ?? null,
     detectionLimitMax: detectionLimitMax ?? null,
     label: label || null,
+    substance: substance || null,
     author: req.token.id,
     dateInscription: new Date(),
   };
--- a/api/controllers/v1/sensor-configuration/update.js
+++ b/api/controllers/v1/sensor-configuration/update.js
@@ -102,7 +102,10 @@ module.exports = async (req, res) => {
     effectiveQkCode = resolvedQuantityKind.code;
   } else {
     const qk = await TQuantityKind.findOne({ id: config.quantityKind });
-    effectiveQkCode = qk?.code;
+    if (!qk) {
+      return res.serverError('Could not resolve existing quantity kind.');
+    }
+    effectiveQkCode = qk.code;
   }
 
   // 7d. Substance validation
@@ -150,7 +153,8 @@ module.exports = async (req, res) => {
   if (shouldAutoClear) {
     updateData.substance = null;
   } else if (substance !== undefined) {
-    updateData.substance = substance;
+    updateData.substance =
+      SensorConfigurationService.normalizeSubstance(substance);
   }
 
   // 9. Update the record
--- a/api/controllers/v1/sensor-configuration/update.js
+++ b/api/controllers/v1/sensor-configuration/update.js
@@ -28,6 +28,7 @@ module.exports = async (req, res) => {
   const detectionLimitMin = req.param('detectionLimitMin');
   const detectionLimitMax = req.param('detectionLimitMax');
   const label = req.param('label');
+  const substance = req.param('substance');
 
   // 4. If no recognized updatable field is present → 400
   const hasUpdatableField =
@@ -38,11 +39,12 @@ module.exports = async (req, res) => {
     resolution !== undefined ||
     detectionLimitMin !== undefined ||
     detectionLimitMax !== undefined ||
-    label !== undefined;
+    label !== undefined ||
+    substance !== undefined;
 
   if (!hasUpdatableField) {
     return res.badRequest(
-      'You must provide at least one updatable field (quantityKind, unit, precisionUpper, precisionLower, resolution, detectionLimitMin, detectionLimitMax, label).'
+      'You must provide at least one updatable field (quantityKind, unit, precisionUpper, precisionLower, resolution, detectionLimitMin, detectionLimitMax, label, substance).'
     );
   }
 
@@ -71,11 +73,10 @@ module.exports = async (req, res) => {
   }
 
   // 6. Validate quantityKind if provided
+  let resolvedQuantityKind;
   if (quantityKind !== undefined) {
-    const existingQuantityKind = await TQuantityKind.findOne({
-      id: quantityKind,
-    });
-    if (!existingQuantityKind) {
+    resolvedQuantityKind = await TQuantityKind.findOne({ id: quantityKind });
+    if (!resolvedQuantityKind) {
       return res.badRequest('You must provide a valid quantity kind.');
     }
   }
@@ -95,6 +96,41 @@ module.exports = async (req, res) => {
     );
   }
 
+  // 7c. Determine effective quantity kind code for substance validation
+  let effectiveQkCode;
+  if (resolvedQuantityKind) {
+    effectiveQkCode = resolvedQuantityKind.code;
+  } else {
+    const qk = await TQuantityKind.findOne({ id: config.quantityKind });
+    effectiveQkCode = qk?.code;
+  }
+
+  // 7d. Substance validation
+  if (substance !== undefined) {
+    const substanceError = SensorConfigurationService.validateSubstance(
+      substance,
+      effectiveQkCode
+    );
+    if (substanceError) {
+      return res.badRequest(substanceError);
+    }
+  } else if (
+    quantityKind !== undefined &&
+    SensorConfigurationService.isSubstanceRequired(effectiveQkCode) &&
+    !config.substance
+  ) {
+    return res.badRequest(
+      'Substance is required for Concentration or IsotopeDelta quantity kinds.'
+    );
+  }
+
+  // 7e. Auto-clear substance when QK changes to non-substance type
+  const shouldAutoClear =
+    quantityKind !== undefined &&
+    !SensorConfigurationService.isSubstanceRequired(effectiveQkCode) &&
+    config.substance != null &&
+    substance === undefined;
+
   // 8. Build update set: only provided fields + reviewer/dateReviewed
   const updateData = {
     reviewer: req.token.id,
@@ -111,6 +147,11 @@ module.exports = async (req, res) => {
   if (detectionLimitMax !== undefined)
     updateData.detectionLimitMax = detectionLimitMax;
   if (label !== undefined) updateData.label = label || null;
+  if (shouldAutoClear) {
+    updateData.substance = null;
+  } else if (substance !== undefined) {
+    updateData.substance = substance;
+  }
 
   // 9. Update the record
   await TSensorConfiguration.updateOne({ id }).set(updateData);
--- a/api/models/TSensorConfiguration.js
+++ b/api/models/TSensorConfiguration.js
@@ -72,6 +72,13 @@ module.exports = {
       maxLength: 300,
     },
 
+    substance: {
+      type: 'string',
+      allowNull: true,
+      columnName: 'substance',
+      maxLength: 100,
+    },
+
     // Audit fields
     author: {
       allowNull: false,
--- a/api/models/TTimeSeries.js
+++ b/api/models/TTimeSeries.js
@@ -128,6 +128,13 @@ module.exports = {
       maxLength: 100,
     },
 
+    substance: {
+      type: 'string',
+      allowNull: true,
+      columnName: 'substance',
+      maxLength: 100,
+    },
+
     timezoneOffset: {
       type: 'string',
       allowNull: true,
--- a/api/services/SensorConfigurationService.js
+++ b/api/services/SensorConfigurationService.js
@@ -1,4 +1,42 @@
+/**
+ * Codes of quantity kinds that require the substance field.
+ */
+const SUBSTANCE_REQUIRING_CODES = ['Concentration', 'IsotopeDelta'];
+
+/**
+ * Determines if a quantity kind code requires a substance.
+ * @param {string} code - The quantity kind code
+ * @returns {boolean}
+ */
+const isSubstanceRequired = (code) => SUBSTANCE_REQUIRING_CODES.includes(code);
+
+/**
+ * Validates substance field against the effective quantity kind.
+ * Returns an error message string or null if valid.
+ *
+ * @param {string|null|undefined} substance - The substance value from the request
+ * @param {string} quantityKindCode - The effective quantity kind code
+ * @returns {string|null} Error message or null
+ */
+const validateSubstance = (substance, quantityKindCode) => {
+  if (substance != null && substance.length > 100) {
+    return 'The substance must not exceed 100 characters.';
+  }
+  if (isSubstanceRequired(quantityKindCode)) {
+    if (substance == null || substance.trim().length === 0) {
+      return 'Substance is required for Concentration or IsotopeDelta quantity kinds.';
+    }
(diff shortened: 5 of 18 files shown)


Paul-AUB (changes requested):
The feature logic is solid: clean service layer, consistent validation, and impressive property-based test coverage. Two SQL-level issues need attention before this is production-safe.

### Issues (Must Fix)

1. **[sql/9_04_2026_06_03_scientific_observations_ddl.sql]** No ALTER TABLE migration for the `substance` column. The `CREATE TABLE IF NOT EXISTS` blocks are no-ops on environments where these tables already exist (production, staging). The column will silently be absent there, causing ORM errors at runtime. A dedicated migration file is required, following the established pattern in `sql/9_07_2026_06_13_device_serial_number_sensor_label.sql`:
   ```sql
   ALTER TABLE t_sensor_configuration ADD COLUMN IF NOT EXISTS substance varchar(100) NULL;
   ALTER TABLE t_time_series          ADD COLUMN IF NOT EXISTS substance varchar(100) NULL;
   ```

2. **[sql/9_05_2026_06_03_scientific_observations_enumerations.sql:37-46]** The enumeration script reassigns IDs 18–22 to different quantity kind codes (e.g. `LightIntensity` shifts from id=22 to id=18, `IsotopeDelta` from id=26 to id=22). If these rows already exist in a deployed environment, two things break. First, any `t_sensor_configuration` rows that reference the old IDs (e.g. `id_quantity_kind = 22` for LightIntensity) now point to the wrong quantity kind. Second, the INSERT `(17, 'Concentration', …)` will hit a **primary key conflict** on id=17 (which already holds `NitrateConcentration`); `ON CONFLICT (code) DO NOTHING` only resolves conflicts on the `code` unique constraint, not on the `id` primary key, so the entire transaction will error. A proper incremental migration (DELETE the retired codes, INSERT the new ones, UPDATE FK references) is needed for any environment that ran the previous version of `9_05`.

3. **[test/fixtures/tsensorconfiguration.json:47-58]** Fixture #4 (`Channel 3 - δ¹⁸O Isotope Delta`) uses `quantityKind: 3` (IsotopeDelta) but has no `substance` field, violating the invariant introduced by this PR. The fixture represents a valid legacy state (the column is nullable at the DB level), but it's worth either adding a `"substance": null` explicitly for clarity or documenting that this fixture intentionally represents pre-migration data. As-is, it will load silently with `substance = null`, and any integration test that queries it through the new business-rule layer could produce confusing results.

4. **[api/controllers/v1/sensor-configuration/update.js:100-106]** `effectiveQkCode` can be `undefined` if `TQuantityKind.findOne({ id: config.quantityKind })` returns null (data-corruption scenario). `isSubstanceRequired(undefined)` returns `false`, silently skipping substance validation entirely. Consider a guard:
   ```js
   if (!effectiveQkCode) {
     return res.serverError('Could not resolve existing quantity kind.');
   }
   ```

5. **[api/controllers/v1/sensor-configuration/create.js:80 and api/services/observation-import/EntityBuilder.js:84,206]** `substance || null` is used instead of `substance ?? null`. Because `||` coerces any falsy value (including `''`) to `null`, while `??` only coerces `undefined`/`null`. Since `validateSubstance` rejects empty strings before the data object is built, this is currently harmless, but it is inconsistent with the `??` used in `converters.js:689` and could be surprising if the validation logic ever changes. Consider using `?? null` throughout.

6. **[api/services/SensorConfigurationService.js:4,173]** `SUBSTANCE_REQUIRING_CODES` is exported but not imported anywhere else in the codebase. If it is not part of an intentional public API, keeping it unexported reduces the surface of the module and avoids the implicit promise of a stable export.

### Nitpicks (Optional)

7. **[PR description — "How" section]** The description says "adds Concentration quantity kind and **mg/L** unit", but the unit actually added in the enumeration SQL is `micromole (µM)` at id=18 (`mg/L` already existed at id=16). Minor inaccuracy that could mislead readers of the git log.

8. **[assets/swaggerV1.yaml — SensorConfigurationCreate schema]** The `substance` field in `SensorConfigurationCreate` is missing `nullable: true`. While the description explains when it must/must not be present, adding `nullable: true` would make the schema contract explicit for clients relying on the OpenAPI spec to generate code.

ClemRz:
Thanks for the thorough review @Paul-AUB! Here's what was addressed:

### Fixed

1. **Missing ALTER TABLE migration** — Created `sql/9_08_2026_06_16_substance_column_and_concentration_qk.sql` with the substance columns, QK consolidation (removes old specific `*Concentration` codes, reassigns IDs with no gaps), and view update. The `9_04` and `9_05` files are reverted to match develop — all changes go through the incremental migration.

2. **ID reassignment in enumerations** — Handled by the migration file. The base `9_05` is unchanged; the migration performs the UPDATE/DELETE/reassignment for deployed environments.

3. **Fixture #4 missing explicit substance** — Added `"substance": null` for clarity.

4. **Unguarded `effectiveQkCode`** — Added a guard returning `res.serverError()` if the quantity kind lookup fails.

5. **`||` vs `??` inconsistency** — Changed to `substance ?? null` in `create.js` and `EntityBuilder.js` to match `converters.js`.

8. **Swagger missing `nullable: true`** — Added to `SensorConfigurationCreate.substance`.

### Not addressed

6. **Unexport `SUBSTANCE_REQUIRING_CODES`** — Intentionally exported; it's imported by tests for verification against the service logic.

7. **PR description inaccuracy (mg/L vs µM)** — You're right, the unit added is `micromole (µM)`, not mg/L. Noted for the record but no code impact.

urien:
J'ai fait un commentaire sur la PR front, qui impacte le back
https://github.com/GrottoCenter/grottocenter-front/issues/1399

Paul-AUB (changes requested):
All items from the first round have been addressed — the migration file exists, the fixture is clear, the guard is in place, and `??` is used consistently. Two new issues surfaced on closer inspection of the migration.

### Issues (Must Fix)

1. **[sql/9_08_2026_06_16_substance_column_and_concentration_qk.sql, step 2d]** The FK update ordering will cause the migration to fail on any deployed environment. The sequence `UPDATE t_sensor_configuration SET id_quantity_kind = 18 WHERE id_quantity_kind = 22` executes before `UPDATE t_quantity_kind SET id = 18 WHERE id = 22`. Because the FK `t_sensor_configuration_t_quantity_kind_fk` is NOT DEFERRABLE (no CASCADE, no DEFERRABLE clause in the DDL), PostgreSQL checks the constraint immediately at statement level. At that moment, `id = 18` does not yet exist in `t_quantity_kind` (it was deleted in step 2c) → FK violation, transaction is aborted. The fix is to add `SET CONSTRAINTS t_sensor_configuration_t_quantity_kind_fk DEFERRED;` immediately before step 2d begins (which is valid here because we're inside a `BEGIN … COMMIT` block).

2. **[sql/9_08_2026_06_16_substance_column_and_concentration_qk.sql]** Existing rows in `t_time_series` are not updated. The table stores a denormalized `quantity_kind_code varchar(100)` string (confirmed in the DDL: no FK, purely a snapshot). After this migration runs, rows that were created for `NitrateConcentration`, `NitriteConcentration`, `AmmoniumConcentration`, `PhosphateConcentration`, or `SilicateConcentration` sensor configs still carry those old codes. The `v_measurement_wide` view joins via `LEFT JOIN t_quantity_kind qk ON qk.code = ts.quantity_kind_code`; those rows will now produce `NULL` for every `qk.*` column (including `si_to_display_factor`, `display_symbol`, `symbol_si`), silently breaking all BI calculations for historical data. The migration needs:
   ```sql
   UPDATE t_time_series SET quantity_kind_code = 'Concentration'
   WHERE quantity_kind_code IN (
     'NitrateConcentration', 'NitriteConcentration',
     'AmmoniumConcentration', 'PhosphateConcentration', 'SilicateConcentration'
   );
   ```

3. **[Design — substance field]** The reviewer `urien` opened [grottocenter-front#1399](https://github.com/GrottoCenter/grottocenter-front/issues/1399) requesting a controlled substance list backed by PubChem or ChEBI rather than a free-text field, to avoid naming inconsistencies across users. The backend currently accepts any string up to 100 chars. Before this API ships, consider whether the string contract is intentional (open-ended, curated by UX) or whether a future `t_substance` reference table would require a breaking API change. If the plan is eventually to constrain to a reference table, documenting that intent now (or adding a `GET /api/v1/substances` stub) would reduce migration friction later.

ClemRz:
Addressed both items from round 2:

**1. FK violation during ID reassignment** — Fixed by ALTERing `t_sensor_configuration_t_quantity_kind_fk` to `DEFERRABLE INITIALLY IMMEDIATE` and calling `SET CONSTRAINTS ... DEFERRED` before the reassignment block. The constraint is checked at `COMMIT` when all IDs are consistent.

**2. Stale `quantity_kind_code` in `t_time_series`** — Added `UPDATE t_time_series SET quantity_kind_code = 'Concentration' WHERE quantity_kind_code IN ('NitrateConcentration', 'NitriteConcentration', 'AmmoniumConcentration', 'PhosphateConcentration', 'SilicateConcentration')` so the `v_measurement_wide` view join continues to resolve correctly for historical data.

**3. Controlled substance list (suggestion)** — Acknowledged. The free-text field is intentional for now; a `t_substance` reference table can be introduced later without a breaking API change since the field is already optional/nullable. Tracked via grottocenter-front#1399.

All tests passing (2843 pass, 0 fail).

ClemRz:
Re: suggestion #3 (controlled substance list) — opened #1681 to track the follow-up work on a `t_substance` reference table backed by PubChem/ChEBI. The current free-text field ships as-is; no breaking change will be needed when the reference table is added later.

Paul-AUB (dismissed):
All items from the two previous review rounds have been addressed correctly. The migration now handles the FK deferral, the `t_time_series` backfill is in place, the `??` operator is used consistently, and the guard on `effectiveQkCode` is solid. This is a well-structured feature with excellent test coverage (property-based + integration). A few remaining observations below.

1. **[api/services/SensorConfigurationService.js:26]** `validateSubstance` rejects all-whitespace strings via `substance.trim().length === 0`, but the value is stored as-is in both controllers (`substance ?? null` in create, `updateData.substance = substance` in update). A substance like `"  NO₃⁻  "` would pass validation and be stored with leading/trailing spaces, producing malformed labels in `v_measurement_wide` (e.g., `Concentration [ NO₃⁻ ] (µM)`). Consider trimming before storing, or add a `substance.trim()` call in `validateSubstance` before the length check, and return the trimmed value (or trim in the controllers before passing to `validateSubstance`).

2. **[sql/9_08_2026_06_16_substance_column_and_concentration_qk.sql, step 2b2]** The migration sets `substance = NULL` for all existing `NitrateConcentration`, `NitriteConcentration`, etc. time series (they never had this column, so it's always NULL). As a result, historical concentration data loses its specific substance identity after migration — an operator looking at archived time series for nitrate can no longer tell it apart from phosphate or silicate ones by substance. This is an accepted trade-off but it is not explicitly called out in the PR description or migration comments. Worth a brief comment like: `-- NOTE: historical time series will have substance = NULL; users must re-specify substance on their sensor configs going forward.`

3. **[sql/9_08_2026_06_16_substance_column_and_concentration_qk.sql:53-57]** The `t_sensor_configuration_t_quantity_kind_fk` constraint is left as `DEFERRABLE INITIALLY IMMEDIATE` after the migration, permanently diverging from the `NOT DEFERRABLE` DDL definition in `9_04`. At runtime, `INITIALLY IMMEDIATE` is functionally equivalent (checked at statement time unless explicitly deferred), so this causes no bugs. However, the schema diff will surprise anyone running `\d t_sensor_configuration` or comparing against the DDL. Consider restoring the original constraint definition at the end of the migration block — one additional `ALTER TABLE … DROP CONSTRAINT / ADD CONSTRAINT` without the `DEFERRABLE` clause.

### Nitpicks (Optional)

4. **[test/integration/4_routes/SensorConfigurations/substance.test.js]** The update suite covers substance change, auto-clear via QK change, and rejecting Concentration without substance. One scenario not covered: `PATCH { substance: null }` on a config that currently has a non-substance-requiring QK (explicit clear when there is nothing to clear). This is a no-op path but adding it would close the branch coverage gap in the update controller's `else if (substance !== undefined)` block.

5. **[test/integration/1_services/observation-import/EntityBuilder.substance.integration.test.js:11-55]** The inline test view `v_measurement_wide_test` omits the `geom` column present in the production view. This is an intentional simplification (the comment explains the test DB limitation). Minor documentation improvement: note that `geom` is excluded to keep the test view minimal, so a future reviewer doesn't wonder if it was accidentally dropped.

ClemRz:
## 🚀 Deployment Procedure

This PR includes a SQL migration that must be run **before** deploying the new application code.

### Steps

1. **Put the API in maintenance mode** (or schedule during low-traffic window)
2. **Run the migration script** against the production database:
   ```sql
   \i sql/9_08_2026_06_16_substance_column_and_concentration_qk.sql
   ```
   This script is transactional (`BEGIN`/`COMMIT`) — it either fully applies or rolls back on error.
3. **Deploy the application code**
4. **Verify**:
   - `GET /api/v1/devices/:id/configurations` returns the `substance` field
   - `SELECT * FROM v_measurement_wide LIMIT 5` works and shows `substance` / compound labels
   - `SELECT COUNT(*) FROM t_quantity_kind` returns 22 (no gaps)

### What the migration does

| Step | Action |
|------|--------|
| 1 | Adds `substance` column (nullable varchar(100)) to `t_sensor_configuration` and `t_time_series` |
| 2a | Renames `NitrateConcentration` (id 17) → generic `Concentration` |
| 2b | Migrates sensor configs from retired QK IDs (18–21) to the generic (17) |
| 2b2 | Updates denormalized `quantity_kind_code` in `t_time_series` for retired codes |
| 2c | Deletes retired specific concentration codes (18–21) |
| 2d | Reassigns IDs 22–26 → 18–22 to eliminate gaps (with deferred FK) |
| 2f–2g | Restores FK constraint to NOT DEFERRABLE |
| 3 | Recreates `v_measurement_wide` view with substance-aware compound labels |

### ⚠️ Important notes

- **Order matters**: migration must run before code deployment. The new code references `Concentration` (not the old specific codes) and expects the `substance` column to exist.
- **Historical data**: existing time series that used specific concentration codes will have `substance = NULL`. Users must re-specify substance on their sensor configurations going forward.
- **Rollback**: if needed, reverse the view change and re-add the old QK codes. The `substance` columns can be left in place (nullable, ignored by old code).

### 🔍 Post-migration audit

Run the following queries to identify sensor configurations and time series that need manual substance assignment:

```sql
-- Sensor configurations with Concentration/IsotopeDelta QK but no substance set
SELECT
  sc.id AS config_id,
  sc.id_device AS device_id,
  qk.code AS quantity_kind,
  sc.substance,
  sc.date_inscription,
  c.nickname AS author
FROM t_sensor_configuration sc
JOIN t_quantity_kind qk ON qk.id = sc.id_quantity_kind
LEFT JOIN t_caver c ON c.id = sc.id_author
WHERE qk.code IN ('Concentration', 'IsotopeDelta')
  AND sc.substance IS NULL
  AND sc.is_deleted = false
ORDER BY sc.id_device, sc.id;

-- Time series with Concentration/IsotopeDelta but no substance (historical data)
SELECT
  ts.id AS time_series_id,
  ts.quantity_kind_code,
  ts.substance,
  ts.measurement_count,
  ts.start_date,
  ts.end_date,
  o.cave_name,
  o.point_label,
  sc.id AS config_id
FROM t_time_series ts
JOIN t_observation o ON o.id = ts.id_observation
LEFT JOIN t_sensor_configuration sc ON sc.id = ts.id_sensor_configuration
WHERE ts.quantity_kind_code IN ('Concentration', 'IsotopeDelta')
  AND ts.substance IS NULL
  AND ts.is_deleted = false
ORDER BY o.cave_name, ts.start_date;
```

These results identify records that lost their specific substance identity during the consolidation. A domain expert should update `substance` on the affected sensor configurations, then backfill time series accordingly.

ClemRz:
Addressed all items from round 3:

**1. Trim substance before storing** — Added `normalizeSubstance()` helper in `SensorConfigurationService` that trims whitespace (and returns null for empty/null input). Used consistently in `create.js`, `update.js`, and `EntityBuilder.js`. Property tests updated to expect trimmed values.

**2. Historical data comment** — Added a `-- NOTE:` comment in the migration clarifying that historical time series will have `substance = NULL` after migration.

**3. Restore FK to NOT DEFERRABLE** — Added step 2f (`SET CONSTRAINTS ... IMMEDIATE`) followed by 2g (drop + re-add without `DEFERRABLE`). Schema now matches the original DDL definition after migration completes.

**4. Test for PATCH { substance: null } on non-substance QK** — Added integration test covering the no-op path.

**5. Geom exclusion note** — Added documentation in the test file header explaining the intentional omission.

All tests passing (lint + full suite).

PR #1683, opened 2026-06-17
fix(dbSync): update archiver usage for v8 API

ClemRz:
## 🤔 What

- Fix `TypeError: archiver is not a function` in automatic weekly dbSync
- Add integration tests for the `makeDbSync` file export path

## 🤷‍♂️ Why

The `archiver` package was upgraded to v8 (`^8.0.0` in package.json), which introduced a breaking API change. v8 no longer exports a factory function — it exports classes (`ZipArchive`, `TarArchive`, etc.).

The automatic dbSync (runs every Monday at 2 AM UTC) calls `makeDbSync(true)` which hits the archiver code path. The manual resync script defaults to `makeDbSync(false)`, skipping file export entirely — that's why manual resyncs still worked.

Last successful automatic run: June 1, 2026.

## 🔍 How

- Replace `const archiver = require('archiver')` with `const { ZipArchive } = require('archiver')`
- Replace `archiver('zip')` with `new ZipArchive()`
- Add `DbSync.test.js` covering the file export path with stubbed dependencies (Azure upload, Typesense) so this class of breakage is caught by CI in the future

## 🧪 Testing

```bash
npm run test -- --grep "DbSync"
```

All 5 new tests pass. Full suite (2882 tests) also passes.

## 📸 Previews

N/A — backend-only fix.

### DIFF
--- a/api/dbSync/dbSync.js
+++ b/api/dbSync/dbSync.js
@@ -173,7 +173,7 @@ async function makeDbSync(isFileExportEnabled = true) {
   let archiveP;
   if (isFileExportEnabled) {
     if (!FileService.isCredentials) {
-      sails.log.warn('[dbSync] DB sync aborded, no azure credentials supplied');
+      sails.log.warn('[dbSync] DB sync aborted, no azure credentials supplied');
       return;
     }
     archive = new ZipArchive();
--- a/api/dbSync/dbSync.js
+++ b/api/dbSync/dbSync.js
@@ -1,6 +1,6 @@
 const fs = require('fs');
 const { pipeline, Duplex, Readable } = require('stream');
-const archiver = require('archiver');
+const { ZipArchive } = require('archiver');
 const CommonService = require('../services/CommonService');
 const FileService = require('../services/FileService');
 const syncUtils = require('./utils');
@@ -176,7 +176,7 @@ async function makeDbSync(isFileExportEnabled = true) {
       sails.log.warn('[dbSync] DB sync aborded, no azure credentials supplied');
       return;
     }
-    archive = archiver('zip');
+    archive = new ZipArchive();
     const { promise } = pipelineAsync(
       archive,
       // fs.createWriteStream(syncUtils.EXPORT_FILE_NAME) // For debug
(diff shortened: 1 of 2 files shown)


Paul-AUB (dismissed):
The fix is minimal and correct — destructuring `ZipArchive` from the v8 archiver export and using `new ZipArchive()` is exactly the right migration. The public API used afterward (`append`, `finalize`, `pointer`) is preserved in v8, so no other callsites need changing.

The integration tests are well-structured with proper sinon setup/teardown. A few points worth considering:

1. **[test/integration/1_services/DbSync.test.js:10]** The outer `describe` block is titled "makeDbSync with file export enabled", but it also houses the "skip file export" (line 82) and "abort when no Azure credentials" (line 89) tests. Consider splitting into two sibling `describe` blocks — one for `isFileExportEnabled = true` and one for `isFileExportEnabled = false` — to make the test tree self-documenting and to avoid confusing a reader who runs tests filtered by description.

2. **[test/integration/1_services/DbSync.test.js:92-98]** The "should create typesense collections" test verifies `createTimestampedCollection` and `switchCollectionAlias` are called 7 times, but does not assert on `importDocuments`. Since the stub at line 60 is anonymous (no reference stored), there is no way to assert it was invoked. Capturing it (`importDocumentsStub = sinon.stub(...)`) and asserting `importDocumentsStub.callCount >= 7` would close the loop on the full Typesense import path.

### Nitpicks (Optional)

3. **[test/integration/1_services/DbSync.test.js:19-20 / 48-50]** Saving and restoring `FileService.isCredentials` via a `before`/`afterEach` pair works fine here, but using `sinon.stub` (or a getter spy) for the boolean property would align this test with how the other properties (`upload`, `setMetadata`) are handled and make the restoration automatic via `sinon.restore()`.

4. **[api/dbSync/dbSync.js:176]** Minor typo in the existing log message: `"aborded"` → `"aborted"`. Unrelated to this PR, but an easy one-character fix if you're touching the line anyway.

ClemRz:
Addressed the review in 82c48f5f:

1. **Describe block split** — done. Tests are now organized into "with file export enabled" and "with file export disabled" sibling blocks.
2. **`importDocumentsStub` captured** — done. Added explicit assertion that it's not called when queries return 0 rows (correct behavior for the empty-data stub scenario).
3. **`sinon.stub` for `isCredentials`** — skipped. Current save/restore pattern is clear and works fine; switching to a getter spy would add complexity for no functional gain here.
4. **Typo `aborded` → `aborted`** — fixed.

PR #1687, opened 2026-06-17
fix(entrance): preserve h_location.id_entrance on permanent delete

### LINKED ISSUE #1685: bug(api): [23502] hlocation - null id_entrance on permanent delete
**Describe the bug**
When permanently deleting an entrance without merging into another entrance (`isPermanent=1` without `entityId`), the delete controller attempts to null out `id_entrance` in the `h_location` history table. However, `h_location.id_entrance` has a NOT NULL constraint, so the UPDATE fails with PostgreSQL error 23502.

The code at `api/controllers/v1/entrance/delete.js` sets `target = null` when there is no merge target, then executes:
```sql
UPDATE h_location SET id_entrance = $1 WHERE id_entrance = $2
```
with `$1 = null`, violating the constraint.

**To Reproduce**
Based on the trace logs:
1. Authenticate as a Moderator
2. `DELETE /api/v1/entrances/159489?isPermanent=1`
3. The entrance has associated locations (id 11049) with history rows in `h_location`

**Input Data**

- **Endpoint:** `DELETE /api/v1/entrances/159489?isPermanent=1`
- **Request Body:**
```json
{}
```
- **URL/Query Parameters:** `isPermanent=1` (path param `id=159489`)
- **User:** id 32183, nickname "Stochus", groups: Moderator, User

**Expected behavior**
The API should handle permanent deletion of entrances that have `h_location` history rows by either:
1. Deleting the `h_location` rows (since the entrance is being permanently deleted, not merged), or
2. Using a different strategy to preserve audit history without violating NOT NULL constraints (e.g., a sentinel value or making the column nullable)

**Error Details**
- Error Type: OperationalError (E_QUERY_FAILED)
- Occurrences: 1
- Time Range: 2026-06-17T16:36:07Z to 2026-06-17T16:36:07Z
- Representative Trace ID: 019ed670-9ba8-760e-a26e-fabbd8070034
- Stack Trace:
```
OperationalError: Query failed: null value in column "id_entrance" of relation "h_location" violates not-null constraint
    at Object.queryFailed (sails-hook-orm/lib/datastore-method-utils/help-send-native-query.js:100:14)
    ...
  raw.error: severity: 'ERROR', code: '23502'
    detail: "Failing row contains (11049, 32183, null, ...)"
    schema: 'public', table: 'h_location', column: 'id_entrance'
```

**Additional context**
- Service: `api/controllers/v1/entrance/delete.js`
- Model: `hlocation` (history table for TLocation)
- DB Error Code: `23502` (not_null_violation)
- Root cause: Line `CommonService.query('UPDATE h_location SET id_entrance = $1 WHERE id_entrance = $2', [target, entranceId])` where `target` is `null` because there is no merge destination.
- Fix suggestion: When `shouldMergeInto` is false (pure delete, not merge), delete the `h_location` rows instead of nulling the FK, or make `h_location.id_entrance` nullable.

ClemRz:
## 🤔 What

- Stop nulling `h_location.id_entrance` (and other h_ table FKs) during permanent entrance deletion without merge
- Retain original entrance references in history rows for auditability
- Closes #1685

## 🤷‍♂️ Why

When permanently deleting an entrance without merging into another, the code was setting `h_location.id_entrance = NULL`. This violated the `NOT NULL` constraint on `h_location.id_entrance`, causing a 500 error (PostgreSQL error 23502).

Nulling the FK also destroys the audit trail — after deletion you can't query which entrance a history record belonged to.

## 🔍 How

The FK constraints from h_ tables to `t_entrance` were already dropped in migration `1_2026_05_05_drop_history_parent_fk.sql`. This means h_ rows can safely point to a non-existent `t_entrance` row without blocking the hard delete.

The fix splits the h_ UPDATE block into two branches:
- **Merge path** (`shouldMergeInto = true`): reassign h_ references to the surviving entrance (existing behavior, unchanged)
- **Pure delete path** (`shouldMergeInto = false`): skip h_ UPDATEs entirely, preserving the original `id_entrance` value as an orphaned-but-valid historical reference

No schema migration needed. The `NOT NULL` constraint stays in place, which is correct — a location history row should always record which entrance it belonged to.

## 🧪 Testing

- `npm run test -- --grep "Delete - Auditability"` validates the fix directly
- Full test suite passes (2882 tests, 0 failures)

## 📸 Previews

N/A

### DIFF
--- a/api/controllers/v1/entrance/delete.js
+++ b/api/controllers/v1/entrance/delete.js
@@ -180,57 +180,72 @@ module.exports = async (req, res) => {
       ),
     ]);
 
-    // Clean up FK references to this entrance in history tables.
-    // H-rows are preserved for auditability — only the FK is nulled/reassigned
-    // so the hard delete of t_entrance can proceed.
+    // Clean up FK references to this entrance in T-models (secondary 'exit' FK)
+    // and, when merging, reassign H-model references to the surviving entrance.
+    //
+    // When NOT merging (pure delete), h_ rows intentionally keep their original
+    // id_entrance value. The FK constraints have been dropped (see migration
+    // 1_2026_05_05_drop_history_parent_fk.sql), so orphaned references are safe
+    // and preserve full traceability for audit purposes.
+    //
     // Raw SQL is required because Waterline silently fails on composite-PK models.
     //
     // IMPORTANT: This block must cover every h_ table that has an id_entrance
     // or id_exit column. If a new sub-entity type is added to subEntityDelete
     // above, a matching h_ UPDATE must be added here.
-    await Promise.all([
-      // T-models: secondary 'exit' FK
+    const tModelExitUpdates = [
       TDescription.update({ exit: entranceId }).set({ exit: target }),
       TRigging.update({ exit: entranceId }).set({ exit: target }),
       TComment.update({ exit: entranceId }).set({ exit: target }),
-      // H-models: all entrance/exit references
-      CommonService.query(
-        'UPDATE h_location SET id_entrance = $1 WHERE id_entrance = $2',
-        [target, entranceId]
-      ),
-      CommonService.query(
-        'UPDATE h_description SET id_entrance = $1 WHERE id_entrance = $2',
-        [target, entranceId]
-      ),
-      CommonService.query(
-        'UPDATE h_description SET id_exit = $1 WHERE id_exit = $2',
-        [target, entranceId]
-      ),
-      CommonService.query(
-        'UPDATE h_rigging SET id_entrance = $1 WHERE id_entrance = $2',
-        [target, entranceId]
-      ),
-      CommonService.query(
-        'UPDATE h_rigging SET id_exit = $1 WHERE id_exit = $2',
-        [target, entranceId]
-      ),
-      CommonService.query(
-        'UPDATE h_comment SET id_entrance = $1 WHERE id_entrance = $2',
-        [target, entranceId]
-      ),
-      CommonService.query(
-        'UPDATE h_comment SET id_exit = $1 WHERE id_exit = $2',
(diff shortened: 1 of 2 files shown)


Paul-AUB (dismissed):
The fix is correct and well-motivated. Splitting the h_ UPDATE block on `shouldMergeInto` neatly eliminates the NOT NULL violation without touching the constraint itself, and the audit trail rationale is sound given that the FK constraints were already dropped in the prior migration. Two minor items below.

1. **[api/controllers/v1/entrance/delete.js:29]** The `subEntityDelete` JSDoc still reads _"preserved for auditability with FK references nulled via raw SQL separately."_ After this PR the pure-delete path no longer nulls anything — it intentionally preserves the original FK value. Consider updating the comment to match, e.g. _"preserved for auditability; in merge paths the FK is reassigned via raw SQL, in delete paths the original value is kept as an orphaned reference."_

2. **[test/integration/4_routes/Entrances/delete-auditability.test.js:273-278]** The test verifies that `h_location.id_entrance` is preserved, which directly covers the reported bug. Consider adding parallel assertions for the other orphaned FK columns (`h_description.id_entrance`, `h_rigging.id_entrance`, `h_comment.id_entrance`, `h_history.id_entrance`, `h_name.id_entrance`) to give the same confidence for every h_ table skipped in the pure-delete branch. The before/after hooks already insert rows for all of them.

ClemRz:
Both review suggestions addressed in cce7a39f:

1. Updated the `subEntityDelete` JSDoc to accurately describe the merge vs delete behavior (no longer mentions "nulled via raw SQL").
2. Added parallel `id_entrance` preservation assertions for `h_description`, `h_rigging`, `h_comment`, `h_history`, and `h_name` alongside the existing `h_location` check.

PR #1688, opened 2026-06-17
fix(validation): pre-validate string lengths before DB transactions

### LINKED ISSUE #1686: bug(api): [IMPORT_TRANSACTION_ERROR] tobservation - name exceeds maxLength on import
**Describe the bug**
The observation import endpoint returns a 500 error when the `observationName` field in the import profile exceeds 200 characters. The Waterline validation catches the length violation, but the error is returned as a 500 (IMPORT_TRANSACTION_ERROR) instead of a 400 validation error. The user should receive a clear validation message before the import transaction begins.

In this case, the `observationName` was 220 characters long (20 characters over the 200 max limit), derived from the document title describing time series data.

**To Reproduce**
Based on the trace logs:
1. Authenticate as a User
2. `POST /api/v1/observations/import` with a multipart form including a profile where `observationName` exceeds 200 characters
3. The server starts the import transaction, attempts to create the observation record, and fails on the `name` column validation

**Input Data**

- **Endpoint:** `POST /api/v1/observations/import`
- **Request Body (profile excerpt):**
```json
{
  "deviceId": 18,
  "caveId": 20847,
  "pointLabel": "Inconnu",
  "latitude": 43.735011,
  "longitude": 3.933706,
  "observationName": "Time series of type chemistry in Le Lez (Méditerranée) basin - MEDYCYSS observatory - KARST observatory network - OZCAR Critical Zone network Research Infrastructure - Source Lirou // Chemical survey station - NO3 values",
  "documentTitle": "Time series of type chemistry in Le Lez (Méditerranée) basin - MEDYCYSS observatory - KARST observatory network - OZCAR Critical Zone network Research Infrastructure - Source Lirou // Chemical survey station - NO3 values",
  "dataQuality": "raw",
  "samplingIntervalSeconds": 1209600
}
```
- **User:** id 427, nickname "Frédéric Urien", groups: User

**Expected behavior**
The API should validate the `observationName` length (and other string fields) during profile validation — before beginning the database transaction — and return a 400 Bad Request with a clear message indicating which field exceeds the maximum length and by how much.

**Error Details**
- Error Type: IMPORT_TRANSACTION_ERROR (Waterline validation)
- Occurrences: 2
- Time Range: 2026-06-17T05:38:30Z to 2026-06-17T05:38:56Z
- Representative Trace ID: 019ed416-f32a-76b9-9190-eafa03ea0c5c (reference_id)
- Error message:
```
Invalid new record.
Details:
  Could not use specified `name`. Violated one or more validation rules:
  • Value was -20 characters longer than the configured maximum length (200)
```

**Additional context**
- Service: `api/services/observation-import/` (import pipeline)
- Model: `tobservation` (TObservation.name field, maxLength: 200)
- Error Code: Waterline validation (maxLength)
- The `observationName` field value is 220 characters — the user-provided document title is used verbatim as the observation name without truncation or pre-validation
- Fix suggestion: Add length validation for `observationName` (and `documentTitle` if applicable) in the profile validation step of the import pipeline, returning a 400 with actionable feedback before entering the transaction

ClemRz:
## 🤔 What

- Add a generic `stringLengthValidation` utility (`api/utils/stringLengthValidation.js`) for pre-validating string field lengths against model `maxLength` constraints
- Validate `observationName` (≤200), `pointLabel` (≤200), and `documentTitle` (≤300) in the observation import `ProfileValidator` — before any DB transaction starts
- Refactor `nameValidation.js` to delegate to the generic utility (same API, no breaking changes)
- Closes #1686

## 🤷‍♂️ Why

When a string field exceeds its model's `maxLength`, Waterline catches it inside the DB transaction and throws — surfacing as a 500 `IMPORT_TRANSACTION_ERROR`. Users get an opaque server error instead of a clear 400 explaining which field is too long and by how much.

This is an API-wide pattern problem: controllers either hardcode limits (drifting from model definitions) or skip length checks entirely. The new utility provides a single reusable function that any controller or service can use to validate string lengths before hitting the DB.

## 🔍 How

1. **`api/utils/stringLengthValidation.js`** — Generic utility with:
   - `validateStringLength(fieldName, value, maxLength)` → `{ field, message }` or `null`
   - `validateStringLengths(fieldMap)` → validates multiple fields, returns error array

2. **`ProfileValidator.js`** — Added rule #11 that calls `validateStringLengths` on the three profile fields that flow into constrained DB columns. Violations are now caught in Stage 1 (profile validation) and returned as HTTP 400 with `IMPORT_VALIDATION_ERROR`.

3. **`nameValidation.js`** — Refactored to delegate to the generic utility. Existing callers (cave, entrance, massif, organization controllers) are unaffected.

## 🧪 Testing

- 20 new tests added (all passing):
  - `test/integration/2_utils/stringLengthValidation.test.js` — unit tests for the generic utility
  - New `describe('string field length validation')` block in `ProfileValidator.test.js`
- Full test suite: 2902 passing, 0 failing
- Linter: clean

## 📸 Previews

N/A (backend-only change)

### DIFF
--- a/api/services/observation-import/ProfileValidator.js
+++ b/api/services/observation-import/ProfileValidator.js
@@ -2,6 +2,7 @@ const { COMPONENT_TYPES } = require('./TimestampConverter');
 const isBlank = require('../../utils/isBlank');
 const isNonBlankString = require('../../utils/isNonBlankString');
 const isValidId = require('../../utils/isValidId');
+const { validateStringLengths } = require('../../utils/stringLengthValidation');
 
 const VALID_NUMBER_LOCALES = ['en', 'fr'];
 const VALID_DATA_QUALITIES = ['raw', 'validated', 'suspect', 'rejected'];
@@ -279,6 +280,16 @@ const validate = (profile) => {
       );
     }
   }
+
+  // 11. String field length validation against model maxLength constraints.
+  //     Catches length violations early (before the DB transaction) so users
+  //     get a 400 instead of a 500.
+  const lengthErrors = validateStringLengths({
+    observationName: { value: profile.observationName, maxLength: 200 },
+    pointLabel: { value: profile.pointLabel, maxLength: 200 },
+    documentTitle: { value: profile.documentTitle, maxLength: 300 },
+  });
+  lengthErrors.forEach((err) => errors.push(err.message));
 
   return errors;
 };
--- a/api/services/observation-import/ProfileValidator.js
+++ b/api/services/observation-import/ProfileValidator.js
@@ -284,10 +284,23 @@ const validate = (profile) => {
   // 11. String field length validation against model maxLength constraints.
   //     Catches length violations early (before the DB transaction) so users
   //     get a 400 instead of a 500.
+  //     We trim observationName and documentTitle before measuring because
+  //     EntityBuilder trims them before saving — validating the raw value
+  //     would produce false positives for strings with trailing whitespace.
+  const trimIfString = (v) => (typeof v === 'string' ? v.trim() : v);
   const lengthErrors = validateStringLengths({
-    observationName: { value: profile.observationName, maxLength: 200 },
-    pointLabel: { value: profile.pointLabel, maxLength: 200 },
-    documentTitle: { value: profile.documentTitle, maxLength: 300 },
+    observationName: {
+      value: trimIfString(profile.observationName),
+      maxLength: 200, // TName.name maxLength
+    },
+    pointLabel: {
+      value: profile.pointLabel, // not trimmed by EntityBuilder
+      maxLength: 200, // TPoint.label maxLength
+    },
+    documentTitle: {
+      value: trimIfString(profile.documentTitle),
+      maxLength: 300, // TDescription.title maxLength
+    },
   });
   lengthErrors.forEach((err) => errors.push(err.message));
 
--- a/api/utils/nameValidation.js
+++ b/api/utils/nameValidation.js
@@ -1,6 +1,6 @@
 const { validateStringLength } = require('./stringLengthValidation');
 
-const NAME_MAX_LENGTH = 200;
+const NAME_MAX_LENGTH = 200; // TName.name maxLength
 
 /**
  * Validates that a name string does not exceed the DB column limit.
--- a/api/utils/nameValidation.js
+++ b/api/utils/nameValidation.js
@@ -1,3 +1,5 @@
+const { validateStringLength } = require('./stringLengthValidation');
+
 const NAME_MAX_LENGTH = 200;
 
 /**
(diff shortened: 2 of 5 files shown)


Paul-AUB (dismissed):
Clean, well-scoped fix that correctly addresses the 500→400 regression. The generic utility is well-designed and the test coverage is thorough. A few items worth discussing before merge.

1. **[api/services/observation-import/ProfileValidator.js:287-290 / api/services/observation-import/EntityBuilder.js:37-38, 515]** Trim discrepancy for `observationName` and `documentTitle`. The validation runs against the **raw** profile value, but `EntityBuilder` saves the **trimmed** version:
   - `observationName`: saved as `profile.observationName.trim()` (EntityBuilder line 515)
   - `documentTitle`: saved via `deriveDocumentTitle()` which trims the value (EntityBuilder lines 37-38)

   Consequence: a value like `"a".repeat(199) + "  "` (201 raw chars, 199 trimmed chars) would trigger a 400 error from this validation even though the trimmed value would have saved successfully. This is a false positive. `pointLabel` is consistent (no trimming in EntityBuilder, no trim in validation).

   Consider validating the trimmed value where a trim is applied downstream — or trimming the value early in profile parsing so validation and saving operate on the same string.

2. **[api/utils/stringLengthValidation.js:1-60 / api/services/observation-import/ProfileValidator.js:287-290]** The `maxLength` values (200, 200, 300) are hardcoded literals. If a Waterline model changes its constraint, the validator silently drifts. The issue description acknowledges this as an API-wide pattern problem, but since this utility is positioned as the canonical pre-validation layer, consider co-locating the constants with the models or exporting them so they can be imported here:
   ```js
   // e.g. from the model or a shared constants file
   const { NAME_MAX_LENGTH } = require('../../models/TName'); // 200
   ```
   At minimum, a comment linking each constant to its model column would make future maintenance easier.

3. **[api/utils/nameValidation.js:12-15]** The refactoring changes the user-visible error message for cave, entrance, massif, organization, and name update controllers. Old format: `"Name is too long (X characters). Maximum is 200."` → new format: `"Name exceeds maximum length of 200 characters (got X, Y over limit)."`. The PR description says "same API, no breaking changes" — the function signature is unchanged, but the message text is different. No existing tests pin the message body (they assert only on HTTP 400), so tests won't catch a frontend that matches on the old string. Worth a callout in the PR or changelog if any client relies on this wording.

### Nitpicks (Optional)

4. **[test/integration/1_services/observation-import/ProfileValidator.test.js:195-206]** The null/undefined test passes `observationName: null, pointLabel: undefined, documentTitle: null` — consider adding a case with whitespace-only strings (e.g., `"   "`) to document the current behavior and guard against any future pre-trim logic being added.

5. **[api/utils/stringLengthValidation.js:45-64]** The JSDoc block at the top of the file (lines 1-20) is a usage example rather than API documentation. It's helpful for discoverability, but it duplicates the `@param`/`@returns` JSDoc on each function. If the project convention is to keep both, no action needed — otherwise consider consolidating.

ClemRz:
Thanks for the thorough review @Paul-AUB! Addressed the feedback:

**Item 1 (Trim discrepancy)** — Fixed. `observationName` and `documentTitle` are now trimmed before length validation, matching EntityBuilder's downstream behavior. `pointLabel` is intentionally left untrimmed since EntityBuilder doesn't trim it.

**Item 2 (Hardcoded constants)** — Added model-linking comments (`// TName.name maxLength`, `// TPoint.label maxLength`, `// TDescription.title maxLength`) next to each constant for traceability.

**Item 3 (Error message wording)** — Acknowledged. No client in this ecosystem matches on the old message string, and the function signature is unchanged. No code change needed.

**Item 4 (Whitespace-only test)** — Added three tests: whitespace-only strings don't trigger errors, trim-before-validate prevents false positives on `observationName`/`documentTitle`, and `pointLabel` is correctly NOT trimmed.

**Item 5 (JSDoc)** — Skipped, stylistic only.

All 2933 tests passing, lint clean.

PR #1691, opened 2026-06-19
feat(substance): add t_substance reference table with PubChem integration

ClemRz:
## 🤔 What

Replace free-text `substance` varchar field on sensor configurations and time series with a normalized `t_substance` reference table linked via FK.

- Add `t_substance` table (name, formula, cas_number, external_id, external_source)
- Pre-seed 22 speleology-relevant substances with PubChem CIDs
- Add `id_substance` FK to `t_sensor_configuration` and `t_time_series`
- Rename `substance` column → `substance_label` (denormalized for BI)
- Add `GET /api/v1/substances` (public search) and `POST /api/v1/substances` (auth-required create)
- Implement `SubstanceService` (local ILIKE search + PubChem fallback for authenticated users)
- Implement `PubChemService` (autocomplete + property lookup, 5s timeout, graceful degradation)
- Validate `idSubstance` in sensor config create/update (QK coupling, auto-clear on QK change)
- Propagate `id_substance` + `substance_label` through import pipeline to time series
- Update `v_measurement_wide` view with `id_substance` and bracket label format
- Return populated substance object in sensor config and device GET responses
- Add diagnostic SQL script for manual migration of historical free-text values

## 🤷‍♂️ Why

The existing free-text `substance` field caused inconsistencies (e.g., "NO₃⁻" vs "Nitrate" vs "NO3-") and prevented reliable filtering/grouping in BI dashboards. A normalized reference table enables:

- Consistent substance identification across the platform
- PubChem integration for discovering new substances
- Proper FK relationships for joins and data integrity
- Substance-level filtering in downstream views

## 🔍 How

- 3 SQL migrations: table creation + seed → column rename + FK addition → view update with `id_substance`
- `SubstanceService.createOrFind` handles race conditions via `E_UNIQUE` catch-and-retry
- ILIKE search escapes wildcards (`%`, `_`, `\`) to prevent pattern injection
- `PubChemService` uses `AbortController` for 5s timeout with graceful `[]` fallback
- Sensor config controllers accept `idSubstance` (integer FK) instead of free-text string
- `EntityBuilder.processColumn` resolves substance from populated object, integer FK, or null
- `toSensorConfiguration` converter returns substance as a full object (id, name, formula, etc.)
- Controllers wrap service calls in try/catch for proper error responses

## 🧪 Testing

- Run `npm run dev:clean` to rebuild Docker containers (migration includes column renames)
- `npm run test` — all substance-related tests pass (98 tests across models, services, routes)
- Property-based tests cover: externalId/source consistency, search criteria, PubChem auth gate, creation idempotence, view label formatting, substance propagation in import pipeline, validateSubstance logic
- One pre-existing flaky test in `Cavers/ban-unban.property.test.js` is unrelated

## 📸 Previews

N/A (API-only changes)


Paul-AUB (changes requested):
This is a well-structured PR with solid test coverage and a clean two-layer design (local ILIKE + PubChem fallback). The ILIKE wildcard escaping, `AbortController` timeout, and `E_UNIQUE` race-condition catch are all done correctly. A few issues in the SQL migration and service layer deserve attention.

The diff visible in this review covered the core service/controller/model files. Test files were analyzed structurally from the file list.

### Issues (Must Fix)

1. **[sql/9_09_2026_06_18_substance_reference_table.sql:33]** Duplicate unique index on `t_substance.name`. The `CONSTRAINT t_substance_name_key UNIQUE (name)` on line 26 already creates an implicit B-tree unique index in PostgreSQL. The subsequent `CREATE UNIQUE INDEX IF NOT EXISTS t_substance_name_idx ON t_substance (name)` creates a second, identical unique index on the same column. This doubles write overhead on every INSERT/UPDATE to this table and wastes storage. Drop the explicit `CREATE UNIQUE INDEX` — the constraint alone is sufficient.

2. **[sql/9_09_2026_06_18_substance_reference_table.sql:40-137]** DDL operations run outside any transaction after the `COMMIT` on line 36. The sequence `DROP VIEW → RENAME COLUMN → CREATE VIEW → ADD COLUMN → ADD CONSTRAINT` has no BEGIN/COMMIT wrapper. If any step fails mid-migration (e.g., `CREATE VIEW` fails due to a typo, or `ADD CONSTRAINT` fails due to data mismatch), the database is left in a partially migrated state: the view is permanently dropped and columns are renamed, but the view is not recreated. PostgreSQL DDL is transactional — wrap lines 40–137 in `BEGIN; ... COMMIT;`.

3. **[sql/9_09_2026_06_18_substance_reference_table.sql:22]** `external_source varchar(50) DEFAULT 'PubChem'` is a footgun. Any INSERT that omits this column (direct SQL scripts, future code paths, or Waterline if `externalSource` is `undefined` rather than `null`) will silently mark a substance as having a PubChem source even without an `external_id`. This contradicts the seed rows like `Silicate` and `δ¹⁸O` that correctly have `NULL`. Change to `DEFAULT NULL`.

4. **[api/controllers/v1/substance/create.js:21] and [api/services/SubstanceService.js:91]** `externalSource` is computed in both the controller and the service. The controller sets `data.externalSource` but the service re-derives it from `data.externalId`, silently ignoring `data.externalSource`. This is dead code in the controller and will become a bug if a second external source (e.g., ChEBI) is ever added: the controller could pass a different source, but the service would always force 'PubChem'. Consider letting the service own this logic and removing the derivation from the controller, or passing `externalSource` as a first-class parameter and trusting it in the service.

5. **[api/controllers/v1/substance/create.js]** `formula` (model max 100), `casNumber` (model max 20), and `externalId` (model max 50) have no explicit length validation. If these are exceeded, Waterline throws an opaque internal error rather than a clean 400. The `name` field shows the intended pattern — apply the same treatment to the other fields.

6. **[sql/9_09_2026_06_18_substance_reference_table.sql:26]** The UNIQUE constraint on `name` is case-sensitive, but all application lookups use `LOWER(name) = LOWER($1)`. A concurrent direct DB insert (or a future code path that bypasses the service) could insert both "Nitrate" and "NITRATE" as distinct rows — the unique constraint would not catch this, but the application would see them as duplicates. Consider replacing the case-sensitive constraint with a functional unique index: `CREATE UNIQUE INDEX t_substance_name_lower_idx ON t_substance (LOWER(name))` and dropping `t_substance_name_key`. This aligns the DB constraint with the application's lookup semantics.

### Nitpicks (Optional)

7. **[api/controllers/v1/substance/create.js:43-55]** The create response takes two distinct paths: new substances use `res.status(201); return res.json(substance)`, while existing duplicates go through `ControllerService.treat()`. If `ControllerService.treat()` applies any response normalization or envelope wrapping, the two paths produce different response shapes for the same endpoint. Consider routing both through the same path (e.g., set `res.status(created ? 201 : 200)` then call `ControllerService.treat()` unconditionally).

8. **[api/services/SubstanceService.js:8-15]** `mapRow` uses raw SQL snake_case keys (`row.cas_number`, `row.external_id`) while `findById` returns a Waterline object with camelCase keys (`casNumber`, `externalId`). The two paths are correctly consumed in different places, but the naming asymmetry within the same service is a subtle maintenance hazard. A shared `mapWaterlineRow` helper or consistently using Waterline throughout would remove the ambiguity.

ClemRz:
Thanks for the thorough review @Paul-AUB! All issues and suggestions addressed in the latest push:

| # | Finding | Resolution |
|---|---------|------------|
| 1 | Duplicate unique index | Removed — the functional index (item 6) now serves as the sole uniqueness enforcement |
| 2 | DDL outside transaction | Sections 3–5 wrapped in `BEGIN; ... COMMIT;` |
| 3 | `DEFAULT 'PubChem'` footgun | Changed to `DEFAULT NULL` |
| 4 | Dead `externalSource` in controller | Removed — service owns this logic exclusively |
| 5 | Missing length validation | Added 400 checks for `formula` (100), `casNumber` (20), `externalId` (50) |
| 6 | Case-sensitive UNIQUE vs ILIKE lookups | Replaced with `CREATE UNIQUE INDEX ... (LOWER(name))` in both migration and test DB |
| 7 | Inconsistent response paths | Unified to `res.status(created ? 201 : 200); res.json(substance)` |
| 8 | `mapRow` naming asymmetry | Skipped — both patterns are correctly consumed in their respective contexts |

All tests passing (2947/2947), lint clean.

Paul-AUB (approved):
All 8 findings from the previous review have been correctly addressed. Verified against the latest diff:

| # | Finding | Status |
|---|---------|--------|
| 1 | Duplicate unique index | Resolved — `CONSTRAINT t_substance_name_key` removed; functional index `t_substance_name_lower_idx` is now the sole uniqueness enforcement |
| 2 | DDL outside transaction | Resolved — sections 3–5 (rename, FK additions, view recreate) are wrapped in `BEGIN; ... COMMIT;` |
| 3 | `DEFAULT 'PubChem'` footgun | Resolved — `external_source` now defaults to `NULL` |
| 4 | Dead `externalSource` in controller | Resolved — controller no longer derives `externalSource`; service owns this at line 91 of `SubstanceService.js` |
| 5 | Missing length validation | Resolved — `formula` (100), `casNumber` (20), `externalId` (50) all validated in `create.js` |
| 6 | Case-sensitive UNIQUE vs ILIKE | Resolved — `CREATE UNIQUE INDEX t_substance_name_lower_idx ON t_substance (LOWER(name))` aligns DB constraint with application lookup semantics |
| 7 | Inconsistent response paths | Resolved — unified to `res.status(created ? 201 : 200); return res.json(substance)` |
| 8 | `mapRow` naming asymmetry | Accepted as-is (both call sites correctly consume their respective shapes) |

Additional spot-checks from this review pass:

- `SensorConfigurationService.validateSubstance` correctly returns `{ error, substance }` and does a live `findById` lookup before accepting a substance FK.
- `EntityBuilder.processColumn` handles all three substance states (populated object, integer FK, null) correctly.
- `toSensorConfiguration` converter returns substance as a full object when populated.
- The `isAuthenticated = !!req.token` check in `find.js` works correctly because token parsing runs at the HTTP middleware level (not the policy), so authenticated callers to the public endpoint still receive `req.token` and get the PubChem fallback.
- Transaction structure in the migration is sound: table creation + functional index in one transaction, all schema changes + FK constraints in a second transaction.

This is a solid, well-tested feature. Approving.

PR #1693, opened 2026-06-19
refactor(unit): move SI conversion factors from quantity kind to unit

ClemRz:
## 🤔 What

Restructure how SI conversion factors and display units are managed in the scientific observation domain.

- Move `siToDisplayFactor` / `siToDisplayOffset` from `TQuantityKind` to `TUnit`
- Add `displayUnit` FK on `TQuantityKind` pointing to the canonical human-friendly unit (e.g., °C for Temperature, hPa for Pressure)
- Add `dimension` column on `TUnit` grouping physically interconvertible units
- Guard `value_si` and `value_display` in `v_measurement_wide`: NULL when import unit and display unit have incompatible dimensions
- Consolidate substance-specific quantity kinds (`CO2Concentration`, `DissolvedOxygen`, `TotalDissolvedSolids`, `RadonConcentration`) into generic `Concentration` with substance FK
- Compact `t_quantity_kind` IDs to eliminate gaps
- Fix `ON CONFLICT` in substance seed migration to match expression index

## 🤷‍♂️ Why

- **Conversion factors on the unit**: a unit knows how to convert itself to/from SI regardless of which physical quantity is being measured. The factor/offset is a property of the unit, not the quantity kind.
- **Canonical display unit**: `v_measurement_wide` needs a single preferred unit per quantity kind (always °C for Temperature, always hPa for Pressure) so Superset dashboards show consistent columns regardless of what individual importers chose.
- **Dimension safety**: importing CO₂ in ppm (dimensionless ratio) and displaying in µM (molar concentration) is physically meaningless without substance-specific molar mass. The dimension column lets the view detect incompatible conversions and return NULL instead of nonsensical numbers.

## 🔍 How

- **SQL migration** (`9_12`): adds `factor_to_si`, `offset_to_si`, `dimension` columns to `t_unit`; adds `id_display_unit` FK to `t_quantity_kind`; populates all values; consolidates QKs; reassigns IDs using deferred constraints (same pattern as `9_08`); recreates `v_measurement_wide` with dimension guard.
- **Model layer**: `TUnit` gains `siToDisplayFactor`, `siToDisplayOffset`, `dimension`; `TQuantityKind` gains `displayUnit` FK, drops old conversion columns.
- **SIConverter**: reads factor/offset from `sensorConfig.unit` instead of `sensorConfig.quantityKind`.
- **Services**: `DeviceService` and `SensorConfigurationService` resolve the nested `displayUnit` on quantity kinds.
- **View logic**: `value_si` and `value_display` are only computed when `iu.dimension = du.dimension`; both dimension columns are exposed for debugging/filtering.
- **Superset dataset**: updated to include `id_substance`, `substance_label`, `import_dimension`, `display_dimension`.

## 🧪 Testing

All tests passing (129 relevant + full suite). Property-based tests cover SI conversion formula correctness with renamed fields and dimension semantics.

```bash
npm run test
```

## 📸 Previews

N/A — backend-only change.


Paul-AUB (changes requested):
Thorough, well-documented refactor — moving conversion factors onto the unit is the right modeling call, the migration comments are excellent, and the dimension guard is a thoughtful safety mechanism. A few correctness concerns are worth addressing before merge, mostly around the symbol-based join in the recreated view and the `value_si` NULL semantics.

### Issues (Must Fix)

1. **[sql/9_12…unit_conversion_factors.sql:415]** The recreated view joins the import unit by symbol: `LEFT JOIN t_unit iu ON iu.symbol = ts.unit_symbol`. `t_unit` only has a UNIQUE constraint on `code` (see `9_04…ddl.sql:42`), **not** on `symbol`. If two units ever share a symbol, this join fans out and silently duplicates every matching measurement row in the view — directly inflating the `COUNT(value)` metrics that the Superset pivot chart relies on. This is not hypothetical: the updated fixture `test/fixtures/tunit.json` already contains two units with symbol `‰` (id 3 `PER_MIL` and id 29 `per_mil`). Production seed symbols are currently unique, but nothing enforces that. Consider adding `CONSTRAINT t_unit_symbol_key UNIQUE (symbol)` (and de-duplicating the fixture), so the `iu` join cannot multiply rows.

2. **[test/fixtures/tunit.json:18-41]** Related to #1 — the fixture defines `micromole` (id 18) plus two `‰` entries (id 3, id 29) with overlapping symbols. Beyond the join risk above, having both `PER_MIL` and `per_mil` map to the same symbol makes any symbol-based lookup ambiguous and the fixture inconsistent with the unique production seed. Please collapse these to a single row.

3. **[sql/9_12…unit_conversion_factors.sql:363-370]** `value_si` is now NULLed whenever `iu.dimension <> du.dimension`. `value_display` depending on the display unit's dimension is correct, but `value_si` is the raw stored SI value and is independent of which display unit the quantity kind happens to prefer — NULLing it discards valid data. Concretely, every consolidated `Concentration` measurement (the former CO₂/ppm, DissolvedOxygen/mg·L⁻¹, TDS/mg·L⁻¹, Radon/Bq·m⁻³ rows) now points at the `µM` display unit (`molar_concentration`), so `import_dimension <> display_dimension` for all of them and **both** `value_si` and `value_display` come back NULL. The numbers are still in `value`, but the SI column that dashboards may already consume goes dark for that whole class of data. If the intent is only to avoid mislabeling `value_si` against `qk.symbol_si` (mol/L), consider keeping `value_si` populated and guarding only `value_display`, or document this behavior change explicitly since it affects existing rows.

4. **[sql/9_12…unit_conversion_factors.sql:199-223]** `id_display_unit` is set from a hardcoded list of 22 quantity-kind codes and then immediately constrained `SET NOT NULL`. Any quantity-kind row present in the DB but missing from this list will make the `ALTER … SET NOT NULL` fail the whole migration. A defensive `UPDATE … SET id_display_unit = <fallback> WHERE id_display_unit IS NULL` before the constraint (or an assertion that all rows are covered) would make the migration more robust against drift.

5. **[sql/9_12…unit_conversion_factors.sql:83-86]** `milligram_per_liter` is given `factor_to_si = 1` (identity) and dimension `mass_concentration`, while true SI mass concentration is kg·m⁻³. The comment acknowledges this is an SI-adjacent convention, which is fine, but it means `value_si` for mass-concentration units is not actually in SI base units. Worth a one-line note in the view or model so a future reader doesn't assume `value_si` is always strict SI.

### Nitpicks (Optional)

6. **[api/services/mapping/converters.js:670]** `source.quantityKind.displayUnit instanceof Object` works, but `typeof x === 'object' && x !== null` is the more conventional guard elsewhere in the codebase and avoids edge cases with cross-realm objects. Minor consistency point.

7. **[sql/9_12…unit_conversion_factors.sql:15-17, api/models/TUnit.js]** The DB columns are `factor_to_si` / `offset_to_si` but the model exposes them as `siToDisplayFactor` / `siToDisplayOffset` — opposite directional naming. It's clearly documented, but the inversion is a long-term footgun for anyone reading raw SQL alongside the model. Consider aligning the names (in either direction) in a follow-up.

8. **[api/services/observation-import/SIConverter.js]** Nice cleanup dropping the stale `3.8/3.9/3.10` step-number comments along with the rename. No action needed.

ClemRz:
Thanks for the thorough review @Paul-AUB! Addressed in the latest push:

### Issues (Fixed)

**Item 1 — Symbol uniqueness**: Added `UNIQUE (symbol)` constraint on `t_unit` (both migration and `test/customSQL.js`). The join is now safe.

**Item 2 — Duplicate fixture**: Removed the duplicate `PER_MIL` (id 3). Kept only `per_mil` (id 29) matching the production seed.

### Suggestions (Addressed)

**Item 4 — Defensive fallback**: Added `UPDATE ... SET id_display_unit = (matching SI unit) WHERE id_display_unit IS NULL` before the `SET NOT NULL`.

**Item 5 — mg/L comment**: Expanded the comment to clearly state that `value_si` for mass-concentration units is in mg/L (not kg/m³).

### Also fixed

**Item 7 — Column name alignment**: Renamed the DB columns from `factor_to_si`/`offset_to_si` to `si_to_display_factor`/`si_to_display_offset`. Since these columns are brand new in this migration, there's no backward compat concern. SQL and JS names now both read "si to display" — no ambiguity.

### Not changing

**Item 3 — `value_si` NULL**: Kept the NULL guard. The reasoning: when someone imports 400 ppm for Concentration, `value_si = 0.0000004` (a dimensionless ratio), but `unit_si` says "mol/L". Showing that number next to the mol/L label is misleading — it's not actually a molar concentration. `value` (in the original import unit) remains available for those rows. The `import_dimension` / `display_dimension` columns make the mismatch visible.

**Item 6**: `instanceof Object` is the pattern used throughout this converter file. Switching one instance to `typeof` would be inconsistent.

**Item 8**: 👍

Paul-AUB (approved):
Re-reviewed after the `fix(unit): address code review feedback` push (e5be4d94). All previously flagged Must-Fix items are resolved, and the two declined suggestions come with sound reasoning. Verified across SQL, model, fixtures, and test setup — approving.

### Verified Fixes

- **Item 1 (symbol fanout)** ✅ `UNIQUE (symbol)` is now enforced both in the migration (`9_12…:26-27`) and in the test harness (`test/customSQL.js`), so the `LEFT JOIN t_unit iu ON iu.symbol = ts.unit_symbol` can no longer multiply measurement rows.
- **Item 2 (duplicate fixture)** ✅ `test/fixtures/tunit.json` now has a single `‰` entry (`per_mil`, id 29); the duplicate `PER_MIL` (id 3) is gone.
- **Item 4 (NOT NULL robustness)** ✅ The defensive `UPDATE … SET id_display_unit = (SELECT … WHERE symbol = symbol_si) WHERE id_display_unit IS NULL` (`9_12…:226-228`) now runs before `SET NOT NULL`, so unmapped QK drift won't abort the migration.
- **Item 5 (mg/L convention)** ✅ The expanded comment (`9_12…:84-86`) clearly states `value_si` for mass-concentration units is in mg/L, not strict SI kg/m³.
- **Item 7 (column naming)** ✅ DB columns renamed to `si_to_display_factor` / `si_to_display_offset`, matching the model's `siToDisplayFactor` / `siToDisplayOffset`. No stale `factor_to_si` / `offset_to_si` references remain anywhere in the repo.

### Reasonably Declined

- **Item 3 (`value_si` NULL guard)**: Kept, with the rationale that a number labeled against a mismatched SI unit is more misleading than an absent one, and `value` plus the `import_dimension` / `display_dimension` columns preserve/expose the underlying data. Acceptable as a deliberate semantic choice.
- **Item 6 (`instanceof Object`)**: Kept for consistency with the rest of `converters.js`. Fair.

### Note (non-blocking)

The `ALTER TABLE t_unit ADD CONSTRAINT t_unit_symbol_key UNIQUE (symbol)` will fail at migration time if the live `t_unit` data ever contains duplicate symbols. Current production seed symbols are unique, so this is fine — just worth keeping in mind as a deploy-time precondition.

Nice, well-documented refactor. 👍

PR #1696, opened 2026-06-21
feat(sso): add product-agnostic SSO token endpoint and Superset security manager

### LINKED ISSUE #1694: Seamless Superset authentication via JIT user provisioning

Currently, only administrators can manually create user accounts in the Superset instance (bi.grottocenter.org). There is no integration between Grottocenter authentication and Superset — users cannot access the BI dashboards without an admin creating their account first. This doesn't scale and creates friction for users who want to explore data.

Seamless, Just-In-Time (JIT) user provisioning in Superset, triggered on the user's first attempt to access bi.grottocenter.org from the Grottocenter UI. The user should not have to re-authenticate.

**Proposed flow:**

1. Authenticated user clicks a "BI" button in grottocenter.org
2. Front-end calls the GC API: `POST /api/v1/superset/auth-token` (requires valid GC auth token)
3. API generates a **single-use, short-lived token** (e.g., 30s TTL) tied to the user
4. Front-end redirects to `bi.grottocenter.org` passing this token (delivery mechanism TBD — cookie on `.grottocenter.org`, POST form, or query param)
5. Superset's custom security manager calls back to the GC API to **validate and consume** the single-use token: `POST /api/v1/superset/validate-token`
6. GC API validates the token (checks TTL, single-use, user status including ban check) and returns user info (email, first name, last name)
7. If the user doesn't exist in Superset, it is created with the `GC_Creator` role
8. If the user already exists, their session is established directly
9. Superset session duration: **1 hour** (short session to limit exposure if a user is banned in GC between sessions)

**Key design decisions:**

- **Identity mapping:** email is used as the Superset username (more stable than nickname, which users can change)
- **Single-use token:** mitigates replay attacks and token leakage; the token is consumed on validation
- **Short Superset session (1h):** limits the window where a banned GC user could still access Superset
- **Ban/revocation awareness:** the validate-token endpoint checks the user's ban status and token blacklist before returning user info — a banned user's single-use token will be rejected
- **No account created until first access:** with ~5.5k GC accounts and a small active fraction, JIT provisioning avoids polluting Superset with thousands of unused accounts
- **Unauthenticated users cannot access Superset:** the BI button only appears for logged-in users, and the auth-token endpoint requires a valid GC session

**API-side scope (this repo):**

- [ ] `POST /api/v1/superset/auth-token` — issue a single-use, short-lived token for the authenticated user
- [ ] `POST /api/v1/superset/validate-token` — validate and consume the token, return user info (email, first name, last name) or reject if expired/used/banned
- [ ] Storage for single-use tokens (in-memory with TTL, or DB row with `used` flag)
- [ ] Swagger update for new endpoints

**Out of scope (tracked separately):**

- Front-end "BI" button and redirect logic (grottocenter-front)
- Superset custom security manager in `superset_config.py` (can be a follow-up or handled in the Docker config)
- Token delivery mechanism decision (cookie vs POST form vs query param) — to be finalized during design

| Alternative | Why not chosen |
|---|---|
| Full OAuth2 provider in GC API | Overkill — no third-party clients, no scopes needed. Adds complexity (authorization code grant, token introspection, etc.) for a single consumer. |
| Shared JWT validation (Superset validates GC token directly with shared secret) | Doesn't respect GC's ban/revocation mechanism — Superset can't check the blacklist. Also couples Superset to GC's token format. |
| Header-based auth via reverse proxy | Requires a proxy layer that validates GC JWTs. More infrastructure. Doesn't handle JIT provisioning natively. |

**Additional context**

- Superset instance: bi.grottocenter.org (self-hosted on Azure, deployed via GitHub Actions)
- Current GC auth: JWT-based with `TOKEN_SALT`, includes `iat`-based revocation via `t_token_blacklist`
- Superset uses Flask-AppBuilder security — custom `SecurityManager` class can override login behavior
- All provisioned users get the existing `GC_Creator` role in Superset

ClemRz:
## 🤔 What

Implements product-agnostic SSO token issuance and a Superset custom security manager for seamless JIT authentication.

- New endpoint: `POST /api/v1/sso/auth-token` — issues a 30s JWT signed with a per-product salt
- Product registry pattern: add future products by setting an env var (no code changes)
- Superset custom security manager at `/login/sso` — verifies JWT, provisions user, establishes session
- JTI replay protection (in-memory, 30s eviction)
- Synthetic email (`{id}@grottocenter.org`) to avoid PII exposure
- GC_Creator role whitelist permissions script (runs on every boot)
- Closes #1694

## 🤷‍♂️ Why

Users currently need an admin to manually create their Superset account. This enables seamless access: click a button in Grottocenter → authenticated in Superset in a new tab, no separate credentials needed.

## 🔍 How

**GC API side:**
- `SsoService.js` — Product Registry maps `"superset"` → `SSO_SALT_SUPERSET` env var. Builds JWT payload with `sub`, `aud`, synthetic `email`, `firstName`, `lastName`, `jti`, `iat`, `exp`.
- Controller fetches the full caver record (for name/surname), delegates to SsoService.
- Route gated by `tokenAuth` policy (ban/revocation handled by existing `parseAuthToken` middleware).
- Dev fallback salt (gated by `NODE_ENV=development`) matches Superset's docker-compose default.

**Superset side:**
- `gc_security_manager.py` — registers `/login/sso` as a raw Flask route (bypasses FAB auth + CSRF).
- Verifies signature (HS256), audience (`"superset"`), iat freshness (30s), jti uniqueness.
- JIT provisions user with `GC_Creator` role or updates name if changed (never modifies roles on existing users).
- `gc_creator_permissions.py` — whitelist approach, runs on every boot to ensure the role has exactly the right permissions.
- `jti_store.py` — thread-safe in-memory store with automatic 30s eviction.

**Front-end integration** (not in this PR — tracked in grottocenter-front#1413):
- Hidden HTML form with `target="_blank"` submits the JWT to `/login/sso` in a new tab.

## 🧪 Testing

**Node.js (27 tests):**
```bash
npm run test -- --grep "SsoService"
```
- Unit tests: salt resolution, payload construction, token issuance
- Property-based tests (fast-check): payload correctness, signing round-trip, jti uniqueness, invalid product rejection, missing salt
- Integration tests (supertest): happy path, 400/401/500 error cases

**Python (21 tests):**
```bash
python3 -m pytest docker/superset/tests/ -v
```
- JTI store: add/contains/eviction
- JWT verification: signature, audience, expiry, iat freshness
- Property-based tests (hypothesis): verification gate, replay protection, error page safety

**Manual E2E:**
1. `npm run dev` (with or without `SSO_SALT_SUPERSET` — falls back in dev mode)
2. Rebuild Superset: local docker-compose passes `SUPERSET_SSO_SECRET` via env
3. Open `test-sso.html` in browser, click the button — Superset opens authenticated in a new tab

## 📸 Previews

N/A — backend change. The front-end "BI" button will be added in a separate PR (grottocenter-front#1413).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,6 +68,10 @@ config/local.js
 node_modules
 bower_components
 
+# Python
+__pycache__/
+*.pyc
+
 
 
 
--- a/api/controllers/v1/sso/auth-token.js
+++ b/api/controllers/v1/sso/auth-token.js
@@ -1,0 +1,29 @@
+const SsoService = require('../../../services/SsoService');
+
+module.exports = async (req, res) => {
+  const { product } = req.body || {};
+
+  // req.token only has { id, groups, nickname } — fetch full caver for name/surname
+  const caver = await TCaver.findOne({ id: req.token.id });
+  if (!caver) {
+    return res.forbidden('Access denied.');
+  }
+
+  const result = SsoService.issueToken(caver, product);
+
+  if (result.error) {
+    if (result.status === 500) {
+      return res.serverError(result.error);
+    }
+    return res.badRequest(result.error);
+  }
+
+  const params = { controllerMethod: 'SsoController.authToken' };
+  return ControllerService.treat(
+    req,
+    null,
+    { token: result.token },
+    params,
+    res
+  );
+};
--- a/api/services/SsoService.js
+++ b/api/services/SsoService.js
@@ -1,0 +1,113 @@
+/**
+ * SsoService.js
+ *
+ * @description :: Product-agnostic SSO token issuance for cross-origin
+ *                 Just-In-Time authentication with external products.
+ *                 Each registered product has its own signing secret
+ *                 resolved from a dedicated environment variable.
+ */
+
+const crypto = require('crypto');
+const jwt = require('jsonwebtoken');
+
+/**
+ * Product registry mapping product identifiers to their
+ * corresponding environment variable names for signing secrets.
+ * To onboard a new product: add an entry here and set the env var.
+ */
+const PRODUCT_REGISTRY = {
+  superset: 'SSO_SALT_SUPERSET',
+};
+
+const SSO_TOKEN_TTL = 30; // seconds
+
+// Local dev fallback (matches SUPERSET_SSO_SECRET default in superset_config.py).
+// Only used when NODE_ENV=development and the env var is not set.
+const DEV_FALLBACK_SALT = 'aR4nd0mSs0DevSalt_change_in_prod';
+
+/**
+ * Resolve the signing salt for a given product.
+ * @param {string} product - Product identifier from request body
+ * @returns {{ salt: string } | { error: string, status: number }}
+ */
+function resolveSalt(product) {
+  if (!product || typeof product !== 'string') {
+    return {
+      error: "The 'product' field is required and must be a string.",
+      status: 400,
+    };
+  }
+
+  const envVarName = PRODUCT_REGISTRY[product];
+  if (!envVarName) {
+    return {
+      error: `Product '${product}' is not supported.`,
+      status: 400,
+    };
+  }
+
+  const salt = process.env[envVarName];
+  if (!salt || !salt.trim()) {
+    // In development, fall back to a shared dev secret (must match Superset side)
+    if (process.env.NODE_ENV === 'development') {
+      return { salt: DEV_FALLBACK_SALT };
+    }
+    sails.log.error(
+      `SsoService: signing secret not configured for product '${product}' (env var: ${envVarName})`
+    );
+    return {
+      error: 'Server configuration error.',
+      status: 500,
+    };
+  }
+
+  return { salt };
+}
+
+/**
+ * Build the SSO token payload for a caver and product.
+ * Uses synthetic email ({id}@grottocenter.org) to avoid PII exposure.
+ *
+ * @param {Object} caver - Caver record (must have id; may have name, surname)
+ * @param {string} product - Product identifier (used as `aud` claim)
+ * @returns {Object} JWT payload
+ */
+function buildPayload(caver, product) {
+  return {
+    sub: caver.id,
+    aud: product,
+    email: `${caver.id}@grottocenter.org`,
+    firstName: caver.name || '',
+    lastName: caver.surname || '',
+    jti: crypto.randomUUID(),
+  };
+}
+
+/**
+ * Issue an SSO token for the given caver and product.
+ * @param {Object} caver - Caver record (id, name, surname)
+ * @param {string} product - Product identifier
+ * @returns {{ token: string } | { error: string, status: number }}
+ */
+function issueToken(caver, product) {
+  const saltResult = resolveSalt(product);
+  if (saltResult.error) {
+    return saltResult;
+  }
+
+  const payload = buildPayload(caver, product);
+  const token = jwt.sign(payload, saltResult.salt, {
+    algorithm: 'HS256',
+    expiresIn: SSO_TOKEN_TTL,
+  });
+
+  return { token };
+}
+
+module.exports = {
+  PRODUCT_REGISTRY,
+  SSO_TOKEN_TTL,
+  resolveSalt,
+  buildPayload,
+  issueToken,
+};
--- a/assets/swaggerV1.yaml
+++ b/assets/swaggerV1.yaml
@@ -64,6 +64,8 @@ tags:
     description: Scientific observation data import
   - name: Substances
     description: Substance reference table for chemical species and isotopes
+  - name: SSO
+    description: Single Sign-On token issuance for external product integration
 
 paths:
   '/health':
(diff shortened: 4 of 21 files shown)


Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[docker/superset/jti_store.py + docker/superset/gc_security_manager.py:119-122]** TOCTOU race condition in JTI replay protection. `contains()` and `add()` each acquire the lock independently. Two concurrent requests carrying the same JTI can both pass `contains()` before either call reaches `add()`, letting both requests authenticate. The fix is an atomic `add_if_new(jti) -> bool` method on `JtiStore` that checks and adds in a single lock acquisition:
   ```python
   def add_if_new(self, jti: str) -> bool:
       with self._lock:
           self._evict_stale()
           if jti in self._store:
               return False
           self._store[jti] = time.time()
           return True
   ```
   Then in `_handle_sso_login` replace the two-step check with:
   ```python
   if not _jti_store.add_if_new(jti):
       return _render_error("This token has already been used.")
   ```

2. **[docker/superset/gc_security_manager.py:38]** The module-level `_jti_store` is an in-memory dict, scoped to a single Python process. Superset in production typically runs under Gunicorn with multiple worker processes (each forks its own address space). A token presented to worker A is marked as used in A's store, but worker B's store knows nothing about it, allowing a replay if the second request lands on a different worker. For genuine replay protection in production this needs a shared backend (e.g., Redis via `SET NX EX`). At minimum, a comment should document this limitation so operators know to run single-worker or add Redis.

3. **[docker/superset/gc_security_manager.py:98]** `verify_aud: False` disables PyJWT's built-in audience validation and relies entirely on the manual `payload.get("aud") != EXPECTED_AUDIENCE` check on line 106. This works, but it silently skips the library's standard gate. Consider passing `audience="superset"` to `jwt.decode()` and removing the `verify_aud: False` option — PyJWT will then raise `InvalidAudienceError` automatically, and the manual check becomes a redundant safety net rather than the only gate.

4. **[docker/superset/gc_creator_permissions.py:50]** `("all_datasource_access", "all_datasource_access")` grants JIT-provisioned users access to **all** datasources in the Superset instance. If Superset ever hosts datasources that should not be exposed to all GC users (e.g., internal admin views or restricted datasets), every provisioned user will have access. Consider whether row-level security or per-datasource permissions would be more appropriate.

5. **[docker/superset/gc_creator_permissions.py:17-19]** `("can_write", "Dashboard")` and `("can_write", "Chart")` allow all JIT-provisioned users to create and modify dashboards and charts. If the goal is for regular cavers to view curated dashboards rather than author them, this is more permissive than intended. The existing `can_read` entries would be sufficient for a viewer role.

6. **[test/integration/1_services/SsoService.test.js:35-40]** The "should return error 500 when env var is not set" test relies on `NODE_ENV !== 'development'`. If the integration test suite runs with `NODE_ENV=development` (common for local test runs), `resolveSalt` will return the dev fallback salt instead of a 500, causing this test to fail unexpectedly. Consider explicitly setting `NODE_ENV=test` (or a non-development value) in `before()` for this describe block, or restructuring the test to stub `process.env.NODE_ENV`.

### Nitpicks (Optional)

7. **[docker/superset/gc_security_manager.py:109-112]** There is a double freshness check: PyJWT's `verify_exp: True` already rejects any token whose `exp` is in the past, and since `exp = iat + 30`, this covers the same 30-second window. The manual `iat` check on lines 110–112 is effectively redundant. It could be kept as defense-in-depth (it guards against tokens with a manipulated `exp`), but it's worth a comment explaining why both exist.

8. **[docker/superset/superset_config.py:21]** `SUPERSET_SSO_SECRET` falls back to an empty string `''` rather than a dev default, while the docker-compose hardcodes `aR4nd0mSs0DevSalt_change_in_prod`. This means if `SUPERSET_SSO_SECRET` env var is not set in the container, `_handle_sso_login` will log an error and return a generic error page — which is safe but could be confusing during local development without docker-compose. Adding a Python-side dev fallback (mirroring the Node.js approach) or documenting the requirement to always set this env var would improve the developer experience.

ClemRz (commented):
Thanks for the thorough review @Paul-AUB! Addressed in f67a75dc:

**Fixed:**

1. **TOCTOU race** — Added atomic `add_if_new(jti) -> bool` on `JtiStore`. The security manager now uses a single call instead of separate `contains()` + `add()`.
2. **Multi-worker limitation** — Added comment documenting that each Gunicorn worker has its own JTI store. Acceptable for current traffic + 30s TTL. Redis upgrade path noted.
3. **`verify_aud: False`** — Now uses PyJWT's built-in `audience="superset"` parameter. `InvalidAudienceError` is caught explicitly.
6. **NODE_ENV test dependency** — All 500-path unit tests now explicitly set `process.env.NODE_ENV = 'test'` and restore after.
7. **Double freshness check** — Added defense-in-depth comment explaining why both `exp` (PyJWT) and `iat` (manual) checks coexist.
8. **Empty string fallback** — Added comment explaining the intentional fail-safe behavior in `superset_config.py`.

**Intentional (no change):**

4. **`all_datasource_access`** — Required for users to create charts against any dataset. If we ever host restricted datasets, we'll revisit with per-datasource permissions.
5. **`can_write` on Dashboard/Chart** — Explicitly intended. GC_Creator users are researchers who author their own dashboards, not just viewers.

Paul-AUB (dismissed):
The previous issues have all been addressed cleanly. The atomic `add_if_new`, PyJWT audience validation, NODE_ENV guard in the `resolveSalt()` tests, defense-in-depth comment, and multi-worker documentation are all in place. A few smaller items worth noting:

1. **[test/integration/1_services/SsoService.test.js:147-155]** The `issueToken()` describe block has a "should return error when salt is not configured" test that does not set `NODE_ENV = 'test'` before asserting `result.status === 500`. The fix from this iteration was applied to the `resolveSalt()` tests above it (lines 36-62) but not here. If this test file runs outside Sails's `test` environment (e.g., standalone with `NODE_ENV=development`), `issueToken` will hit the dev-fallback branch and return a signed token instead of the expected error — the assertion fails silently. Adding the same `process.env.NODE_ENV = 'test'` guard used in the `resolveSalt()` tests would make the three 500-path tests consistent.

2. **[docker/superset/gc_security_manager.py:185-187]** The user-name update commits via `db.session.commit()` directly rather than through `self.update_user(user)` (inherited from Flask-AppBuilder's `SecurityManager`). Both work because `user` is a tracked entity in the current session, but `update_user()` is the idiomatic path through the security manager's abstraction and handles any session edge cases the parent class might add in future upgrades.

### Nitpicks (Optional)

3. **[docker/superset/gc_security_manager.py:92-94]** `_render_error("An unexpected error occurred.")` returns HTTP 400 for the "SUPERSET_SSO_SECRET not configured" case. A missing server secret is a 5xx condition — returning 400 classifies it as a client fault in error monitors. Returning 500 for this specific branch (while keeping 400 for token-level rejections) would improve observability.

4. **[docker/superset/gc_security_manager.py:167, 179]** `first_name or "Caver"` fires whenever `firstName` is an empty string, which happens when `caver.name` is `null` or `""` in GC (since `buildPayload` coerces it to `''`). This means nameless cavers appear as "Caver" in Superset — probably intentional, but a one-line comment would save the next reader from second-guessing it.

ClemRz:
Addressed in 2b73fe1c:

1. **`issueToken()` missing-salt test** — Now guards `NODE_ENV = 'test'`, consistent with `resolveSalt()` tests.
2. **User-name update** — Uses `self.update_user(user)` (idiomatic FAB path) instead of raw `db.session.commit()`.
3. **Missing `SUPERSET_SSO_SECRET`** — Returns HTTP 500 instead of 400 (server config error, not client fault).
4. **"Caver" fallback** — Added comments in both create and update paths explaining the empty-name coercion.

PR #1703, opened 2026-06-26
chore(search): remove entrances array from caves Typesense index

### LINKED ISSUE #1682: chore(search): remove entrances array from caves Typesense index after frontend migration
## Summary

Once the frontend has migrated to use `nbEntrances` (see GrottoCenter/grottocenter-front#1402), the `entrances` array can be removed from the caves Typesense index entirely.

## What to change

1. **`api/dbSync/entities/cave.js`** — Remove the `entrances` join from `processRows` (currently kept with a comment: "Keep until frontend migrates from entrances.length to nbEntrances").
2. **`api/services/CaveService.js`** — In `updateInSearch`, stop mapping `entrances` into the search document.
3. **Typesense schema** — Confirm the `entrances` field is no longer referenced, then trigger a full re-sync after deploy.

## Why

The `entrances` array on caves was only indexed so the frontend could derive a count via `.length`. With `nbEntrances` as a first-class scalar, the array is redundant — it wastes index RAM and inflates payloads.

## Blocked by

- GrottoCenter/grottocenter-front#1402 must be deployed first.

## Related

- API PR: #1676

ClemRz:
## 🤔 What

- Remove the `entrances` array join from the caves Typesense sync pipeline (`api/dbSync/entities/cave.js`)
- Closes #1682

## 🤷‍♂️ Why

The `entrances` array was only indexed so the frontend could derive a count via `.length`. With `nbEntrances` as a first-class scalar (added in #1676), the array is redundant — it wastes index RAM and inflates payloads.

The frontend has migrated to use `nbEntrances` (GrottoCenter/grottocenter-front#1402).

## 🔍 How

Removed the `entrances` join from `processRows` in `api/dbSync/entities/cave.js`. The Typesense schema and `CaveService.updateInSearch` already only reference the `nbEntrances` scalar — no changes needed there.

After deploy, a full re-sync should be triggered to drop the stale `entrances` field from existing documents in Typesense.

## 🧪 Testing

All existing tests pass (261 passing, 0 failing). No new tests needed — this is a removal of dead code.

## 📸 Previews

N/A

### DIFF
--- a/api/dbSync/entities/cave.js
+++ b/api/dbSync/entities/cave.js
@@ -29,19 +29,6 @@ const query = `
 
 async function* processRows(source) {
   for await (const rows of source) {
-    // Keep until frontend migrates from entrances.length to nbEntrances
-    const joins = [
-      {
-        table: 't_entrance',
-        foreignField: 'id_cave',
-        rows,
-        localField: 'entrances',
-        fields: ['id'],
-        transform: (e) => e.id,
-      },
-    ];
-
-    await Promise.all(joins.map((e) => exportUtils.joinMany(e)));
     for (const row of rows) yield row;
   }
 }


Paul-AUB (approved):
Clean, targeted removal. The change is correct and complete for the batch sync path.

PR #1708, opened 2026-06-27
feat(import): add async CSV import queue with job status endpoint

### LINKED ISSUE #1574: feat(entrance): make CSV import asynchronous using pg-boss job queue
## Problem

The CSV import endpoint (`POST /api/v1/entrances/import-rows`) processes all rows synchronously within a single HTTP request. For each row it performs:

- Column validation
- Author resolution (DB lookup or creation)
- Duplicate detection
- Cave creation (with Typesense indexing, notifications)
- Entrance creation (with sensitivity check, Typesense indexing, notifications)
- Optional alternate name creation
- Coordinates snapshot invalidation

For large imports (hundreds or thousands of rows), this keeps the HTTP connection open for a long time, risking:
- Client-side timeouts
- Proxy/load balancer timeouts (Azure App Service has a 230s limit)
- Poor UX — the user has no progress feedback

## Proposal

Leverage the `pg-boss` job queue infrastructure introduced in the `feat/async-geocoding-queue` branch (`EnrichmentQueueService`) to make CSV imports asynchronous:

1. **Accept & validate** — The `import-rows` endpoint validates the payload structure, checks admin permissions, deduplicates rows by `id` + `attributionName`, splits them into chunks, and returns immediately with a batch/job ID.
2. **Enqueue** — Each chunk is enqueued as a separate job on a `csv-import` queue, grouped under a common batch ID. Chunks can be processed in parallel for faster throughput.
3. **Process in background** — Workers process chunks concurrently. Each job handles its subset of rows (validation, author resolution, cave/entrance creation, Typesense indexing). Progress is tracked per chunk.
4. **Generic job status endpoint** — A new `GET /api/v1/jobs/:jobId` endpoint allows the client to poll for progress on any async job (not just CSV imports). It returns job type, status (`pending`, `active`, `completed`, `failed`), progress metadata, and the result payload on completion.
5. **Report generation** — Once all chunks complete, the backend generates result CSV files (success, duplicates, failures), stores them in Azure Blob Storage, and attaches download URLs to the job result.
6. **Completion notification** — A notification is created via the existing `NotificationService` with links to the generated report files. The user sees it in the UI, and receives an email if they've opted in.

This generic approach means any future async operation (bulk updates, exports, etc.) gets status tracking and notifications for free without adding feature-specific endpoints.

## Benefits

- No HTTP timeout risk regardless of import size
- Progress visibility for the user
- Reuses existing `pg-boss` infrastructure — no new dependencies
- Failed jobs get automatic retries via pg-boss retry policies

## Scope

- [ ] Create a `csv-import` queue with chunk-based parallel processing
- [ ] Refactor `import-rows` controller to deduplicate, split into chunks, enqueue, and return immediately
- [ ] Implement chunk worker logic (row processing, progress tracking)
- [ ] Add batch aggregation step (wait for all chunks, generate report CSVs, store in Azure Blob Storage)
- [ ] Add a generic job status endpoint (`GET /api/v1/jobs/:jobId`) with access control (users see only their own jobs)
- [ ] Send a completion notification via `NotificationService` with report file download links (in-app + opt-in email)
- [ ] Update the front-end to poll for progress and download reports from the provided URLs
- [ ] Update Swagger spec

## Related

- `api/controllers/v1/entrance/import-rows.js` — current controller
- `api/services/EnrichmentQueueService.js` — pg-boss queue pattern to follow

ClemRz:
## 🤔 What

Converts the synchronous `POST /api/v1/entrances/import-rows` endpoint into an async queue-based pipeline and adds a new `GET /api/v1/jobs/:batchId` endpoint for progress polling.

- Refactor import-rows to return 202 with a batch ID immediately
- Add `CSVImportQueueService` with pg-boss backed chunk processing
- Add `TJobBatch` model and `t_job_batch` SQL migration
- Add job status endpoint with ownership/admin access control
- Generate CSV reports (success/duplicates/failures) uploaded to Azure Blob
- Send email notification on import completion (translated in 15 locales)
- Add affinity-based chunking to keep duplicate-key rows in the same chunk

Closes #1574

## 🤷‍♂️ Why

The synchronous endpoint times out on large imports (thousands of rows). Moving to async processing:
- Unblocks the HTTP request immediately (202 response)
- Processes rows in parallel across 4 workers
- Provides real-time progress feedback via polling
- Generates downloadable CSV reports for successes/duplicates/failures
- Notifies the user by email when done

## 🔍 How

**Architecture:**
1. Controller validates the payload (mandatory columns on first row) and enqueues the batch
2. `CSVImportQueueService.createBatch()` splits rows into chunks using affinity-based grouping (rows sharing the same dedup key always land in the same chunk)
3. Each chunk is a pg-boss job processed by one of 4 concurrent workers (`teamSize: 1, teamConcurrency: 4`)
4. After each chunk completes, a separate completion-check job fires to avoid a race condition (job state must be committed before querying)
5. When all chunks are done, results are aggregated, CSV reports uploaded to Azure, and the user is notified

**Key design decisions:**
- `teamSize: 1` so pg-boss stores the handler return value as job output (with teamSize > 1, output is discarded by pg-boss v12)
- Completion check runs as a separate queue to avoid reading stale job state
- `ENTRANCE_MANDATORY_COLUMNS` extracted to `csvHelper.js` as single source of truth
- `sails.services.coordinatessnapshotservice` used instead of `require()` to get the Sails-loaded instance

## 🧪 Testing

```bash
# Run unit + integration tests
npm run test -- --grep "CSVImportQueueService"
npm run test -- --grep "Job features"
npm run test -- --grep "Import rows"

# End-to-end smoke test (requires running dev server + Docker)
./scripts/test-async-import.sh
```

Tested locally with 2000 rows (1700 fresh + 300 duplicates) — completes in ~170s with correct classification (1700 successes, 300 duplicates, 0 failures).

## 📸 Previews

**Polling response (in progress):**
```json
{
  "batchId": "6e5f37de-...",
  "status": "active",
  "progress": { "totalChunks": 41, "completedChunks": 20, "processedRows": 980 }
}
```

**Completed response:**
```json
{
  "status": "completed",
  "progress": { "totalRows": 2000, "processedRows": 2000, "successes": 1700, "duplicates": 300, "failures": 0 },
  "result": { "reportUrls": { "success": "https://...", "duplicates": "https://..." }, "summary": { "successes": 1700, "duplicates": 300, "failures": 0 } }
}
```


ClemRz:
## Front-end Integration Guide

The CSV import endpoint has changed from synchronous (200 with results) to asynchronous (202 with batch ID + polling). Here's what needs to change in the front-end:

### API Changes

**Before (synchronous):**
```
POST /api/v1/entrances/import-rows → 200 { successfulImport: [], failureImport: [], total: {...} }
```

**After (asynchronous):**
```
POST /api/v1/entrances/import-rows → 202 { batchId, totalRows, totalChunks, statusUrl }
GET /api/v1/jobs/:batchId           → 200 { batchId, type, status, progress, result }
```

### Implementation Steps

1. **Submit the import** — `POST /api/v1/entrances/import-rows` with `{ data: [...rows] }`. Expect a `202` response:
   ```json
   {
     "batchId": "uuid",
     "totalRows": 2000,
     "totalChunks": 41,
     "statusUrl": "/api/v1/jobs/uuid"
   }
   ```

2. **Poll for progress** — `GET /api/v1/jobs/:batchId` until `status` is `"completed"` or `"failed"`:
   ```json
   {
     "batchId": "uuid",
     "type": "csv-import",
     "status": "active",       // "pending" | "active" | "completed" | "failed"
     "progress": {
       "totalChunks": 41,
       "completedChunks": 20,
       "totalRows": 2000,
       "processedRows": 980,
       "successes": 900,
       "duplicates": 50,
       "failures": 30
     },
     "result": null             // populated only when status = "completed"
   }
   ```

3. **Show results** — When `status === "completed"`, `result` contains:
   ```json
   {
     "reportUrls": {
       "success": "https://...signed-url...",    // null if no successes
       "duplicates": "https://...signed-url...", // null if no duplicates
       "failures": "https://...signed-url..."    // null if no failures
     },
     "summary": {
       "successes": 1700,
       "duplicates": 300,
       "failures": 0
     }
   }
   ```

4. **Handle errors:**
   - `400` — empty data array or missing mandatory columns (immediate, no batch created)
   - `403` — user is not admin
   - `500` — job queue unavailable or batch creation failed
   - `status: "failed"` — all chunks failed during processing

### UX Recommendations

- Show a progress bar using `processedRows / totalRows` (or `completedChunks / totalChunks` for chunk granularity)
- Poll every 2–3 seconds
- When complete, show the summary and offer download links for the report CSVs (if URLs are non-null)
- Report URLs are SAS-signed and expire after 7 days
- The user also gets an email notification with the same summary and links

### Access Control

- Only the batch initiator can view their own job status
- Moderators and Admins can view any batch
- Non-authorized users get `403`

### New Notification Type: `IMPORT_COMPLETE`

This PR adds a new notification type `IMPORT_COMPLETE` (id: 8) to `t_notification_type`. The front-end notification system needs to handle this new type:

- **When:** Sent to the user who initiated the CSV import batch, once all chunks have been processed
- **Display:** Should render as an in-app notification (e.g., "Your CSV import has completed") with a link to view the results/download reports
- **Distinction from existing types:** Unlike `CREATE`/`UPDATE`/`DELETE` which relate to entity CRUD operations, `IMPORT_COMPLETE` is a job-completion event — it won't have the usual entity link (cave, entrance, etc.) but instead relates to a batch job

ClemRz:
## Dev Environment: Local File Storage

This PR also includes a local file storage mechanism for development (commit `feat(storage): add local file storage for dev environment`).

### What it does

When `AZURE_KEY` is not set (local dev), file uploads are stored on disk in `.local-uploads/` instead of Azure Blob Storage — mirroring how the email helper logs to console instead of sending via SES.

### How it works

- **Documents & reports** are written to `.local-uploads/documents/`
- **DB exports** are written to `.local-uploads/db-exports/`
- Files are served back via a `/local-uploads/*` middleware on the local server
- URLs in API responses point to `http://localhost:<port>/local-uploads/...`
- `.local-uploads/` is gitignored
- `nodemon.json` whitelists only `api/`, `config/`, and `app.js` to prevent restarts when files are written

### No changes needed to test locally

Just run `npm run dev` without `AZURE_KEY` — uploads work automatically. No more "Missing credential" warnings for document operations.

Paul-AUB (changes requested):
This is a well-structured PR with a clear design rationale. The separation between chunk processing, completion checking, and aggregation is clean, and the affinity-based chunking to preserve duplicate detection is a solid approach. A few issues need addressing before merge.

> **Scope note:** 2760 additions; analysis focused on `CSVImportQueueService.js`, the controller, model, migration, and config. The test files and Swagger doc were reviewed at a higher level.

---

### Issues (Must Fix)

1. **[api/services/CSVImportQueueService.js:377–403]** `checkBatchCompletion` is not idempotent, creating a duplicate-notification window. When the last two chunks complete at nearly the same time, both enqueue a completion-check job before either runs. The completion queue processes them sequentially, but *both* see `allDone = true` (all chunks are already committed to `pgboss.job` as `completed`). Both then call `aggregateBatch`, which uploads reports, creates a `TNotification`, and sends an email — resulting in the user receiving two emails and two in-app notifications. Fix: add an atomic test-and-set guard before calling `aggregateBatch`:
   ```js
   const guard = await TJobBatch.updateOne({ id: batchId, status: 'active' }).set({ status: 'aggregating' });
   if (!guard) return; // another worker already took this
   await module.exports.aggregateBatch(batchId, jobs);
   ```
   This requires adding `'aggregating'` to the `isIn` validator and the SQL `CHECK` constraint (or using a DB enum).

2. **[CODE_REVIEW.md]** This self-review file should not be merged into the repository. It lists some items as open issues that the final implementation actually addresses (e.g., the race condition is mitigated via the completion queue), making the document misleading. The relevant discussion belongs in PR comments or the linked issue, not in version-controlled source. Remove this file before merging.

---

3. **[api/services/CSVImportQueueService.js:411–452]** `aggregateBatch` has no error handling around the Azure upload step. If `generateAndUploadReports` throws (e.g., transient Azure failure), the exception propagates up through `checkBatchCompletion` → `processCompletionCheck`, the completion-check job fails, and `TJobBatch.status` stays stuck at `'active'` permanently — no retry, no error state, no notification. Consider wrapping the upload in try/catch and setting `status: 'failed'` on error so the user is informed:
   ```js
   try {
     const reportUrls = await module.exports.generateAndUploadReports(...);
     // ... rest of aggregation
   } catch (err) {
     sails.log.error('CSVImportQueueService: aggregation failed:', err);
     await TJobBatch.updateOne({ id: batchId }).set({ status: 'failed', completedAt: new Date() });
   }
   ```

4. **[api/models/TJobBatch.js:1–77]** Add `autoUpdatedAt: false` to the model. The model explicitly customizes `createdAt` (good practice for a table that uses snake_case column names), but doesn't opt out of Waterline's automatic `updatedAt` management. The SQL migration has no `updated_at` column. While `schema: true` in the global config may prevent Waterline from inserting the column, this relies on implicit behaviour that could break silently if the global config changes. Being explicit is safer and consistent with the pattern already used for `createdAt`.

5. **[api/services/CSVImportQueueService.js:597–599]** The email subject `'CSV Import Complete'` is hardcoded English. The `locale` parameter is correctly passed to `sendEmail`, so the EJS template body is fully translated, but the subject line will be English for all 15 locales. Add a locale-aware i18n key for the subject (e.g., `__('csv_import_complete_subject')`).

---

### Nitpicks (Optional)

6. **[api/services/CSVImportQueueService.js:346–353]** The outer `catch` in `processOneChunk` swallows unexpected chunk-level errors and returns an empty `result` (`{ successes: [], duplicates: [], failures: [] }`). From the user's perspective, rows in that chunk silently disappear — the completed batch will show fewer total rows than `totalRows` with no failure entries to explain the discrepancy. Consider pushing all remaining unprocessed rows as failures when the outer catch fires, or re-throwing so pg-boss can retry the job (given `retryLimit: 3` is already configured).

7. **[scripts/test-async-import.sh:38]** `TOKEN_SALT || 'aR4nd0mT0kenSalt'` provides a hardcoded fallback secret. Acceptable for a local dev script, but the fallback risks accidentally generating valid tokens against a non-dev environment if `TOKEN_SALT` is unset. Consider removing the fallback and requiring the variable explicitly.

ClemRz:
Thanks for the thorough review @Paul-AUB! Pushed fixes in 14fb03f0. Here's what was addressed:

**Fixed:**

1. **Race condition in `checkBatchCompletion`** — Added an atomic test-and-set guard (`updateOne({id, status:'active'}).set({status:'aggregating'})`). Only the first completion-check worker transitions the batch; subsequent ones get `null` and bail out. Added `'aggregating'` to the model validator and Swagger enum.

2. **`CODE_REVIEW.md` removed** — Agreed, doesn't belong in version control.

3. **Error handling in `aggregateBatch`** — Wrapped `generateAndUploadReports` in try/catch. On failure, batch transitions to `'failed'` with a `completedAt` timestamp so the user knows something went wrong.

6. **Silent row loss in `processOneChunk`** — The outer catch now pushes all remaining unprocessed rows as failures with the error message, so no rows silently vanish from batch results.

7. **TOKEN_SALT fallback removed** — Script now exits with an error if the env var is unset.

**Not changed (with rationale):**

4. **`autoUpdatedAt: false`** — This setting is not supported at the model level in Sails 1.0 (causes a boot error). The global config already opts out of auto-timestamps by not defining `updatedAt` in the global attributes block, so there's no risk of Waterline injecting the column.

5. **Email subject i18n** — The `sendEmail` helper already passes `emailSubject` through `i18n.__()` (see `api/helpers/send-email.js:95`), and translations for `'CSV Import Complete'` exist in all 15 locale files. No change needed.

Paul-AUB (changes requested):
Re-review after `d1c2a3e4`. The previously flagged must-fix items are well addressed:

- **Race condition** in `checkBatchCompletion` is now guarded by an atomic `updateOne({ id, status: 'active' }).set({ status: 'aggregating' })` test-and-set, with `'aggregating'` added to both the model `isIn` validator and the Swagger enum. 👍
- **`CODE_REVIEW.md`** removed.
- **Error handling** in `aggregateBatch` now wraps `generateAndUploadReports` in try/catch and transitions the batch to `'failed'`.
- **Silent row loss** in `processOneChunk`'s outer catch now pushes remaining rows as failures.
- **`TOKEN_SALT` fallback** removed from the smoke-test script.
- Email subject i18n confirmed fine — `send-email.js:95` runs `emailSubject` through `__()` and `'CSV Import Complete'` exists in all 15 locales.

One remaining issue and a few suggestions below.

### Issues (Must Fix)

1. **[sql/2_2025_06_26_job_batch_table.sql]** The migration filename deviates from the active convention in a way that risks the table never being created on existing databases. All recent migrations use the `9_XX_2026_MM_DD_*.sql` pattern (latest: `9_15_2026_06_23_*`), whereas this file is `2_2025_06_26_*`. Concerns:
   - **Ordering risk:** the postgres docker init runs files lexicographically on a *fresh* volume, so the dev/test flow happens to work (deps in `0_`/`1_` precede `2_`). But any production migration runner that tracks a high-water-mark by filename would see `2_2025...` sort *before* the already-applied `9_15_2026...` and skip it entirely — leaving `t_job_batch` uncreated and the feature broken in prod.
   - **Category mismatch:** per `sql/README.md`, prefix `2_` is for *seeding fixed values*, while `1_` is for table DDL. This file does `CREATE TABLE` + `ALTER TABLE ... ADD COLUMN` (DDL) in addition to the notification-type seed.
   - **Wrong year:** the date is `2025_06_26` but it should be `2026_06_26`.

   Consider renaming to `9_16_2026_06_26_job_batch_table.sql` to match the established recent pattern and remove any ordering ambiguity.

2. **[api/services/CSVImportQueueService.js:432-445]** `aggregateBatch` only aggregates `state === 'completed'` job outputs. Since `processOneChunk` catches internally and (almost) never throws, chunk jobs normally complete — but if a chunk job ends up `failed` via worker crash or `expireInHours: 4` expiry, that chunk's rows vanish from the reports and the summary, so `processedRows`/summary totals will be silently lower than `totalRows` with no failure entries explaining the gap. Consider detecting failed chunks during aggregation (e.g., recording the count of missing rows or appending synthetic failure entries) so the report reconciles against `totalRows`.

3. **[config/locales/*.json]** The 10 new keys are appended at the end of each locale file, but the files are maintained in sorted order (and `scripts/sort-translation.js` exists for exactly this). Consider running the sort script so `"CSV Import Complete"` etc. land in their alphabetical positions and the diff stays minimal on future edits.

4. **[api/services/CSVImportQueueService.js:564-594]** Reports are uploaded via `getReportsContainerClient()`, which reuses the *documents* container — and `FileService.document.getUrl` notes that container "allows anonymous access". If so, the 7-day SAS token added in `uploadReport` is effectively moot (the blob is readable without it), and access control reduces to the unguessability of the batch UUID in the path. That's probably acceptable given report contents (line numbers + messages), but consider a private container or documenting the rationale so it isn't mistaken for SAS-gated access.

### Nitpicks (Optional)

5. **[api/controllers/v1/job/find.js:6-22]** The `notFound` check runs before the authorization check, so an authenticated non-owner can distinguish "batch does not exist" (404) from "exists but not yours" (403). Low risk given batch IDs are UUIDs, but returning 404 for the unauthorized case too would avoid leaking existence.

6. **[config/http.js:378-380]** The path-traversal guard uses `filePath.startsWith(uploadsRoot)` without a trailing separator, so a sibling dir like `.local-uploads-x` would pass the prefix test. `path.resolve` already neutralizes `../` sequences (they resolve outside the root and get rejected), so this is purely theoretical and dev-only (`AZURE_KEY` unset) — comparing against `uploadsRoot + path.sep` would close it.

ClemRz:
Pushed fixes in 64ad635c. All 6 items from the re-review addressed:

1. **Migration renamed** to `9_16_2026_06_26_job_batch_table.sql` (correct prefix, year, sequence)
2. **Failed chunks detected** in `aggregateBatch` — synthetic failure entry appended so reports reconcile against `totalRows`
3. **Locale files sorted** alphabetically via `scripts/sort-translation.js`
4. **Container access documented** — JSDoc on `uploadReport` explains the SAS + anonymous access rationale
5. **404 for unauthorized** — job find returns 404 (not 403) to prevent leaking batch existence; test + Swagger updated
6. **Path-traversal guard fixed** — `uploadsRoot` now includes trailing `path.sep`

Lint clean, 3155 tests passing.

Paul-AUB (dismissed):
Re-review after `ce74198b`. All previously flagged must-fix items are verifiably addressed:

- **Race condition** in `checkBatchCompletion` — atomic `updateOne({ id, status: 'active' }).set({ status: 'aggregating' })` guard present; `'aggregating'` added to the model `isIn` and the Swagger enum (`swaggerV1.yaml:7413`). 👍
- **`CODE_REVIEW.md`** removed from the tree.
- **`aggregateBatch` error handling** — `generateAndUploadReports` wrapped in try/catch, batch transitions to `'failed'` with `completedAt`.
- **Silent row loss** — `processOneChunk`'s outer catch now pushes remaining rows as failures (`CSVImportQueueService.js:353-364`).
- **Failed-chunk reconciliation** — synthetic failure entry appended in `aggregateBatch` when `failedChunkCount > 0`.
- **Migration** renamed to `9_16_2026_06_26_job_batch_table.sql` (correct prefix/year).
- **Locale files** sorted (new keys land alphabetically, e.g. `en.json:10`).
- **Container access** rationale documented in the `uploadReport` JSDoc.
- **404 for unauthorized** in `job/find.js` (avoids existence leak).
- **Path-traversal guard** now uses `uploadsRoot + path.sep` (`config/http.js:377`).

Well-structured feature. No blocking issues remain — a few robustness suggestions and nitpicks below.

1. **[api/services/CSVImportQueueService.js:401-404]** `checkBatchCompletion` derives `allDone` from `jobs.every(...)` over the pg-boss rows currently visible for the batch, but never verifies that all `totalChunks` chunk jobs are actually present. The correctness of the whole completion flow therefore rests entirely on the `status: 'active'` guard being set *after* the `createBatch` enqueue loop (line 166). That ordering creates two theoretical windows:
   - **Premature aggregation:** while `createBatch` is still enqueuing, only a subset of chunk jobs exist. If every enqueued-so-far job happens to be `completed`, `allDone` is `true`. Today this is caught only because `status` is still `'pending'` and the `active`→`aggregating` guard returns `null`.
   - **Stuck batch:** conversely, if every chunk finishes and fires its completion check *before* `createBatch` flips the status to `'active'`, all those checks bail on the guard, and nothing re-triggers aggregation — the batch is stuck at `'active'` forever with no further trigger.

   In practice both are prevented by the ~2s `newJobCheckInterval` poll (workers can't pick up a chunk before the status flip), so this is low-risk. But it's fragile to rely on scheduler timing. Consider making the completion condition explicit — e.g. `allDone && jobs.length === batch.totalChunks` — so the invariant is enforced in code rather than by enqueue/poll timing, and the guard becomes defense-in-depth rather than the sole safety net.

2. **[api/services/CSVImportQueueService.js:713-726]** `getBatchProgress` increments `completedChunks` only for jobs in state `'completed'`; chunks that end up `'failed'` (worker crash / `expireInHours` expiry) are never counted. A front-end progress bar driven by `completedChunks / totalChunks` will then stall below 100% even once the batch has finished as `completed`/`failed`. Consider counting `failed` chunks toward `completedChunks` (or exposing a separate `failedChunks` count) so the progress reconciles against `totalChunks`.

### Nitpicks (Optional)

3. **[api/services/CSVImportQueueService.js:346]** `sails.services.coordinatessnapshotservice.invalidate()` is called without `await` or a `.catch()`. If `invalidate()` is async and rejects, it surfaces as an unhandled rejection rather than being logged with the chunk context. Consider awaiting it (inside the existing try) or attaching a `.catch(sails.log.error)`.

4. **[api/services/CSVImportQueueService.js:390-399]** `checkBatchCompletion` (and `getBatchProgress`) query `pgboss.job` filtered on `data->>'batchId'`, a JSONB extraction with no supporting index. `checkBatchCompletion` runs once per chunk completion, so a large batch triggers `totalChunks` such scans. Fine at current volumes, but worth keeping in mind if import sizes or batch concurrency grow.

ClemRz:
Pushed fixes in 97840a2e addressing the remaining suggestions from the latest review:

1. **Explicit chunk count verification** — `checkBatchCompletion` now fetches the batch and verifies `jobs.length >= batch.totalChunks` before evaluating completion, making the invariant explicit in code rather than relying on enqueue/poll timing.
2. **Failed chunks counted in progress** — `getBatchProgress` now increments `completedChunks` for both `completed` and `failed` states, so the front-end progress bar reconciles against `totalChunks`.
3. **Unhandled rejection fixed** — `coordinatessnapshotservice.invalidate()` now has a `.catch(sails.log.error)` to prevent unhandled promise rejections.

Tests added for the new edge cases (batch not found, fewer chunks than expected, failed chunks in progress). 3233 tests passing, lint clean.

Paul-AUB (dismissed):
Re-review after `97840a2e`. All previously flagged items are verifiably addressed:

- **Explicit chunk count guard** — `checkBatchCompletion` now checks `jobs.length < batch.totalChunks` before evaluating completion (`CSVImportQueueService.js:412`). The invariant is explicit in code rather than relying on enqueue/poll timing. 👍
- **Failed chunks in progress** — `getBatchProgress` counts both `'completed'` and `'failed'` chunk states toward `completedChunks` (`CSVImportQueueService.js:727`), so the front-end progress bar always reconciles against `totalChunks`. 👍
- **Unhandled rejection** — `coordinatessnapshotservice.invalidate()` now chains `.catch(sails.log.error)` (`CSVImportQueueService.js:346–353`). 👍

The core architecture is solid: affinity-based chunking, atomic test-and-set guard for aggregation, explicit completion-count verification, and full error paths that prevent silent row loss. No blocking issues remain.

---

1. **[api/services/CSVImportQueueService.js:419–426]** When `allFailed === true` the batch transitions to `'failed'` and returns — no email and no in-app notification reach the user. The success path calls `notifyCompletion` which sends both, creating an asymmetry: a user whose import partially succeeds gets notified, but one whose entire import fails does not (they discover it only via polling). Consider calling a slim failure-notification helper in the `allFailed` branch and in the `aggregateBatch` error-catch at line 495. The email template already renders a `failures` count, so a failure email would be minimal effort to add.

2. **[api/services/CSVImportQueueService.js:419–426 vs 431–435]** The `allFailed` branch updates `status: 'failed'` without the atomic guard applied to the `aggregateBatch` path. Under horizontal scaling (multiple API instances each running a completion worker), two workers could both enter this branch and write `status: 'failed'` twice. The result is idempotent — both writes are identical and no side-effect is triggered — so this is safe today. Worth noting for any future extension (e.g. failure email) that would need the guard added here too.

---

### Nitpicks (Optional)

3. **[api/services/CSVImportQueueService.js:655–660]** `TNotification.create` sets both `notifier` and `notified` to `batch.initiator`. For CRUD entity events `notifier` is the human actor; here there is no actor (system event), so self-notification is a reasonable fallback. A one-line comment would help future maintainers who will otherwise wonder if this is a bug.

Paul-AUB (changes requested):
Re-review after `dcfe3f76`. The four fixes from `grottocenter-front#1464` (notification on the failed-batch path, deterministic `processedRows`/`failures` for failed chunks using `rows.length`, and the `success` → `successes` rename across the service, Swagger, email template, smoke-test script, and fixtures) all look correct and consistent — nice catch renaming it consistently everywhere including the shell script and EJS template.

One issue reintroduces a bug from the very first review round in a new code path.

### Issues (Must Fix)

1. **[api/services/CSVImportQueueService.js:419-436]** The new "all chunks failed" branch in `checkBatchCompletion` calls `notifyCompletion` without the atomic test-and-set guard used a few lines below for the normal completion path (lines 438-447: `updateOne({ id, status: 'active' }).set({ status: 'aggregating' })`). This is exactly the race this guard was added to prevent (see the original review's Issue #1): when the last two chunks of a batch fail at nearly the same time, both enqueue a completion-check job before either one runs; the completion queue processes them sequentially, but both see `allFailed = true` (all chunks are already committed as `failed`), and both call `notifyCompletion`, which unconditionally creates a `TNotification` (line ~665) and sends an email (line ~678) with no idempotency check of its own. Result: the user gets two in-app notifications and two emails when an import fails entirely, mirroring the bug already fixed for the success/partial-success path.

   The new test (`CSVImportQueueService.test.js`, "should set status to failed and notify when all jobs failed") only exercises a single call to `checkBatchCompletion` and doesn't cover concurrent invocation, so it doesn't catch this.

   Suggested fix — reuse the same guard before the `allFailed` branch's side effects, e.g.:
   ```js
   const allFailed = jobs.every((j) => j.state === 'failed');
   if (allFailed) {
     const guard = await TJobBatch.updateOne({
       id: batchId,
       status: 'active',
     }).set({ status: 'failed', completedAt: new Date(), result: resultPayload });
     if (!guard) return; // another worker already handled this
     await module.exports.notifyCompletion(batchId, resultPayload);
     return;
   }
   ```
   (building `resultPayload` before the guard, as it already is).

### Nitpicks (Optional)

2. **[api/services/CSVImportQueueService.js:751-758]** `getBatchProgress`'s failed-chunk branch trusts `data.rows.length` from the job payload to count failures/processedRows. This is consistent with the `aggregateBatch` reconciliation logic, but relies on the job's `data` column still containing the original `rows` array at read time — worth a short comment noting that assumption if pg-boss's job data retention/pruning behavior ever changes (`expireInHours: 4` currently keeps it available long enough in practice).

ClemRz:
Addressed in `7e6c29f4`:

**Issue #1 (Must Fix):** Applied the atomic test-and-set guard to the `allFailed` branch — `TJobBatch.updateOne({ id: batchId, status: 'active' }).set({ status: 'failed', ... })` now prevents concurrent completion-check workers from both calling `notifyCompletion`.

**Nitpick #2:** Added a clarifying comment in `getBatchProgress` documenting that the `data.rows.length` read relies on pg-boss retaining the job's `data` column within the `expireInHours: 4` retention window.

Paul-AUB (approved):
Re-review after `7e6c29f4`. Both items from the previous round are verifiably addressed:

- **Issue #1 (Must Fix) — atomic guard on the `allFailed` branch** — `checkBatchCompletion` now wraps the failed-batch transition in the same test-and-set guard used by the aggregation path: `TJobBatch.updateOne({ id: batchId, status: 'active' }).set({ status: 'failed', ... })`, with `resultPayload` built before the guard and `notifyCompletion` gated behind `if (!guard) return` (`CSVImportQueueService.js:419-442`). Two completion-check jobs firing concurrently for an all-failed batch can no longer both call `notifyCompletion`, so the duplicate-notification/email race is closed — consistent with the guard already protecting the success path. 👍

- **Nitpick #2 — retention assumption comment** — `getBatchProgress` now documents that the failed-chunk row count read from `data.rows.length` relies on pg-boss retaining the job's `data` column within the `expireInHours: 4` window (`CSVImportQueueService.js:757-760`). 👍

The `dcfe3f76 → 7e6c29f4` diff is scoped to exactly these two changes (10 insertions, 2 deletions in a single file); no unrelated changes were bundled in.

The feature is in good shape: affinity-based chunking preserves duplicate detection, both completion paths (success and all-failed) are now guarded against concurrent workers, failed chunks are reconciled against `totalRows`, and error paths prevent silent row loss. No blocking issues remain.

PR #1709, opened 2026-06-28
feat(notification): add batch mark-as-read endpoint

### LINKED ISSUE #1559: feat(notification): batch mark notifications as read
## Summary

Add the ability to mark multiple notifications as read in a single request — either a specific set of notification IDs, or all unread notifications at once.

Currently, the only way to mark a notification as read is one at a time via `POST /api/v1/notifications/:notificationId/read`. Users with many unread notifications (e.g. after a vacation or a burst of activity in subscribed areas) have no way to clear them efficiently.

### What exists today

- `POST /api/v1/notifications/:notificationId/read` — marks a single notification as read by setting `dateReadAt` to the current timestamp. Validates ownership (`notified === req.token.id`).
- `GET /api/v1/notifications/unread/count` — returns the count of unread notifications for the authenticated user.
- `GET /api/v1/notifications` — paginated list of all notifications for the authenticated user.
- `TNotification` model has a `dateReadAt` field (nullable timestamp). A notification is "unread" when `dateReadAt` is `null`.
- All notification routes require the `tokenAuth` policy.

### What's missing

There is no endpoint to mark multiple notifications as read at once. The front-end (GrottoCenter/grottocenter-front) needs a batch endpoint to support:
1. Marking a selected set of notifications as read (checkbox selection).
2. A "mark all as read" action that clears all unread notifications.

### New endpoint: `PUT /api/v1/notifications/read`

A single new endpoint that handles both use cases via the request body:

| Body field | Type | Required | Description |
|---|---|---|---|
| `ids` | `number[]` | No | Specific notification IDs to mark as read. If omitted or empty, **all** unread notifications for the authenticated user are marked as read. |

**Behavior:**

- If `ids` is provided and non-empty: mark only those notifications as read (set `dateReadAt = now()` where `dateReadAt IS NULL`). Validate that every ID belongs to the authenticated user (`notified === req.token.id`). Return `403 Forbidden` if any ID doesn't belong to the user. Ignore IDs that are already read (idempotent).
- If `ids` is omitted or empty: mark **all** unread notifications for the authenticated user as read.
- Use `PUT` since this is an idempotent update operation on existing resources.
- Return `204 No Content` on success (consistent with the existing single mark-as-read endpoint).

### Route & policy registration

- Add `'PUT /api/v1/notifications/read': 'v1/notification/mark-as-read-batch'` to `config/routes.js`.
- Add `'v1/notification/mark-as-read-batch': 'tokenAuth'` to `config/policies.js`.

### Implementation notes

- The controller should delegate to a thin service method or use Waterline directly, consistent with the existing `mark-as-read.js` pattern.
- For the "mark all" case, use `TNotification.update({ notified: req.token.id, dateReadAt: null }, { dateReadAt: new Date() })` — no need to fetch IDs first.
- For the "mark selected" case, first verify ownership of all requested IDs in a single query, then batch-update.
- The existing single `POST /api/v1/notifications/:notificationId/read` endpoint remains unchanged for backward compatibility.

## Acceptance criteria

- [ ] `PUT /api/v1/notifications/read` with `{ ids: [1, 2, 3] }` marks only those notifications as read, if they belong to the authenticated user and are currently unread.
- [ ] `PUT /api/v1/notifications/read` with `{}` or `{ ids: [] }` marks all unread notifications for the authenticated user as read.
- [ ] Returns `403 Forbidden` if any provided ID does not belong to the authenticated user.
- [ ] Returns `204 No Content` on success.
- [ ] Already-read notifications in the `ids` list are silently skipped (idempotent).
- [ ] Requires `tokenAuth` policy.
- [ ] The existing single mark-as-read endpoint (`POST /api/v1/notifications/:notificationId/read`) continues to work unchanged.
- [ ] Integration tests cover: batch by IDs, mark all, ownership validation, idempotency, and unauthenticated access.

## Related

- Front-end issue: GrottoCenter/grottocenter-front#1303 (to be created)
- Existing single mark-as-read: `api/controllers/v1/notification/mark-as-read.js`

ClemRz:
## 🤔 What

- Add `PUT /api/v1/notifications/read` endpoint for batch marking notifications as read
- Closes #1559

## 🤷‍♂️ Why

Users with many unread notifications (after a vacation or a burst of activity in subscribed areas) have no efficient way to clear them. The front-end (GrottoCenter/grottocenter-front#1427) needs a batch endpoint to support a "mark all as read" action and future checkbox selection.

## 🔍 How

- **Controller** (`api/controllers/v1/notification/mark-as-read-batch.js`):
  - If `ids` is provided and non-empty: validates they are positive integers, enforces a max of 1000 IDs, verifies ownership via a single `count()` query, then batch-updates only unread ones
  - If `ids` is omitted or empty: marks all unread notifications for the authenticated user as read
  - Returns 403 for both non-owned and non-existent IDs (prevents ID enumeration)
  - Returns 204 No Content on success (idempotent)
- **Route**: `PUT /api/v1/notifications/read` → `v1/notification/mark-as-read-batch`
- **Policy**: `tokenAuth` (JWT required)
- **Swagger**: OpenAPI 3.0 documentation added

## 🧪 Testing

9 integration tests covering:
- Unauthenticated access (401)
- Mark all unread (empty body / empty ids array)
- Mark specific IDs only
- Ownership validation (403 for other user's notifications)
- Non-existent IDs (403)
- Idempotency (already-read notifications silently skipped)
- Invalid input (400 for non-integer/negative IDs)
- Max length exceeded (400 for >1000 IDs)

All 3097 existing tests pass with no regressions.

### DIFF
--- a/api/controllers/v1/notification/mark-as-read-batch.js
+++ b/api/controllers/v1/notification/mark-as-read-batch.js
@@ -1,0 +1,46 @@
+const MAX_IDS = 1000;
+
+module.exports = async (req, res) => {
+  const ids = req.body && Array.isArray(req.body.ids) ? req.body.ids : [];
+
+  if (ids.length > 0) {
+    // Validate that all provided IDs are positive integers
+    const allValid = ids.every((id) => Number.isInteger(id) && id > 0);
+    if (!allValid) {
+      return res.badRequest('ids must be an array of positive integers.');
+    }
+
+    if (ids.length > MAX_IDS) {
+      return res.badRequest(
+        `ids array exceeds the maximum allowed length of ${MAX_IDS}.`
+      );
+    }
+
+    // Verify ownership: all requested IDs must belong to the authenticated user.
+    // Returns 403 for both "not yours" and "doesn't exist" to prevent ID enumeration.
+    const ownedCount = await TNotification.count({
+      id: { in: ids },
+      notified: req.token.id,
+    });
+
+    if (ownedCount !== ids.length) {
+      return res.forbidden(
+        'You cannot mark as read notifications that do not belong to you.'
+      );
+    }
+
+    // Batch update only the unread ones
+    await TNotification.update({
+      id: { in: ids },
+      dateReadAt: null,
+    }).set({ dateReadAt: new Date() });
+  } else {
+    // Mark all unread notifications for the authenticated user as read
+    await TNotification.update({
+      notified: req.token.id,
+      dateReadAt: null,
+    }).set({ dateReadAt: new Date() });
+  }
+
+  return res.status(204).send();
+};
--- a/api/controllers/v1/notification/mark-as-read-batch.js
+++ b/api/controllers/v1/notification/mark-as-read-batch.js
@@ -1,14 +1,17 @@
 const MAX_IDS = 1000;
 
 module.exports = async (req, res) => {
-  const ids = req.body && Array.isArray(req.body.ids) ? req.body.ids : [];
+  const rawIds = req.body && Array.isArray(req.body.ids) ? req.body.ids : [];
 
-  if (ids.length > 0) {
+  if (rawIds.length > 0) {
     // Validate that all provided IDs are positive integers
-    const allValid = ids.every((id) => Number.isInteger(id) && id > 0);
+    const allValid = rawIds.every((id) => Number.isInteger(id) && id > 0);
     if (!allValid) {
       return res.badRequest('ids must be an array of positive integers.');
     }
+
+    // De-duplicate so that repeated IDs don't cause a false 403
+    const ids = [...new Set(rawIds)];
 
     if (ids.length > MAX_IDS) {
       return res.badRequest(
--- a/assets/swaggerV1.yaml
+++ b/assets/swaggerV1.yaml
@@ -7089,6 +7089,45 @@ paths:
         '500':
           description: The signing secret is not configured for the requested product (server misconfiguration).
 
+  '/notifications/read':
+    put:
+      tags:
+        - notifications
+      summary: Batch mark notifications as read
+      description: >-
+        Marks multiple notifications as read for the authenticated user.
+        If `ids` is provided and non-empty, only those specific notifications are marked as read.
+        If `ids` is omitted or empty, all unread notifications for the authenticated user are marked as read.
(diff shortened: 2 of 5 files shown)


Paul-AUB (dismissed):
Clean, well-scoped implementation that matches the issue spec closely. The ownership check via a single `count()` query is an efficient approach, and returning 403 for both non-owned and non-existent IDs neatly prevents ID enumeration. Route registration (PUT vs. the existing POST — no method collision), policy, Swagger, and the 9 integration tests all look solid. A few non-blocking points below.

1. **[mark-as-read-batch.js:24-33]** Duplicate IDs in the request can produce a false `403`. If the client sends `{ ids: [1, 1, 2] }`, `count()` returns `2` (distinct rows) while `ids.length` is `3`, so the strict `ownedCount !== ids.length` check rejects a request whose IDs all legitimately belong to the user. Consider de-duplicating up front, e.g. `const ids = [...new Set(rawIds)]`, before the length/ownership comparisons. This also makes the `MAX_IDS` ceiling reflect distinct IDs.

2. **[mark-as-read-batch.js:24]** `mark-as-read.js` casts the token id with `Number(req.token.id)` when comparing ownership, whereas here `notified: req.token.id` is passed to Waterline raw. It mirrors `count-unread.js`, so it's consistent and the tests pass — just flagging the divergence from `mark-as-read.js` in case a future change relies on a numeric type. No action needed if Waterline coercion is being relied on intentionally.

### Nitpicks (Optional)

3. **[mark-as-read-batch.js]** Unlike `count-unread.js` (which wraps its logic in `try/catch` + `ErrorService`), this controller has no explicit error handling. It follows the `mark-as-read.js` style and Sails will surface a 500 on rejection, so this is fine — noting only for consistency awareness across the notification controllers.

ClemRz:
Thanks for the review @Paul-AUB!

**Item 1 (duplicate IDs)** — Good catch. Fixed in 0108e725: the raw input is now de-duplicated with `new Set()` before the length check and ownership query. Validation still runs on the raw array so invalid values are caught regardless.

**Item 2 (token id coercion)** — Acknowledged, intentionally consistent with `count-unread.js`. Leaving as-is.

**Item 3 (no try/catch)** — Acknowledged, follows `mark-as-read.js` style. Leaving as-is.

PR #1710, opened 2026-06-28
fix(rateLimit): exempt all relation DELETE routes from IP rate limiter

### LINKED ISSUE #1707: Rate limit too strict on relation/association DELETE endpoints
# Rate limit too strict on relation/association DELETE endpoints

## Problem

The current DELETE rate limiter ([`config/rateLimit/rateLimiter.js`](config/rateLimit/rateLimiter.js)) applies a single uniform limit to **all** HTTP DELETE requests:

- Regular users: **1 DELETE / hour**
- Leaders & Moderators: **20 DELETEs / hour**
- Administrators: unlimited

This is too aggressive for a subset of DELETE endpoints that do **not destroy data** — they only remove a link between two existing entities. A user managing their explored entrances, leaving an organization, or an admin cleaning up document associations could exhaust their hourly quota with a single action.

## Affected endpoints

All of the following endpoints only remove an association row. No entity is deleted.

| Endpoint | Action |
|---|---|
| `DELETE /api/v1/entrances/:entranceId/cavers/:caverId` | Remove explored entrance from a caver |
| `DELETE /api/v1/caves/:caveId/organizations/:organizationId` | Remove explored cave from an organization |
| `DELETE /api/v1/cavers/:caverId/organizations/:organizationId` | Remove a caver from an organization |
| `DELETE /api/v1/cavers/:caverId/groups/:groupId` | Remove a caver from a group |
| `DELETE /api/v1/countries/:id/organizations/:organizationId` | Unlink an organization from a country |
| `DELETE /api/v1/countries/:countryId/regions/:regionId/organizations/:organizationId` | Unlink an organization from a region |
| `DELETE /api/v1/massifs/:id/organizations/:organizationId` | Unlink an organization from a massif |
| `DELETE /api/v1/entrances/:entranceId/documents/:documentId` | Unlink a document from an entrance |
| `DELETE /api/v1/caves/:caveId/documents/:documentId` | Unlink a document from a cave |
| `DELETE /api/v1/massifs/:massifId/documents/:documentId` | Unlink a document from a massif |

## Expected behavior

Relation DELETEs should have their own, more permissive rate limit — separate from the one that protects destructive DELETEs (deleting a cave, an entrance, a caver account, etc.).

Suggested thresholds (open for discussion):

| Role | Relation DELETE limit |
|---|---|
| Regular user | ~20 / hour ou ~30 / hour |
| Leader & Moderator | ~100 / hour |
| Administrator | unlimited |

## Proposed solution

1. Add a dedicated `relationDeleteRateLimit` instance in `config/rateLimit/rateLimiter.js` with relaxed thresholds, driven by new env vars (e.g. `USER_RELATION_DELETE_RATE_LIMIT`, `RELATION_DELETE_RATE_LIMIT`).
2. Apply this limiter **only** to the affected routes in `config/http.js`, before the existing `deleteRateLimit` which remains unchanged for destructive DELETEs.
3. Add integration tests in `test/integration/2_utils/rateLimiter.test.js` covering the new limiter.

ClemRz:
## 🤔 What

- Exempt all 10 relation/association DELETE endpoints from the restrictive `deleteRateLimit` middleware.
- Address review feedback from #1706 (authenticated test variant, negative assertion, comment accuracy).
- Closes #1707

## 🤷‍♂️ Why

The `deleteRateLimit` allows regular users only 1 DELETE request per hour per IP. This is far too aggressive for endpoints that only remove an association row (e.g., unlinking a document from an entrance, removing a caver from an organization). These endpoints don't destroy data — they remove a link between two existing entities. A user managing their associations could exhaust their hourly quota with a single action.

## 🔍 How

Added a `RELATION_DELETE_PATTERNS` array listing all 10 relation DELETE path regexes, and an `isRelationDelete()` helper used in the `deleteRateLimit.skip` function. When a request path matches any of the patterns, the destructive-action rate limit is skipped. The `generalRateLimit` still applies to these routes, so they're not unprotected.

Also addressed Paul's review suggestions from #1706:
1. Added a test variant with `req.token` set (regular user) to mirror the real authenticated call path.
2. Added a negative assertion: requests to `/api/v1/entrances/42/cavers/7/extra` (extending beyond the pattern) are still rate-limited (429).
3. Fixed the inline comment: "tokenAuth ensures the caller is authenticated, and the controller enforces ownership/authorization."

## 🧪 Testing

```bash
```

All 15 rate limiter tests pass, including the 3 new ones.

## 📸 Previews

N/A

### DIFF
--- a/config/rateLimit/rateLimiter.js
+++ b/config/rateLimit/rateLimiter.js
@@ -94,6 +94,10 @@ const deleteMax = (req) => {
  * These routes only remove a link between two entities — no data is destroyed.
  * They are exempt from the restrictive DELETE rate limit.
  *
+ * MAINTENANCE: When adding a new relation DELETE route in config/routes.js,
+ * add a corresponding pattern here so the new route is not subject to the
+ * restrictive 1-per-hour DELETE rate limit.
+ *
  * Covered routes:
  *   DELETE /api/v1/entrances/:entranceId/cavers/:caverId
  *   DELETE /api/v1/caves/:caveId/organizations/:organizationId
--- a/config/rateLimit/rateLimiter.js
+++ b/config/rateLimit/rateLimiter.js
@@ -89,6 +89,39 @@ const deleteMax = (req) => {
   return USER_DELETE_MAX;
 };
 
+/**
+ * Patterns matching relation/association DELETE endpoints.
+ * These routes only remove a link between two entities — no data is destroyed.
+ * They are exempt from the restrictive DELETE rate limit.
+ *
+ * Covered routes:
+ *   DELETE /api/v1/entrances/:entranceId/cavers/:caverId
+ *   DELETE /api/v1/caves/:caveId/organizations/:organizationId
+ *   DELETE /api/v1/cavers/:caverId/organizations/:organizationId
+ *   DELETE /api/v1/cavers/:caverId/groups/:groupId
+ *   DELETE /api/v1/countries/:id/organizations/:organizationId
+ *   DELETE /api/v1/countries/:countryId/regions/:regionId/organizations/:organizationId
+ *   DELETE /api/v1/massifs/:id/organizations/:organizationId
+ *   DELETE /api/v1/entrances/:entranceId/documents/:documentId
+ *   DELETE /api/v1/caves/:caveId/documents/:documentId
+ *   DELETE /api/v1/massifs/:massifId/documents/:documentId
+ */
+const RELATION_DELETE_PATTERNS = [
+  /^\/api\/v1\/entrances\/\d+\/cavers\/\d+$/,
+  /^\/api\/v1\/caves\/\d+\/organizations\/\d+$/,
+  /^\/api\/v1\/cavers\/\d+\/organizations\/\d+$/,
+  /^\/api\/v1\/cavers\/\d+\/groups\/\d+$/,
+  /^\/api\/v1\/countries\/\d+\/organizations\/\d+$/,
+  /^\/api\/v1\/countries\/\d+\/regions\/\d+\/organizations\/\d+$/,
+  /^\/api\/v1\/massifs\/\d+\/organizations\/\d+$/,
+  /^\/api\/v1\/entrances\/\d+\/documents\/\d+$/,
+  /^\/api\/v1\/caves\/\d+\/documents\/\d+$/,
+  /^\/api\/v1\/massifs\/\d+\/documents\/\d+$/,
+];
+
+const isRelationDelete = (path) =>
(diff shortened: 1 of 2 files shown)


Paul-AUB (dismissed):
Solid, well-scoped change. All 10 `RELATION_DELETE_PATTERNS` were cross-checked against `config/routes.js` and each one maps to a real `DELETE` route with matching parameter arity. Since `deleteRateLimit` runs as a global middleware (`config/http.js` order array), `req.path` correctly resolves to the full `/api/v1/...` path, so the anchored regexes behave as intended. The new tests are a good improvement (authenticated variant + negative boundary assertion). No blocking issues found.

2. **Regression coverage for destructive DELETEs** — `config/rateLimit/rateLimiter.js:109-120` The new patterns correctly don't match single-segment destructive routes (e.g. `DELETE /api/v1/entrances/:id`), and the `/extra` negative test is a nice guard. Consider also adding one assertion that a genuinely destructive DELETE path is *still* rate-limited (returns 429), so a future broadening of `RELATION_DELETE_PATTERNS` can't silently exempt a destructive route.

3. **Pattern/route drift** — `RELATION_DELETE_PATTERNS` duplicates the relation-route list that also lives in `config/routes.js`. As relation endpoints are added, the two can drift (a new relation route would silently inherit the strict 1/h limit until someone remembers to update this array). Worth a short comment pointing maintainers here, or a follow-up to derive the list from a single source.

### Nitpicks (Optional)

4. **Trailing slash** — `config/rateLimit/rateLimiter.js:110-119` The `$`-anchored patterns won't match a trailing-slash variant (e.g. `/api/v1/caves/1/documents/50/`), which would fall through to the strict limiter. Real-world risk is low (clients don't send trailing slashes here), just noting the edge.

5. **Test setup duplication** — `test/integration/2_utils/rateLimiter.test.js:` The unauthenticated and authenticated exemption tests share most of their app/agent setup and the same `RELATION_PATHS` loop. Could be parameterized to reduce duplication, but it's perfectly readable as-is.

ClemRz:
Thanks for the thorough review @Paul-AUB! Addressed in 8a021f0b:

**Suggestion 2 (Regression coverage):** Added a test asserting that `DELETE /api/v1/entrances/42` (destructive single-segment route) is still rate-limited. This catches any future over-broadening of `RELATION_DELETE_PATTERNS`.

**Suggestion 3 (Pattern/route drift):** Added a `MAINTENANCE:` comment in the JSDoc above `RELATION_DELETE_PATTERNS` reminding to update the array when adding new relation DELETE routes.

**Suggestions 4 & 5 (trailing slash, test duplication):** Acknowledged but skipped — low risk and readable as-is per your own notes.

PR #1711, opened 2026-06-28
fix(search): resolve null entrances in cave search results

ClemRz:
## 🤔 What

- Fix `entrances` field in cave search results returning `[null, null, ...]` instead of actual entrance IDs
- Add regression tests for `toSimpleCave` entrance ID extraction

## 🤷‍♂️ Why

The quick-search endpoint (`POST /api/v1/search`) returns `null` values in the `entrances` array for cave results. This happens because the production Typesense index stores entrance IDs as plain integers (e.g., `[20, 21, 25877]`), but the `toSimpleCave` converter assumed they would always be Waterline objects with an `.id` property. Calling `.id` on a number yields `undefined`, which JSON serializes as `null`.

## 🔍 How

Changed `source.entrances?.map((e) => e.id)` to `source.entrances?.map((e) => e?.id ?? e)` in `toSimpleCave`. This handles both formats:
- Waterline objects (from DB queries with `.populate()`) → extracts `.id`
- Plain IDs (from Typesense indexed documents) → passes through as-is

## 🧪 Testing

```bash
npm run test -- --grep "toSimpleCave"
```

Four new test cases cover:
1. Waterline objects → extracts IDs
2. Plain numeric IDs (Typesense format) → passes through
3. String IDs → passes through
4. Absent `entrances` field → returns `undefined`

### DIFF
--- a/api/services/mapping/converters.js
+++ b/api/services/mapping/converters.js
@@ -79,7 +79,7 @@ const c = {
     temperature: source.temperature,
     isDiving: source.isDiving,
     nbEntrances: source.nbEntrances ?? source.entrances?.length ?? 0,
-    entrances: source.entrances?.map((e) => e.id),
+    entrances: source.entrances?.map((e) => e?.id ?? e),
   }),
 
   toCaver: (source, meta) => {
--- a/api/services/mapping/converters.js
+++ b/api/services/mapping/converters.js
@@ -79,7 +79,9 @@ const c = {
     temperature: source.temperature,
     isDiving: source.isDiving,
     nbEntrances: source.nbEntrances ?? source.entrances?.length ?? 0,
-    entrances: source.entrances?.map((e) => e?.id ?? e),
+    entrances: source.entrances
+      ?.map((e) => e?.id ?? e)
+      .filter((e) => e != null),
   }),
 
   toCaver: (source, meta) => {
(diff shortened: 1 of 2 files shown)


Paul-AUB (dismissed):

1. **[test/integration/1_services/Converters.test.js]** The `null`-in-array edge case is not covered. With `e?.id ?? e`, a `null` entry evaluates as `null?.id → undefined`, then `undefined ?? null → null`, so nulls pass through silently. This matches the old broken behavior and could mask corrupt Typesense documents. Consider adding a test that documents the current behavior (or asserts that nulls are filtered out, if that's the desired outcome).

### Nitpicks (Optional)

2. **[test/integration/1_services/Converters.test.js]** Consider adding a test for `entrances: []` (empty array). It's trivially handled by the current code, but it documents the contract and prevents future regressions if anyone refactors the optional-chaining logic.

ClemRz:
Both review items addressed in 6c950a24:

1. **Null-in-array edge case** — Added `.filter((e) => e != null)` after the `.map()` in `toSimpleCave` so null/undefined entries from corrupt Typesense documents are stripped rather than passing through silently. Added a regression test.

2. **Empty array contract test** — Added a test for `entrances: []` documenting it returns `[]`.

All 3107 tests pass, lint clean.

PR #1713, opened 2026-06-30
Feat/1684 massif sensitivity status lock

### LINKED ISSUE #1684: Admin-controlled lock on entrance and massif sensitivity
# Requirements Document

## Introduction

Add an `isSensitiveLocked` boolean field to both entrances and massifs in the GrottoCenter API. When an administrator locks the sensitivity of an entrance or massif, only administrators can subsequently change its `isSensitive` value. The lock is a deliberate, symmetric admin action — it can freeze sensitivity in either state (on or off). Additionally, massif cascade operations (marking a massif as sensitive) must skip entrances whose sensitivity is locked.

## Glossary

- **TEntrance**: The Waterline model (`api/models/TEntrance.js`) representing a cave entrance in the `t_entrance` table.
- **TMassif**: The Waterline model (`api/models/TMassif.js`) representing a massif in the `t_massif` table.
- **EntranceService**: The service (`api/services/EntranceService.js`) containing entrance business logic including data coercion and creation.
- **MassifService**: The service (`api/services/MassifService.js`) containing massif business logic including sensitivity cascade propagation.
- **mark-sensitive**: The controller (`api/controllers/v1/massif/mark-sensitive.js`) that sets a massif as sensitive and cascades to child entrances.
- **unmark-sensitive**: The controller (`api/controllers/v1/massif/unmark-sensitive.js`) that removes sensitivity from a massif (no cascade to entrances).
- **preview-sensitive**: The controller (`api/controllers/v1/massif/preview-sensitive.js`) that returns a count of entrances that would be affected by a sensitivity cascade.
- **isSensitiveLocked**: A boolean column on `t_entrance` and `t_massif`. When `true`, only administrators can change the `isSensitive` value of that entity. Defaults to `false` (unlocked).
- **Cascade_Operation**: The process by which marking a massif as sensitive propagates `isSensitive = true` to all non-sensitive entrances geographically within the massif polygon.
- **Administrator**: A user whose JWT token groups include the `ADMINISTRATOR` role as checked by `RightService.hasGroup`.

### Requirement 1: Add isSensitiveLocked column to entrance and massif tables

**User Story:** As a database administrator, I want `is_sensitive_locked` columns on `t_entrance` and `t_massif`, so that the lock state is persisted and queryable.

#### Acceptance Criteria

1. THE `t_entrance` table SHALL have a column `is_sensitive_locked` of type `boolean`, `NOT NULL`, with a default value of `false`.
2. THE `t_massif` table SHALL have a column `is_sensitive_locked` of type `boolean`, `NOT NULL`, with a default value of `false`.
3. THE migration script SHALL be idempotent (using `IF NOT EXISTS` or equivalent safe patterns).
4. THE TEntrance Waterline model SHALL include an `isSensitiveLocked` attribute mapped to the `is_sensitive_locked` column with `type: 'boolean'`, `allowNull: false`, `defaultsTo: false`.
5. THE TMassif Waterline model SHALL include an `isSensitiveLocked` attribute mapped to the `is_sensitive_locked` column with `type: 'boolean'`, `allowNull: false`, `defaultsTo: false`.

### Requirement 2: Return isSensitiveLocked in API responses

**User Story:** As a frontend developer, I want the API to include `isSensitiveLocked` in entrance and massif responses, so that the UI can render lock state indicators and enforce client-side logic.

#### Acceptance Criteria

1. THE `toEntrance` converter SHALL include `isSensitiveLocked` in its output, sourced from `source.isSensitiveLocked` or `source.is_sensitive_locked`.
2. THE `toMassif` converter SHALL include `isSensitiveLocked` in its output, sourced from `source.isSensitiveLocked`.
3. WHEN `isSensitiveLocked` is `null` or `undefined` on the source object (e.g., old data before migration), THE converters SHALL default the output value to `false`.
4. THE `GET /api/v1/entrances/:id` response SHALL include the `isSensitiveLocked` field.
5. THE `GET /api/v1/massifs/:id` response (via `toMassif`) SHALL include the `isSensitiveLocked` field.

### Requirement 3: Entrance update respects lock state

**User Story:** As a caver, I want the API to reject my attempt to change `isSensitive` on a locked entrance, so that administrator-locked sensitivity cannot be overridden.

#### Acceptance Criteria

1. IF an entrance has `isSensitiveLocked` set to `true` and the request sender is not an Administrator, THEN THE `PUT /api/v1/entrances/:id` endpoint SHALL respond with `403 Forbidden` when the request body contains an `isSensitive` value different from the current persisted value.
2. IF an entrance has `isSensitiveLocked` set to `true` and the request sender is an Administrator, THEN THE `PUT /api/v1/entrances/:id` endpoint SHALL allow changes to `isSensitive`.
3. IF an entrance has `isSensitiveLocked` set to `false` or `undefined`, THEN THE existing sensitivity permission logic SHALL apply unchanged (non-admins can mark as sensitive but cannot unmark).
4. THE `PUT /api/v1/entrances/:id` endpoint SHALL allow administrators to set `isSensitiveLocked` to `true` or `false`.
5. THE `PUT /api/v1/entrances/:id` endpoint SHALL reject (ignore or 403) attempts by non-administrators to change `isSensitiveLocked`.

### Requirement 4: Entrance creation respects isSensitiveLocked

**User Story:** As a developer, I want the entrance creation flow to accept `isSensitiveLocked` from administrators, so that administrators can lock sensitivity at creation time.

#### Acceptance Criteria

1. WHEN an Administrator sends `isSensitiveLocked: true` in a `POST /api/v1/entrances` request, THE created entrance SHALL have `isSensitiveLocked` set to `true`.
2. WHEN a non-Administrator sends `isSensitiveLocked: true` in a `POST /api/v1/entrances` request, THE API SHALL ignore the field and create the entrance with `isSensitiveLocked` set to `false`.
3. WHEN `isSensitiveLocked` is omitted from the request body, THE created entrance SHALL default to `isSensitiveLocked: false`.

### Requirement 5: Massif sensitivity lock enforcement

**User Story:** As an administrator, I want the massif sensitivity endpoints to respect the lock state, so that the sensitivity of a locked massif cannot be changed until it is unlocked.

#### Acceptance Criteria

1. IF a massif has `isSensitiveLocked` set to `true`, THEN THE `POST /api/v1/massifs/:id/mark-sensitive` endpoint SHALL respond with `403 Forbidden` with a message indicating the massif sensitivity is locked.
2. IF a massif has `isSensitiveLocked` set to `true`, THEN THE `POST /api/v1/massifs/:id/unmark-sensitive` endpoint SHALL respond with `403 Forbidden` with a message indicating the massif sensitivity is locked.
3. THE `PUT /api/v1/massifs/:id` endpoint SHALL allow administrators to set `isSensitiveLocked` to `true` or `false`.
4. THE `PUT /api/v1/massifs/:id` endpoint SHALL strip `isSensitiveLocked` from the update payload if the request sender is not an Administrator (same pattern as `isSensitive` being stripped).

### Requirement 6: Massif cascade skips locked entrances

**User Story:** As an administrator performing a massif cascade, I want the cascade to skip entrances whose sensitivity is locked, so that manually locked entrances remain unchanged.

#### Acceptance Criteria

1. WHEN `MassifService.propagateSensitivityToEntrances` runs, THE query SHALL exclude entrances where `is_sensitive_locked = true` from the set of entrances to update.
2. THE `POST /api/v1/massifs/:id/mark-sensitive` response SHALL include a `skippedLockedCount` field indicating how many entrances were skipped due to their sensitivity being locked.
3. THE `POST /api/v1/massifs/:id/mark-sensitive` response SHALL continue to include the `count` field indicating how many entrances were actually updated.
4. WHEN entrance auto-inheritance is triggered during entrance creation (point falls within a sensitive massif), THE auto-inheritance SHALL still apply regardless of the entrance's lock state because the entrance is new and has no prior lock.

### Requirement 7: Preview endpoint reports locked entrance count

**User Story:** As an administrator previewing a cascade, I want to know how many entrances are locked and will be skipped, so that I can make an informed decision before confirming.

#### Acceptance Criteria

1. THE `GET /api/v1/massifs/:id/preview-sensitive` response SHALL include a `lockedCount` field indicating how many entrances within the massif have `isSensitiveLocked = true` and `isSensitive = false` (i.e., locked in non-sensitive state and thus would be skipped).
2. THE `GET /api/v1/massifs/:id/preview-sensitive` response SHALL continue to include the `count` field indicating the number of non-sensitive, non-locked entrances that would be affected.
3. IF all non-sensitive entrances in the massif are locked, THE `count` field SHALL be `0` and `lockedCount` SHALL reflect the total locked non-sensitive entrance count.

### Requirement 8: Massif update endpoint handles isSensitiveLocked

**User Story:** As an administrator, I want to set or unset the sensitivity lock on a massif via the update endpoint, so that I can control whether the massif's sensitivity can be changed.

#### Acceptance Criteria

1. WHEN an Administrator sends `isSensitiveLocked` in the `PUT /api/v1/massifs/:id` request body, THE endpoint SHALL persist the value to the `is_sensitive_locked` column.
2. WHEN a non-Administrator sends `isSensitiveLocked` in the `PUT /api/v1/massifs/:id` request body, THE endpoint SHALL strip the field (not persist it).
3. THE `MassifService.getConvertedDataFromClientRequest` SHALL extract `isSensitiveLocked` from the request using the same `coerceBool` pattern used for `isSensitive`.

### Requirement 9: Update Swagger/OpenAPI specification

**User Story:** As a frontend developer or API consumer, I want the OpenAPI spec to document the `isSensitiveLocked` field, so that I can understand the API contract without reading the source.

#### Acceptance Criteria

1. THE `assets/swaggerV1.yaml` SHALL document `isSensitiveLocked` as a boolean field in the entrance request and response schemas.
2. THE `assets/swaggerV1.yaml` SHALL document `isSensitiveLocked` as a boolean field in the massif request and response schemas.
3. THE `assets/swaggerV1.yaml` SHALL document the `skippedLockedCount` field in the `mark-sensitive` response schema.
4. THE `assets/swaggerV1.yaml` SHALL document the `lockedCount` field in the `preview-sensitive` response schema.
5. THE `assets/swaggerV1.yaml` SHALL document the `403` response for `mark-sensitive` and `unmark-sensitive` when the massif sensitivity is locked.

dawoldo:
## 🤔 What

Adds an `isSensitiveLocked` boolean to **entrances** and **massifs**. When an administrator locks the sensitivity of an entity, only administrators can subsequently change its `isSensitive` value. The lock is a deliberate, symmetric admin action; it can freeze sensitivity in either state (on or off). Closes  #1684.

- New `is_sensitive_locked` column (`NOT NULL DEFAULT false`) on `t_entrance`, `h_entrance`, `t_massif`, `h_massif`
- `isSensitiveLocked` exposed in entrance and massif API responses
- Entrance & massif sensitivity changes are blocked for non-admins when the entity is locked
- Massif sensitivity cascade (`mark-sensitive`) now skips locked entrances and reports how many were skipped
- `preview-sensitive` reports how many entrances would be skipped due to the lock

## 🤷‍♂️ Why

Marking an entrance/massif as sensitive hides its coordinates from non-privileged users. Until now, a non-admin could still flip an entrance's sensitivity back, and there was no way for an admin to freeze the sensitivity state of an entity (e.g. an entrance that must stay non-sensitive even though it sits inside a sensitive massif, or one that must stay sensitive regardless of edits). `isSensitiveLocked` gives administrators that explicit control.

## 🔍 How

The feature mirrors the existing `isSensitive` plumbing end-to-end.

- **Migration** (`sql/9_16_2026_06_29_entrance_massif_is_sensitive_locked.sql`): idempotent `ADD COLUMN` on the four tables, plus a `CREATE OR REPLACE` of `histo_update_entrance()` and `histo_update_massif()` (based on their current production bodies) so the history triggers archive the new column.
- **Models / converters**: `isSensitiveLocked` attribute on `TEntrance`/`TMassif`; `toEntrance`/`toMassif` emit it, defaulting `null`/`undefined` to `false` for pre-migration rows.
- **Entrance create**: the field is admin-only — stripped for non-admins (defaults to `false`). A locked entrance is **exempt from massif sensitivity auto-inheritance** at creation, mirroring the cascade skip.
- **Entrance update**: a non-admin attempting to change `isSensitive` on a locked entrance gets **403**; non-admin lock changes are stripped; admins are unaffected.
- **Massif `mark-sensitive` / `unmark-sensitive`**: return **403** when the massif is locked. The cascade query excludes `is_sensitive_locked = true` entrances; `mark-sensitive` returns a new `skippedLockedCount`.
- **Massif `preview-sensitive`**: `count` now reflects only non-sensitive, non-locked entrances; a new `lockedCount` reports how many would be skipped.
- **Massif update**: admins can set `isSensitiveLocked`; it is stripped for everyone else (same pattern as `isSensitive`).

## 🧪 Testing

- **Converters**: `isSensitiveLocked` is emitted and defaults to `false`.
- **Entrance create route**: non-admin value ignored → `false`; admin value persisted; omitted → `false`.
- **Entrance update route**: locked + non-admin sensitivity change → 403; admin allowed; non-admin lock change ignored; admin unlock works.
- **Massif mark-sensitive route**: 403 when locked; cascade skips locked entrances and returns correct `count` + `skippedLockedCount`.
- **Massif unmark-sensitive route**: 403 when locked.
- **Massif preview-sensitive route**: `count` excludes locked entrances and `lockedCount` is reported.
- **Massif update route**: lock persisted for admin, stripped for non-admin.

Migration verified against a local `grottoce` DB: columns present on all four tables, idempotent re-run is a no-op, and both history triggers fire correctly with the new column.

## 📸 Previews

_N/A — backend only._

### DIFF
--- a/api/controllers/v1/entrance/create.js
+++ b/api/controllers/v1/entrance/create.js
@@ -1,5 +1,6 @@
 const ControllerService = require('../../../services/ControllerService');
 const EntranceService = require('../../../services/EntranceService');
+const RightService = require('../../../services/RightService');
 const { validateNameLength } = require('../../../utils/nameValidation');
 
 module.exports = async (req, res) => {
@@ -29,6 +30,15 @@ module.exports = async (req, res) => {
     dateInscription: new Date(),
     isOfInterest: false,
   };
+
+  // Only administrators can lock sensitivity at creation time
+  const isAdmin = RightService.hasGroup(
+    req.token.groups,
+    RightService.G.ADMINISTRATOR
+  );
+  if (!isAdmin) {
+    delete cleanedData.isSensitiveLocked;
+  }
 
   const nameDescLocData =
     EntranceService.getConvertedNameFromClientRequest(req);
--- a/api/controllers/v1/entrance/update.js
+++ b/api/controllers/v1/entrance/update.js
@@ -40,7 +40,7 @@ module.exports = async (req, res) => {
       newIsSensitiveValue !== currentEntrance.isSensitive;
     if (currentEntrance.isSensitiveLocked && wantsSensitiveChange) {
       return res.forbidden(
-        `The sensitivity of this entrance is locked. Only an administrator can change it.`
+        'The sensitivity of this entrance is locked. Only an administrator can change it.'
       );
     }
   }
--- a/api/controllers/v1/entrance/update.js
+++ b/api/controllers/v1/entrance/update.js
@@ -29,6 +29,22 @@ module.exports = async (req, res) => {
 
   // Check if sensitive change is permitted
   const newIsSensitiveValue = cleanedData.isSensitive;
+
+  if (!isAdmin) {
+    // Non-admins cannot change the sensitivity lock
+    delete cleanedData.isSensitiveLocked;
+
+    // When sensitivity is locked, only admins may change isSensitive (either way)
+    const wantsSensitiveChange =
+      newIsSensitiveValue !== undefined &&
+      newIsSensitiveValue !== currentEntrance.isSensitive;
+    if (currentEntrance.isSensitiveLocked && wantsSensitiveChange) {
+      return res.forbidden(
+        `The sensitivity of this entrance is locked. Only an administrator can change it.`
+      );
+    }
+  }
+
   if (
     !isAdmin &&
     newIsSensitiveValue === false &&
--- a/api/controllers/v1/massif/mark-sensitive.js
+++ b/api/controllers/v1/massif/mark-sensitive.js
@@ -24,6 +24,15 @@ module.exports = async (req, res) => {
     return res.notFound({ message: `Massif of id ${massifId} not found.` });
   }
 
+  // Sensitivity is frozen while the massif is locked
+  if (massif.isSensitiveLocked) {
+    return res.forbidden('The sensitivity of this massif is locked.');
+  }
+
+  // Count entrances that will be skipped because their sensitivity is locked
+  const skippedLockedCount =
+    await MassifService.countLockedUnsensitiveEntrances(massifId);
+
   // Idempotency: skip if already sensitive
   if (massif.isSensitive) {
     const updatedMassif = await MassifService.getPopulatedMassif(massifId);
@@ -32,6 +41,7 @@ module.exports = async (req, res) => {
       null,
       {
         count: 0,
+        skippedLockedCount,
         massif: toMassif(updatedMassif),
       },
       { controllerMethod: 'MassifController.mark-sensitive' },
@@ -66,6 +76,7 @@ module.exports = async (req, res) => {
       null,
       {
         count: updatedEntranceIds.length,
+        skippedLockedCount,
         massif: toMassif(updatedMassif),
       },
       { controllerMethod: 'MassifController.mark-sensitive' },
--- a/api/controllers/v1/massif/mark-sensitive.js
+++ b/api/controllers/v1/massif/mark-sensitive.js
@@ -29,11 +29,7 @@ module.exports = async (req, res) => {
     return res.forbidden('The sensitivity of this massif is locked.');
   }
 
-  // Count entrances that will be skipped because their sensitivity is locked
-  const skippedLockedCount =
-    await MassifService.countLockedUnsensitiveEntrances(massifId);
-
-  // Idempotency: skip if already sensitive
+  // Idempotency: skip if already sensitive (no entrances are touched)
   if (massif.isSensitive) {
     const updatedMassif = await MassifService.getPopulatedMassif(massifId);
     return ControllerService.treat(
@@ -41,13 +37,17 @@ module.exports = async (req, res) => {
       null,
       {
         count: 0,
-        skippedLockedCount,
+        skippedLockedCount: 0,
         massif: toMassif(updatedMassif),
       },
       { controllerMethod: 'MassifController.mark-sensitive' },
       res
     );
   }
+
+  // Count entrances that will be skipped because their sensitivity is locked
+  const skippedLockedCount =
+    await MassifService.countLockedUnsensitiveEntrances(massifId);
 
   try {
     // Set the massif as sensitive and get IDs of entrances that were updated
--- a/api/controllers/v1/massif/preview-sensitive.js
+++ b/api/controllers/v1/massif/preview-sensitive.js
@@ -24,11 +24,13 @@ module.exports = async (req, res) => {
 
   try {
     const count = await MassifService.countUnsensitiveEntrances(massifId);
+    const lockedCount =
+      await MassifService.countLockedUnsensitiveEntrances(massifId);
 
     return ControllerService.treat(
       req,
       null,
-      { count },
+      { count, lockedCount },
       { controllerMethod: 'MassifController.preview-sensitive' },
       res
     );
--- a/api/controllers/v1/massif/unmark-sensitive.js
+++ b/api/controllers/v1/massif/unmark-sensitive.js
@@ -23,6 +23,11 @@ module.exports = async (req, res) => {
     return res.notFound({ message: `Massif of id ${massifId} not found.` });
   }
 
+  // Sensitivity is frozen while the massif is locked
+  if (massif.isSensitiveLocked) {
+    return res.forbidden('The sensitivity of this massif is locked.');
+  }
+
   // Idempotency: skip if already non-sensitive
   if (!massif.isSensitive) {
     const updatedMassif = await MassifService.getPopulatedMassif(massifId);
--- a/api/controllers/v1/massif/update.js
+++ b/api/controllers/v1/massif/update.js
@@ -1,6 +1,7 @@
 const ControllerService = require('../../../services/ControllerService');
 const MassifService = require('../../../services/MassifService');
 const NotificationService = require('../../../services/NotificationService');
+const RightService = require('../../../services/RightService');
 const { toMassif } = require('../../../services/mapping/converters');
 const { validateNameLength } = require('../../../utils/nameValidation');
 
@@ -28,6 +29,15 @@ module.exports = async (req, res) => {
   }
   // Sensitivity is managed exclusively via mark-sensitive / unmark-sensitive
   delete cleanedData.isSensitive;
+
+  // Only administrators can change the sensitivity lock
+  const isAdmin = RightService.hasGroup(
+    req.token.groups,
+    RightService.G.ADMINISTRATOR
+  );
+  if (!isAdmin) {
+    delete cleanedData.isSensitiveLocked;
+  }
 
   // Validate name length
   const nameText = req.body.name?.text;
--- a/api/models/HEntrance.js
+++ b/api/models/HEntrance.js
@@ -98,6 +98,13 @@ module.exports = {
       defaultsTo: false,
     },
 
+    isSensitiveLocked: {
+      type: 'boolean',
+      allowNull: false,
+      columnName: 'is_sensitive_locked',
+      defaultsTo: false,
+    },
+
     hasBat: {
       type: 'boolean',
       allowNull: false,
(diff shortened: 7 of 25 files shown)


ClemRz (dismissed):

- **[api/models/HEntrance.js] & [api/models/HMassif.js]** The SQL migration adds `is_sensitive_locked` to both `h_entrance` and `h_massif` and the trigger functions archive the value correctly. However the Waterline models `HEntrance` and `HMassif` don't declare the `isSensitiveLocked` attribute. This means `HEntrance.find()` / `HMassif.find()` won't return the field, so the history snapshot endpoints (e.g. `get-snapshots`) will lack `isSensitiveLocked` even though the `toEntrance` converter already handles it. Consider adding:
  ```js
  isSensitiveLocked: {
    type: 'boolean',
    allowNull: false,
    columnName: 'is_sensitive_locked',
    defaultsTo: false,
  },
  ```
  to both history models for completeness.

- **[api/controllers/v1/massif/mark-sensitive.js:30-31]** `countLockedUnsensitiveEntrances` is called before the idempotency check. When the massif is already sensitive, this extra DB query still fires even though the response is "already sensitive, nothing changed." The overhead is marginal but consider moving it after the `if (massif.isSensitive)` early-return, or alternatively, include `skippedLockedCount` in the idempotent response only when relevant (i.e. skip the query entirely for the no-op path). Up to you — it's technically correct as-is.

### Nitpicks (Optional)

- **[test/integration/1_services/GeoLocNetworks.property.test.js]** The refactoring from `function() { this.timeout(...) }` to arrow + `.timeout(...)` is a nice cleanup and aligns with project conventions. Just confirming this is intentional — it's unrelated to the sensitivity lock feature but welcome.

- **[api/controllers/v1/entrance/update.js:40]** The template literal for the 403 message uses backticks but doesn't interpolate anything. A regular string would work too, though this is purely cosmetic.

ClemRz (approved):

- **[api/controllers/v1/entrance/update.js:40-45]** The new lock check correctly blocks non-admins from toggling `isSensitive` on a locked entrance. However, a subtle interaction exists: after the lock check passes (line 45), the existing code at line 47 still checks `!isAdmin && newIsSensitiveValue === false`. For a locked entrance where `isSensitive` is currently `false`, a non-admin sending `isSensitive: false` will pass the lock check (because the value didn't change) and then also pass the "unmark" check (because the value is the same as current). This is correct behavior — just worth confirming the two guards compose intentionally. No action needed if this was deliberate.

- **[api/controllers/v1/massif/preview-sensitive.js]** The `count` field in the preview response now excludes locked entrances (via the updated `COUNT_UNSENSITIVE_ENTRANCES_IN_MASSIF` query). This means `count + lockedCount` represents all non-sensitive entrances in the massif. Consider documenting this relationship in the Swagger description (e.g., "count + lockedCount = total non-sensitive entrances within the massif") to help API consumers understand the invariant.

- **[sql/9_16_2026_06_29_entrance_massif_is_sensitive_locked.sql]** The trigger `histo_update_entrance()` is correctly replicated from the previous migration (`8_02_...`), with `is_sensitive_locked` added. Note: there are nine boolean columns (`has_bat`, `danger_*`, etc.) that exist on `h_entrance` but are still not archived by this trigger — this is a pre-existing gap (not introduced here), but worth tracking as a future cleanup item.

### Nitpicks (Optional)

- **[test/integration/4_routes/Entrances/create.test.js]** The test "should not auto-inherit massif sensitivity when the entrance is locked" creates a `TMassif` directly (bypassing the normal name creation). This works for the test but leaves an orphaned massif record with no name. Consider adding a `TName.create` for the massif if other tests in the suite assert on massif names, though it doesn't affect this test's correctness.

- **[api/services/MassifService.js:131]** The comment on `COUNT_UNSENSITIVE_ENTRANCES_IN_MASSIF` could be updated to reflect its new filtering behavior (it now excludes locked entrances too). Currently, the query name alone doesn't convey that locked entrances are excluded. A brief comment like "Counts non-sensitive, non-locked entrances..." would improve clarity for future readers.

PR #1716, opened 2026-07-03
Fix/1208 legislation guidelines

dawoldo:
## 🤔 What

Adds the missing permanent (hard) delete mode to the guideline delete endpoint, gated to administrators.

- `DELETE /v1/guidelines/{id}` now accepts an `isPermanent` query parameter
- Default behavior is unchanged: a reversible soft-delete (moderator or admin)
- `isPermanent=true` fully removes the guideline, its history snapshots, and its entity associations — **irreversible, administrators only**
- Updated Swagger docs and added integration tests for both delete modes

## 🤷‍♂️ Why

The issue's requirements now as  for a way to permanently remove a guideline, while keeping the destructive operation restricted to administrators (consistent with the permanent-delete pattern already used for devices and sensor configurations).

## 🔍 How

In [api/controllers/v1/guideline/delete.js](api/controllers/v1/guideline/delete.js):

- **Parameter parsing:** `isPermanent` is treated as permanent only for `true`/`'true'`/`'1'`. Explicit falsy values (`'0'`/`'false'`) and an absent param fall back to soft-delete — avoiding the `!!req.param(...)` pitfall where `isPermanent=0` would read as truthy. The web client sends `?isPermanent=1`.
- **Authorization:** permanent deletion returns `403` for non-admins; soft-delete remains available to moderators.
- **Permanent delete (two-phase):**
  1. If the guideline isn't already soft-deleted, `destroyOne` fires the `change_guideline` trigger (soft-delete) and the `t_last_change` 'delete' row is attributed to the admin via `RecentChangeService.setDeleteRestoreAuthor`.
  2. FK-referencing children (`j_guideline_country`, `j_guideline_region`, `j_guideline_massif`) and history rows (`HGuideline`) are removed, then a second `destroyOne` performs the hard delete.
- **Response:** after a hard delete the row no longer exists, so instead of re-querying (which would return `null`), the already-fetched record is reused with `isDeleted` set to `true`.

Swagger ([assets/swaggerV1.yaml](assets/swaggerV1.yaml)) updated to document the `isPermanent` param, the admin restriction, and the refined `403`/`404` responses.

## 🧪 Testing

Integration tests in [test/integration/4_routes/Guidelines/delete.test.js](test/integration/4_routes/Guidelines/delete.test.js) cover both paths.

### DIFF
--- a/api/controllers/v1/guideline/delete.js
+++ b/api/controllers/v1/guideline/delete.js
@@ -54,19 +54,16 @@ module.exports = async (req, res) => {
       );
     }
 
-    // Remove FK-referencing children before the hard delete
-    await sails.sendNativeQuery(
-      'DELETE FROM j_guideline_country WHERE id_guideline = $1',
-      [guidelineId]
-    );
-    await sails.sendNativeQuery(
-      'DELETE FROM j_guideline_region WHERE id_guideline = $1',
-      [guidelineId]
-    );
-    await sails.sendNativeQuery(
-      'DELETE FROM j_guideline_massif WHERE id_guideline = $1',
-      [guidelineId]
-    );
+    // Remove FK-referencing children before the hard delete. Unlike
+    // device/delete.js and sensor-configuration/delete.js, we don't pre-check
+    // for children and return res.conflict(): the only rows pointing at
+    // t_guideline are these junction tables (countries/regions/massifs), which
+    // we own and delete outright. If a future FK to t_guideline is added that
+    // isn't cleared here, the phase-2 hard delete will surface a raw DB error
+    // instead of a friendly 409 — add the corresponding cleanup (or a guard) then.
+    await JGuidelineCountry.destroy({ guideline: guidelineId });
+    await JGuidelineRegion.destroy({ guideline: guidelineId });
+    await JGuidelineMassif.destroy({ guideline: guidelineId });
     await HGuideline.destroy({ t_id: guidelineId });
 
     await TGuideline.destroyOne({ id: guidelineId }); // Phase 2: hard-delete
@@ -90,6 +87,10 @@ module.exports = async (req, res) => {
 
   // The row is gone after a hard delete, so reuse the already-fetched record
   // rather than re-querying (which would return null). Reflect the deleted state.
+  // The populated countries/regions/massifs arrays on this object still reflect
+  // the pre-delete state; this matches the previous behaviour (which re-fetched
+  // right after the soft-delete) and relies on the `change_guideline` trigger
+  // not cascading to the junction rows on soft-delete — an implicit contract.
   guideline.isDeleted = true;
 
   return ControllerService.treatAndConvert(
--- a/api/controllers/v1/guideline/delete.js
+++ b/api/controllers/v1/guideline/delete.js
@@ -25,7 +25,11 @@ module.exports = async (req, res) => {
     });
   }
 
-  const isPermanent = !!req.param('isPermanent');
+  // The web client sends `?isPermanent=1`; accept the common truthy encodings
+  // ('1'/'true', or a real boolean from a JSON body) while treating explicit
+  // falsy values ('0'/'false') and an absent param as a soft delete. A bare
+  // `!!req.param(...)` would wrongly treat `isPermanent=0`/`false` as permanent.
+  const isPermanent = [true, 'true', '1'].includes(req.param('isPermanent'));
 
   if (isPermanent) {
     // Permanent (irreversible) deletion is gated to administrators; moderators
--- a/api/controllers/v1/guideline/delete.js
+++ b/api/controllers/v1/guideline/delete.js
@@ -25,27 +25,73 @@ module.exports = async (req, res) => {
     });
   }
 
-  if (guideline.isDeleted) {
-    return res.notFound({
-      message: `Guideline of id ${guidelineId} is already deleted.`,
-    });
-  }
-  await TGuideline.destroyOne({ id: guidelineId }); // Trigger soft-deletes
-  const deletedGuideline = await GuidelineService.getGuideline(guidelineId);
+  const isPermanent = req.param('isPermanent') === 'true';
 
-  // The DB trigger `change_guideline` automatically inserts a `t_last_change` row on soft-delete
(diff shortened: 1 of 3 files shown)


ClemRz (dismissed):

- **[api/controllers/v1/guideline/delete.js:32]** The `isPermanent` parsing introduces a third pattern (`[true, 'true', '1'].includes(...)`) that diverges from both existing conventions in this codebase:
  - Most controllers use `!!req.param('isPermanent')` (truthy check)
  - `device/delete.js` and `sensor-configuration/delete.js` use `req.param('isPermanent') === 'true'` (strict equality)

  The rationale in the comments is sound (avoiding the `!!` pitfall with `'0'`), but the inconsistency means `isPermanent=1` hard-deletes a guideline while having no effect on a device or sensor-configuration. Consider aligning with `device/delete.js` (`=== 'true'` only) for consistent API behavior, or — if the `'1'` case is genuinely needed — open a follow-up to unify all controllers.

- **[api/controllers/v1/guideline/delete.js:56–67]** The raw SQL deletes for `j_guideline_country`, `j_guideline_region`, and `j_guideline_massif` are correct and parameterized, but the model already has Waterline associations (`countries`, `regions`, `massifs`) via through-tables that have their own models (`JGuidelineCountry`, `JGuidelineRegion`, `JGuidelineMassif`). You could use:
  ```js
  await JGuidelineCountry.destroy({ guideline: guidelineId });
  await JGuidelineRegion.destroy({ guideline: guidelineId });
  await JGuidelineMassif.destroy({ guideline: guidelineId });
  ```
  This keeps the code at the ORM layer (consistent with `HGuideline.destroy({ t_id: guidelineId })` on line 68) and avoids coupling to physical table/column names. The raw approach works fine too — this is a style preference for consistency within the same function.

- **[api/controllers/v1/guideline/delete.js:39–55]** The two-phase soft-delete + hard-delete pattern here doesn't include an FK guard that `device/delete.js` and `sensor-configuration/delete.js` do. Those controllers check for children *between* phases and return `res.conflict()` if hard-delete would violate referential integrity. Since guidelines own junction rows (countries, regions, massifs) that are explicitly deleted on lines 56–68, there's no true FK risk here. But if a future FK is added pointing *to* `t_guideline` that isn't covered by the explicit deletions, the hard delete would fail with a raw DB error rather than a friendly 409. A comment noting this design choice would help future maintainers.

- **[api/controllers/v1/guideline/delete.js:73]** In the soft-delete path, `guideline.isDeleted = true` is mutated on the originally-fetched (populated) object. This works because `toSimpleGuideline` only reads the properties it needs. However, the original code re-queried via `GuidelineService.getGuideline(guidelineId)` after soft-delete. The new approach (reusing the stale object) means the `countries`, `regions`, and `massifs` arrays returned in the response will still reflect pre-delete state (populated objects), which is consistent with what the old code did. Just flagging that this behavioral equivalence relies on the trigger *not* cascading deletes to junction rows during soft-delete — which is the current behavior but is an implicit contract.

- **[test/integration/4_routes/Guidelines/delete.test.js:86]** The test for "hard delete a guideline along with its history and associations" doesn't assert on `j_guideline_region`. The controller deletes from all three junction tables, but the test only verifies `j_guideline_country` and `j_guideline_massif`. Consider adding the region association to the test setup (e.g., `await TGuideline.addToCollection(guideline.id, 'regions', ['FR-01'])`) and verifying it's cleaned up.

### Nitpicks (Optional)

- **[api/controllers/v1/guideline/delete.js]** The `GuidelineService` import is now only used for `getGuideline()` at line 22. The file no longer uses it for a post-delete re-query, which is fine — just noting the import is still justified.

- **[test/integration/4_routes/Guidelines/delete.test.js:91]** The cleanup in the "403 for moderator" test uses `TGuideline.destroy({ id: guideline.id })` which soft-deletes. This leaves soft-deleted test records in the DB. Since the test template is rebuilt per shard, this isn't a real issue, but it differs from the hard-delete tests that clean up completely.

ClemRz (approved):

- **[api/controllers/v1/guideline/delete.js:32]** The `isPermanent` parsing uses `[true, 'true', '1'].includes(req.param('isPermanent'))`, which differs from `device/delete.js` and `sensor-configuration/delete.js` that use `req.param('isPermanent') === 'true'`. This means `?isPermanent=1` hard-deletes a guideline but has no effect on a device or sensor-configuration — inconsistent API behavior for consumers. The rationale in the comments is sound, but consider either aligning with the existing `=== 'true'` pattern for consistency, or opening a follow-up issue to unify all permanent-delete controllers to accept `'1'` as well. As it stands, the Swagger spec declares `type: boolean`, so `?isPermanent=true` is the canonical form and supporting `'1'` is an undocumented extension.

- **[api/controllers/v1/guideline/delete.js:46–57]** In the permanent-delete branch, if `!guideline.isDeleted`, the controller performs a soft-delete (phase 1) and then calls `RecentChangeService.setDeleteRestoreAuthor`. Neither `device/delete.js` nor `sensor-configuration/delete.js` call `RecentChangeService` at all during permanent deletion. This means hard-deleting a guideline (that wasn't already soft-deleted) leaves a `t_last_change` "delete" row attributed to the admin, which then has its referenced entity fully removed. Consider whether this orphaned `t_last_change` row is desirable. If the hard-delete removes all traces of the entity, should the `t_last_change` entry also be cleaned up? (This may be intentional for audit purposes — worth a comment if so.)

### Nitpicks (Optional)

- **[test/integration/4_routes/Guidelines/delete.test.js:91]** The cleanup for the "403 for moderator" test uses `await TGuideline.destroy({ id: guideline.id })`, which soft-deletes via the trigger. Since the test database is rebuilt per shard this isn't a real problem, but using a native query (`DELETE FROM t_guideline WHERE id = $1`) or adding a second `destroy` call would truly remove it — matching the thoroughness of the other tests.

- **[test/integration/4_routes/Guidelines/delete.test.js:101]** The test "should hard delete a guideline along with its history and associations" creates a guideline but doesn't explicitly verify that the history snapshot from the phase-1 soft-delete was created before asserting it's gone. Adding a brief comment like `// Phase 1 creates a snapshot; phase 2 removes it` would make the test's expectations clearer for future readers.

PR #1719, opened 2026-07-05
feat(document): add image thumbnail generation on upload

### LINKED ISSUE #1468: [IMAGES] Server-side Thumbnail Generation for Images
### Context

Currently, the frontend displays image thumbnails (e.g. in `ImageThumbnail`) by loading the full-resolution image using `completePath`, even for small preview cards (~240×180).

While `loading="lazy"` helps defer offscreen images, all visible thumbnails still trigger full image downloads.

---

### Problem

This approach has significant performance implications:

- High-resolution images (often 5MB+) are downloaded even when displayed at small sizes
- Increased page load time, especially for documents containing multiple images
- Higher bandwidth consumption (client + server)
- Poor UX on slow or mobile networks

This becomes particularly noticeable in views listing multiple documents with images.

---

### Current Behavior

- Frontend requests: `completePath`
- Backend returns: original full-resolution image
- No resizing or optimization occurs server-side

---

### Proposed Solution

Introduce **server-side thumbnail generation** as part of the image upload pipeline.

#### Key idea

When an image is uploaded:
1. Store the original image (existing behavior)
2. Generate one or more resized versions (thumbnails)
3. Store and expose their paths via the API

#### Example

```json
{
  "completePath": "/images/abc123.jpg",
  "thumbnailPath": "/images/thumbnails/abc123_240.jpg"
}
```

### Link in frontend
https://github.com/GrottoCenter/grottocenter-front/pull/1174#pullrequestreview-3816137826

ClemRz:
## 🤔 What

- Add synchronous server-side thumbnail generation for image uploads
- Generate up to 3 WebP variants (480px, 1280px, 1920px) when an image is uploaded via `FileService.document.create()`
- Expose thumbnail URLs in the API response via a `thumbnails` object on each file
- Add a backfill script for existing images (`scripts/backfill-thumbnails.js`)
- Closes #1468

## 🤷‍♂️ Why

The frontend currently downloads full-resolution images for every display context (cards, mobile, lightbox). This wastes bandwidth and degrades performance, especially on mobile. Serving appropriately-sized WebP variants per context reduces page load time significantly.

## 🔍 How

- **New `ThumbnailService`** (`api/services/ThumbnailService.js`): wraps `sharp` for resize + WebP conversion. Pure functions for path computation and variant selection, plus an `async generate()` that handles Azure upload with all-or-nothing semantics.
- **SQL migration** adds 3 nullable `VARCHAR(1000)` columns to `t_file` (`thumbnail_small`, `thumbnail_medium`, `thumbnail_large`).
- **`FileService.document.create()`** calls `ThumbnailService.generate()` after the original upload succeeds. Uses `foundFormat[0].mimeType` (not the pre-existing buggy `const { mimeType } = foundFormat` destructuring from the array).
- **`FileService.document.delete()`** cleans up thumbnail blobs (best-effort).
- **`toFile` converter** returns a `thumbnails` object (or `null` if no thumbnails exist).
- **Backfill script** lifts Sails, queries files without thumbnails via `.populate('fileFormat')`, processes in configurable batches with `--dry-run` and `--batch-size` flags.
- **Graceful degradation**: if thumbnail generation fails, the upload still succeeds with null thumbnails.

## 🧪 Testing

- 45 new tests (property-based + unit + integration), all passing
- Property tests validate: no-upscaling invariant, path computation pattern, converter output shape
- Integration tests validate: image upload produces thumbnails, non-image has null, sharp failure doesn't break upload, delete cleans up blobs
- Full test suite: 3155 passing, 2 pre-existing failures (unrelated timing tests)
- Requires `AZURE_KEY` env var to test actual thumbnail generation against Azure

## 📸 Previews

API response shape for an image file with thumbnails:
```json
{
  "fileName": "cave-entrance.jpg",
  "completePath": "…/12345-cave-entrance.jpg",
  "thumbnails": {
    "small": "…/thumbnails/small/12345-cave-entrance.webp",
    "medium": "…/thumbnails/medium/12345-cave-entrance.webp",
    "large": "…/thumbnails/large/12345-cave-entrance.webp"
  }
}
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -72,6 +72,7 @@ module.exports = {
     TSubstance: 'readonly',
     SubstanceService: 'readonly',
     PubChemService: 'readonly',
+    ThumbnailService: 'readonly',
     TType: 'readonly',
     TTokenBlacklist: 'readonly',
     TConversation: 'readonly',
--- a/api/models/TFile.js
+++ b/api/models/TFile.js
@@ -62,5 +62,26 @@ module.exports = {
       columnName: 'id_document',
       model: 'TDocument',
     },
+
+    thumbnailSmall: {
+      type: 'string',
+      allowNull: true,
+      columnName: 'thumbnail_small',
+      maxLength: 1000,
+    },
+
+    thumbnailMedium: {
+      type: 'string',
+      allowNull: true,
+      columnName: 'thumbnail_medium',
+      maxLength: 1000,
+    },
+
+    thumbnailLarge: {
+      type: 'string',
+      allowNull: true,
+      columnName: 'thumbnail_large',
+      maxLength: 1000,
+    },
   },
 };
--- a/api/services/FileService.js
+++ b/api/services/FileService.js
@@ -86,6 +86,26 @@ module.exports = {
 
   isCredentials: !!credentials,
 
+  /**
+   * Test-only: Retrieve the current module-level credentials object.
+   * Used by tests to save original credentials before injecting mocks.
+   * @returns {object|null}
+   */
+  getCredentialsForTest() {
+    return credentials;
+  },
+
+  /**
+   * Test-only: Override the module-level credentials object.
+   * Allows integration tests to inject a mock blob client without
+   * reimplementing create/delete. Call with `null` to restore original.
+   * @param {object|null} mockCredentials
+   */
+  setCredentialsForTest(mockCredentials) {
+    credentials = mockCredentials;
+    this.isCredentials = !!mockCredentials;
+  },
+
   document: {
     getUrl(path) {
       // The documents container allow anonymous access
--- a/api/services/FileService.js
+++ b/api/services/FileService.js
@@ -126,7 +126,7 @@ module.exports = {
       if (foundFormat.length === 0) {
         throw new FileError(INVALID_FORMAT, name);
       }
-      const { mimeType } = foundFormat;
+      const { mimeType } = foundFormat[0];
 
       if (!credentials) {
         noCredentialWarning('Document upload', {
@@ -147,6 +147,21 @@ module.exports = {
         }
       }
 
+      let thumbnailPaths = { small: null, medium: null, large: null };
+      const fileMimeType = foundFormat[0].mimeType;
+      if (credentials && ThumbnailService.isProcessable(fileMimeType)) {
+        try {
+          thumbnailPaths = await ThumbnailService.generate(
+            file.buffer,
+            pathName,
+            credentials.documentsBlobClient
+          );
+        } catch (err) {
+          sails.log.error('Thumbnail generation failed:', err);
+          // Continue — thumbnails are optional
+        }
+      }
+
       const param = {
         dateInscription: new Date(),
         fileName: name,
@@ -154,6 +169,9 @@ module.exports = {
         fileFormat: foundFormat[0].id,
         path: pathName,
         isValidated,
+        thumbnailSmall: thumbnailPaths.small,
+        thumbnailMedium: thumbnailPaths.medium,
+        thumbnailLarge: thumbnailPaths.large,
       };
       if (fetchResult) {
         let createQuery = TFile.create(param);
@@ -187,6 +205,27 @@ module.exports = {
             destroyedRecord.path
           );
         await blockBlobClient.delete({ deleteSnapshots: 'include' });
+
+        // Best-effort cleanup of thumbnail blobs
+        const thumbnailPaths = [
+          destroyedRecord.thumbnailSmall,
+          destroyedRecord.thumbnailMedium,
+          destroyedRecord.thumbnailLarge,
+        ].filter(Boolean);
+        await Promise.all(
+          thumbnailPaths.map(async (thumbPath) => {
+            try {
+              const thumbBlobClient =
+                credentials.documentsBlobClient.getBlockBlobClient(thumbPath);
+              await thumbBlobClient.delete({ deleteSnapshots: 'include' });
+            } catch (err) {
+              sails.log.error(
+                `Failed to delete thumbnail blob ${thumbPath}:`,
+                err
+              );
+            }
+          })
+        );
       }
       return destroyedRecord;
     },
--- a/api/services/ThumbnailService.js
+++ b/api/services/ThumbnailService.js
@@ -1,0 +1,141 @@
+/**
+ * ThumbnailService.js
+ *
+ * @description :: Service for generating image thumbnails using sharp.
+ *   Encapsulates resize + WebP conversion logic and Azure upload for variants.
+ */
+
+const sharp = require('sharp');
+const path = require('path');
+
+/**
+ * MIME types that sharp can process for thumbnail generation.
+ * Excludes SVG (rasterization unreliable), PCX, DXF (unsupported by sharp).
+ * Note: image/webp is excluded because webp is not registered in t_file_format,
+ * so webp uploads are rejected before reaching this code.
+ */
+const PROCESSABLE_MIME_TYPES = new Set([
+  'image/jpeg',
+  'image/png',
+  'image/gif',
+  'image/tiff',
+  'image/x-ms-bmp',
+]);
+
+const VARIANTS = [
+  { name: 'small', width: 480 },
+  { name: 'medium', width: 1280 },
+  { name: 'large', width: 1920 },
+];
+
+const WEBP_QUALITY = 80;
+
+module.exports = {
+  PROCESSABLE_MIME_TYPES,
+  VARIANTS,
+  WEBP_QUALITY,
+
+  /**
+   * Determine if a MIME type is processable for thumbnails.
+   * @param {string} mimeType
+   * @returns {boolean}
+   */
+  isProcessable(mimeType) {
+    return PROCESSABLE_MIME_TYPES.has(mimeType);
+  },
+
+  /**
+   * Compute the thumbnail blob path for a given variant and original path.
+   * @param {string} variant - 'small' | 'medium' | 'large'
+   * @param {string} originalPath - e.g. '12345-cave-entrance.jpg'
+   * @returns {string} e.g. 'thumbnails/small/12345-cave-entrance.webp'
+   */
+  computeThumbnailPath(variant, originalPath) {
+    const ext = path.extname(originalPath);
+    const stem = originalPath.slice(0, originalPath.length - ext.length);
+    return `thumbnails/${variant}/${stem}.webp`;
+  },
+
+  /**
+   * Determine which variants should be generated based on original width.
+   * Only returns variants whose target width is strictly less than the original.
+   * @param {number} originalWidth - Width of the source image in pixels
+   * @returns {Array<{name: string, width: number}>}
+   */
+  getApplicableVariants(originalWidth) {
+    return VARIANTS.filter((v) => v.width < originalWidth);
+  },
+
+  /**
+   * Resize an image buffer to the target width, output as WebP.
+   * Preserves aspect ratio (no crop, no upscaling).
+   * Preserves alpha channel for PNG inputs and animation for GIF inputs.
+   * @param {Buffer} inputBuffer - The original image buffer
+   * @param {number} targetWidth - The target width in pixels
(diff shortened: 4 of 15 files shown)


ClemRz:
## Frontend Integration Guide

### API Response Shape

Every file object in document responses now includes a `thumbnails` field:

```json
{
  "id": 1630,
  "fileName": "cave-entrance.jpg",
  "completePath": "https://grottocenter.blob.core.windows.net/documents/12345-cave-entrance.jpg",
  "thumbnails": {
    "small": "https://grottocenter.blob.core.windows.net/documents/thumbnails/small/12345-cave-entrance.webp",
    "medium": "https://grottocenter.blob.core.windows.net/documents/thumbnails/medium/12345-cave-entrance.webp",
    "large": "https://grottocenter.blob.core.windows.net/documents/thumbnails/large/12345-cave-entrance.webp"
  }
}
```

### When `thumbnails` is `null`

- The file is not an image (PDF, CSV, etc.)
- Thumbnail generation failed (corrupted file, Azure error)
- The file was uploaded before backfill ran

**Fallback:** always use `completePath` when `thumbnails` is `null`.

### When individual variants are `null`

If the original image is smaller than a variant's target width, that variant is `null` (no upscaling). For example, a 1000px wide image will have:
- `small`: URL (480px < 1000px ✓)
- `medium`: `null` (1280px > 1000px, skipped)
- `large`: `null` (1920px > 1000px, skipped)

### Recommended Usage by Context

| Context | Use variant | Fallback chain |
|---------|-------------|----------------|
| Card preview (desktop) | `thumbnails.small` | `thumbnails.medium` → `completePath` |
| Mobile full-width card | `thumbnails.medium` | `thumbnails.large` → `completePath` |
| Lightbox / full view | `thumbnails.large` | `completePath` |
| Download / deep zoom | `completePath` (always) | — |

### Example `srcSet` Implementation

```jsx
const getImageSources = (file) => {
  if (!file.thumbnails) {
    return { src: file.completePath };
  }

  const { small, medium, large } = file.thumbnails;
  const srcSet = [
    small && `${small} 480w`,
    medium && `${medium} 1280w`,
    large && `${large} 1920w`,
  ].filter(Boolean).join(', ');

  return {
    src: small || medium || large || file.completePath,
    srcSet: srcSet || undefined,
    sizes: '(max-width: 480px) 100vw, (max-width: 1280px) 50vw, 33vw',
  };
};
```

### Notes

- All thumbnails are WebP format (broadly supported, ~30% smaller than JPEG at equivalent quality)
- URLs are public (no auth token needed), same as `completePath`
- Old images will get thumbnails after the backfill script runs post-deployment
- No API contract changes needed — `thumbnails` is purely additive

Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[api/services/FileService.js:129](api/services/FileService.js#L129)** The pre-existing `const { mimeType } = foundFormat;` bug (destructuring from the array returned by `TFileFormat.find(...).limit(1)` instead of `foundFormat[0]`) is still present and still used for the *original* file's Azure upload (`blobContentType: mimeType` at line 129) and in the `noCredentialWarning` call. The PR description explicitly calls out this bug ("Uses `foundFormat[0].mimeType` … not the pre-existing buggy `const { mimeType } = foundFormat` destructuring") and the new thumbnail code correctly reads `foundFormat[0].mimeType`, but the original upload path was left unfixed — every uploaded original file (image or not) is still stored with `blobContentType: undefined`. Since this PR already touches this exact code and the description implies it's resolved, consider fixing the original destructuring too (`const { mimeType } = foundFormat[0];`) so both call sites are consistent and correct.

2. **[test/integration/1_services/FileService.thumbnails.test.js:1148-1241](test/integration/1_services/FileService.thumbnails.test.js#L1148-L1241)** This "integration" suite entirely replaces `FileService.document.create` and `FileService.document.delete` with hand-written copies inside `before()`, restoring the originals in `after()`. As a result, none of the assertions in this file (thumbnail generation on create, cleanup on delete, PDF/SVG null-thumbnails, sharp-failure fallback) actually exercise the real `FileService.document.create`/`delete` implementations — they exercise a parallel reimplementation that must be kept in sync by hand. A regression in the real `FileService.js` (e.g. the bug in finding #1) would not be caught by this suite. Consider instead stubbing only the Azure `BlobServiceClient`/`credentials` at the module level (or injecting a container client) so the real `FileService.document.create`/`delete` functions run under test.

3. **[test/integration/1_services/ThumbnailService.property.test.js:1628-1644](test/integration/1_services/ThumbnailService.property.test.js#L1628-L1644)** Property 3 ("Converter thumbnail output construction") defines a local `buildExpectedThumbnails` helper that duplicates the `toFile` converter's thumbnail-building logic instead of calling `converters.toFile` directly. This tests that the copy is self-consistent, not that the actual converter in `api/services/mapping/converters.js` behaves this way. Consider importing `converters` and asserting against `converters.toFile(source).thumbnails` instead.

4. **[api/services/ThumbnailService.js:233-241](api/services/ThumbnailService.js#L233-L241)** `generate()`'s "all-or-nothing" semantics generate buffers first and then upload with `Promise.all`, but if one upload rejects, blobs from the other (already-resolved) uploads in that `Promise.all` batch remain in Azure Blob Storage while the DB record stores `null` for all variants (per the catch block). This leaves orphaned thumbnail blobs with no reference in `t_file`, which will never be cleaned up (including by `FileService.document.delete`, since it only deletes paths read from `thumbnailSmall/Medium/Large`, which are null). Not a correctness blocker for the upload flow, but worth a note or follow-up since storage cost/cleanup was a stated design goal.

5. **[scripts/backfill-thumbnails.js:1036](scripts/backfill-thumbnails.js#L1036)** `succeeded += 1` runs whenever no exception is thrown, including when `ThumbnailService.generate` internally caught an error and returned all-null paths (per its own try/catch). Since a real generation failure (corrupted image, unsupported variant, Azure hiccup within `generate`) is swallowed inside `ThumbnailService.generate` rather than thrown, the backfill script's own `failed` counter will under-report actual thumbnail-generation failures, and the final summary can claim "Succeeded: N" for files that ended up with no thumbnails at all. Consider having `generate()` optionally surface whether it succeeded (e.g. return a `hadError` flag) so the backfill summary is accurate.

### Nitpicks (Optional)

6. **[assets/swaggerV1.yaml:8150-8345](assets/swaggerV1.yaml#L8150-L8345)** Beyond adding `thumbnails`, this hunk also removes/replaces unrelated `File` schema fields (`dateReviewed`, `document`, `fileFormat`, `path`) with `id`, `isValidated`, `completePath` — a broader documentation fix unrelated to thumbnail generation. Likely a legitimate correction (matching what `toFile` actually returns), but worth calling out separately in the PR description/commit history since it's bundled with the thumbnail feature.

ClemRz:
Thanks for the thorough review @Paul-AUB! Here's how each item was addressed in `16a277ac`:

**1. `const { mimeType } = foundFormat` bug** — This was already fixed in the PR (`foundFormat[0]` destructuring at line 130). The diff shows the change from `develop`.

**2. Tests reimplementing `FileService.document.create/delete`** — Fixed. Added `setCredentialsForTest()` to `FileService` so tests can inject a mock blob client and exercise the real `create`/`delete` code paths. The test no longer contains a hand-written copy of the function.

**3. Property test using local `buildExpectedThumbnails` helper** — Fixed. Now imports and calls the real `toFile` converter from `api/services/mapping/converters.js`.

**4. Orphaned blobs on partial upload failure** — Added a comment documenting the trade-off. Orphaned blobs from partial `Promise.all` failures are rare and low-cost vs. the complexity of per-variant rollback. A periodic cleanup job can address this if needed.

**5. Backfill `succeeded` counter** — Already handled correctly: the script only increments `succeeded` when at least one thumbnail path is non-null, and increments `skipped` otherwise. The `ThumbnailService.generate` catch block returns all-null paths (doesn't throw), and the backfill checks for that explicitly.

**6. Swagger schema changes** — Acknowledged. The removals (`dateReviewed`, `document`, `fileFormat`, `path`) and additions (`id`, `isValidated`, `completePath`) align the schema with what `toFile` actually returns. This is a documentation correction bundled with the feature.

Paul-AUB (dismissed):
Re-reviewed after commit `16a277ac` ("fix(document): address PR review feedback on thumbnail generation"), which addresses the changes requested in the previous review.

### Verified fixes

1. **[api/services/FileService.js:140](api/services/FileService.js#L140)** The `const { mimeType } = foundFormat;` destructuring bug is now fixed to `const { mimeType } = foundFormat[0];`, matching the correct pattern already used for `fileMimeType`. Confirmed in the diff (`-const { mimeType } = foundFormat;` / `+const { mimeType } = foundFormat[0];`).
2. **[test/integration/1_services/FileService.thumbnails.test.js](test/integration/1_services/FileService.thumbnails.test.js)** now injects mock Azure credentials via the new `FileService.setCredentialsForTest()` and calls the real `FileService.document.create`/`delete` throughout — no more hand-written reimplementation. This closes the gap where a regression in the real code wouldn't have been caught.
3. **[test/integration/1_services/ThumbnailService.property.test.js:5,184,209](test/integration/1_services/ThumbnailService.property.test.js#L5)** now imports and asserts against the real `converters.toFile`, instead of a local `buildExpectedThumbnails` duplicate.
4. **[api/services/ThumbnailService.js:120-126](api/services/ThumbnailService.js#L120-L126)** the orphaned-blob trade-off for partial `Promise.all` upload failures is now documented with an explanatory comment. Acceptable as a documented, low-cost trade-off rather than a fix.

5. **[scripts/backfill-thumbnails.js:154-168](scripts/backfill-thumbnails.js#L154-L168)** The `succeeded`/`skipped` split is technically correct as clarified (a file only counts as `succeeded` when at least one thumbnail path is non-null). However, a genuine `ThumbnailService.generate()` failure (corrupted image, Azure hiccup, etc.) is caught internally and returns all-null paths without throwing, so it still lands in the `skipped` bucket and is reported to the operator as "Skipped (too small)" — which is misleading for files that actually failed rather than being legitimately too small to need thumbnails. Consider having `generate()` distinguish "no applicable variants" from "an error occurred" (e.g. a `hadError` flag) so the backfill summary counts and log message stay accurate.
6. **[test/integration/1_services/FileService.thumbnails.test.js:12-14,38-44](test/integration/1_services/FileService.thumbnails.test.js#L12-L44)** `before()` only saves `FileService.isCredentials` (a boolean) as `originalCredentials`, then `after()` restores `credentials` to `{ documentsBlobClient: mockContainerClient }` (the mock) whenever that boolean was `true` — never the real original credentials object, which isn't saved anywhere. In an environment where `AZURE_KEY` is set (the PR description notes this is needed "to test actual thumbnail generation against Azure"), running this suite would leave the module-level `credentials` permanently pointed at the mock client for the rest of the test run. Low risk in CI (where `AZURE_KEY` is typically unset), but worth tightening if thumbnail tests are ever run with real Azure credentials locally.

### Nitpicks (Optional)

7. **[assets/swaggerV1.yaml:8150-8177](assets/swaggerV1.yaml#L8150-L8177)** As previously noted, the `File` schema changes bundle an unrelated documentation correction (removing `dateReviewed`/`document`/`fileFormat`/`path`, adding `id`/`isValidated`/`completePath`) alongside the `thumbnails` addition. Acknowledged as intentional in the author's response — no action needed, just flagging for the commit history record.

No new Must-Fix issues found in the delta. Approving.

ClemRz:
Addressed review suggestions #5 and #6 from the second review in `04af2aae`:

**#5 — Backfill `succeeded`/`skipped` accuracy:**
- `ThumbnailService.generate()` now returns a `hadError: boolean` flag alongside the thumbnail paths.
- The backfill script uses `hadError` to distinguish "image too small for any variant" (counted as `skipped`) from "generation actually failed" (counted as `failed` with a warning log). The summary is now accurate.

**#6 — Test credential restore:**
- Added `FileService.getCredentialsForTest()` to retrieve the actual credentials object before overriding.
- The test's `after()` hook now restores the real original credentials (not the mock), so subsequent test suites aren't affected when `AZURE_KEY` is set.

**Bonus:** Fixed 3 pre-existing `func-names` lint warnings in `GeoLocNetworks.property.test.js` (switched to arrow functions with chained `.timeout()`).

PR #1722, opened 2026-07-08
fix(entrance): use raw SQL for h_name queries and filter redundant name changes

ClemRz:
## 🤔 What

- Replace Waterline `HName.find()` with raw SQL for `h_name` queries in `EntranceService.getHEntrancesWithName`
- Add `filterToActualNameChanges` to `TemporalNameResolver` to suppress redundant name-change snapshots
- Add property-based test (Property 5) and integration test coverage for the filtering logic
- Add fixture (`hname.json` id=4) to exercise the filter in route tests

## 🤷‍♂️ Why

`h_name` has a composite primary key `(id, date_reviewed)` but the Waterline model declares `primaryKey: 'id'`. When a `t_name` row is updated multiple times, `h_name` accumulates rows sharing the same `id` with different `date_reviewed`. Waterline deduplicates by `id`, silently losing history records.

Additionally, non-name edits (e.g. bumping `date_reviewed`) trigger the history trigger, producing `h_name` rows with unchanged names. These generated spurious name-change snapshots in the timeline.

## 🔍 How

1. **Raw SQL with parameterized queries** — bypass Waterline's deduplication by querying `h_name` directly with `$N` placeholders. The entrance query joins `t_caver` for author/reviewer; the cave query intentionally skips that join since `resolveNameAtDate` only needs `name` and `dateReviewed`.

2. **`filterToActualNameChanges`** — pure function that sorts records by `dateReviewed`, then keeps only those whose `name` differs from the next record's name (or the current `TName` for the last entry). Defensive sort ensures correctness regardless of caller ordering.

3. **Type safety** — `entranceId` is explicitly coerced to integer before passing to the SQL query.

## 🧪 Testing

- `npm run test` — all existing tests pass (the `login-banned` failures are a pre-existing Docker clock-drift issue unrelated to this change)
- Property 5 validates the filter with randomized inputs (100 runs, small name pool to force duplicates)
- Route test for `GET /api/v1/entrances/1/snapshots` asserts exactly 2 name-change snapshots and verifies the filtered-out name doesn't appear

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/api/services/EntranceService.js
+++ b/api/services/EntranceService.js
@@ -121,12 +121,38 @@ module.exports = {
       : true; // No need to call hasRight if it's not a sensitive entrance
 
     // Fetch all HName records for the entrance (temporal resolution)
-    const entranceHNames = await HName.find({
-      entrance: entranceId,
-      isMain: true,
-    })
-      .populate('author')
-      .populate('reviewer');
+    // Use raw SQL because h_name has a composite PK (id, date_reviewed) but
+    // the Waterline model declares primaryKey: 'id'. When the same t_name row
+    // is updated multiple times, h_name contains multiple rows with the same id
+    // but different date_reviewed. Waterline deduplicates by id, losing records.
+    const numericEntranceId =
+      typeof entranceId === 'string' ? parseInt(entranceId, 10) : entranceId;
+    const entranceHNamesRaw = await CommonService.query(
+      `SELECT h.id, h.name, h.is_main, h.date_inscription, h.date_reviewed,
+              h.id_entrance, h.id_cave, h.id_author, h.id_reviewer,
+              a.nickname AS author_nickname, r.nickname AS reviewer_nickname
+       FROM h_name h
+       LEFT JOIN t_caver a ON a.id = h.id_author
+       LEFT JOIN t_caver r ON r.id = h.id_reviewer
+       WHERE h.id_entrance = $1 AND h.is_main = true
+       ORDER BY h.date_reviewed`,
+      [numericEntranceId]
+    );
+    const entranceHNames = (entranceHNamesRaw.rows || []).map((row) => ({
+      id: row.id,
+      name: row.name,
+      isMain: row.is_main,
+      dateInscription: row.date_inscription,
+      dateReviewed: row.date_reviewed,
+      entrance: row.id_entrance,
+      cave: row.id_cave,
+      author: row.id_author
+        ? { id: row.id_author, nickname: row.author_nickname }
+        : null,
+      reviewer: row.id_reviewer
+        ? { id: row.id_reviewer, nickname: row.reviewer_nickname }
+        : null,
+    }));
 
     // Fetch all current TName records for the entrance (full names collection)
     const allEntranceNames = await TName.find({ entrance: entranceId });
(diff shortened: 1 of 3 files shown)


Paul-AUB (dismissed):
Thorough, well-reasoned change. The root cause (Waterline deduplicating `h_name` by the mis-declared `primaryKey: 'id'` on a composite-PK table) is clearly explained, the raw SQL fix is correctly parameterized, and the new filtering logic is backed by a property test plus a route test. Nice detail that the manual row mapping reduces `author`/`reviewer` to exactly `{ id, nickname }`, which is all `toSimpleCaver` (used by the snapshot converter) reads — no field regression there. The `.rows` destructuring and `CommonService.query` usage are consistent with the rest of the codebase.

No blocking issues found. A few suggestions and nitpicks below.

1. **[EntranceService.js:180-187]** Type-coercion is inconsistent between the two queries. The entrance query explicitly coerces `entranceId` to an integer (`numericEntranceId`), but the cave query passes `caveIds` straight through. In practice `caveIds` are DB-derived integers so it works today, but for symmetry and defensiveness consider coercing them the same way (e.g. `caveIds.map((id) => parseInt(id, 10))`), or drop the coercion in the entrance query if it's deemed unnecessary — either way, making the two paths consistent aids readability.

2. **[EntranceService.js:130-155, 181-199]** The two raw queries each hand-map `h_name` columns to camelCase objects, duplicating the column→property mapping and bypassing the Waterline model entirely. If the `h_name` schema changes, both blocks (and the model) must be updated in lockstep. Consider extracting a small shared mapper (e.g. `mapHNameRow(row)`) so the entrance and cave paths stay in sync, with the cave path simply omitting author/reviewer.

### Nitpicks (Optional)

3. **[EntranceService.js:138, 185]** The SQL `ORDER BY date_reviewed` is effectively redundant since `filterToActualNameChanges` and `resolveNameAtDate` both re-sort in JS. Harmless and arguably aids readability of the raw result, so keeping it is fine — just noting it isn't load-bearing.

4. **[TemporalNameResolver.js:97-100]** `new Date(x).getTime()` is recomputed for each comparison during the sort. Negligible for the expected record counts, but if these arrays ever grow large a precomputed key (e.g. Schwartzian transform) would avoid repeated `Date` parsing.

5. **[EntranceService.js:137]** `is_main = true` is filtered in SQL, so the mapped `isMain` field is always `true`. It's carried through harmlessly for object-shape parity with the old Waterline records — just flagging it as effectively constant.

ClemRz:
Addressed review feedback in deab3bcf:

1. **NaN guard** — Added `if (Number.isNaN(numericEntranceId)) return [];` after the `parseInt` coercion in `getHEntrancesWithName`. Defensive against non-numeric `:id` params since the route lacks `validateId`.

2. **Nullish coalescing** — Changed `currentName || ''` to `currentName ?? ''` in `filterToActualNameChanges` so only `null`/`undefined` are coerced (not other falsy values).

3. **Shared row mapper** — Extracted `mapHNameRow(row, { includeAuthorReviewer })` helper in `EntranceService.js` to DRY the two `h_name` column→camelCase mapping blocks. The entrance path passes `{ includeAuthorReviewer: true }`, the cave path uses the default (omits author/reviewer).

Items acknowledged but not acted on (reviewer flagged as informational/harmless):
- Dynamic placeholder safety note on `caveIds` — DB-derived integers, no risk
- SQL `ORDER BY` redundancy — aids readability, kept as-is
- Schwartzian transform for sort — negligible for expected record counts
- `isMain` always `true` — carried for object-shape parity

PR #1725, opened 2026-07-09
fix(entrance): consistent snapshot semantic and fix history triggers

### LINKED ISSUE #1723: History triggers missing columns — snapshot data loss
## Bug: History triggers missing columns — entrance snapshots have incorrect data

Two history triggers are incomplete, causing data loss in the snapshot timeline.

### 1. `histo_update_entrance` missing boolean columns

The trigger (`sql/0_triggers.sql`) does not copy these columns to `h_entrance`:

- `has_bat`, `danger_flooding`, `danger_co2`, `danger_rockfall`, `danger_pollution`
- `need_clean_gear`, `need_stay_on_trail`, `has_rules`, `is_touristic`
- `precision`

They were added to both `t_entrance` and `h_entrance` but the trigger was never updated. All h_entrance rows get the column default (`false`) instead of the actual values.

**Impact**: The UI shows all boolean flags as if they changed in the last edit (false diff). Historical boolean data is permanently lost for existing snapshots.

### 2. `histo_update_name` does not record who made the rename

The trigger copies `OLD.id_reviewer` (reviewer of the previous version, typically NULL) instead of `NEW.id_reviewer` (the person performing the rename).

**Impact**: All name-change snapshots show "Modifié par un auteur inconnu". Cannot attribute name changes to a user.

### Verification

```sql
-- Booleans all false in history despite being true in t_entrance
SELECT date_reviewed, has_bat, danger_flooding, danger_co2, danger_rockfall,
       danger_pollution, need_clean_gear, need_stay_on_trail, has_rules, is_touristic
FROM h_entrance WHERE id = 148366 ORDER BY date_reviewed;

SELECT has_bat, danger_flooding, danger_co2, danger_rockfall, danger_pollution,
       need_clean_gear, need_stay_on_trail, has_rules, is_touristic
FROM t_entrance WHERE id = 148366;

-- id_reviewer always NULL in h_name
SELECT id, name, id_reviewer, date_reviewed
FROM h_name WHERE id_entrance = 148366 AND is_main = true ORDER BY date_reviewed;
```

### Proposed fix

1. Update `histo_update_entrance` to include the missing columns in the INSERT
2. Update `histo_update_name` to store `NEW.id_reviewer` instead of `OLD.id_reviewer`

### LINKED ISSUE #1724: Entrance snapshot name semantic inconsistent with other properties
### Context

All snapshot endpoints (description, comment, rigging, location, history, guideline, document) follow the same pattern: every property in the JSON comes from the h_ table row, which stores the **OLD values** (state before the change at `date_reviewed`). This is consistent with the copy-on-write trigger design.

### Problem

The entrance snapshot endpoint (`GET /api/v1/entrances/:id/snapshots`) introduced temporal name resolution that breaks this consistency:

| Property | Source | Semantic |
|---|---|---|
| `latitude`, `longitude`, `cave`, `altitude`, booleans, etc. | h_entrance columns | OLD value (state before the change) |
| `name` | `resolveNameAtDate(snapshot.id)` | Name **active** at that time (resolved from h_name timeline) |
| `caveName` | `resolveNameAtDate(snapshot.id)` | Cave name **active** at that time |
| `names` | Current TName records | **Current** state (not historical at all) |
| `isNameChangeSnapshot.name` | `record.name` from h_name | OLD name being replaced |

Within the same array element, some properties answer "what was the state before this edit?" while others answer "what was the state at this point in time?" — and `names` answers "what is the state right now?".

### Affected code

- `api/services/EntranceService.js` — `getHEntrancesWithName()` resolves `entrance.name` via `TemporalNameResolver.resolveNameAtDate()` and assigns `entrance.names = allEntranceNames` (current TName records)
- `api/services/TemporalNameResolver.js` — `buildNameChangeSnapshots()` uses `record.name` (OLD value) for the synthetic snapshot's `name`

### Expected behavior

Every element in the `entrances` array should follow the same semantic for all its properties. Either:

**Option A — "State before the change" (match the dominant pattern):** Remove temporal name resolution. Let `name` come from h_name raw OLD values, like every other h_ snapshot type does for its own-table properties.

**Option B — "State at that point in time" (full temporal resolution):** Resolve all properties to what they were at that specific timestamp. This is a larger refactor but provides a more useful timeline for the UI.

Either way, `names` should not return current TName records — it should either be the historical names at that time or be omitted.

### Related

- #1723 (trigger bugs — missing booleans and reviewer)

ClemRz:
## 🤔 What

- Fix entrance snapshot `names` field exposing current TName records instead of historical state
- Fix `histo_update_entrance` trigger missing 11 columns (`has_bat`, `danger_flooding`, `danger_co2`, `danger_rockfall`, `danger_pollution`, `need_clean_gear`, `need_stay_on_trail`, `has_rules`, `is_touristic`, `is_sensitive_locked`, `precision`)
- Fix `histo_update_name` trigger storing `OLD.id_reviewer` (always NULL) instead of `NEW.id_reviewer` (who made the change)
- Add temporal language resolution for snapshot entries
- Closes #1723
- Closes #1724

## 🤷‍♂️ Why

The entrance snapshot endpoint had inconsistent semantics: most properties followed "state before the change" (from h_entrance OLD values), but `names` returned the current TName records and `language` depended on them. This broke the front-end diff rendering and leaked current state into historical snapshots.

The two trigger bugs meant:
1. All boolean flags in h_entrance were always `false` regardless of actual state — causing the UI to show them as "changed" on every entrance
2. Name-change snapshots never showed who performed the rename (always "unknown author")

## 🔍 How

**Snapshot semantic consistency:**
- `entrance.names = []` on all h_entrance snapshots (empty, consistent with "no historical names data")
- Language resolved temporally via new `resolveLanguageAtDate()` using h_name records (same algorithm as name resolution)
- `getMainLanguage()` now falls back to `source.language` when `names` array is empty

**Trigger fixes (SQL migration `1_2026_07_09_fix_history_triggers.sql`):**
- `histo_update_entrance`: added all missing boolean columns + `precision` + `is_sensitive_locked` to the INSERT statement
- `histo_update_name`: changed `OLD.id_reviewer` → `NEW.id_reviewer` in the VALUES clause

Note: existing h_entrance rows cannot be retroactively fixed — the boolean data was never captured. Only future snapshots will be correct.

## 🧪 Testing

- `npm run dev:clean` to rebuild containers with the new triggers
- `npm run test` — all passing
- Manual testing via UI: create entrance, edit booleans, rename, move cave — verify snapshots via `GET /api/v1/entrances/:id/snapshots`
- Verify booleans are now correctly captured in new h_entrance rows after `dev:clean`
- Verify `id_reviewer` is populated on h_name rows after a rename

## 📸 Previews

N/A — API-only changes, no UI modifications.


Paul-AUB (changes requested):
Thorough review focused on the two trigger fixes (#1723) and the snapshot semantic changes (#1724). The JavaScript changes are clean and well-tested, but the SQL trigger fixes have a load-ordering problem that silently reverts both fixes on any from-scratch database.

### Issues (Must Fix)

1. **[sql/1_2026_07_09_fix_history_triggers.sql + sql/0_triggers.sql]** The `histo_update_entrance` boolean/precision fix is overwritten on a fresh DB build. The `sql/` directory is mounted to `/docker-entrypoint-initdb.d` (`docker/docker-compose.yml:18`), so on a clean init Postgres runs the files in **alphabetical order**. `histo_update_entrance` is redefined by three later-sorted files:
   - `sql/8_02_2026_04_25_entrance_insert_trigger.sql:37` — INSERT ends at `id_geology`, no booleans / precision / is_sensitive_locked
   - `sql/9_16_2026_06_29_entrance_massif_is_sensitive_locked.sql:51` — adds `is_sensitive_locked` but **still omits** `has_bat`, `danger_*`, `need_*`, `has_rules`, `is_touristic`, `precision`

   Because `9_16…` sorts after `1_2026_07_09…`, it is the **final** definition on a fresh `npm run dev:clean`, so the boolean columns and `precision` are still never captured — the #1723 fix is effectively dead for any database built from scratch. (Existing production DBs that apply this migration incrementally, after `9_16` already ran, are unaffected; and the CI test DB only loads `0_initDatabase.sql`, so tests won't catch this.) Consider either renaming this migration so it sorts **after** `9_16` (e.g. a higher prefix), or updating `8_02` and `9_16` so their trigger bodies also carry the complete column set.

2. **[sql/1_2026_07_09_fix_history_triggers.sql + sql/0_triggers.sql]** Same root cause for the `histo_update_name` `id_reviewer` fix. `sql/9_04_2026_06_03_scientific_observations_ddl.sql:437` redefines `histo_update_name()` and still uses `OLD.id_reviewer` (line 465). It sorts after this PR's migration, so on a fresh init the reviewer fix is reverted too — name-change snapshots would again record `OLD.id_reviewer`. The fix in `9_04` needs to be reconciled, or the corrected definition must be guaranteed to run last.

3. **[api/services/EntranceService.js:232-247]** This resolves the primary ask of #1724 (`names` no longer leaks current `TName` records — good). However the within-element semantic is still mixed: `name`, `caveName` and `language` are resolved to "state at that point in time" (temporal), while `latitude`, `longitude`, booleans, etc. come from the h_entrance OLD row ("state before the change"). Issue #1724 framed this as choosing Option A (all raw OLD) or Option B (full temporal). This lands in between. It may be an acceptable deliberate trade-off, but consider adding a short note (in code or the issue) documenting that the mixed semantic is intentional, so the next reader doesn't re-open the same question.

4. **[api/services/TemporalNameResolver.js:54-69]** `resolveLanguageAtDate` duplicates the exact future-record filter/sort of `resolveNameAtDate` (lines 22-41). Consider extracting a small private helper that returns the active h_name record for a snapshot date, then have both functions read `.name` / `.language` from it — keeps the temporal algorithm defined in one place.

### Nitpicks (Optional)

5. **[api/services/TemporalNameResolver.js:150-173]** `buildNameChangeSnapshots` sets `names: []` but never sets `language`, so with the new `getMainLanguage` fallback (`source.language ?? null`) these synthetic snapshots always report `language: null`. This matches prior behaviour, but if the UI displays language on name-change rows, consider resolving it here too for consistency with the h_entrance snapshots.

ClemRz (commented):
All 5 review items addressed. Here's what changed:

## Must Fix (items 1 & 2) — SQL load ordering

Renamed `sql/1_2026_07_09_fix_history_triggers.sql` → `sql/9_17_2026_07_09_fix_history_triggers.sql` so it sorts after `8_02` and `9_16` on fresh `dev:clean` builds. Also added `id_observation` to `histo_update_name` (matches the `9_04` schema).

## Suggestions (items 3 & 4)

- Added a comment documenting the intentional mixed semantic (temporal for names/language, raw OLD for coordinates/booleans)
- Extracted `findActiveRecordAtDate` helper in `TemporalNameResolver.js` to DRY the shared filter/sort logic

## Nitpick (item 5)

Added `language: record.language ?? null` to `buildNameChangeSnapshots` output.

## Additional fixes discovered during testing

- **`TName.reviewer` on rename**: The update controller now sets `reviewer: req.token.id` on the TName row during a rename, so `histo_update_name`'s `NEW.id_reviewer` is actually populated.
- **Atomic move-to-cave**: Replaced `removeFromCollection`/`addToCollection` with a single `TEntrance.updateOne` — produces one h_entrance snapshot instead of two (no phantom NULL-cave intermediate).
- **UTC timestamp fix**: `machinepack-postgresql` bundles its own `pg` instance. We now configure both the top-level and nested `pg` to serialize/parse dates in UTC. No-op on production (already UTC) — only fixes local dev with non-UTC clocks.
- **`isEnrichmentSnapshot` flag**: Added `is_enrichment` column to `h_entrance`. The enrichment job sets `SET LOCAL app.is_enrichment = 'true'` in its transaction, and the trigger reads it — deterministic, no heuristics. The converter passes it through as `isEnrichmentSnapshot` for the front-end.
- **`docs/snapshot-types.md`**: Front-end integration guide explaining the three snapshot types and recommended UI treatment.

## Test results

- **Lint**: ✅
- **Tests**: 3192 passing, 0 failing
- **Manual curl test suite**: all snapshot flags, booleans, reviewer, and chronological ordering verified on a fresh `dev:clean` environment

Paul-AUB (dismissed):
All five items from the previous review have been properly addressed. The rename to `9_17_2026_07_09_fix_history_triggers.sql` confirms the sort-ordering fix — `9_17` now sorts after `8_02`, `9_04`, and `9_16`, so the corrected trigger definitions are guaranteed to be the final ones on any fresh `dev:clean` build. The `findActiveRecordAtDate` refactor, the mixed-semantic comment, `language` on name-change snapshots, and the `reviewer` population on rename are all cleanly implemented.

The additional fixes (atomic move-to-cave via `TEntrance.updateOne`, UTC timestamp fix, `isEnrichmentSnapshot` flag via `SET LOCAL`) are solid, and `docs/snapshot-types.md` is a good reference for the front-end.

1. **[app.js:37 + config/datastores.js:304]** Both entry points (`node app.js` and `sails lift`) require `machinepack-postgresql/node_modules/pg` by reaching directly into a transitive dependency's internal `node_modules`. If `machinepack-postgresql` is ever updated to use a peer `pg` (or flattens its dependency tree), this `require` path will break at startup with no graceful fallback. Consider wrapping it in a try/catch to degrade gracefully (e.g. log a warning and skip the patch if the nested path isn't found), or resolve the path via `require.resolve('machinepack-postgresql/node_modules/pg')` so the error message is actionable.

2. **[app.js:35-43 + config/datastores.js:302-310]** The six-line pg UTC-fix block is duplicated verbatim in both files. A single `require('./config/pg-utc-fix')` in both places would prevent the two copies from drifting. Low priority since the fix is intentional and documented, but worth noting for long-term maintainability.

### Nitpicks (Optional)

3. **[docs/snapshot-types.md:377]** The example response in the docs shows `"reviewer": { "id": 2, "nickname": "MelvildMode" }` on the **second** (post-enrichment) snapshot but `"reviewer": null` on the first. Per the trigger semantics, the enrichment job itself doesn't supply a reviewer, so both snapshots would have `"reviewer": null` in practice (the h_entrance row stores `OLD.id_reviewer`, which is set by the previous human edit). The example may mislead front-end developers about when `reviewer` is populated on enrichment snapshots.

ClemRz:
Addressed items 1 & 2 from the latest review:

**Item 1 (fragile nested require) + Item 2 (duplication):** Extracted the pg UTC-fix into a shared `config/pg-utc-fix.js` module and promoted `pg` from devDependencies to dependencies (it's required in production code). Both `app.js` and `config/datastores.js` now just `require('./config/pg-utc-fix')` / `require('./pg-utc-fix')`.

The nested `machinepack-postgresql/node_modules/pg` path remains because the two pg instances are genuinely different versions (8.22 vs 8.11, not deduped by npm). The shared module documents why this is necessary.

**Item 3 (nitpick — reviewer in docs example):** Skipped — the second snapshot in the example is a regular user edit (no `isEnrichmentSnapshot` flag), not an enrichment snapshot. Its `reviewer` being populated is correct: `OLD.id_reviewer` reflects whoever last edited the entrance before that trigger fired.

All tests passing (3192 pass, 0 fail). Lint clean.

PR #1734, opened 2026-07-21
fix(search): populate name in KML/GPX exports with field mapping

### LINKED ISSUE #1728: Bug: gestion des noms dans les fichiers d'export
Lana m'a indiqué que les fichier gpx et kml ne contenaient pas le nom des cavités
En regardant le nom est bien présent mais pas dans le champ name

Image

Si c'est possible il serait sans doute souhaitable que le nom soit de la champ name de manière a faciliter l'usage du fichier dans d'autres applications.
Je n'ai pas fait de test pour vérifier que la localisation du nom dans le fichier était réellement un pb

ClemRz:
## 🤔 What

- Fix KML and GPX exports missing entrance name in the `<name>` element when custom columns are selected
- Closes #1728

## 🤷‍♂️ Why

When exporting entrances to KML or GPX with field mapping (custom column selection), the `<name>` element was empty because the field-mapping logic only carried over `id`, coordinates, and the user-selected aliased fields. Tools like Google Earth, QGIS, and Marble rely on `<name>` to label placemarks/waypoints — without it, exported points appear unlabeled.

## 🔍 How

In the controller's field-mapping path (`advanced-search-export.js`), the `name` field from the original Typesense document is now always included in the mapped document when the format is `kml` or `gpx`. This ensures the serializers can populate the standard `<name>` element regardless of which columns the user selected.

GeoJSON is unaffected — it doesn't use a structural `<name>` element and the field-selection contract remains unchanged.

## 🧪 Testing

1. Start the local environment (`npm run dev:up && npm run dev`)
2. Export entrances to KML with custom columns:
   ```bash
   curl -s -X POST 'http://localhost:1337/api/v1/advanced-search/export?format=kml' \
     -d '{"entity":"entrances","query":"*","columns":["name","country"],"columnsName":["Nom","Pays"]}'
   ```
3. Verify each `<Placemark>` has a non-empty `<name>` element
4. Repeat with `format=gpx` and verify each `<wpt>` has a non-empty `<name>` element
5. Verify GeoJSON export still excludes the raw `name` field from properties when field mapping is active

### DIFF
--- a/api/controllers/v1/search/advanced-search-export.js
+++ b/api/controllers/v1/search/advanced-search-export.js
@@ -250,6 +250,7 @@ module.exports = async (req, res) => {
         docsToSerialize = filtered.map((doc) => {
           // Only include coordinates and id for geometry construction;
           // aliased fields are the sole user-visible properties.
+          // KML/GPX also need the entrance name for the <name> element.
           const mapped = {
             id: doc.id,
             latitude: doc.latitude,
@@ -257,6 +258,9 @@ module.exports = async (req, res) => {
           };
           if (doc.altitude != null) {
             mapped.altitude = doc.altitude;
+          }
+          if (format === 'kml' || format === 'gpx') {
+            mapped.name = doc.name;
           }
           fieldMapping.forEach(({ key, alias }) => {
             mapped[alias] = resolveField(doc, key);
--- a/test/integration/4_routes/AdvancedSearchExport/export-gpx.test.js
+++ b/test/integration/4_routes/AdvancedSearchExport/export-gpx.test.js
@@ -144,6 +144,31 @@ describe('Geo Export - GPX', () => {
         });
     });
 
+    it('should populate <name> even when field mapping is active', (done) => {
+      const SearchService = require('../../../../api/services/SearchService');
+      sinon.stub(SearchService, 'collectionSearch').resolves({
+        hits: entranceHits,
+        found: 2,
+      });
+
+      supertest(sails.hooks.http.app)
+        .post('/api/v1/advanced-search/export?format=gpx')
+        .send({
+          query: 'test',
+          entity: 'entrances',
+          columns: ['country'],
+          columnsName: ['Pays'],
+        })
+        .expect(200)
+        .end((err, res) => {
+          if (err) return done(err);
+          const body = res.text;
+          should(body).match(/<name>Cave A<\/name>/);
+          should(body).match(/<name>Cave B<\/name>/);
+          return done();
+        });
+    });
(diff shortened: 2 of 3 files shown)


Paul-AUB (dismissed):
## Review

Focused, well-scoped fix. The change correctly targets the field-mapping path in `advanced-search-export.js`, which was the only path that dropped `name`; the non-mapped path already passes raw documents through untouched. Both KML and GPX serializers list `name` in their `GEOMETRY_FIELDS`, so it renders in `<name>`/`<name>` and is excluded from `<ExtendedData>`/`<desc>` — no duplication. The `format === 'kml' || format === 'gpx'` guard is necessary rather than incidental: `geojson.js` does **not** include `name` in its `GEOMETRY_FIELDS`, so including `name` unconditionally would leak it into GeoJSON `properties` and break the field-selection contract. The PR description's claim that GeoJSON is unaffected is accurate.

No must-fix issues found.

1. **[test/integration/4_routes/AdvancedSearchExport/export-kml.test.js / export-gpx.test.js]** No regression test accompanies the fix. The GeoJSON suite already exercises the field-mapping path (`export-geojson.test.js:212`, `columns: ['name', 'country']`), but the KML/GPX suites only assert `<name>` on the non-mapped path (`export-kml.test.js:138`). Since the bug lived specifically in the field-mapping branch, consider adding a case that POSTs with `columns`/`columnsName` (that exclude a `Name` alias) and asserts each `<Placemark>`/`<wpt>` still carries a non-empty `<name>`. This locks in the fix against future refactors of the mapping logic.

### Nitpicks (Optional)

2. **[api/controllers/v1/search/advanced-search-export.js:262]** The controller now encodes per-format knowledge of which serializers want `name`. This is fine given the current serializer design, but it does couple the controller to serializer internals — if a future format also needs `name`, this list must be updated in lockstep. A longer-term alternative would be to always pass `name` and let each serializer decide (by adding/removing `name` from its own `GEOMETRY_FIELDS`), centralizing the "does this format surface name?" decision in one place. Not worth changing for this PR.

ClemRz:
Thanks for the review @Paul-AUB!

**Suggestion #1 (regression tests):** Good catch — added regression tests in both `export-kml.test.js` and `export-gpx.test.js` that exercise the field-mapping path (`columns`/`columnsName`) and assert `<name>` is still populated in each `<Placemark>`/`<wpt>`. See 9f8aa1e0.

**Nitpick #2 (controller coupling):** Agreed it's fine for now. If we add more geo formats we can revisit centralizing the "does this format need name?" decision in the serializers themselves.

PR #1738, opened 2026-07-23
feat(auth): add anti-bot protection to signup endpoint

### LINKED ISSUE #1727: Verify Cloudflare Turnstile token and rate-limit `/signup`
# Verify Cloudflare Turnstile token and rate-limit `/signup`

Companion of the front-end ticket  https://github.com/GrottoCenter/grottocenter-front/issues/1462

## Context

Bots are massively creating accounts through the public sign-up form. The front-end companion ticket introduces a **Cloudflare Turnstile** widget on the sign-up form (https://github.com/GrottoCenter/grottocenter-front/issues/1462 — see the *Recommendation* section) which produces a **token** and sends it along with the sign-up payload. That token is worthless unless the back-end verifies it: a bot that bypasses the front-end widget can still POST directly to `/signup` today. This ticket covers the back-end half.

## Scope

Changes to the grottocenter-api repository, on the `POST /api/<version>/signup` endpoint (mirrors [apiRoutes.js:25](packages/web-app/src/conf/apiRoutes.js#L25) on the front). Three layers are added on top of the existing sign-up handler:

1. **Turnstile token verification** (primary defense)
2. **Honeypot field rejection** (cheap complement)
3. **Rate limiting** on the endpoint (defense in depth)

The existing e-mail verification stays untouched.

### Configuration

Add two secrets to the API environment (never committed):

- `TURNSTILE_SECRET_KEY` — the secret key issued by Cloudflare for the site key used on the front. Different keys per environment (dev, staging, prod). Cloudflare provides always-pass / always-fail test keys for CI.
- (optional) `TURNSTILE_ENABLED` — a boolean flag to disable the check locally / in tests without redeploying.

### Handler changes

The front now sends an extra field in the JSON body — proposed name: `captchaToken` (string). The back-end must:

1. **Reject early** if `captchaToken` is missing or empty → `400 Bad Request` with an error code the front can translate (e.g. `CAPTCHA_MISSING`).
2. **POST** to Cloudflare's verification endpoint:
   ```
   POST https://challenges.cloudflare.com/turnstile/v0/siteverify
   Content-Type: application/x-www-form-urlencoded

   secret=<TURNSTILE_SECRET_KEY>&response=<captchaToken>&remoteip=<client-ip>
   ```
   The `remoteip` field is optional but recommended — it lets Cloudflare cross-check the token against the IP that solved the challenge.
3. **Parse the JSON response** — the `success` field is a boolean; if `false`, inspect `error-codes` for logging (`timeout-or-duplicate`, `invalid-input-response`, etc.) and reject with `400` + error code `CAPTCHA_INVALID`.
4. **Only then**, proceed with the existing sign-up logic (validation, user creation, verification e-mail).

Tokens are **single-use** and expire ~5 minutes after issuance — do not cache them, do not accept the same token twice.

Recommended: extract this into a small middleware/service (`verifyTurnstileToken(token, ip)`) so it can be reused later on `/login` and `/forgotPassword`.

### Error handling

- **Cloudflare unreachable / 5xx**: log and reject with `503` (do not fail open — a bot could exploit a Cloudflare outage to flood us). If this becomes a real availability concern, we can revisit and fail open with elevated rate-limits, but not by default.
- **Timeout on the outbound call**: set an aggressive timeout (~3 seconds); same handling as unreachable.

## 2. Honeypot field

The front will include a hidden input in the form (proposed name: `website`). Real users will never fill it; naive scrapers fill every field.

- If `body.website` is present **and non-empty**, reject with `400` — but **return the same generic success shape** as a valid sign-up would (or a plausible generic 400) so that the bot cannot easily learn which field is the honeypot.
- Log the attempt with the client IP for monitoring.
- Absolutely trivial to implement (a few lines) and catches the dumbest wave "for free".

## 3. Rate limiting on `/signup`

Independently of Turnstile, throttle abusive callers.

Suggested policy (tune with real traffic data):

| Key | Window | Limit |
| --- | --- | --- |
| Client IP | 10 min | 5 sign-up attempts |
| Client IP | 24 h | 20 sign-up attempts |
| E-mail domain (part after `@`) | 1 h | 10 sign-ups from the same domain |

On limit hit, return `429 Too Many Requests` with a `Retry-After` header.

### Implementation notes (framework-agnostic)

The grottocenter-api historical stack is **Sails.js / Node.js**; the common libraries there are:

- [`express-rate-limit`](https://github.com/express-rate-limit/express-rate-limit) — simplest, in-memory or Redis-backed
- [`rate-limiter-flexible`](https://github.com/animir/node-rate-limiter-flexible) — more knobs, supports multiple keys

If the API is behind Cloudflare / a reverse proxy, make sure to:

- Trust the proxy so `req.ip` returns the real client IP (`app.set('trust proxy', ...)` in Express, or the Sails equivalent), otherwise the limiter buckets everyone under the proxy's IP.
- Prefer the `CF-Connecting-IP` header if the API sits behind Cloudflare directly.

For persistence across API instances (if the API runs in more than one pod/process), the limiter store must be shared — Redis is the pragmatic choice.

## Response contract with the front-end

To keep the front simple and translatable, agree on machine-readable error codes:

| HTTP status | Body `code` | Meaning | Front-end handling |
| --- | --- | --- | --- |
| `400` | `CAPTCHA_MISSING` | No token in payload | Show a generic "please retry the challenge" message |
| `400` | `CAPTCHA_INVALID` | Token rejected by Cloudflare | Reset the widget, ask user to retry |
| `429` | `RATE_LIMITED` | Too many attempts | Show a "try again later" message; respect `Retry-After` |
| `400` | existing validation codes | Unchanged | Unchanged |

The existing `500` fallback keeps its current meaning.

## Testing

- Unit tests around `verifyTurnstileToken` using Cloudflare's [test keys](https://developers.cloudflare.com/turnstile/troubleshooting/testing/) — always-pass secret `1x0000000000000000000000000000000AA` and always-fail secret `2x0000000000000000000000000000000AA`.
- Integration test on `/signup`:
  - happy path with the always-pass secret
  - missing token → `400 CAPTCHA_MISSING`
  - invalid token → `400 CAPTCHA_INVALID`
  - honeypot filled → rejected
  - rate limit exceeded → `429 RATE_LIMITED`
- Ensure e-mail verification e-mails are **not** sent when the request is rejected upstream.

## Rollout

1. Deploy with `TURNSTILE_ENABLED=false` first — verify no regression on legitimate sign-ups.
2. Flip `TURNSTILE_ENABLED=true` in staging, run manual checks with the front branch that emits the token.
3. Flip in production. Monitor:
   - sign-up conversion rate
   - count of `CAPTCHA_INVALID` rejections (proxy for bot volume caught)
   - count of `RATE_LIMITED` rejections
   - Cloudflare Turnstile dashboard (challenges served, solve rate)

## Follow-ups (out of scope for this ticket)

- Extend `verifyTurnstileToken` to `POST /login` (protects against credential stuffing) and `POST /forgotPassword` (protects against user enumeration & mail bombing).
- Consider adding a background job to purge unverified accounts older than N days — bots that squeaked through email verification would leave dangling rows otherwise.

## References

- Front-end companion: https://github.com/GrottoCenter/grottocenter-front/issues/1462
- Cloudflare Turnstile docs: https://developers.cloudflare.com/turnstile/
- Turnstile server-side validation: https://developers.cloudflare.com/turnstile/get-started/server-side-validation/
- Turnstile test keys: https://developers.cloudflare.com/turnstile/troubleshooting/testing/

ClemRz:
## 🤔 What

Add two defense layers to `POST /api/v1/signup` to block automated bot registrations:
- **Honeypot field rejection** — catches naive bots that auto-fill a hidden `website` field
- **Cloudflare Turnstile token verification** — validates the front-end CAPTCHA token via the Cloudflare siteverify API

Closes #1727

## 🤷‍♂️ Why

The signup endpoint is being abused by bots creating fake accounts in bulk. The front-end companion ticket introduces a Turnstile widget, but the back-end must independently verify the token — otherwise bots can POST directly to the endpoint.

## 🔍 How

- **Defense ordering**: honeypot first (zero-cost string check), then Turnstile (external API call). Each layer short-circuits on rejection.
- **HoneypotGuard** (`api/services/HoneypotGuard.js`): stateless check using existing `isNonBlankString` utility. Returns deceptive 204 on trap (bot thinks signup succeeded).
- **TurnstileService** (`api/services/TurnstileService.js`): reusable service with `isEnabled()`, `validateConfig()`, and `verifyToken()`. Uses native `fetch` with `AbortSignal.timeout(3000)`. Fail-closed design (any error → `CAPTCHA_SERVICE_UNAVAILABLE`).
- **Bootstrap validation** (`config/bootstrap.js`): app refuses to start if `TURNSTILE_ENABLED=true` without `TURNSTILE_SECRET_KEY`. Logs warning when disabled.
- **Controller preamble** (`api/controllers/v1/auth/sign-up.js`): inserts both checks before existing logic. Existing signup flow unchanged.
- Configurable via `TURNSTILE_ENABLED` and `TURNSTILE_SECRET_KEY` env vars.

## 🧪 Testing

- Property tests (6 properties, `fast-check` with 100 runs each)
- Unit tests for both services (HoneypotGuard: 9, TurnstileService: 18)
- Integration tests via supertest (7 route-level tests covering defense ordering, rejections, and happy paths)
- All anti-bot tests pass. Full suite: same 17 pre-existing failures unrelated to this feature (now fixed — `global.sails` restoration).

To run just the anti-bot tests:
```bash
npm run test -- --grep "HoneypotGuard|TurnstileService|anti-bot|Property 5"
```

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,30 @@ Example:
 docker exec grotto-postgres psql -U root -d grottoce -c "SELECT id, id_country, iso_3166_2 FROM t_entrance WHERE id_country = 'US' LIMIT 5;"
 ```
 
+### Cloudflare Turnstile (anti-bot CAPTCHA)
+
+The signup endpoint is protected by Cloudflare Turnstile. Turnstile requires a domain registered in the Cloudflare dashboard, so it **cannot be tested end-to-end on localhost with real keys** — the widget will fail to load and the front-end won't be able to generate a token.
+
+Two options for local development:
+
+**Option 1 — disable Turnstile** (CAPTCHA check is skipped entirely):
+
+Set in `docker/.env`:
+```
+TURNSTILE_ENABLED=false
+```
+
+**Option 2 — use Cloudflare's always-pass test keys** (widget renders and tokens always validate):
+
+Set in `docker/.env`:
+```
+TURNSTILE_ENABLED=true
+TURNSTILE_SECRET_KEY=1x0000000000000000000000000000000AA
+```
+And use site key `1x0000000000000000000000000000000AA` on the front-end side.
+
+For staging and production, `TURNSTILE_ENABLED=true` must be set alongside the real `TURNSTILE_SECRET_KEY`. The app refuses to start if `TURNSTILE_ENABLED=true` without a secret key configured.
+
 ### Refresh materialized views (development only)
 
 The database uses materialized views for performance optimization. These views need to be refreshed periodically to reflect the latest data. In development, you can manually refresh them using the provided cron scripts:
--- a/README.md
+++ b/README.md
@@ -119,23 +119,7 @@ docker exec grotto-postgres psql -U root -d grottoce -c "SELECT id, id_country, 
 
 The signup endpoint is protected by Cloudflare Turnstile. Turnstile requires a domain registered in the Cloudflare dashboard, so it **cannot be tested end-to-end on localhost with real keys** — the widget will fail to load and the front-end won't be able to generate a token.
 
-Two options for local development:
-
-**Option 1 — disable Turnstile** (CAPTCHA check is skipped entirely):
-
-Set in `docker/.env`:
-```
-TURNSTILE_ENABLED=false
-```
-
-**Option 2 — use Cloudflare's always-pass test keys** (widget renders and tokens always validate):
-
-Set in `docker/.env`:
-```
-TURNSTILE_ENABLED=true
-TURNSTILE_SECRET_KEY=1x0000000000000000000000000000000AA
-```
-And use site key `1x0000000000000000000000000000000AA` on the front-end side.
+See `docker/sample.env` for the two local development options (disable entirely, or use Cloudflare's always-pass test keys). The environment variables must be **exported in your shell** before running `npm run dev` — they are not read from `docker/.env`, which is only used by docker-compose for the database and search containers.
 
 For staging and production, `TURNSTILE_ENABLED=true` must be set alongside the real `TURNSTILE_SECRET_KEY`. The app refuses to start if `TURNSTILE_ENABLED=true` without a secret key configured.
 
--- a/api/controllers/v1/auth/sign-up.js
+++ b/api/controllers/v1/auth/sign-up.js
@@ -1,8 +1,46 @@
 const AuthService = require('../../../services/AuthService');
 const CaverService = require('../../../services/CaverService');
+const HoneypotGuard = require('../../../services/HoneypotGuard');
 const LanguageService = require('../../../services/LanguageService');
+const TurnstileService = require('../../../services/TurnstileService');
 
 module.exports = async (req, res) => {
+  // --- Anti-bot defense layers (order matters) ---
+
+  // Layer 1: Honeypot
+  const honeypotResult = HoneypotGuard.check(req.body);
+  if (honeypotResult.trapped) {
+    sails.log.warn('[AntiBot:Honeypot] Bot trapped', {
+      ip: req.ip,
+      website: honeypotResult.value,
+    });
+    return res.ok(); // Deceptive response — res.ok() returns 204 to mimic normal signup success
+  }
+
+  // Layer 2: Turnstile
+  if (TurnstileService.isEnabled()) {
+    const turnstileResult = await TurnstileService.verifyToken(
+      req.body.captchaToken,
+      req.ip
+    );
+    if (!turnstileResult.pass) {
+      sails.log.warn('[AntiBot:Turnstile] Rejected', {
+        ip: req.ip,
+        errorCode: turnstileResult.errorCode,
+      });
+      const statusMap = {
+        CAPTCHA_MISSING: 400,
+        CAPTCHA_INVALID: 400,
+        CAPTCHA_SERVICE_UNAVAILABLE: 503,
+      };
+      return res.status(statusMap[turnstileResult.errorCode] || 500).json({
+        error: turnstileResult.errorCode,
+      });
+    }
+  }
+
+  // --- Existing signup logic (unchanged) ---
+
   // Check params
   let email = req.param('email');
   if (!email || !CaverService.isARealCaver(email)) {
--- a/api/controllers/v1/auth/sign-up.js
+++ b/api/controllers/v1/auth/sign-up.js
@@ -8,11 +8,11 @@ module.exports = async (req, res) => {
   // --- Anti-bot defense layers (order matters) ---
 
   // Layer 1: Honeypot
-  const honeypotResult = HoneypotGuard.check(req.body);
+  const honeypotResult = HoneypotGuard.check(req.allParams());
   if (honeypotResult.trapped) {
     sails.log.warn('[AntiBot:Honeypot] Bot trapped', {
       ip: req.ip,
-      website: honeypotResult.value,
+      website: String(honeypotResult.value).slice(0, 200),
     });
     return res.ok(); // Deceptive response — res.ok() returns 204 to mimic normal signup success
   }
@@ -20,7 +20,7 @@ module.exports = async (req, res) => {
   // Layer 2: Turnstile
   if (TurnstileService.isEnabled()) {
     const turnstileResult = await TurnstileService.verifyToken(
-      req.body.captchaToken,
+      req.param('captchaToken'),
       req.ip
     );
     if (!turnstileResult.pass) {
@@ -28,14 +28,10 @@ module.exports = async (req, res) => {
         ip: req.ip,
         errorCode: turnstileResult.errorCode,
       });
-      const statusMap = {
-        CAPTCHA_MISSING: 400,
-        CAPTCHA_INVALID: 400,
-        CAPTCHA_SERVICE_UNAVAILABLE: 503,
-      };
-      return res.status(statusMap[turnstileResult.errorCode] || 500).json({
-        error: turnstileResult.errorCode,
-      });
+      if (turnstileResult.errorCode === 'CAPTCHA_SERVICE_UNAVAILABLE') {
+        return res.status(503).json({ error: turnstileResult.errorCode });
+      }
+      return res.badRequest({ error: turnstileResult.errorCode });
     }
   }
 
--- a/api/services/HoneypotGuard.js
+++ b/api/services/HoneypotGuard.js
@@ -1,0 +1,26 @@
+/**
+ * HoneypotGuard.js
+ *
+ * @description :: Stateless utility that inspects the request body for a
+ *                 hidden decoy `website` field. Bots that auto-fill all form
+ *                 fields will populate this invisible field, allowing cheap
+ *                 detection without external API calls.
+ */
+
+const isNonBlankString = require('../utils/isNonBlankString');
+
+module.exports = {
+  /**
+   * Check whether the honeypot field indicates bot activity.
+   *
+   * @param {Object} body - The parsed request body
+   * @returns {{ trapped: boolean, value: string|undefined }}
+   */
+  check(body) {
+    const value = body && body.website;
+    if (isNonBlankString(value)) {
+      return { trapped: true, value };
+    }
+    return { trapped: false };
+  },
+};
--- a/api/services/HoneypotGuard.js
+++ b/api/services/HoneypotGuard.js
@@ -17,7 +17,12 @@ module.exports = {
    * @returns {{ trapped: boolean, value: string|undefined }}
    */
   check(body) {
-    const value = body && body.website;
+    const rawValue = body && body.website;
+    if (rawValue === undefined || rawValue === null) {
+      return { trapped: false };
+    }
+    // Coerce to string so non-string types (arrays, objects) also trigger the trap
+    const value = String(rawValue);
     if (isNonBlankString(value)) {
       return { trapped: true, value };
     }
--- a/api/services/TurnstileService.js
+++ b/api/services/TurnstileService.js
@@ -61,6 +61,12 @@ module.exports = {
       return { pass: false, errorCode: 'CAPTCHA_MISSING' };
     }
 
+    // Cloudflare Turnstile tokens are bounded at ~2 KB; reject oversized values
+    // before making an outbound call — cheap defense against log/payload abuse.
+    if (token.length > 2048) {
+      return { pass: false, errorCode: 'CAPTCHA_INVALID' };
+    }
+
     const secret = process.env.TURNSTILE_SECRET_KEY;
 
     try {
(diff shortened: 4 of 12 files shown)


ClemRz:
Réponse au commentaire de Paul concernant le test en localhost :

C'est un comportement attendu — Cloudflare Turnstile exige un domaine enregistré dans le dashboard Cloudflare pour émettre des tokens. Sur `localhost`, le widget ne peut pas se charger avec de vraies clés (d'où la capture d'écran avec l'erreur Cloudflare), et l'API renvoie bien 400 (`CAPTCHA_MISSING`) puisque aucun token n'est transmis.

Pour tester en local, deux options sont documentées dans `docker/sample.env` et dans le README (commit poussé à l'instant) :

**Option 1 — désactiver Turnstile** (la vérification CAPTCHA est entièrement ignorée) :
```
TURNSTILE_ENABLED=false
```

**Option 2 — utiliser les clés de test Cloudflare** (always-pass, utilisables sur n'importe quel domaine) :
```
TURNSTILE_ENABLED=true
TURNSTILE_SECRET_KEY=1x0000000000000000000000000000000AA
```
Côté front, utiliser la site key `1x0000000000000000000000000000000AA` (ou `1x00000000000000000000AA` — les deux sont valides selon la doc Cloudflare).

Le staging reste l'environnement adapté pour un test end-to-end avec de vraies clés.

Paul-AUB:
### Issues (Must Fix)

1. **[README.md:29 & docker/sample.env:25]** The always-pass **site key** documented for the front-end is wrong. Cloudflare's always-pass **secret** is `1x0000000000000000000000000000000AA`, but the always-pass **site key** is `1x00000000000000000000AA` (see https://developers.cloudflare.com/turnstile/troubleshooting/testing/). As written, a front-end developer copy-pasting the value will get a widget that fails to load and never issues a token, defeating the whole "Option 2" workflow. Suggested wording:

    ```
    TURNSTILE_ENABLED=true
    TURNSTILE_SECRET_KEY=1x0000000000000000000000000000000AA
    # Pair with front-end site key: 1x00000000000000000000AA
    ```

2. **[README.md:19 & docker/sample.env:14-15]** The docs tell the reader to set `TURNSTILE_ENABLED` / `TURNSTILE_SECRET_KEY` in `docker/.env`, but that file is only consumed by `docker-compose` for the DB and Typesense containers (see [scripts/developmentSetup.js:63-66](scripts/developmentSetup.js#L63-L66)). The Sails process launched by `npm run dev` runs on the host, has no `dotenv`/`--env-file` hook, and reads `process.env` from the shell only. As a result, editing `docker/.env` has no effect on `TurnstileService.isEnabled()` / `validateConfig()`. Either (a) reword the docs to say "export these in your shell before `npm run dev`", or (b) add a dotenv loader (e.g. `node -r dotenv/config app.js`) that reads `docker/.env` — otherwise anyone following the README will see the bootstrap keep failing or Turnstile stay in whatever state their shell dictates.

3. **[api/controllers/v1/auth/sign-up.js:11,23]** The rest of this controller reads inputs via `req.param('...')` (lines 45, 56, 65, 76, 97, 103); the new anti-bot layer instead reaches directly into `req.body`. `req.param()` merges body/query/route params and is the local convention. Consider switching to `req.param('captchaToken')` and passing `req.allParams()` (or `req.body || {}`) to `HoneypotGuard.check()` for consistency and robustness against query-string clients.

4. **[api/controllers/v1/auth/sign-up.js:36-38]** The Turnstile rejection returns `{ error: 'CAPTCHA_MISSING' }` via `res.status().json()`, while every other error in the same controller returns `{ message: '...' }` via `res.badRequest(...)`. That divergence forces the front-end to key off two response shapes for the same endpoint, and it also bypasses [api/responses/badRequest.js](api/responses/badRequest.js) / `sails.helpers.logResponse` used everywhere else. Consider either passing the code through `res.badRequest({ error: 'CAPTCHA_MISSING' })` (Sails will still emit it as 400 with `formatResponseData`) or documenting the shape explicitly on the OpenAPI spec.

5. **[config/bootstrap.js:117]** `TurnstileService.validateConfig()` is called at the very end of `bootstrap()`, after `dbSync.ensureSearchDbIsPopulated()`, blacklist cache, enrichment queue, CSV import queue, and two coordinate-snapshot loads. If Turnstile is misconfigured, the app crashes after a full slow startup. Consider hoisting the call to the top of the function (before any I/O) so misconfig fails fast.

6. **[api/services/TurnstileService.js:82-93]** The outbound POST forwards `token` verbatim to Cloudflare with no upper bound (only the Sails body-parser limit gates it). Cloudflare Turnstile tokens are bounded at ~2 KB. Consider a fast-path `token.length > 2048 → CAPTCHA_INVALID` before the outbound call, to make an obvious abuse pattern cheap to reject.

7. **[api/controllers/v1/auth/sign-up.js:31-38]** `statusMap` already contains an entry for every value `TurnstileService.verifyToken()` can return, so `|| 500` is unreachable. Either drop it, or log a warning when an unknown code is seen so future additions to the service don't silently degrade to opaque 500s.

### Nitpicks (Optional)

8. **[api/controllers/v1/auth/sign-up.js:13-16]** `sails.log.warn('[AntiBot:Honeypot] Bot trapped', { …, website: honeypotResult.value })` logs an attacker-controlled string of unbounded length (up to the body-parser cap). Consider truncating to ~200 chars before logging (`value: honeypotResult.value.slice(0, 200)`) — cheap defense against log-flood abuse.

9. **[docker/sample.env:23-30]** `TURNSTILE_ENABLED=false` appears once inside the comment block ("Option 1 — disable entirely") and once as an active setting at the bottom. The commented one is redundant with the active line; consider dropping it to avoid future drift.

10. **[test/integration/1_services/HoneypotGuard.property.test.js:19,44 & TurnstileService.property.test.js:37,71,150,181,201 & sign-up.property.test.js:35]** The `describe` titles all begin with `Feature: signup-anti-bot-protection, Property N: …`. That prefix is repeated on every describe and is very noisy in the mocha reporter. Consider nesting the same way [test/integration/4_routes/Auth/sign-up-anti-bot.test.js](test/integration/4_routes/Auth/sign-up-anti-bot.test.js) already does: `describe('signup-anti-bot-protection', () => { describe('Property N: …', () => { it(...) }) })`.

11. **[README.md:9-32]** The block duplicates verbatim the block in [docker/sample.env:14-30](docker/sample.env#L14-L30). One of them is enough; consider having README link to `docker/sample.env` (or vice versa) to prevent drift.

12. **[api/services/TurnstileService.js:65-93]** The 4xx-with-`success:false` branch (`if (!response.ok)` → try/catch parse) exists for a theoretical case Cloudflare's siteverify doesn't actually document (invalid tokens come back as 200 with `success:false`). Consider dropping that branch and treating any non-2xx as `CAPTCHA_SERVICE_UNAVAILABLE` — simpler, and there's no test that currently exercises the 4xx-with-body path anyway.

ClemRz:
All 12 items addressed in 71d8b67a. Summary:

**Must Fix**
1. ✅ Correct always-pass site key in `docker/sample.env` and `README.md` — front-end site key is now `1x00000000000000000000AA`, secret key unchanged at `1x0000000000000000000000000000000AA`
2. ✅ Docs now clarify that `TURNSTILE_ENABLED` / `TURNSTILE_SECRET_KEY` must be **exported in the shell** before `npm run dev` — `docker/.env` is only consumed by docker-compose. `README.md` deduplication (item 11) folded into same fix: the Turnstile section now links to `docker/sample.env` instead of duplicating the block.

**Suggestions**
3. ✅ `req.body` → `req.allParams()` / `req.param('captchaToken')` throughout the anti-bot preamble
4. ✅ CAPTCHA 400 rejections now go through `res.badRequest({ error: ... })` (same JSON shape, goes through `formatResponseData` + `logResponse`). 503 stays as `res.status(503).json()` — no custom 503 response exists in the project.
5. ✅ `TurnstileService.validateConfig()` hoisted to the very top of `bootstrap()`, before any I/O
6. ✅ `token.length > 2048 → CAPTCHA_INVALID` guard added before the outbound fetch
7. ✅ `statusMap` + unreachable `|| 500` removed; replaced with an explicit `if (errorCode === 'CAPTCHA_SERVICE_UNAVAILABLE')` branch
12. ✅ Non-2xx handling in `verifyToken` collapsed to a single `CAPTCHA_SERVICE_UNAVAILABLE` return (the 4xx-with-body parse branch was dead code)

**Nitpicks**
8. ✅ Honeypot log value truncated to 200 chars before logging
9. ✅ Redundant `TURNSTILE_ENABLED=false` line in the `sample.env` comment block removed
10. ✅ Property test `describe` titles restructured: outer `signup-anti-bot-protection` / inner `Property N: …` across all three property test files. The mock `req` in `sign-up.property.test.js` also gained `allParams: () => req.body` to align with the controller change.

Paul-AUB (dismissed):
Re-reviewed after commit 71d8b67a. All 12 items from my previous review round are properly addressed:

- Always-pass site key corrected in `README.md` / `docker/sample.env` (item 1)
- Docs now clarify env vars must be exported in the shell, not read from `docker/.env`, and the README/sample.env duplication is resolved via a link instead (items 2, 11)
- `req.body` → `req.allParams()` / `req.param('captchaToken')` for consistency with the rest of the controller (item 3)
- CAPTCHA 400s now go through `res.badRequest(...)` (item 4)
- `TurnstileService.validateConfig()` hoisted to the top of `bootstrap()`, before any I/O (item 5)
- Token length guard (`> 2048` → `CAPTCHA_INVALID`) added before the outbound fetch (item 6)
- Dead `statusMap` / unreachable `|| 500` replaced with an explicit branch (item 7)
- Honeypot log value truncated to 200 chars (item 8)
- Redundant `TURNSTILE_ENABLED=false` line removed from the sample.env comment block (item 9)
- Property test `describe` titles restructured to nest under `signup-anti-bot-protection` (item 10)
- Dead 4xx-with-body parsing branch in `verifyToken` collapsed into the single non-2xx → `CAPTCHA_SERVICE_UNAVAILABLE` path (item 12)

I did not find any new blocking issues on this pass.

1. **[api/services/HoneypotGuard.js:20-21]** `isNonBlankString` requires `typeof val === 'string'`, so a `website` value sent as a non-string JSON type (e.g. `{"website": ["spam"]}` or `{"website": {"a": 1}}`) silently bypasses the honeypot instead of being trapped, since `HoneypotGuard.check` only reads `body.website` without coercing it to a string first. This is a minor gap given Turnstile is the primary defense layer, but a trivial `String(value)` coercion (mirroring what the controller already does when logging, line 15 of `sign-up.js`) would close it.

### Nitpicks (Optional)

2. **[api/controllers/v1/auth/sign-up.js:13,27]** Passing a plain object as the second argument to `sails.log.warn(...)` is a new pattern in this codebase — the only other `sails.log.warn` call with extra context (`api/dbSync/dbSync.js:176`) uses a single formatted string. Not a functional issue (Sails' default logger will just stringify/inspect the object), but worth confirming it renders as expected in whatever log aggregation is used in production.

Note: I was unable to run the test suite locally in this pass (`node_modules` not installed in the review environment), so this review is based on static inspection of the diff and surrounding code, not a live test run.

ClemRz:
Second-round review addressed in cd4d8282.

### Suggestion 1 — Non-string `website` values bypass honeypot (fixed)

`HoneypotGuard.check()` now coerces `rawValue` to `String` before calling `isNonBlankString`, so payloads like `{"website": ["spam"]}` or `{"website": {"a": 1}}` are trapped instead of silently passing through. `null` and `undefined` are short-circuited before the coercion to preserve their existing "not trapped" semantics.

4 new unit tests added: array value, plain object value, `null`, and `undefined`.

### Suggestion 2 — `sails.log.warn(msg, obj)` pattern (no change)

Sails' Winston-backed logger handles the second argument correctly — it's inspected and appended to the log line. No functional issue, no change needed.

PR #1741, opened 2026-07-26
fix(csvImport): fix batch status=failed on successful import

ClemRz:
## Summary

Fixes the bug reported in [grottocenter-front#1464](https://github.com/GrottoCenter/grottocenter-front/pull/1464) where a successful CSV entrance import displayed `status: "failed"` with the message "L'import a échoué. 0 lignes n'ont pas pu être importées."

Production logs (July 23–24) confirmed two root causes.

### Root cause 1 — `processOneChunk` crashes after a successful import

```
TypeError: Cannot read properties of undefined (reading 'catch')
    at Object.processOneChunk (.../CSVImportQueueService.js:348:11)
```

`sails.services.coordinatessnapshotservice` is `undefined` in the pg-boss worker context (workers run outside Sails' normal request lifecycle). The crash happens after rows have already been imported, causing the pg-boss job to be marked `failed` despite successful work. The fix replaces the `sails.services.*` reference with a direct `require()`, consistent with all other service calls in `processOneChunk`.

### Root cause 2 — `uploadReport` crashes on SAS token generation

```
TypeError: Invalid sharedKeyCredential, userDelegationKey or accountName.
    at generateBlobSASQueryParametersInternal (.../BlobSASSignatureValues.js:50:11)
    at Object.uploadReport (.../CSVImportQueueService.js:638:22)
```

`generateBlobSASQueryParameters` performs an `instanceof` check on the credential. When `FileService` is required across different Node require-cache contexts (which happens with Sails' module loader), the `StorageSharedKeyCredential` instance fails the check even though the object is structurally valid. The documents container already allows anonymous read access (noted in the original code comment), so SAS tokens were never necessary here. The fix drops SAS generation entirely and returns the plain blob URL directly, removing the fragile credential dependency.

### Why the batch showed `status: "failed"` despite `successes: 1`

Because `aggregateBatch` caught the `uploadReport` error and set `status = "failed"` without persisting the result payload. `getBatchProgress` independently queries pg-boss job outputs and correctly shows `successes: 1` — the two fields were sourced independently and could diverge. A third commit addresses this: report upload failures no longer set `status: "failed"`; the batch is marked `completed` with `reportUrls: null` so the front-end always receives accurate row counts.

## What

Three commits:

1. **`aggregateBatch` resilience**: when `generateAndUploadReports` throws, fall back to `reportUrls: null` and mark the batch `completed` instead of `failed`. Summary counts are always persisted.
2. **`processOneChunk` crash fix**: replace `sails.services.coordinatessnapshotservice` with `require('./CoordinatesSnapshotService')`.
3. **`uploadReport` SAS fix**: drop `generateBlobSASQueryParameters` entirely. Return `blockBlobClient.url` directly. Remove now-unused `BlobSASPermissions` and `generateBlobSASQueryParameters` imports.

New unit tests cover the `aggregateBatch` upload-failure path and happy path.

## Related

- Front-end bug report: [grottocenter-front#1464 (comment)](https://github.com/GrottoCenter/grottocenter-front/pull/1464#issuecomment-5064689743)
- Original async import PR: [grottocenter-api#1708](https://github.com/GrottoCenter/grottocenter-api/pull/1708)

### DIFF
--- a/api/services/CSVImportQueueService.js
+++ b/api/services/CSVImportQueueService.js
@@ -505,14 +505,13 @@ module.exports = {
       );
     } catch (err) {
       sails.log.error(
-        `CSVImportQueueService: aggregation failed for batch ${batchId}:`,
+        `CSVImportQueueService: report generation failed for batch ${batchId} (rows were already imported):`,
         err
       );
-      await TJobBatch.updateOne({ id: batchId }).set({
-        status: 'failed',
-        completedAt: new Date(),
-      });
-      return;
+      // Report upload failed, but the rows were already imported successfully.
+      // Mark the batch as completed with null reportUrls so the front-end
+      // can show accurate import counts instead of a misleading "failed" status.
+      reportUrls = { successes: null, duplicates: null, failures: null };
     }
 
     const resultPayload = {
--- a/api/services/CSVImportQueueService.js
+++ b/api/services/CSVImportQueueService.js
@@ -530,45 +530,57 @@ module.exports = {
 
   /**
    * Generate CSV reports and upload to Azure Blob Storage.
+   * Each report is uploaded independently; a failure on one does not prevent
+   * the others from being uploaded. URLs for successfully uploaded reports are
+   * returned; failed ones stay null and a warning is logged.
    * @param {string} batchId
    * @param {Array} successes
    * @param {Array} duplicates
    * @param {Array} failures
-   * @returns {Object} { success: url|null, duplicates: url|null, failures: url|null }
+   * @returns {Object} { successes: url|null, duplicates: url|null, failures: url|null }
    */
   async generateAndUploadReports(batchId, successes, duplicates, failures) {
     const reportUrls = { successes: null, duplicates: null, failures: null };
 
     const prefix = `csv-import-reports/${batchId}`;
 
-    if (successes.length > 0) {
-      const csv = module.exports.toCSV(
-        ['line', 'caveId', 'entranceId', 'latitude', 'longitude'],
-        successes
-      );
-      reportUrls.successes = await module.exports.uploadReport(
-        prefix,
-        'successes.csv',
-        csv
-      );
-    }
+    const uploads = [
+      {
+        key: 'successes',
+        rows: successes,
+        columns: ['line', 'caveId', 'entranceId', 'latitude', 'longitude'],
+        filename: 'successes.csv',
+      },
+      {
+        key: 'duplicates',
+        rows: duplicates,
+        columns: ['line', 'message'],
+        filename: 'duplicates.csv',
+      },
+      {
+        key: 'failures',
+        rows: failures,
+        columns: ['line', 'message'],
+        filename: 'failures.csv',
+      },
+    ];
 
-    if (duplicates.length > 0) {
-      const csv = module.exports.toCSV(['line', 'message'], duplicates);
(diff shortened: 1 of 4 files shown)


Paul-AUB (dismissed):

1. **[api/services/CSVImportQueueService.js:539-575]** `generateAndUploadReports` uploads `successes.csv`, `duplicates.csv`, and `failures.csv` sequentially and returns a single object only on full success. If, say, the successes upload succeeds but the duplicates upload then throws, the whole call rejects and `aggregateBatch`'s catch block discards the already-uploaded `successes` URL, replacing it with `null` (line 510). The successes report blob still exists in storage but its URL is lost to the front-end/notification. Not a regression introduced by this PR, but since this is the exact code path being hardened here, it might be worth having `generateAndUploadReports` catch upload errors per-report (returning whatever URLs did succeed) rather than all-or-nothing.

2. **[api/services/CSVImportQueueService.js:601-611]** `FileService.getSharedKeyCredential()` and the `isLocalStub` dev-mode branch are removed from `uploadReport`, but the `getSharedKeyCredential` method itself is left in `FileService.js` and now appears unused anywhere in the codebase. Consider removing it (or confirming another caller still needs it) to avoid dead code.

### Nitpicks (Optional)

3. **[api/services/CSVImportQueueService.js:539]** The JSDoc `@returns` tag above `generateAndUploadReports` still describes `{ success: url|null, ... }` (line 537) — the key is actually `successes`, not `success`. Pre-existing typo, unrelated to this PR, but easy to fix while in the area.

---

Verified the two root-cause fixes (direct `require` for `CoordinatesSnapshotService` instead of `sails.services.*`, and dropping SAS token generation in favor of the plain blob URL) are consistent with the existing pattern already used elsewhere in this file (e.g. `EntranceCSVImportService`, `CaveService`, `LanguageService` are all required the same way). The `aggregateBatch` resilience fix correctly ensures the batch is marked `completed` (not `failed`) when only report upload fails, while the pre-existing `allFailed`-chunks branch in `checkBatchCompletion` correctly continues to use `status: 'failed'` for genuine import failures — this PR doesn't regress that distinction. New unit tests cover both the upload-failure and happy-path branches of `aggregateBatch` and follow the existing sinon stub/restore conventions used throughout the test file.

I wasn't able to run the test suite locally (it requires a live Postgres connection for the Sails bootstrap, unavailable in this environment), so testing was done by manual code reading rather than execution.

ClemRz:
All three suggestions addressed in the latest commit:

1. **Per-report error isolation** — `generateAndUploadReports` now wraps each upload in its own `try/catch`. A failure on `duplicates.csv` no longer discards a successfully uploaded `successes.csv` URL; it logs a warning and leaves that key as `null` while the others are returned normally.

2. **Dead code removed** — `getSharedKeyCredential` deleted from `FileService.js`. Confirmed no other callers exist (`sharedKeyCredential` and the related Azure SDK imports stay — still used by `getSignedReadUrl` for the private db-exports container).

3. **JSDoc typo fixed** — `{ success: url|null }` → `{ successes: url|null }` in the `@returns` tag of `generateAndUploadReports`.

Also fixed a pre-existing `func-names` lint warning in `CSVImportQueueService.property.test.js` (unnamed `function` on `describe` block) — switched to arrow function on `describe` and chained `.timeout(30000)` directly on the `it()` call, which is the correct pattern.

Issue #1737, opened 2026-07-23
bug(api): [23503] tgrotto - FK violation j_document_grotto_author on permanent delete


ClemRz:
**Describe the bug**
Permanently deleting an organization (grotto) fails with a 500 error due to a foreign key violation. The `DELETE /api/v1/organizations/:id?isPermanent=1` endpoint cleans up several related tables (documents, explored caves, notifications, names) but does NOT reassign or remove rows from `j_document_grotto_author` before calling `TGrotto.destroyOne()`. PostgreSQL then rejects the delete because `j_document_grotto_author` still references the organization.

**To Reproduce**
Based on the trace logs:
1. Authenticate as an Administrator/Moderator
2. `DELETE /api/v1/organizations/1836?isPermanent=1`
3. The organization has documents linked via `j_document_grotto_author`
4. The API returns 500

**Input Data**
The following input triggered the error:

- **Endpoint:** `DELETE /api/v1/organizations/1836?isPermanent=1`
- **Request Body:**
```json
{}
```
- **URL/Query Parameters:** `id=1836`, `isPermanent=1`
- **User:** ID 60181 (Paul AUBERTIN), roles: Administrator, Moderator, Leader

**Expected behavior**
The API should handle related `j_document_grotto_author` rows (reassign to merge target or remove) before permanently deleting the organization, and return a successful response.

**Error Details**
- Error Type: `OperationalError [AdapterError]`
- Occurrences: 1
- Time Range: 2026-07-23T00:40:04Z to 2026-07-23T00:40:04Z
- Representative Trace ID: `019f8c6a-4846-7183-b5bb-b1feed3d1673`
- Stack Trace:
```
OperationalError [AdapterError]: Unexpected error from database adapter: update or delete on table "t_grotto" violates foreign key constraint "j_document_grotto_author_t_caver_fk" on table "j_document_grotto_author"
    at _afterCounting (waterline/lib/waterline/methods/destroy-one.js:167:17)
    ...
    adapterMethodName: 'destroy',
    modelIdentity: 'tgrotto',
    raw: {
      code: '23503',
      detail: 'Key (id)=(1836) is still referenced from table "j_document_grotto_author".',
      constraint: 'j_document_grotto_author_t_caver_fk',
      table: 'j_document_grotto_author'
    }
```

**Additional context**
- Service: `api/controllers/v1/organization/delete.js`
- Model: `tgrotto`
- DB Error Code: `23503`
[...]

PR #1743, opened 2026-07-28
fix(organizations): re-point explored caves on duplicate merge

### LINKED ISSUE #1795: Merging duplicate organizations silently destroys the deleted org's explored caves
**Describe the bug**

When two organizations are duplicates, a moderator permanently deletes one and passes `entityId` to merge it into the survivor. Documents, `editor` / `library` references and `redirectTo` pointers are all re-pointed to the survivor — but **explored caves are not**.

`api/controllers/v1/organization/delete.js` runs `JGrottoCaveExplorer.destroy({ grotto: organizationId })` unconditionally, so the deleted organization's `j_grotto_cave_explorer` rows are dropped even on a merge. The result is silent data loss: merging two duplicate records destroys the exploration history of the one being deleted, even though both describe the same real-world organization.

No error is raised and nothing is logged — the merge appears to succeed.

**To Reproduce**

1. Create organizations A and B (duplicates of each other).
2. Give A two explored caves.
3. As a moderator: `DELETE /api/v1/organizations/<A>?isPermanent=true&entityId=<B>`
4. B does not gain A's explored caves. The rows are gone.

Note: permanent deletion returns `501` if the organization still has partner networks, partner entrances or cavers attached, so an org without those is needed to reach this code path.

**Expected behavior**

On a merge (`entityId` present), the deleted organization's explored-cave relationships should be re-pointed to the surviving organization, mirroring how documents and `editor`/`library` links are already handled. Without `entityId`, dropping them is correct and should stay unchanged.

**Additional context**

- Controller: `api/controllers/v1/organization/delete.js`
- Join table: `j_grotto_cave_explorer`, composite PK `(id_cave, id_grotto)`

The composite PK matters: a blind `UPDATE ... SET id_grotto = <survivor>` raises a unique violation as soon as both organizations explore the same cave — which is precisely the expected situation for duplicates. Shared rows must be deduplicated before re-pointing the remainder, and the two statements must be atomic, or a failure between them deletes the shared caves without handing the rest over.

**Related** — other open gaps in this same permanent-delete path, not covered by this issue:

- #1737 — `j_document_grotto_author` FK violation on permanent delete
- #1783 — `j_organization_massif` FK violation on permanent delete
- #1757 — `E_UNIQUE` on `reafectField` during merge-delete

dawoldo:
## 🤔 What

Fix the permanent deletion of an organization so that its explored caves are preserved when the deletion is a merge into another organization.

- `DELETE /api/v1/organizations/:id?isPermanent=true&entityId=<survivor>` now re-points the deleted org's `j_grotto_cave_explorer` rows to the surviving org.
- Without an `entityId` (plain permanent delete), behaviour is unchanged: the relationships are dropped.
- Added integration test coverage for the merge case, including the duplicate-cave edge case.

## 🤷‍♂️ Why

When two organizations are duplicates, moderators permanently delete one while passing `entityId` to merge it into the survivor. Documents, `editor`/`library` references and `redirectTo` pointers were already re-pointed to the survivor, but explored caves were not: `JGrottoCaveExplorer.destroy({ grotto: organizationId })` ran unconditionally.

The result was silent data loss — merging two duplicate organizations destroyed the exploration history of the one being deleted, even though it belonged to the same real-world organization.

## 🔍 How

In `api/controllers/v1/organization/delete.js`, the unconditional `destroy` is now branched on `shouldMergeInto`:

- **No merge target** → unchanged `JGrottoCaveExplorer.destroy({ grotto: organizationId })`.
- **Merge target** → two native queries:
  1. `DELETE` the deleted org's rows for caves the survivor *already* explores.
  2. `UPDATE ... SET id_grotto = <survivor>` for the remaining rows.

Step 1 is required because `j_grotto_cave_explorer` has a composite primary key `(id_cave, id_grotto)`: a blind `UPDATE` would raise a unique violation as soon as both organizations explore the same cave — precisely the situation for duplicates. Deduplicating first, then re-pointing, keeps exactly one row per cave for the survivor.

Raw SQL (`sails.sendNativeQuery`) is used rather than the Waterline model because this is a set-based update on a join table; doing it through the ORM would mean fetching and re-creating rows one by one.

## 🧪 Testing

The integration test in `test/integration/4_routes/Organization/delete.test.js` covers the merge path.

Manual check:

1. Create two organizations A and B.
2. Give A two explored caves, one of which B also explores.
3. `DELETE /api/v1/organizations/<A>?isPermanent=true&entityId=<B>` as a moderator.
4. B should now explore both caves, each exactly once, and no row should reference A.

Note: permanent deletion still returns `501` if the organization has partner networks, partner entrances or cavers — use an org without those to reach this code path.

## 📸 Previews

N/A — API-only change, no UI surface.

Closes #1795

### DIFF
--- a/api/controllers/v1/organization/delete.js
+++ b/api/controllers/v1/organization/delete.js
@@ -74,8 +74,30 @@ module.exports = async (req, res) => {
       return res.status(501).send();
     }
 
-    // Remove explored cave/entrance relationships
-    await JGrottoCaveExplorer.destroy({ grotto: organizationId });
+    // Explored caves: when merging into a surviving organization, re-point the
+    // deleted duplicate's relationships to it instead of dropping them. The
+    // join table's PK is (id_cave, id_grotto), so first delete the deleted
+    // org's rows for caves the survivor already explores to avoid a PK
+    // collision, then re-point the remainder. Without a merge target, drop them.
+    if (shouldMergeInto) {
+      await sails.sendNativeQuery(
+        `DELETE FROM j_grotto_cave_explorer d
+           WHERE d.id_grotto = $1
+             AND EXISTS (
+               SELECT 1 FROM j_grotto_cave_explorer k
+               WHERE k.id_grotto = $2 AND k.id_cave = d.id_cave
+             )`,
+        [organizationId, mergeIntoId]
+      );
+      await sails.sendNativeQuery(
+        `UPDATE j_grotto_cave_explorer
+           SET id_grotto = $2
+           WHERE id_grotto = $1`,
+        [organizationId, mergeIntoId]
+      );
+    } else {
+      await JGrottoCaveExplorer.destroy({ grotto: organizationId });
+    }
 
     if (organization.documents.length > 0) {
       if (shouldMergeInto) {
--- a/api/controllers/v1/organization/delete.js
+++ b/api/controllers/v1/organization/delete.js
@@ -75,26 +75,36 @@ module.exports = async (req, res) => {
     }
 
     // Explored caves: when merging into a surviving organization, re-point the
-    // deleted duplicate's relationships to it instead of dropping them. The
-    // join table's PK is (id_cave, id_grotto), so first delete the deleted
-    // org's rows for caves the survivor already explores to avoid a PK
-    // collision, then re-point the remainder. Without a merge target, drop them.
+    // deleted org's relationships to it instead of dropping them. The join
+    // table's PK is (id_cave, id_grotto), so first delete the deleted org's
+    // rows for caves the survivor already explores to avoid a PK collision,
+    // then re-point the remainder. Both statements run in a transaction: a
+    // failure between them would delete the shared caves without ever handing
+    // the rest over to the survivor. Without a merge target, drop them all.
     if (shouldMergeInto) {
-      await sails.sendNativeQuery(
-        `DELETE FROM j_grotto_cave_explorer d
-           WHERE d.id_grotto = $1
-             AND EXISTS (
-               SELECT 1 FROM j_grotto_cave_explorer k
-               WHERE k.id_grotto = $2 AND k.id_cave = d.id_cave
-             )`,
-        [organizationId, mergeIntoId]
-      );
-      await sails.sendNativeQuery(
-        `UPDATE j_grotto_cave_explorer
-           SET id_grotto = $2
-           WHERE id_grotto = $1`,
-        [organizationId, mergeIntoId]
-      );
+      await sails.getDatastore().transaction(async (db) => {
+        await sails
+          .getDatastore()
+          .sendNativeQuery(
+            `DELETE FROM j_grotto_cave_explorer d
+               WHERE d.id_grotto = $1
+                 AND EXISTS (
+                   SELECT 1 FROM j_grotto_cave_explorer k
+                   WHERE k.id_grotto = $2 AND k.id_cave = d.id_cave
+                 )`,
+            [organizationId, mergeIntoId]
+          )
+          .usingConnection(db);
+        await sails
+          .getDatastore()
+          .sendNativeQuery(
+            `UPDATE j_grotto_cave_explorer
+               SET id_grotto = $2
+               WHERE id_grotto = $1`,
+            [organizationId, mergeIntoId]
+          )
+          .usingConnection(db);
+      });
     } else {
       await JGrottoCaveExplorer.destroy({ grotto: organizationId });
     }
--- a/api/controllers/v1/organization/delete.js
+++ b/api/controllers/v1/organization/delete.js
@@ -56,7 +56,13 @@ module.exports = async (req, res) => {
       .set({ [field]: replacement });
   }
 
-  const deletePermanently = !!req.param('isPermanent');
+  // The web client sends `?isPermanent=1`; accept the common truthy encodings
+  // ('1'/'true', or a real boolean) while treating explicit falsy values
+  // ('0'/'false') and an absent param as a soft delete. A bare `!!req.param(...)`
+  // would wrongly treat `isPermanent=0`/`false` as permanent.
+  const deletePermanently = [true, 'true', '1'].includes(
+    req.param('isPermanent')
+  );
   const mergeIntoId = parseInt(req.param('entityId'), 10);
   let shouldMergeInto = !Number.isNaN(mergeIntoId);
   let mergeIntoEntity;
--- a/api/controllers/v1/organization/delete.js
+++ b/api/controllers/v1/organization/delete.js
@@ -5,6 +5,7 @@ const RightService = require('../../../services/RightService');
 const { toOrganization } = require('../../../services/mapping/converters');
 const NameService = require('../../../services/NameService');
 const RecentChangeService = require('../../../services/RecentChangeService');
+const CommonService = require('../../../services/CommonService');
 
 module.exports = async (req, res) => {
   const hasRight = RightService.hasGroup(
@@ -83,27 +84,23 @@ module.exports = async (req, res) => {
     // the rest over to the survivor. Without a merge target, drop them all.
     if (shouldMergeInto) {
       await sails.getDatastore().transaction(async (db) => {
-        await sails
-          .getDatastore()
-          .sendNativeQuery(
-            `DELETE FROM j_grotto_cave_explorer d
-               WHERE d.id_grotto = $1
-                 AND EXISTS (
-                   SELECT 1 FROM j_grotto_cave_explorer k
-                   WHERE k.id_grotto = $2 AND k.id_cave = d.id_cave
-                 )`,
-            [organizationId, mergeIntoId]
-          )
-          .usingConnection(db);
-        await sails
-          .getDatastore()
-          .sendNativeQuery(
-            `UPDATE j_grotto_cave_explorer
-               SET id_grotto = $2
-               WHERE id_grotto = $1`,
-            [organizationId, mergeIntoId]
-          )
-          .usingConnection(db);
+        await CommonService.query(
+          `DELETE FROM j_grotto_cave_explorer d
+             WHERE d.id_grotto = $1
+               AND EXISTS (
+                 SELECT 1 FROM j_grotto_cave_explorer k
+                 WHERE k.id_grotto = $2 AND k.id_cave = d.id_cave
+               )`,
+          [organizationId, mergeIntoId],
+          db
+        );
+        await CommonService.query(
+          `UPDATE j_grotto_cave_explorer
+             SET id_grotto = $2
+             WHERE id_grotto = $1`,
+          [organizationId, mergeIntoId],
+          db
+        );
       });
     } else {
       await JGrottoCaveExplorer.destroy({ grotto: organizationId });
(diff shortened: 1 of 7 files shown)


ClemRz (changes requested):
The fix is correct, well-reasoned, and the test covers the meaningful edge cases. A few notes below.

### Issues (Must Fix)

- **[api/controllers/v1/organization/delete.js:83-97]** The DELETE and the UPDATE run as two separate auto-committed statements. If the UPDATE fails after the DELETE succeeds, the surviving org loses the caves that were shared with the deleted org — they get deleted but never re-pointed. The rest of the permanent-delete block has the same pattern (no transaction wrapping), but this is the only place where two raw statements form an interdependent pair where partial success produces silent data loss.

  Consider wrapping these two in an explicit transaction:
  ```javascript
  await sails.sendNativeQuery('BEGIN');
  try {
    await sails.sendNativeQuery(
      `DELETE FROM j_grotto_cave_explorer d
         WHERE d.id_grotto = $1
           AND EXISTS (
             SELECT 1 FROM j_grotto_cave_explorer k
             WHERE k.id_grotto = $2 AND k.id_cave = d.id_cave
           )`,
      [organizationId, mergeIntoId]
    );
    await sails.sendNativeQuery(
      `UPDATE j_grotto_cave_explorer
         SET id_grotto = $2
         WHERE id_grotto = $1`,
      [organizationId, mergeIntoId]
    );
    await sails.sendNativeQuery('COMMIT');
  } catch (err) {
    await sails.sendNativeQuery('ROLLBACK');
    throw err;
  }
  ```
  Alternatively, the two statements can be collapsed into one atomic query using a CTE:
  ```sql
  WITH duplicates AS (
    DELETE FROM j_grotto_cave_explorer
    WHERE id_grotto = $1
      AND EXISTS (
        SELECT 1 FROM j_grotto_cave_explorer k
        WHERE k.id_grotto = $2 AND k.id_cave = j_grotto_cave_explorer.id_cave
      )
  )
  UPDATE j_grotto_cave_explorer
    SET id_grotto = $2
    WHERE id_grotto = $1
  ```
  The CTE form is simpler and avoids the need for manual transaction management.

- **[api/controllers/v1/organization/delete.js:19]** `organizationId` is the raw string from `req.param('id')` (the `validateId` policy is not applied to this route — only `tokenAuth` is, per `config/policies.js:278`). It ends up as `$1` in both native queries. PostgreSQL will implicitly cast a numeric string to `integer` for a FK column, so this works in practice, but it's a latent footgun: a non-numeric value (bypassing route-level validation) would produce a database-level error rather than a clean `notFound`. Pre-existing issue, not introduced here, but worth noting while the surrounding code is being touched.

- **[test/integration/4_routes/Organization/delete.test.js:109]** The test creates `targetOrg` without `isDeleted: true`, which is correct for a merge target. It would also be useful to assert that the `sharedCave` is present in `survivorCaves` (not just that the count is 2 and both IDs appear). The current `deepEqual` on the sorted ID array already covers this implicitly, so this is just an observation — the test is solid.

- **[assets/swaggerV1.yaml:5818]** The `DELETE /organizations/{id}` spec doesn't document the `isPermanent` or `entityId` query parameters, or the `501` response. This PR doesn't change that surface, but since the explored-caves merge path is now correctly implemented, it's a good moment to fill in the missing parameter documentation.

### Nitpicks (Optional)

- **[api/controllers/v1/organization/delete.js:74-100]** The inline comment block is thorough and helpful. Minor: the phrase "deleted duplicate's relationships" in the comment could be "the deleted org's relationships" to stay consistent with the surrounding code's vocabulary (`shouldMergeInto`, `mergeIntoId`). No action needed.

ClemRz (changes requested):
The transaction concern from the previous review has been addressed — the two-statement sequence is now correctly wrapped in `sails.getDatastore().transaction(async (db) => { ... })` with `.usingConnection(db)`, matching the established pattern elsewhere in the codebase (e.g. `EnrichmentQueueService`, `GeoAssociationService`). The logic itself is sound and the test covers the important PK-collision edge case. One correctness issue to fix before merging.

### Issues (Must Fix)

- **[assets/swaggerV1.yaml:5857]** The `responses` block documents `'204'` for a successful operation, but the controller calls `ControllerService.treatAndConvert` which resolves to `res.ok()` (HTTP **200**). The existing tests also assert `.expect(200)`. This PR touches the `responses` block, so it's the right moment to fix it.

  ```suggestion
        '200':
          description: Successful operation
  ```

- **[api/controllers/v1/organization/delete.js:87-107]** `sails.getDatastore().sendNativeQuery(...).usingConnection(db)` is the correct pattern, but `CommonService.query(sql, values, db)` already wraps it and is used throughout the codebase for raw queries inside transactions (the `connection` third argument threads the connection through). Consider using it for consistency:

  ```javascript
  const CommonService = require('../../../services/CommonService');
  // ...
  await sails.getDatastore().transaction(async (db) => {
    await CommonService.query(
      `DELETE FROM j_grotto_cave_explorer d
         WHERE d.id_grotto = $1
           AND EXISTS (
             SELECT 1 FROM j_grotto_cave_explorer k
             WHERE k.id_grotto = $2 AND k.id_cave = d.id_cave
           )`,
      [organizationId, mergeIntoId],
      db
    );
    await CommonService.query(
      `UPDATE j_grotto_cave_explorer
         SET id_grotto = $2
         WHERE id_grotto = $1`,
      [organizationId, mergeIntoId],
      db
    );
  });
  ```

- **[config/policies.js:278]** The `validateId` policy is now correctly applied to the delete route. Note that `entityId` (a query param, not a path param) is not covered by `validateId`, which only inspects path params ending in `Id` plus the `id` route param. The controller already handles non-numeric `entityId` safely via `parseInt` + `Number.isNaN`, so this is not a bug — just worth being aware of.

### Nitpicks (Optional)

- **[assets/swaggerV1.yaml:5831]** The `id` path parameter is typed as `type: string` while `entityId` is `type: integer`. The actual `id` is an integer FK — this is a pre-existing inconsistency, no action needed here.

ClemRz:
Pushed `a396817e` addressing the outstanding feedback from the 2026-08-12 review. @dawoldo's internship has ended, so this is picked up on their behalf; their commits and authorship are untouched.

**Must-fix — swagger status code.** `DELETE /organizations/{id}` documented `'204'` but the controller answers through `ControllerService.treatAndConvert` → `res.ok()` → **200**, and all seven tests in `Organization/delete.test.js` assert `.expect(200)`. Now documented as `200`.

Worth noting for anyone touching the spec later: `'204'` appears 24 more times in `swaggerV1.yaml`, so this looks like a copy-paste convention rather than a one-off. Only this endpoint was verified and corrected here — the rest may deserve a separate sweep.

**Should-consider — `CommonService.query`.** The two statements now go through `CommonService.query(sql, values, db)` rather than `sails.getDatastore().sendNativeQuery(...).usingConnection(db)`. That helper is exactly this pattern and is used **87 times** across `api/`, so it's clearly the house convention. Behaviour is identical — it's the same call with the connection threaded through.

Everything from the first review round was already in place before the branch was left: the two-statement sequence wrapped in `sails.getDatastore().transaction()`, and `validateId` added to the route in `config/policies.js`.

**Verification** — `npm test` → **3192 passing, 0 failing**, including the merge path and the PK-collision edge case. `eslint` clean. Swagger still parses. Rebase not needed: the branch merges into `develop` with no conflicts.

**Not addressed here**, deliberately — three open bugs live in this same permanent-delete block, and I've left them out to keep this PR to one concern:

- #1737 — `j_document_grotto_author` FK violation on permanent delete (**still live in production**; #1756 was closed as its duplicate, but #1737 itself is unfixed)
- #1783 — `j_organization_massif` FK violation
- #1757 — `E_UNIQUE` on `reafectField` during merge-delete

Whoever picks those up will be editing these same ~40 lines, so landing this PR first avoids a conflict.

Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[api/controllers/v1/organization/delete.js:59 / assets/swaggerV1.yaml:5839]** The newly documented boolean contract says that only `isPermanent=true` performs the irreversible hard delete, but the controller still coerces the raw query value with `!!req.param('isPermanent')`. Query parameters arrive as strings, so both `?isPermanent=false` and the web client's common `?isPermanent=0` encoding are truthy and permanently delete the organization. This is especially risky now that generated clients can expose this parameter as a boolean. Consider parsing only explicit true encodings (for example, `[true, 'true', '1'].includes(...)`, as `guideline/delete.js` does) and adding regression coverage proving that `false`/`0` remain soft deletes.

ClemRz:
@Paul-AUB thanks — fixed in 8d470493. Your Must Fix had two halves and the second one turned out to be more interesting than expected.

**The parsing fix**

`isPermanent` is now parsed with `[true, 'true', '1'].includes(...)`, the `guideline/delete.js` pattern you pointed at.

**The regression coverage, and a false pass**

I first wrote the two tests asserting only `res.body.isDeleted`, then checked they actually fail without the fix. Only one did:

```
✅ should soft delete when isPermanent=false (string)  → FAILED (correctly)
❌ should soft delete when isPermanent=0               → still PASSED
```

`isDeleted` is set in JS on the soft-delete pass that runs *before* the permanent branch, so the response says `isDeleted: true` whether or not the row was then hard-deleted — that assertion can never catch this bug. Both tests now assert the row survives in the database, and both fail on a reverted controller. Worth noting this was the `?isPermanent=0` case specifically, which is what the web client emits (`apiRoutes.js:316`).

**Two test-infrastructure fixes this exposed**

- `histo_delete_grotto` was missing from the test schema, so `TGrotto.destroyOne()` hard-deleted in tests and the survives-in-database assertion couldn't be made at all. Now installed, mirroring `sql/0_triggers.sql:165`. This is a broad change — every grotto test now soft-deletes like production — but the full suite is green.
- `test/seed-database.js` wasn't in `SNAPSHOT_SOURCES`. `customSQL.js` only *defines* the SQL blocks; the seed file decides which ones run. I hit this mid-verification: editing the seed changed template state without invalidating the stamp, and a stale template produced failures unrelated to the code under test. Added.

**Verification:** 3194 passing, 0 failing. Linter clean. Each regression test confirmed to fail against a reverted controller.

**Not addressed, deliberately:** the same `!!req.param('isPermanent')` coercion is in 9 other controllers — `cave`, `massif`, `document`, `description`, `comment`, `history`, `location`, `entrance`, `rigging`. `?isPermanent=0` permanently deletes on all of them. Out of scope here; happy to file a separate ticket if you agree that's the right split.

Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[test/customSQL.js:476 / test/seed-database.js:80]** The current head no longer merges cleanly into `develop`. The base branch added `CREATE_COMMENT_TRIGGERS` and `CREATE_SUB_ENTITY_TRIGGERS` in the same export and seed lists where this PR adds `CREATE_GROTTO_SOFT_DELETE_TRIGGER`; GitHub now reports the PR as `CONFLICTING`, and a local merge-tree check produces conflict markers in both files. Please rebase or merge the latest `develop`, preserving all three trigger exports and all three seed entries, then rerun CI so the resolved result is validated.

Apart that eveything's ok

Paul-AUB (dismissed):

PR #1749, opened 2026-07-30
fix(thumbnail): apply EXIF orientation before resizing

### LINKED ISSUE #1747: Thumbnails lose EXIF orientation: portrait photos are displayed rotated by 90°
# Thumbnails lose EXIF orientation: portrait photos are displayed rotated by 90°

**Repo:** `grottocenter-api`
**Component:** `api/services/ThumbnailService.js`
**Severity:** medium — affects every portrait photo uploaded from a mobile device
**Reported from:** `grottocenter-front` (document detail page)

---

## Summary

Generated thumbnails ignore the source image's `EXIF Orientation` tag. For photos
shot in portrait on a phone, all three WebP variants (`small`, `medium`, `large`)
are written with landscape pixel data and no orientation metadata, so browsers
render them rotated by 90°.

The original file (`completePath`) displays correctly, because it still carries the
EXIF tag and browsers honour it (`image-orientation: from-image` has been the
default since Chrome 81 / Firefox 26 / Safari 13.1). Only the thumbnails are wrong.

## User-visible impact

On document pages, every consumer of the `thumbnails` field shows the image sideways:

- the image gallery (uses `thumbnails.small` via `srcSet`)
- the full-screen lightbox (prefers `thumbnails.large`, falls back to `completePath`)

Files uploaded before the thumbnail feature — and any file whose thumbnail
generation failed — fall back to `completePath` and therefore look correct. This
makes the bug look intermittent, but it is deterministic: it affects exactly the
images whose `EXIF Orientation` tag is not `1`.

## Reproduction

Example document: `/ui/documents/234141`, file
`8499796193519372-20260725_103612.jpg`.

Measured directly from the blobs:

```
ORIGINAL  documents/8499796193519372-20260725_103612.jpg
  stored pixels      : 4080 x 3060   (landscape)
  EXIF Orientation   : 6             (rotate 90° CW)
  rendered by browser: 3060 x 4080   (portrait — correct)

THUMBNAILS  documents/thumbnails/{variant}/8499796193519372-20260725_103612.webp
  small              : 480 x 360     (landscape, no EXIF)
  medium             : 1280 x 960    (landscape, no EXIF)
  large              : 1920 x 1440   (landscape, no EXIF)
  rendered by browser: as stored     (rotated 90° — wrong)
```

All three variants keep the raw landscape pixel layout, and the orientation tag is
gone, so the browser has no cue to correct it.

## Root cause

`sharp` does **not** auto-orient by default, and does not copy input metadata to the
output unless asked. In `api/services/ThumbnailService.js` (`resize()`):

```js
async resize(inputBuffer, targetWidth) {
  return sharp(inputBuffer, { animated: true })
    .resize({ width: targetWidth, withoutEnlargement: true })
    .webp({ quality: WEBP_QUALITY })
    .toBuffer();
}
```

There is no `.rotate()` / `.autoOrient()` call anywhere in the pipeline. Sequence of
events for an `Orientation: 6` photo:

1. `sharp` decodes the raw stored pixels (landscape) and ignores the EXIF tag.
2. `.resize()` scales those landscape pixels.
3. `.webp()` writes the output without the orientation tag.

Result: pixels never rotated, and the metadata that would have told the browser to
rotate them is dropped.

## Proposed fix

`sharp` is pinned at `^0.35.3`, so `.autoOrient()` is available. It must be called
**before** `.resize()`, otherwise the target width is applied to the pre-rotation
dimensions:

```js
async resize(inputBuffer, targetWidth) {
  return sharp(inputBuffer, { animated: true })
    .autoOrient()                                              // apply EXIF orientation, drop the tag
    .resize({ width: targetWidth, withoutEnlargement: true })
    .webp({ quality: WEBP_QUALITY })
    .toBuffer();
}
```

Equivalent alternative: the constructor option
`sharp(inputBuffer, { animated: true, autoOrient: true })`, which applies at decode
time and makes pipeline ordering irrelevant.

GIFs carry no EXIF orientation tag, so this is a no-op for them — no regression on
animated inputs.

> Keeping the EXIF metadata on the output instead (`.keepMetadata()`) would also
> make browsers render correctly, since WebP supports EXIF chunks. Auto-orienting is
> preferable for thumbnails: it produces upright pixels, avoids shipping camera
> metadata (GPS, device) to clients, and keeps the output smaller.

## Secondary issue — variant selection uses the un-oriented width

In `generate()`, the variant filter reads the *stored* width:

```js
const metadata = await sharp(imageBuffer, { ... }).metadata();
const { width: originalWidth } = metadata;
const applicableVariants = this.getApplicableVariants(originalWidth);
```

For an `Orientation: 6` image, `metadata.width` is the long side, not the displayed
width. This is not a rendering bug, but it wastes storage: a portrait photo whose
upright width is 1500 px (stored 2000 × 1500) passes the `1920 < 2000` test, so a
`large` variant is generated — while `withoutEnlargement: true` caps it at 1500 px,
producing a variant identical to the source resolution.

`metadata.autoOrient.width` (available since `sharp` 0.34) returns the upright
dimensions and should be used for the variant filter.

## Migration — existing files need regeneration

Already-stored thumbnails will stay wrong after the fix is deployed:
`scripts/backfill-thumbnails.js` only selects rows where all three thumbnail columns
are `null`:

```js
where: {
  thumbnailSmall: null,
  thumbnailMedium: null,
  thumbnailLarge: null,
}
```

A separate regeneration pass is required. Ideally it should target only images whose
`EXIF Orientation` is present and not `1`, so the whole media library isn't
reprocessed — that check can be done cheaply from the metadata of the downloaded
buffer before deciding to re-encode. Existing thumbnail blobs are overwritten in
place (same computed paths), so no cleanup of old blobs is needed.

## Acceptance criteria

- [ ] A newly uploaded portrait photo (`EXIF Orientation` 6 or 8) produces
      thumbnails whose stored dimensions are portrait, matching the original's
      displayed orientation.
- [ ] Mirrored orientations (2, 4, 5, 7) are handled as well — `.autoOrient()`
      covers flips, a bare 90° rotation would not.
- [ ] Landscape photos and images without an EXIF tag are unchanged (no regression).
- [ ] Animated GIF thumbnails still animate.
- [ ] Variant selection is based on the upright width.
- [ ] Existing incorrectly-oriented thumbnails are regenerated.

## Notes

No frontend change is needed or possible: the frontend cannot know that
compensation is required without fetching and parsing the original's EXIF, which
would defeat the purpose of thumbnails. Frontend consumers for reference:

- `packages/web-app/src/components/common/DocumentsList/utils/imageUtils.js` —
  `getThumbnailSources()`, `getLightboxSrc()`
- `packages/web-app/src/pages/DocumentDetails/Section.jsx` — `FilesSection`
- `packages/web-app/src/components/common/DocumentsList/ImageLightbox.jsx`

ClemRz:
## 🤔 What

Fix portrait photos being displayed rotated 90° in all thumbnail variants.

- Call `.autoOrient()` before `.resize()` in `ThumbnailService.resize()` so EXIF orientation is applied to the pixels before scaling
- Use `metadata.autoOrient.width` instead of `metadata.width` in `generate()` so variant selection is based on the upright dimensions, not the raw stored dimensions
- Add `--fix-orientation` flag to `scripts/backfill-thumbnails.js` to regenerate existing incorrectly-oriented thumbnails (only downloads and re-encodes images with EXIF Orientation ≠ 1)

Closes #1747

## 🤷‍♂️ Why

`sharp` does not auto-orient by default and strips all metadata on output. For photos shot in portrait on a phone (EXIF Orientation 6 or 8), the stored pixels are landscape. Without `.autoOrient()`, all three WebP variants kept that landscape pixel layout with no orientation tag, so browsers had no cue to correct them.

Variant selection was also using the raw stored width, meaning a portrait image stored as 4080×3060 was being compared against the 1920px threshold as if it were landscape — producing a `large` variant that `withoutEnlargement` then capped at 1500px, identical to the source resolution.

## 🔍 How

- `.autoOrient()` rotates/flips the decoded pixels according to the EXIF Orientation tag and then drops the tag. Called before `.resize()` so the target width is applied to the already-upright dimensions.
- `metadata.autoOrient` (available since sharp 0.34, already pinned at 0.35.3) returns `{ width, height }` for the upright image. Falls back to `metadata` directly when no orientation tag is present (images without EXIF, GIFs, PNGs), so there is no regression.
- GIFs carry no EXIF orientation tag — `.autoOrient()` is a no-op for them. Animated GIFs are unaffected.
- The backfill script `--fix-orientation` pass queries all files that already have thumbnails, downloads each one, reads `metadata.orientation`, and skips files where it is absent or equal to 1. Only files that actually need re-encoding are passed to `ThumbnailService.generate()`. Existing thumbnail blobs are overwritten in place (same computed paths).

## 🧪 Testing

Run the existing thumbnail integration tests:

```bash
npm run test -- --grep "Thumbnail"
```

To regenerate existing corrupted thumbnails in production after deploy:

```bash
AZURE_KEY=<key> node scripts/backfill-thumbnails.js --fix-orientation --batch-size 5
```

Use `--dry-run` first to preview the candidate count without making changes.

## 📸 Previews

N/A — the fix affects stored pixel layout, not the API surface.

### DIFF
--- a/api/services/ThumbnailService.js
+++ b/api/services/ThumbnailService.js
@@ -73,13 +73,26 @@ module.exports = {
    * Applies EXIF orientation before resizing so portrait photos shot on a
    * phone are not stored rotated. The orientation tag is dropped from the
    * output (the pixels are already upright, so browsers need no correction).
+   *
+   * `.autoOrient()` is skipped for animated inputs (pages > 1) because sharp's
+   * internal rotate step historically collapses multi-frame images to a single
+   * frame for orientations that require an actual pixel transform. Animated GIFs
+   * rarely carry EXIF orientation tags; skipping `.autoOrient()` for them avoids
+   * a still-WebP regression.
+   *
    * @param {Buffer} inputBuffer - The original image buffer
    * @param {number} targetWidth - The target width in pixels
+   * @param {object} [opts]
+   * @param {boolean} [opts.isAnimated=false] - True when the source has multiple frames
    * @returns {Promise<Buffer>} The resized WebP buffer
    */
-  async resize(inputBuffer, targetWidth) {
-    return sharp(inputBuffer, { animated: true })
-      .autoOrient() // rotate/flip pixels per EXIF Orientation, then drop the tag
+  async resize(inputBuffer, targetWidth, { isAnimated = false } = {}) {
+    const pipeline = sharp(inputBuffer, { animated: true });
+    if (!isAnimated) {
+      // rotate/flip pixels per EXIF Orientation, then drop the tag
+      pipeline.autoOrient();
+    }
+    return pipeline
       .resize({ width: targetWidth, withoutEnlargement: true })
       .webp({ quality: WEBP_QUALITY })
       .toBuffer();
@@ -96,15 +109,21 @@ module.exports = {
    * @param {Buffer} imageBuffer - The original image file buffer
    * @param {string} originalPath - The blob path of the original
    * @param {object} blobClient - Azure container client for uploading
+   * @param {object} [preloadedMeta] - Optional sharp metadata already fetched by the caller.
+   *   When provided the internal `sharp().metadata()` call is skipped, avoiding a
+   *   second decode of the same buffer (useful for callers that need orientation before
+   *   calling generate, e.g. the backfill script).
    * @returns {Promise<{small: string|null, medium: string|null, large: string|null, hadError: boolean}>}
    */
-  async generate(imageBuffer, originalPath, blobClient) {
+  async generate(imageBuffer, originalPath, blobClient, preloadedMeta) {
     const result = { small: null, medium: null, large: null, hadError: false };
 
     try {
-      const metadata = await sharp(imageBuffer, {
-        limitInputPixels: 100 * 1024 * 1024, // 100 megapixels max
-      }).metadata();
+      const metadata =
+        preloadedMeta ||
+        (await sharp(imageBuffer, {
+          limitInputPixels: 100 * 1024 * 1024, // 100 megapixels max
+        }).metadata());
       // Use the oriented (upright) width so portrait photos whose stored pixels
       // are landscape (e.g. EXIF Orientation 6/8) are measured correctly.
       // metadata.autoOrient is available since sharp 0.34 and falls back to
@@ -121,11 +140,15 @@ module.exports = {
         return result;
       }
 
+      // Animated inputs (GIFs with multiple frames) must not go through
+      // .autoOrient() — see resize() JSDoc for the rationale.
+      const isAnimated = (metadata.pages || 1) > 1;
+
       // Generate all buffers first (fail-fast before uploading)
       const buffers = await Promise.all(
         applicableVariants.map(async (variant) => ({
           name: variant.name,
-          buffer: await this.resize(imageBuffer, variant.width),
+          buffer: await this.resize(imageBuffer, variant.width, { isAnimated }),
           path: this.computeThumbnailPath(variant.name, originalPath),
         }))
       );
--- a/api/services/ThumbnailService.js
+++ b/api/services/ThumbnailService.js
@@ -70,12 +70,16 @@ module.exports = {
    * Resize an image buffer to the target width, output as WebP.
    * Preserves aspect ratio (no crop, no upscaling).
    * Preserves alpha channel for PNG inputs and animation for GIF inputs.
+   * Applies EXIF orientation before resizing so portrait photos shot on a
+   * phone are not stored rotated. The orientation tag is dropped from the
+   * output (the pixels are already upright, so browsers need no correction).
    * @param {Buffer} inputBuffer - The original image buffer
    * @param {number} targetWidth - The target width in pixels
    * @returns {Promise<Buffer>} The resized WebP buffer
    */
   async resize(inputBuffer, targetWidth) {
     return sharp(inputBuffer, { animated: true })
+      .autoOrient() // rotate/flip pixels per EXIF Orientation, then drop the tag
       .resize({ width: targetWidth, withoutEnlargement: true })
       .webp({ quality: WEBP_QUALITY })
       .toBuffer();
@@ -85,6 +89,9 @@ module.exports = {
    * Generate all applicable thumbnail variants for an image.
    * Uses an all-or-nothing approach: generates all buffers first, then uploads.
    * If any step fails, returns all nulls with hadError: true.
+   *
+   * Variant selection is based on the upright (post-EXIF-orientation) width so
+   * that portrait photos stored in landscape pixel layout are measured correctly.
    *
    * @param {Buffer} imageBuffer - The original image file buffer
    * @param {string} originalPath - The blob path of the original
@@ -98,7 +105,11 @@ module.exports = {
       const metadata = await sharp(imageBuffer, {
         limitInputPixels: 100 * 1024 * 1024, // 100 megapixels max
       }).metadata();
-      const { width: originalWidth } = metadata;
+      // Use the oriented (upright) width so portrait photos whose stored pixels
+      // are landscape (e.g. EXIF Orientation 6/8) are measured correctly.
+      // metadata.autoOrient is available since sharp 0.34 and falls back to
+      // the stored dimensions when no orientation tag is present.
+      const originalWidth = (metadata.autoOrient || metadata).width;
 
       if (!originalWidth) {
         return result;
--- a/scripts/backfill-thumbnails.js
+++ b/scripts/backfill-thumbnails.js
@@ -62,70 +62,63 @@ async function run() {
   });
 
   try {
-    // Query file records without thumbnails in pages to avoid loading all into memory
+    // Paginate through TFile records matching the given `where` clause and collect
+    // processable image files. Explicit sort guarantees stable pagination across
+    // multiple SELECT calls — without it PostgreSQL may reorder rows (autovacuum,
+    // HOT updates) causing records to be missed or processed twice.
     const pageSize = 500;
-    let skip = 0;
-    let page;
-    const imageFiles = [];
 
-    do {
-      page = await TFile.find({
-        where: {
-          thumbnailSmall: null,
-          thumbnailMedium: null,
-          thumbnailLarge: null,
-        },
-      })
-        .populate('fileFormat')
-        .limit(pageSize)
-        .skip(skip);
+    const fetchImageFiles = async (where) => {
+      const files = [];
+      let skip = 0;
+      let page;
+      do {
+        // eslint-disable-next-line no-await-in-loop
+        page = await TFile.find({ where })
+          .populate('fileFormat')
+          .sort('id ASC')
+          .limit(pageSize)
+          .skip(skip);
 
-      page.forEach((f) => {
-        if (!f.fileFormat || !f.fileFormat.mimeType) return;
-        const { mimeType } = f.fileFormat;
-        if (mimeType === 'image/svg+xml') return;
-        if (ThumbnailService.isProcessable(mimeType)) {
-          imageFiles.push(f);
-        }
-      });
+        page.forEach((f) => {
+          if (!f.fileFormat || !f.fileFormat.mimeType) return;
+          const { mimeType } = f.fileFormat;
+          if (mimeType === 'image/svg+xml') return;
+          if (ThumbnailService.isProcessable(mimeType)) {
+            files.push(f);
+          }
+        });
 
-      skip += pageSize;
-    } while (page.length === pageSize);
+        skip += pageSize;
+      } while (page.length === pageSize);
+      return files;
+    };
 
-    console.log(`Found ${imageFiles.length} image files without thumbnails.`);
+    // Collect files that have no thumbnails at all.
+    const missingThumbnailFiles = await fetchImageFiles({
+      thumbnailSmall: null,
+      thumbnailMedium: null,
+      thumbnailLarge: null,
+    });
+    const imageFiles = [...missingThumbnailFiles];
+
+    console.log(
+      `Found ${missingThumbnailFiles.length} image files without thumbnails.`
+    );
 
     // When --fix-orientation is passed, also collect images that already have
     // thumbnails but may have been generated before .autoOrient() was added.
     // We check the EXIF orientation of the downloaded buffer at process time
     // and skip files that are already correctly oriented (orientation == 1 or absent).
     if (fixOrientation) {
-      let orientSkip = 0;
-      let orientPage;
-      do {
-        orientPage = await TFile.find({
-          where: {
-            or: [
-              { thumbnailSmall: { '!=': null } },
-              { thumbnailMedium: { '!=': null } },
-              { thumbnailLarge: { '!=': null } },
-            ],
-          },
-        })
-          .populate('fileFormat')
-          .limit(pageSize)
-          .skip(orientSkip);
-
-        orientPage.forEach((f) => {
-          if (!f.fileFormat || !f.fileFormat.mimeType) return;
-          const { mimeType } = f.fileFormat;
-          if (mimeType === 'image/svg+xml') return;
-          if (ThumbnailService.isProcessable(mimeType)) {
-            imageFiles.push(f);
-          }
-        });
-
-        orientSkip += pageSize;
-      } while (orientPage.length === pageSize);
+      const alreadyThumbnailedFiles = await fetchImageFiles({
+        or: [
+          { thumbnailSmall: { '!=': null } },
+          { thumbnailMedium: { '!=': null } },
+          { thumbnailLarge: { '!=': null } },
+        ],
+      });
+      imageFiles.push(...alreadyThumbnailedFiles);
 
       console.log(
         `Found ${imageFiles.length} total image files to process (including orientation re-check).`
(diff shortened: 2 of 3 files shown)


Paul-AUB:
# Code Review — PR #1749: EXIF orientation handling in thumbnails

Line-by-line review of a small, focused diff that adds EXIF orientation handling to thumbnail generation and introduces a `--fix-orientation` migration mode in the backfill script. Findings are ordered by severity.

---

## 1. `--fix-orientation` cannot repair files whose oriented width is ≤ 480px

**File:** `scripts/backfill-thumbnails.js:236`

**Summary:** No variants are generated, the DB is not updated, and the user keeps seeing the pre-existing, wrongly-oriented thumbnail.

**Failure scenario:** A previously thumbnailed portrait photo has raw dimensions `500x400` landscape with EXIF orientation `6`/`8` (post-orient width = `400`). `ThumbnailService.generate` returns all-`null` with `hadError=false` because `getApplicableVariants(400)` returns `[]`. The backfill hits the `else { skipped += 1 }` branch at line 253, so `TFile.updateOne` is never called. `thumbnailSmall` still points at the old landscape-oriented blob and the user continues to see it wrongly rotated.

**Suggested fix:** To actually repair such records, the script should null out the DB fields (or set them to a regenerated at-native-width variant) when `orientation != 1` and no variant applies.

---

## 2. "Skipped (too small)" summary label is misleading in `--fix-orientation` mode

**File:** `scripts/backfill-thumbnails.js:277`

**Summary:** The counter also aggregates orientation-already-correct skips, so the reported label lies.

**Failure scenario:** Run with `--fix-orientation` on 1000 files, 900 of which have `orientation == 1` (or missing) and are early-returned at line 222. The final line prints `Skipped (too small): 900` even though none were skipped for size. The operator misreads the report as "nothing needed doing" instead of "the vast majority were already correct".

**Suggested fix:** Track "skipped (orientation already correct)" separately from "skipped (too small)" and print both counters.

---

## 3. Every file is decoded by sharp twice in `--fix-orientation` mode

**File:** `scripts/backfill-thumbnails.js:219`

**Summary:** Redundant `sharp().metadata()` call — same buffer decoded twice per file.

**Failure scenario:** For every non-skipped file we call `sharp(imageBuffer).metadata()` here to read orientation, then `ThumbnailService.generate` immediately calls `sharp(imageBuffer, { limitInputPixels: ... }).metadata()` on the same buffer. On a large image set this is measurable wasted CPU.

**Suggested fix:** Capture `meta` once and thread it into `generate`, or read `metadata.orientation` inside `generate` from its already-fetched metadata.

---

## 4. `.autoOrient()` may drop frames of an animated GIF carrying an EXIF orientation tag

**File:** `api/services/ThumbnailService.js:82`

**Summary:** Animated GIFs with a non-trivial EXIF orientation may collapse to a still WebP.

**Failure scenario:** Sharp implements `.autoOrient()` via an internal rotate step; for orientations that require an arbitrary-angle-equivalent transform on animated inputs, sharp has historically collapsed multi-frame images to a single frame. A processable animated GIF (`image/gif` is in `PROCESSABLE_MIME_TYPES`) uploaded with `orientation != 1` would produce a still WebP as its thumbnail instead of an animated one. Rare combination in practice (GIFs seldom carry EXIF orientation), but a concrete regression path introduced by this line.

**Suggested fix:** Skip `.autoOrient()` for animated inputs, or verify frame preservation with the current sharp version before shipping.

---

## 5. Paginated fetch via `.skip(orientSkip)` without an explicit sort risks skipped/duplicated rows

**File:** `scripts/backfill-thumbnails.js:116`

**Summary:** Waterline `.find()` here has no `.sort(...)`, so page boundaries are non-deterministic.

**Failure scenario:** Under PostgreSQL, physical ordering is not guaranteed to be stable between two `SELECT`s (autovacuum, HOT updates, or concurrent writes can reshuffle). If a scheduled task or another operator writes to `TFile` while this collection loop runs, records can be missed (never processed) or double-listed (processed twice). The pre-existing style at line 81 has the same issue, but this PR duplicates the pattern.

**Suggested fix:** Add an explicit `.sort('id ASC')` (or another stable key) to the paginated query.

---

## Overall

This is a small, well-scoped diff. No critical correctness defect was found — the core `.autoOrient()` insertion and the `(metadata.autoOrient || metadata).width` fallback are correct under sharp `^0.35.3` (per `package.json`). The findings above range from a real edge-case gap in the migration flow (#1) down to lower-severity operational/efficiency issues (#3–#5).

Paul-AUB (changes requested):
.

ClemRz:
Thanks for the thorough review @Paul-AUB. All 5 findings addressed in 86b8fa23:

**1 — `--fix-orientation` couldn't repair files with oriented width ≤ 480px**
When `generate()` returns all-null with `hadError: false` (image too small for any variant), the script now nulls out the stale thumbnail DB fields so the front-end stops serving the wrongly-oriented blobs.

**2 — Misleading "Skipped (too small)" counter**
Split into two separate counters: `orientationSkipped` (orientation already correct, no re-encode needed) and `tooSmallSkipped` (image too small / blob too large). The summary only prints `orientationSkipped` when `--fix-orientation` is active.

**3 — Double sharp decode per file in `--fix-orientation` mode**
Fetch `metadata` once before calling `generate()`, then pass it as an optional `preloadedMeta` 4th argument. `generate()` skips its internal `sharp().metadata()` call when a pre-fetched value is provided. The script no longer imports sharp for metadata-only work — it re-uses the same decode path as `generate()`.

**4 — `.autoOrient()` may collapse animated GIF frames**
`resize()` now accepts `{ isAnimated }` option and skips `.autoOrient()` when `true`. `generate()` detects `metadata.pages > 1` and passes `isAnimated: true` through. Static images are unaffected.

**5 — Paginated fetch without explicit sort**
Both `TFile.find()` loops now have `.sort('id ASC')` so page boundaries are stable under concurrent writes or autovacuum reshuffling.

Paul-AUB (changes requested):
Follow-up review after the [previous round of comments](https://github.com/GrottoCenter/grottocenter-api/pull/1749#issuecomment-5140780958). Verified commit `86b8fa23`: findings #2, #3, #4, #5 are correctly resolved, and #1 is resolved for the case it targeted — but the fix for #1 introduces a new regression, detailed below.

### Issues (Must Fix)

1. **[scripts/backfill-thumbnails.js:233-242](scripts/backfill-thumbnails.js#L233-L242)** The orientation-skip check added to fix the previous finding #1 is applied to *every* file in `imageFiles`, including files pulled in by the first query (lines 74-97) that have **no thumbnails at all yet**. Consider: a `TFile` with `thumbnailSmall/Medium/Large` all `null` and a correctly-oriented image (`orientation` absent or `1`) is included via the first (missing-thumbnails) query. When the script is run with `--fix-orientation` — which is exactly the invocation documented in this PR's description for the production migration (`node scripts/backfill-thumbnails.js --fix-orientation --batch-size 5`) — this file hits `if (!orientation || orientation === 1) { orientationSkipped += 1; return; }` and `ThumbnailService.generate()` is never called. The file permanently keeps no thumbnails, even though it was never processed before and needed to be. The script's own header comment (line 8-10) says `--fix-orientation` should still generate thumbnails for images that don't have any, in addition to fixing wrongly-oriented ones, so this contradicts the documented behavior.

   Since files from the first query always have all three thumbnail columns `null`, and files from the second (orientation-recheck) query always have at least one non-null by construction of its `where` clause, the skip check can distinguish them without extra state, e.g.:
   ```js
   const alreadyHasThumbnail =
     file.thumbnailSmall || file.thumbnailMedium || file.thumbnailLarge;
   if (fixOrientation && alreadyHasThumbnail) {
     const { orientation } = meta;
     if (!orientation || orientation === 1) {
       orientationSkipped += 1;
       processed += 1;
       return;
     }
   }
   ```

2. **[test/integration/1_services/ThumbnailService.test.js](test/integration/1_services/ThumbnailService.test.js)** No test exercises `.autoOrient()`, the `isAnimated` skip path, or the `metadata.autoOrient.width` fallback — i.e. none of the behavior this PR actually adds (confirmed via grep across `test/integration/1_services/`: no matches for `orientation`, `autoOrient`, `isAnimated`, or `EXIF`). Given this is a bug-fix PR for a user-visible regression, and that a non-trivial regression slipped into the companion script fix (#1 above), a couple of targeted cases would help — e.g. a fixture JPEG with `Orientation: 6` asserting the output's pixel dimensions are swapped/upright, and a case asserting `resize()` skips `.autoOrient()` when `isAnimated: true`.

3. **[scripts/backfill-thumbnails.js:74-97](scripts/backfill-thumbnails.js#L74-L97) and [scripts/backfill-thumbnails.js:105-138](scripts/backfill-thumbnails.js#L105-L138)** The two pagination loops are near-identical copies (same `populate`/`sort`/`limit`/`skip` shape, same per-record filtering). Worth extracting into a shared helper (e.g. `fetchImageFiles(where)`) to avoid the two drifting apart — the missing `.sort('id ASC')` in one of them was exactly this kind of drift in the previous round.

### Nitpicks (Optional)

4. **[api/services/ThumbnailService.js:77-81](api/services/ThumbnailService.js#L77-L81)** The JSDoc asserts sharp "historically collapses multi-frame images to a single frame" for `.autoOrient()` on animated inputs — worth a link to the relevant sharp issue/changelog entry in a code comment so future maintainers can verify whether it's still true when the `sharp` dependency is upgraded, rather than having to rediscover it.

ClemRz:
Thanks for the thorough follow-up review. Addressed in the amended commit `92f44e04`.

**#1 (orientation-skip regression) — fixed**
The skip check now guards on `alreadyHasThumbnail` before testing orientation, exactly as suggested. Files from the first query (all three thumbnail columns `null`) always flow through to `generate()` regardless of orientation. At the same time, the two near-identical pagination loops were extracted into a shared `fetchImageFiles` helper, which resolves the drift risk raised in suggestion #3.

**#2 (missing tests) — added**
Three targeted cases added to `ThumbnailService.test.js`:
- `resize`: EXIF Orientation 6 fixture — asserts output is portrait (480×640) and orientation tag is absent after `autoOrient`
- `resize`: `isAnimated: true` path — asserts resize still produces correct dimensions when `autoOrient` is skipped
- `generate`: `autoOrient.width` fallback — a 1500×800 raw image with Orientation 6 has an upright width of 800, so only the `small` variant is generated (not `medium`), confirming variant selection uses the upright dimension

**#3 (duplicated pagination loops) — fixed as part of #1**
Covered above.

**#4 (sharp reference link) — left as prose**
Searched the sharp issue tracker and changelog but couldn't find a public issue that specifically documents the multi-frame collapse behaviour for `autoOrient`. Added nothing rather than a fabricated link — the JSDoc prose already conveys the risk clearly enough for a future maintainer to know to check on upgrade.

Paul-AUB (dismissed):
Follow-up review of commit `92f44e04`, which addresses the [previous round's findings](https://github.com/GrottoCenter/grottocenter-api/pull/1749#issuecomment-5147548664) (issue #1: orientation-skip regression on never-thumbnailed files; suggestions #2 and #3: missing tests, duplicated pagination loops).

Verified: the `alreadyHasThumbnail` gate at `scripts/backfill-thumbnails.js:227-238` correctly restores the intended behavior — files from the first (missing-thumbnails) query always flow through to `generate()` regardless of orientation, while files from the second (re-check) query are still skipped when already upright. The `fetchImageFiles` extraction removes the pagination-loop duplication, and the three new tests in `ThumbnailService.test.js` exercise `.autoOrient()`, the `isAnimated` skip path, and the `autoOrient.width` variant-selection fallback. No regression found in the fix for the previous round's Issue #1.

One new, minor edge case introduced by this round's changes:

1. **[scripts/backfill-thumbnails.js:266-279](scripts/backfill-thumbnails.js#L266-L279)** The "clear stale thumbnail paths" branch fires for *any* file that ends up with no generated variants while `--fix-orientation` is set, including files pulled in by the first query that never had thumbnails and simply have a too-small oriented width — not just the "wrong orientation, too small" case it was written for. Concretely: a file with all three thumbnail columns `null`, no EXIF orientation tag, and an oriented width of e.g. 300px is included via `missingThumbnailFiles`; it reaches this branch (no variants applicable, `hadError` false, `fixOrientation` true) and gets a `console.log` saying `"wrong orientation but too small for any variant"` even though its orientation was fine, plus a `TFile.updateOne` writing `null` over fields that are already `null`. Functionally harmless (idempotent no-op write, correctly bucketed under `tooSmallSkipped`), but the log line misreports the reason and adds an unnecessary DB round-trip for files that were never wrongly oriented in the first place. Consider gating this branch on `alreadyHasThumbnail` (mirroring the guard added just above it) so it only fires for the case it was designed for.

### Nitpicks (Optional)

2. **[api/services/ThumbnailService.js:145](api/services/ThumbnailService.js#L145)** and **[scripts/backfill-thumbnails.js:217-219](scripts/backfill-thumbnails.js#L217-L219)** Both the backfill script's standalone metadata fetch and `generate()`'s own fallback fetch omit `{ animated: true }` from the `sharp()` constructor options used to read `metadata.pages`. This matches the pre-existing pattern and sharp does report `pages` from the container header independently of the `animated` decode option, so this is not expected to misbehave — flagging only so it's a conscious choice if anyone revisits `isAnimated` detection later, since it's easy to assume `animated: true` is required here.

Overall this is a solid iteration — the regression from the previous round is fixed correctly, tests were added for the previously-uncovered orientation logic, and the duplicated pagination code was consolidated. The one remaining item is cosmetic/low-severity and not blocking.

ClemRz:
Addressed in `cd1c1eec`.

**Suggestion 1 — fixed**
The "clear stale thumbnail paths" branch is now gated on `alreadyHasThumbnail`, mirroring the guard added in the previous round. Files from the first query (no thumbnails, too small, correct orientation) fall through to the plain `tooSmallSkipped` branch — no DB write, no misleading log line.

**Nitpick 2 — acknowledged, left as-is**
The omission of `{ animated: true }` from the `sharp()` metadata constructor is intentional — sharp reports `pages` from the container header regardless of that option, so `isAnimated` detection is unaffected. Noted for anyone revisiting this later.

PR #1750, opened 2026-07-31
fix(document): replace m2m collections via replaceCollection on update and validate

### LINKED ISSUE #1736: Document edit: author collections cannot be cleared or replaced
# Document edit: author collections cannot be cleared or replaced

## Problem

When a user edits a document to **remove a person author** and add an organization author (`authorsGrotto`), the removed person author remains linked to the document after the modification is validated. The same symptom applies to any change that reduces a many-to-many collection (`authors`, `authorsGrotto`, and by extension `subjects`, `languages`, `isoRegions`, `countries`, `massifs`).

The user's intent — "these are now the only authors" — is silently ignored for whatever was removed.

## Root cause (backend side)

At validation time, [`multiple-validate.js`](api/controllers/v1/document/multiple-validate.js) applies the pending modification via:

```js
await TDocument.updateOne(document.id).set({
  ...cleanedDocumentData,   // contains authors: [ids], authorsGrotto: [ids], ...
  modifiedDocJson: null,
  ...
});
```

Sails/Waterline 1.x does **not** support modifying many-to-many `through` collections via `update().set()`. From the Waterline docs:

> You cannot use `.update()` to modify the associations of a record. Use `.replaceCollection()`, `.addToCollection()`, or `.removeFromCollection()` instead.

As a result, the join tables (`j_document_caver_author`, `j_document_grotto_author`, etc.) are **never touched** by the validation flow — old associations remain, new ones from the pending edit are not written.

This affects the same call site in [`update-with-new-entities.js`](api/controllers/v1/document/update-with-new-entities.js) (which also does `TDocument.updateOne(documentId).set(cleanedData)`).

The pattern is applied correctly elsewhere in the codebase — see [`caver/update.js`](api/controllers/v1/caver/update.js), [`caver/delete.js`](api/controllers/v1/caver/delete.js), [`guideline/update.js`](api/controllers/v1/guideline/update.js), [`caver/set-groups.js`](api/controllers/v1/caver/set-groups.js) — all of which use `replaceCollection`.

## Root cause (transport side)

Even if the backend fix above is applied, a second issue prevents the empty case from ever reaching the DB:

The front sends the document body as `multipart/form-data` (because of file uploads). `FormData` is a flat key/value store and cannot natively express an empty array. When the user removes the last person author, the `authors` field is simply absent from the request body.

Consequently, [`DocumentService.getConvertedDataFromClient`](api/services/DocumentService.js) receives `body.authors === undefined`:

```js
authors: body.authors?.map((a) => a.id),      // → undefined
authorsGrotto: body.authorsGrotto?.map(...),  // → undefined
```

`undefined` is then stored in `modifiedDocJson` and, at validation, passed to `.set()` — which is a no-op for that field. So even after the m2m fix, "clear this collection entirely" still cannot be expressed.

### 1. Use `replaceCollection` for m2m collections at validation time

In `multiple-validate.js` (`validateAndUpdateDocument`) and in `update-with-new-entities.js`:

- Remove the m2m collection fields from the object passed to `TDocument.updateOne(id).set(...)`.
- For each of them, call `TDocument.replaceCollection(id, '<field>').members([...ids])` inside the same transaction.

Affected collections on `TDocument`: `authors`, `authorsGrotto`, `subjects`, `languages`, `isoRegions`, `countries`, `massifs`.

Skip the `replaceCollection` call when the value is `undefined` (not sent by the client). Only call it when the client explicitly provided a value (including an empty array — see next point).

### 2. Accept an explicit "cleared collection" marker

The front cannot send an empty array through `FormData`. We propose the front sends the literal string `'[]'` as the value of a collection field to signal an intentional clear, and leaves the field absent otherwise.

Adapt `getConvertedDataFromClient` to interpret this:

```js
const parseIdList = (v) => {
  if (v === undefined) return undefined;   // untouched
  if (v === '[]') return [];               // explicitly cleared
  return v.map((a) => a.id);
};

// ...
authors: parseIdList(body.authors),
authorsGrotto: parseIdList(body.authorsGrotto),
// (and similarly for subjects, languages, isoRegions, countries, massifs
//  if the front ever needs to clear them)
```

The distinction `undefined` vs `[]` matters:
- `undefined` → don't call `replaceCollection`, keep existing associations.
- `[]` → call `replaceCollection(id, field).members([])`, wipe the collection.
- non-empty array → call `replaceCollection(id, field).members(ids)`, replace.

### 3. Tests

Add integration tests for `PUT /api/v1/documents/:id` followed by validation, covering:

- Replacing `authors` (some removed, some added, some kept) — assert `j_document_caver_author` reflects the new set exactly.
- Clearing `authors` entirely while providing `authorsGrotto` — assert the person authors are gone and the grotto authors are present.
- Same two cases for `authorsGrotto`.
- Untouched collections stay untouched (i.e. no field sent → no change).

## Frontend counterpart

A companion PR on `grottocenter-front` will:

- Update `buildFormData` to emit `'[]'` for empty arrays (currently they are dropped from the FormData).
- Continue to omit the field when the form was not touched (no regression on unchanged fields).

Both PRs need to land together for the feature to work end-to-end.

## Reproduction

1. Open an existing document with at least one person author.
2. Edit the document: remove all person authors, add one organization as author.
3. Have a moderator validate the modification.
4. Reload the document — the removed person author is still listed; the added organization may or may not be listed depending on the code path.

Expected: after step 3, the document has only the organization as author.

ClemRz:
## 🤔 What

Fix silent data loss when editing document many-to-many collections (authors, authorsGrotto, subjects, languages, massifs, isoRegions, countries). Closes #1736.

- Use `replaceCollection` for all 7 m2m associations in `multiple-validate` and `update-with-new-entities` instead of passing them to `.set()`, which Waterline silently ignores
- Wrap `update-with-new-entities` scalar update and `replaceCollection` calls in a single transaction to prevent partial writes
- Add `parseIdList` helper in `DocumentService.getConvertedDataFromClient` to handle the `undefined` / `'[]'` / array distinction for multipart/form-data collection fields
- Fix `getDocumentLanguages` returning `[]` instead of `undefined` when the field is absent, which was silently wiping the language collection on every document edit
- Extract `DOCUMENT_M2M_COLLECTIONS` to `config/constants/document.js` as a single source of truth
- Document the `'[]'` sentinel protocol and add the missing `PUT /documents/{id}/new-entities` entry in `swaggerV1.yaml`
- Add integration tests for replace, clear, and untouched semantics in both update paths

## 🤷‍♂️ Why

Editing a document's authors (or any other m2m collection) had no effect. Waterline's `.updateOne().set()` silently ignores collection fields — they must be handled via `replaceCollection`. As a result, removing an author or replacing one organisation with another was a no-op: the join tables were never touched.

A second issue prevented the "clear all" case from working even after the Waterline fix: `multipart/form-data` cannot represent an empty array, so the front-end omitting a field is indistinguishable from deliberately clearing it. The `'[]'` string sentinel solves this (the companion front-end PR will emit `'[]'` for intentionally cleared collections).

## 🔍 How

**`multiple-validate.js`** — `normalizeAndSplitDocumentData` strips m2m fields from the scalar payload and returns them separately. Inside the existing transaction, `replaceCollection(id, field).members([...]).usingConnection(db)` is called for every field that is not `undefined`. Fields absent from `modifiedDocJson` are left untouched.

**`update-with-new-entities.js`** — Same split-and-replace pattern. The `updateOne` and all `replaceCollection` calls now share a single `sails.getDatastore().transaction` with `.usingConnection(db)` to make the operation atomic.

**`DocumentService.getConvertedDataFromClient`** — New `parseIdList` helper maps `undefined` → `undefined` (untouched), `'[]'` → `[]` (explicit clear), and arrays → mapped id arrays. `getDocumentLanguages` now returns `undefined` instead of `[]` when neither language field is present.

**`config/constants/document.js`** — `DOCUMENT_M2M_COLLECTIONS` exported here; both controllers import it. Adding a new collection to `TDocument` requires a single change.

## 🧪 Testing

Integration tests added in:
- `test/integration/4_routes/Documents/multiple-validate.test.js` — replace authors, clear authors + add grotto authors, leave untouched collections unchanged (all with `after` cleanup)
- `test/integration/4_routes/Documents/update-with-new-entities.test.js` — replace authors, clear authors with `[]`, absent field leaves associations untouched, replace authorsGrotto

All 3300 existing tests pass.

Note: this backend PR needs to land together with its front-end counterpart (updating `buildFormData` to emit `'[]'` for empty arrays) for the end-to-end fix to be complete.

## 📸 Previews

N/A — backend only.

### DIFF
--- a/api/controllers/v1/document/multiple-validate.js
+++ b/api/controllers/v1/document/multiple-validate.js
@@ -2,6 +2,9 @@ const DocumentService = require('../../../services/DocumentService');
 const FileService = require('../../../services/FileService');
 const RightService = require('../../../services/RightService');
 const NotificationService = require('../../../services/NotificationService');
+const {
+  DOCUMENT_M2M_COLLECTIONS,
+} = require('../../../../config/constants/document');
 
 async function markDocumentValidated(
   documentId,
@@ -17,26 +20,37 @@ async function markDocumentValidated(
   });
 }
 
+// The seven m2m associations on TDocument that must be handled via replaceCollection.
+// Waterline silently ignores collection fields passed to .update()/.set().
+// Defined in config/constants/document.js — single source of truth.
+
 // modifiedDocJson may pre-date this fix and contain populated objects instead of plain IDs.
-function normalizeCollectionIds(documentData) {
-  const collectionFields = [
-    'massifs',
-    'authors',
-    'authorsGrotto',
-    'subjects',
-    'languages',
-    'isoRegions',
-    'countries',
-  ];
-  const normalized = { ...documentData };
-  for (const field of collectionFields) {
-    if (Array.isArray(normalized[field])) {
-      normalized[field] = normalized[field].map((item) =>
+// Returns { scalarData, collectionData } where collectionData[field] is either an array of
+// plain ids or undefined (meaning "not sent — keep existing associations").
+function normalizeAndSplitDocumentData(documentData) {
+  const scalarData = { ...documentData };
+  const collectionData = {};
+
+  for (const field of DOCUMENT_M2M_COLLECTIONS) {
+    const value = scalarData[field];
+    // Remove from the scalar payload regardless — .set() cannot handle collections.
+    delete scalarData[field];
+
+    if (value === undefined) {
+      // Field was not sent by the client; leave existing associations untouched.
+      collectionData[field] = undefined;
+    } else if (Array.isArray(value)) {
+      // Normalize: older modifiedDocJson entries may have stored populated objects.
+      collectionData[field] = value.map((item) =>
         typeof item === 'object' && item !== null ? (item.id ?? item) : item
       );
+    } else {
+      // Unexpected value type — treat as "untouched" to be safe.
+      collectionData[field] = undefined;
     }
   }
-  return normalized;
+
+  return { scalarData, collectionData };
 }
 
 async function validateAndUpdateDocument(
@@ -53,7 +67,8 @@ async function validateAndUpdateDocument(
     newFiles,
   } = document.modifiedDocJson;
 
-  const cleanedDocumentData = normalizeCollectionIds(documentData);
+  const { scalarData, collectionData } =
+    normalizeAndSplitDocumentData(documentData);
 
   await sails.getDatastore().transaction(async (db) => {
     // Update associated data not handled by TDocument manually
@@ -64,7 +79,7 @@ async function validateAndUpdateDocument(
 
     await TDocument.updateOne(document.id)
       .set({
-        ...cleanedDocumentData,
+        ...scalarData,
         modifiedDocJson: null,
         dateReviewed: new Date(),
         reviewer: reviewerId,
@@ -74,6 +89,18 @@ async function validateAndUpdateDocument(
         validator: validationAuthor,
       })
       .usingConnection(db);
+
+    // Replace m2m collections for every field that was explicitly sent by the
+    // client (including an empty array, which means "clear all").
+    // Fields not sent (undefined) are left untouched.
+    const collectionPromises = DOCUMENT_M2M_COLLECTIONS.filter(
+      (field) => collectionData[field] !== undefined
+    ).map((field) =>
+      TDocument.replaceCollection(document.id, field)
+        .members(collectionData[field])
+        .usingConnection(db)
+    );
+    await Promise.all(collectionPromises);
 
     const filePromises = [];
     // Files have already been created,
--- a/api/controllers/v1/document/multiple-validate.js
+++ b/api/controllers/v1/document/multiple-validate.js
@@ -93,14 +93,11 @@ async function validateAndUpdateDocument(
     // Replace m2m collections for every field that was explicitly sent by the
     // client (including an empty array, which means "clear all").
     // Fields not sent (undefined) are left untouched.
-    const collectionPromises = DOCUMENT_M2M_COLLECTIONS.filter(
-      (field) => collectionData[field] !== undefined
-    ).map((field) =>
-      TDocument.replaceCollection(document.id, field)
-        .members(collectionData[field])
-        .usingConnection(db)
+    await DocumentService.replaceM2MCollections(
+      document.id,
+      collectionData,
+      db
     );
-    await Promise.all(collectionPromises);
 
     const filePromises = [];
     // Files have already been created,
--- a/api/controllers/v1/document/update-with-new-entities.js
+++ b/api/controllers/v1/document/update-with-new-entities.js
@@ -1,3 +1,4 @@
+const DocumentService = require('../../../services/DocumentService');
 const NotificationService = require('../../../services/NotificationService');
 const {
   DOCUMENT_M2M_COLLECTIONS,
@@ -29,25 +30,6 @@ module.exports = async (req, res) => {
 
   const isArrNotEmpty = (value) => Array.isArray(value) && value.length > 0;
 
-  // For each associated entites :
-  // - check if there are new values
-  // - create the corresponding values
-  // - add the newly created values to the array of collectionData
-  if (isArrNotEmpty(newAuthors)) {
-    const authorParams = newAuthors.map((author) => ({
-      ...author,
-      documents: [documentId],
-    }));
-    const createdAuthors = await TCaver.createEach(authorParams).fetch();
-    const createdAuthorsIds = createdAuthors.map((author) => author.id);
-    // Merge newly created author ids into the authors collection.
-    // collectionData.authors may be undefined (not sent) — start from [] in that case.
-    collectionData.authors = [].concat(
-      collectionData.authors ?? [],
-      createdAuthorsIds
-    );
-  }
-
   if (isArrNotEmpty(newDescriptions)) {
     const missingLanguage = newDescriptions.some((desc) => !desc.language);
     if (missingLanguage) {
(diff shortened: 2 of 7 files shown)


Paul-AUB (changes requested):
Thanks for the thorough writeup and the transaction hardening — the `DOCUMENT_M2M_COLLECTIONS` single-source-of-truth and the `undefined`/`'[]'`/array protocol are a clean solution to the root Waterline problem described in #1736. A couple of gaps remain in the new `update-with-new-entities.js` path and in the swagger docs that are worth addressing before merge.

### Issues (Must Fix)

1. **[api/controllers/v1/document/update-with-new-entities.js:43-49]** When `newAuthors` is used to create additional authors without the client also sending `document.authors`, the merge silently wipes any authors that already exist on the document. `collectionData.authors` starts as `undefined` when the field isn't sent, so `[].concat(collectionData.authors ?? [], createdAuthorsIds)` becomes just the newly-created ids, and this list is no longer `undefined` — so it takes the `replaceCollection` path a few lines later and overwrites the whole `authors` join table with only the new ids, dropping every previously-linked author. The new Caver row is already correctly linked via `documents: [documentId]` at creation time (api/models/TCaver.js:182-186), so the subsequent `replaceCollection` call isn't even needed for that id — but as written it also discards everyone else. This is the exact "silent data loss" class of bug this PR sets out to fix, just reached through the `newAuthors` path instead of the `document.authors` field. Consider either requiring/merging against the document's *current* authors (fetched inside the transaction) when `newAuthors` is used without an explicit `document.authors`, or skip the `replaceCollection` overwrite for `authors` in that case and rely on `addToCollection` for the newly-created ids only.

2. **[assets/swaggerV1.yaml:1669-1697]** The doc for `PUT /documents/{id}/new-entities` states collection fields follow the same `'[]'` sentinel as `PUT /documents/{id}` ("send `'[]'` to clear it"), but `update-with-new-entities.js` never routes through `DocumentService.getConvertedDataFromClient`/`parseIdList` — `collectionData[field]` is taken verbatim from `req.body.document[field]`. If a client follows this doc and sends the literal string `'[]'`, `TDocument.replaceCollection(documentId, field).members('[]')` is called with a string instead of an empty array, which will fail or behave unexpectedly instead of clearing the collection. Since this endpoint is JSON-only (see the integration tests, all using `Content-type: application/json`), clients should just send a real empty array `[]`; the doc should say so instead of referencing the multipart-only `'[]'` sentinel.

3. **[assets/swaggerV1.yaml:1635-1636]** `PUT /documents/{id}` lists `languages` among the fields following the omit/`'[]'`/array convention, but `getConvertedDataFromClient` never reads a `languages` field at all — `languages` is derived from `documentMainLanguage`/`mainLanguage` (api/services/DocumentService.js:24-28), and `getDocumentLanguages` has no `'[]'` handling (an explicit empty string would need to be sent for `mainLanguage`, which isn't documented either). As written, the doc could mislead an API consumer into thinking a `languages` field exists on this endpoint. Worth either removing `languages` from that list or documenting the real `documentMainLanguage`/`mainLanguage` clearing behavior.

4. **[api/controllers/v1/document/update-with-new-entities.js:36-70,74-91]** `TCaver.createEach`/`TDescription.createEach` still run outside the new `sails.getDatastore().transaction()` block. The PR description states the transaction is meant "to prevent partial writes," but if the transaction (scalar update + `replaceCollection` calls) throws after these creates have already committed, the newly created caver/description rows are left behind uncommitted-by-rollback (the caver would even still be linked to the document via its own `documents` collection set at creation). Consider moving both `createEach` calls inside the transaction with `.usingConnection(db)` for a true all-or-nothing write.

5. **[api/controllers/v1/document/multiple-validate.js:96-103] / [api/controllers/v1/document/update-with-new-entities.js:83-90]** The `DOCUMENT_M2M_COLLECTIONS.filter(...).map(...) + Promise.all` block that applies `replaceCollection` is duplicated verbatim between the two controllers. Given the PR already centralizes the field list in `config/constants/document.js`, consider also centralizing this replace logic (e.g. a `DocumentService.replaceM2MCollections(id, collectionData, db)` helper) so future changes to the replace semantics only need to happen in one place.

### Nitpicks (Optional)

6. **[test/integration/4_routes/Documents/multiple-validate.test.js:216,223-224]** `createdMassifIds` is declared and cleaned up in the `after()` hook of the new "Collection replace behaviour" block, but no test in that block creates a `TMassif` or pushes to it — dead code that can be removed (or the `massifs` collection could actually be covered by a test, since it's one of the seven fields this PR fixes).

ClemRz:
Thanks for the thorough review @Paul-AUB. All items addressed in the latest commit (`2cbc519`):

### Issues (Must Fix)

1. **`newAuthors` without `document.authors` silently wiping existing authors** — Fixed. When `document.authors` is absent, the controller now calls `TDocument.addToCollection(documentId, 'authors', createdAuthorsIds)` inside the transaction instead of building a new array from `[]`. This appends the new ids without touching existing associations. Added a regression test to cover this path.

2. **Swagger `new-entities` referencing the `'[]'` sentinel** — Fixed. The endpoint description now correctly documents that, since it's JSON-only, clients send a real empty array `[]` to clear a collection. Explains why this differs from the multipart `PUT /documents/{id}` endpoint.

3. **Swagger `PUT /documents/{id}` listing `languages` as an m2m sentinel field** — Fixed. Removed `languages` from the sentinel list and added a note pointing to the real `documentMainLanguage.id` / `mainLanguage` fields.

4. **`createEach` calls outside the transaction** — Fixed. Both `TCaver.createEach` and `TDescription.createEach` now run inside the `sails.getDatastore().transaction()` block with `.usingConnection(db)`, making the whole operation truly atomic.

5. **Duplicated `replaceCollection` logic** — Fixed. Extracted to `DocumentService.replaceM2MCollections(id, collectionData, db)`. Both controllers now call the shared helper.

### Nitpicks

6. **Dead `createdMassifIds` variable** — Removed.

Paul-AUB (changes requested):
Thanks for the fast turnaround on the previous review round — items 2, 3, 5, and 6 all look correctly addressed (`new-entities` swagger no longer references the `'[]'` sentinel, the `PUT /documents/{id}` doc no longer lists `languages`, `replaceM2MCollections` is now shared, and the dead `createdMassifIds` var is gone). Item 4 (wrapping `createEach` in the transaction) is also done. However, the fix for item 1 introduces a new failure mode that's worth catching before merge.

### Issues (Must Fix)

1. **[api/controllers/v1/document/update-with-new-entities.js:68-77]** When `newAuthors` is provided and `document.authors` is absent, the code now calls `TDocument.addToCollection(documentId, 'authors', createdAuthorsIds)` — but `createEach` a few lines above already links the new caver to the document via `documents: [documentId]` in `authorParams` (line 51), and Waterline populates the through-table row(s) for a collection attribute at `.create()` time. The `j_document_caver_author` junction table has a composite primary key `(id_document, id_caver)` (`sql/0_tables.sql:1054`), so `addToCollection` then tries to insert a row that already exists, which will raise a primary-key violation and abort the transaction. This exact path is exercised by the new test `should preserve existing authors when newAuthors are added without document.authors` (`update-with-new-entities.test.js:258-287`), which sends `newAuthors` with no `document.authors` — it looks like it should currently fail with a 500 rather than the expected `200`. Since the association is already made at creation time, the `addToCollection` call in this branch looks unnecessary — consider dropping it (or only calling it for `newAuthors` ids when `document.authors` was explicitly sent, where `createEach`'s own linkage doesn't apply).

2. **[api/controllers/v1/document/update-with-new-entities.js:89-92]** `scalarData.descriptions` is built from newly created description ids and then passed to `TDocument.updateOne(documentId).set(scalarData)`. `descriptions` is a Waterline collection attribute (`api/models/TDocument.js:273-276`), so — same root cause as the bug this PR fixes for the seven m2m fields — `.set()` silently ignores it; the assignment is a no-op. It happens to be harmless today because `TDescription.createEach` already sets the `document` foreign key directly on each new row, but leaving dead code that mimics the exact anti-pattern this PR is stamping out elsewhere is confusing for future readers. Consider dropping the `scalarData.descriptions` assignment (or, if `document.descriptions` can ever be sent directly by a client to reorder/detach descriptions, handling it explicitly rather than relying on the silent no-op).

3. **[assets/swaggerV1.yaml:1717-1722]** The `newAuthors` description still says "Created ids are merged with `document.authors`," which no longer matches the current implementation when `document.authors` is absent (it now goes through `addToCollection` instead of a merge). Worth updating once the Issue 1 fix lands, so the doc matches the actual behavior for both branches.

ClemRz:
Second round addressed in the amended commit (`55532b6`):

### Issue (Must Fix)

1. **`addToCollection` causing PK violation** — Fixed. Removed the `addToCollection` call entirely. `TCaver.createEach` with `documents: [documentId]` already writes the junction-table row at creation time, so the extra call was redundant and would raise a composite PK violation on `j_document_caver_author`. When `document.authors` is absent, `collectionData.authors` stays `undefined` and `replaceM2MCollections` skips the field — existing authors are untouched, new cavers are linked at creation.

2. **Dead `scalarData.descriptions` assignment** — Removed. `descriptions` is a Waterline collection attribute; `.set()` silently ignores it. The new descriptions are already linked via the `document` FK set in `descParams`. Added a comment to explain why we deliberately don't set it.

3. **Swagger `newAuthors` description stale** — Updated to document both branches: when `document.authors` is present the new ids are merged before `replaceCollection` runs; when absent, existing authors are left untouched and only the new cavers are added via their own junction rows.

**Bonus fix:** Moving `TDescription.createEach` to after `TDocument.updateOne` inside the transaction. Creating a description with a `document` FK before updating the document row causes a deadlock within the same Waterline transaction (FK resolution races with the pending row lock). The document update now runs first to hold the lock, then descriptions are created safely.

Paul-AUB (approved):
Re-reviewed after the two rounds of fixes (`2cbc519`, `55532b6`). All items raised in my previous review are addressed:

- **`newAuthors` silently wiping existing authors** — fixed. `collectionData.authors` is only merged with `createdAuthorsIds` when the client explicitly sent `document.authors`; otherwise it stays `undefined` so `replaceM2MCollections` skips the field and existing associations survive. The earlier `addToCollection` follow-up (which would have caused a PK violation on `j_document_caver_author` since `TCaver.createEach` with `documents: [documentId]` already writes the junction row at creation) was correctly removed rather than patched. Covered by the new "should preserve existing authors when newAuthors are added without document.authors" test.
- **`new-entities` swagger wrongly referencing the `'[]'` sentinel** — fixed; the doc now correctly states this JSON-only endpoint takes a real empty array.
- **`languages` listed as a sentinel field on `PUT /documents/{id}`** — removed from that list; verified `getConvertedDataFromClient` never reads a raw `languages` field on that (multipart) endpoint, it's derived from `documentMainLanguage`/`mainLanguage` via `getDocumentLanguages`, which the swagger text now documents.
- **`createEach` calls outside the transaction** — both `TCaver.createEach` and `TDescription.createEach` now run with `.usingConnection(db)` inside `sails.getDatastore().transaction`, and the description creation was correctly reordered to after the `TDocument.updateOne` to avoid the FK-resolution deadlock described in the commit message.
- **Duplicated `replaceCollection` logic** — extracted to `DocumentService.replaceM2MCollections`, used by both controllers.
- **Dead `createdMassifIds` variable** — no longer present.

I also traced `DOCUMENT_M2M_COLLECTIONS` end-to-end for both write paths (`multiple-validate.js` and `update-with-new-entities.js`) and confirmed `languages` is a genuine m2m collection on `TDocument` (`api/models/TDocument.js:278-282`, through `JDocumentLanguage`), so including it in the constant and in the JSON-only `new-entities` swagger doc is correct — it's just intentionally absent from `getConvertedDataFromClient`'s multipart parsing since that endpoint only exposes it indirectly via `documentMainLanguage`/`mainLanguage`. No new issues found; the transaction wrapping, the `undefined`/`'[]'`/array protocol, and the new integration tests all line up correctly with the described behavior.

Nothing further to flag — looks good to merge on the backend side (pending the companion front-end PR mentioned in the description).

Issue #1745, opened 2026-07-30
Make document authors searchable when the author is an organization


Paul-AUB:
# Make document authors searchable when the author is an organization

**Repository concerned:** `GrottoCenter/grottocenter-api`
**Type:** enhancement / search index gap

---

## Summary

A document can be authored by **persons** (`authors`) and/or by **organizations**
(`authorsOrganization`). The REST document endpoints return both, and the
frontend displays both. But the **document search index only contains
`authors`** — organization authors are absent from it entirely.

Consequence: a document authored solely by an organization (very common for
bulletins, newsletters and society publications) is **unreachable by author
search**, and its author appears **blank** in the search results table.

We had to remove the "Organization author" filter we had built in the frontend
because there is no field to query. We would like to put it back.

---

### 1. No organization-author field exists in the schema

Every plausible field name is rejected by `POST /api/v1/field-search`:

```bash
curl -s -X POST "https://api.grottocenter.org/api/v1/field-search" \
  -H "Content-Type: application/json" \
  -d '{"entity":"documents","field":"authorsOrganization.name","query":"a","filter":{},"size":5}'
```

```json
{"error":"Request failed with HTTP code 404 | Server said: Could not find a field named `authorsOrganization.name` in the schema."}
```

Same error for `authorsGrotto.name` and `organizations.name`.
For comparison, `editor.name` and `authors.nickname` both return `200`.

### 2. The index really has no such field

`POST /api/v1/advanced-search` with `{"entity":"documents"}` returns documents
whose complete key set is:

```
@context, @id, @type, _highlights, _type, authors, cave, creator,
creatorComment, dateInscription, datePublication, dateReviewed,
dateValidation, description, editor, entrances, id, iso3166, language,
languages, library, license, massifs, oldBBS, parent, reviewer, title,
type, validator
```

There is no `authorsOrganization`, no `authorsGrotto`, no organization field of
any kind. Indexed shapes today:

- `authors` → `[{ id, nickname }]`
- `editor` → `{ name, … }` (an organization — so organizations *are* indexable,
  just not in the author role)

### 3. Second, less obvious impact: the results table

The frontend's document results column merges both author kinds
(`packages/web-app/src/components/common/EntityTable/entitiesConfig.jsx`,
`cellsRender.documentAuthors`, wired to the `Author` column):

```js
const names = [
  ...(doc.authors ?? []).map(a => a.nickname),
[...]

Issue #1729, opened 2026-07-19
Complete polymorphic document authors support


Paul-AUB:
# Backend spec — Complete polymorphic document authors support

Companion to [issue.md](https://github.com/GrottoCenter/grottocenter-front/issues/1419#issuecomment-5004658926) (frontend spec for [GrottoCenter/grottocenter-front#1419](https://github.com/GrottoCenter/grottocenter-front/issues/1419)).

> **TL;DR** — The core feature (persons + organizations as document authors) ships **front-only** with zero backend changes. This document lists three optional-but-recommended backend improvements to make the feature "perfect": search parity, an atomic-create convenience, and a data cleanup. Items are independent and can be shipped separately.

---

## Context

The frontend is adding a Zotero-style polymorphic picker that lets a document have both person-authors (via existing `authors[]` / `TCaver`) and organization-authors (via existing `authorsGrotto[]` / `TGrotto`). The full model + join tables + create/update/read/delete paths are already in place — see [issue.md §1](https://github.com/GrottoCenter/grottocenter-front/issues/1419#issuecomment-5004658926).

Three known gaps remain server-side, ordered by priority below.

---

### Problem

`DocumentService.updateInSearch` only ships person-authors to the Typesense `documents` collection. Documents whose only author is an organization are invisible when searching by author name. This also blocks the frontend's "Author" search filter (see [issue.md §7.3](https://github.com/GrottoCenter/grottocenter-front/issues/1419#issuecomment-5004658926)) from working on org-authors.

### Files to change

**1. `api/services/DocumentService.js` — incremental sync**

Around line 29-88 (`updateInSearch`):

- Destructure `authorsGrotto` from `populatedDocument` alongside `authors`.
- Add to the emitted `document`: `authorsOrganization: authorsGrotto?.map((e) => ({ name: e.names?.[0]?.name }))` (mirrors how `editor.name` is resolved at line 75).

**2. `api/dbSync/entities/document.js` — full reindex**

- Add a join block after the existing `j_document_caver_author` block (line 115-123) for `j_document_grotto_author`:

```js
{
  table: 'j_document_grotto_author',
  foreignField: 'id_document',
  rows,
  localField: 'authorsOrganization',
  fields: ['g.id', 'n.name'],
  where: [],
  join: [
    'LEFT JOIN t_grotto g ON g.id = id_grotto',
    'LEFT JOIN t_name n ON n.id_grotto = g.id AND n.is_main = true',
  ],
},
```

[...]

PR #1753, opened 2026-08-04
feat(document): index authorsOrganization in Typesense and rename from authorsGrotto

ClemRz:
## 🤔 What

- Rename the Waterline attribute `authorsGrotto` → `authorsOrganization` on `TDocument`, and update every reference across services, controllers, CSV importer, converters, constants, tests, and Swagger (issues #1729 item E, #1745)
- Index `authorsOrganization` in Typesense — incremental sync via `DocumentService.updateInSearch` and full-reindex via `api/dbSync/entities/document.js` (issue #1729 item A)
- Expose `authorsOrganization` in advanced-search response payloads by removing the `delete data.authorsOrganization` line from `toSearchResult` (issue #1729 item A, #1745)
- Add a SQL migration to rename the misnamed FK constraint `j_document_grotto_author_t_caver_fk` → `j_document_grotto_author_t_grotto_fk` (issue #1729 item D)
- Update Swagger: `authorsOrganization` in the `Document` response schema and POST/PUT request body parameters

Closes items A, D, E from #1729. Remaining items B, C, F are tracked in that issue.

## 🤷‍♂️ Why

The field was internally named `authorsGrotto` (after the `t_grotto` table) while the public API and the product domain use "organization". This asymmetry made the codebase inconsistent and the field was never indexed in Typesense, meaning organization authors were invisible in search results and the Author column was blank in the document results table.

## 🔍 How

**Rename:** One-line change in `TDocument.js` plus a mechanical find-and-replace across all files that referenced `authorsGrotto`. The junction table (`j_document_grotto_author`) and the `JDocumentGrottoAuthor` model are untouched — only the Waterline attribute name and all code referencing it change.

**Typesense indexing:**
- `updateInSearch` now destructures `authorsOrganization` from the populated document and maps each grotto to `{ name: e.names?.[0]?.name }`. The `?? []` fallback ensures Typesense never receives `null` for a `string[]` field.
- `dbSync/entities/document.js` gets a new join block on `j_document_grotto_author` → `t_grotto` → `t_name` (using `n.is_main = true`) and a new schema field `authorsOrganization.name` (`string[]`, `facet: true`, `optional: true`).

**Search results:** removing `delete data.authorsOrganization` from `toSearchResult` in `converters.js` is the entire implementation of exposing the field.

**SQL migration:** wrapped in a `DO $$ BEGIN IF EXISTS ... END $$` block so it is idempotent and matches the project convention.

**Tests:**
- Updated all existing `authorsGrotto` references in `DocumentService.test.js` and route tests
- New property test (`DocumentService.property.test.js`) for the `updateInSearch` mapping invariant using `fast-check`
- New integration test (`advanced-search-document-authors-organization.test.js`) stubbing `SearchService.collectionSearch` to verify the field survives `toSearchResult`
- New test in `update-with-new-entities.test.js` asserting stale clients sending `authorsGrotto` have no effect

## 🧪 Testing

After merging and deploying, a full Typesense reindex is required to make **existing** documents with organization authors searchable (and to fix the null org-author names visible in the validation page):

```bash
node api/dbSync
```

Incremental sync covers only documents validated/restored after the deploy.

## 📸 Previews

N/A — backend-only change.

### DIFF
--- a/api/controllers/v1/document/delete.js
+++ b/api/controllers/v1/document/delete.js
@@ -85,7 +85,7 @@ module.exports = async (req, res) => {
       languages: 'j_document_language',
       massifs: 'j_document_massif',
       subjects: 'j_document_subject',
-      authorsGrotto: 'j_document_grotto_author',
+      authorsOrganization: 'j_document_grotto_author',
       authors: 'j_document_caver_author',
     };
 
@@ -130,7 +130,7 @@ module.exports = async (req, res) => {
     await linkedEntitiesDeleteOrMerge('languages');
     await linkedEntitiesDeleteOrMerge('massifs');
     await linkedEntitiesDeleteOrMerge('subjects');
-    await linkedEntitiesDeleteOrMerge('authorsGrotto');
+    await linkedEntitiesDeleteOrMerge('authorsOrganization');
     await linkedEntitiesDeleteOrMerge('authors');
 
     await sails.sendNativeQuery(
--- a/api/dbSync/entities/document.js
+++ b/api/dbSync/entities/document.js
@@ -117,9 +117,21 @@ async function* processRows(source) {
         foreignField: 'id_document',
         rows,
         localField: 'authors',
-        fields: ['c.id', 'c.nickname'],
+        fields: ['c.nickname'],
         where: [],
         join: ['LEFT JOIN t_caver c ON c.id = id_caver'],
+      },
+      {
+        table: 'j_document_grotto_author j',
+        foreignField: 'j.id_document',
+        rows,
+        localField: 'authorsOrganization',
+        fields: ['n.name'],
+        where: [],
+        join: [
+          'LEFT JOIN t_grotto g ON g.id = j.id_grotto',
+          'LEFT JOIN t_name n ON n.id_grotto = g.id AND n.is_main = true',
+        ],
       },
       {
         table: 'j_document_language',
@@ -276,6 +288,12 @@ module.exports = {
         { name: 'subjects.code', type: 'string[]', optional: true },
         {
           name: 'authors.nickname',
+          type: 'string[]',
+          facet: true,
+          optional: true,
+        },
+        {
+          name: 'authorsOrganization.name',
           type: 'string[]',
           facet: true,
           optional: true,
--- a/api/models/TDocument.js
+++ b/api/models/TDocument.js
@@ -147,7 +147,7 @@ module.exports = {
       model: 'TGrotto',
     },
 
-    authorsGrotto: {
+    authorsOrganization: {
       collection: 'TGrotto',
       via: 'document',
       through: 'JDocumentGrottoAuthor',
--- a/api/services/DocumentCSVImportService.js
+++ b/api/services/DocumentCSVImportService.js
@@ -200,7 +200,7 @@ module.exports = {
       author: authorId,
       // authorComment: '', // TODO Add ?
       authors: creatorsCaverId,
-      authorsGrotto: creatorsGrottoId,
+      authorsOrganization: creatorsGrottoId,
       editor: editorId,
       // library: '', // TODO Add ?
 
--- a/api/services/DocumentService.js
+++ b/api/services/DocumentService.js
@@ -14,6 +14,11 @@ const { DOCUMENT_M2M_COLLECTIONS } = require('../../config/constants/document');
 const normalizeToId = (item) =>
   item != null && typeof item === 'object' ? (item.id ?? item) : item;
 
+// Maps a populated authorsOrganization array to the Typesense-ready shape.
+// Exported so property tests can exercise the actual function rather than a copy.
+const mapAuthorsOrganizationForSearch = (orgs) =>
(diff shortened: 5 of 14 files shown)


Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[code-review.md](code-review.md) (new file, 82 lines)** — This PR adds a `code-review.md` at the repo root that is itself an AI-generated code review of this very commit (it opens with `# Code Review — feat(document): index authorsOrganization...` and references `Commit: 789009f...`). This looks like a local review artifact that was accidentally `git add`-ed instead of being discarded. It isn't documentation and shouldn't live in version control — please remove it before merging.

2. **[test/integration/1_services/DocumentService.property.test.js:4-5](test/integration/1_services/DocumentService.property.test.js#L4-L5)** — `mapAuthorsOrganization` is a hand-copied duplicate of the mapping logic in `DocumentService.js`'s `updateInSearch` (`authorsOrganization?.map((e) => ({ name: e.names?.[0]?.name })) ?? []`, see [api/services/DocumentService.js:86-87](api/services/DocumentService.js#L86-L87)), not imported from the source. The comment even says "Extract the pure mapping function from updateInSearch" but no extraction happened. As written, the property test only validates that the duplicate stays internally consistent with itself — if `updateInSearch`'s real mapping ever changes (e.g. a different fallback or an extra field), this test keeps passing while the claimed "mapping invariant" silently breaks in production. Consider exporting the mapping function from `DocumentService.js` (or an internal helper) and importing it here so the test exercises the actual code path.

3. **[test/integration/1_services/DocumentService.property.test.js:26-34](test/integration/1_services/DocumentService.property.test.js#L26-L34)** — The third test uses `fc.constant([])` as the sole input generator, so `fc.assert` runs the identical case 100 times with no variation — equivalent to a single plain assertion. Consider a plain `it()` here instead of wrapping a constant in a property test.

4. **Backward compatibility of the write field rename** — `getConvertedDataFromClient` ([api/services/DocumentService.js:165](api/services/DocumentService.js#L165)) now only reads `body.authorsOrganization`; a client still sending the old `authorsGrotto` field on create/update gets no error and the field is silently dropped (confirmed by the new test in `update-with-new-entities.test.js`). Issue #1729's Item E proposal suggested keeping `authorsGrotto` as a deprecated alias for one release specifically to avoid this. If the frontend companion change and this API change are guaranteed to deploy atomically, this is fine — but if there's any window where an old frontend build or a third-party API consumer could still send `authorsGrotto`, that data will be silently lost rather than surfaced as an error. Worth confirming the rollout is coordinated.

### Nitpicks (Optional)

5. **PR description vs. code** — The PR description states the `t_name` join in `api/dbSync/entities/document.js` uses `n.is_main IS NOT DISTINCT FROM true`, but the shipped code uses `n.is_main = true` ([api/dbSync/entities/document.js:133](api/dbSync/entities/document.js#L133)). This actually matches the existing convention used by every sibling join in the same file (editor, library, cave, entrances, massifs all use `= true`), so no code change is needed — just flagging the description doesn't match what was implemented.

6. **[assets/swaggerV1.yaml](assets/swaggerV1.yaml)** — The new `authorsOrganization` query parameter documents `id` as `type: integer`, while the neighboring `authors` parameter documents `id` as `type: string`. Likely intentional (different underlying id types), but worth a quick sanity check for consistency.

Paul-AUB (dismissed):
Re-reviewed at `37f65c8` (amended commit, replaces `51b4c7a`).

### Resolved

- **`code-review.md` removed.** The stray AI-generated review artifact at the repo root is gone. This was the only Must Fix item, so the blocker is cleared.

### Bonus fix noticed (not something I had flagged, but worth calling out)

- **[api/dbSync/entities/document.js:120-129](api/dbSync/entities/document.js#L120-L129)** — the full-reindex joins for `authors` and `authorsOrganization` now select only `nickname`/`name`, dropping `c.id` / `j.id_grotto AS id`. This brings the full-reindex path back in line with `updateInSearch`'s incremental mapping (which never included an `id` for either field) and with the Typesense schema (only `authors.nickname` and `authorsOrganization.name` are declared — no `.id` fields). Good catch, resolves a pre-existing inconsistency between the two indexing paths.

### Still open (non-blocking, optional follow-ups from the previous review)

None of these were addressed in this update — restating for the record, but they don't need to hold up merging:

- **[test/integration/1_services/DocumentService.property.test.js:4-5](test/integration/1_services/DocumentService.property.test.js#L4-L5)** — `mapAuthorsOrganization` is still a hand-copied duplicate of `updateInSearch`'s mapping logic rather than imported from `DocumentService.js`, so the property test doesn't actually exercise production code.
- **[test/integration/1_services/DocumentService.property.test.js:26-34](test/integration/1_services/DocumentService.property.test.js#L26-L34)** — third test still uses `fc.constant([])`, giving no property coverage beyond a single assertion.
- **Backward compatibility** — `authorsGrotto` is still silently dropped with no deprecated alias on write; only relevant if the frontend/API rollout isn't guaranteed atomic.

Approving — the blocking issue is resolved and the remaining points are minor/optional.

PR #1770, opened 2026-08-07
fix(massif): set reviewer on update to fix recent-changes feed

### LINKED ISSUE #1769: Recent changes: massif edits always appear as 'created' due to missing reviewer in update controller
## Bug description

Massif edits always appear as "created" in the Recent Changes feed (`GET /api/v1/changes/recent`) instead of "updated".

## Root cause

`api/controllers/v1/massif/update.js` never sets `reviewer` on the row it updates:

```js
// massif/update.js
const cleanedData = MassifService.getConvertedDataFromClientRequest(req);
// ...
await TMassif.updateOne(massifId).set(cleanedData);  // reviewer never included
```

`getConvertedDataFromClientRequest` for massif only returns `caves`, `descriptions`, `documents`, `geogPolygon`, `names`, `isSensitive`, `isSensitiveLocked` — no `reviewer`.

Every other entity update controller sets it explicitly:

```js
// cave/update.js, entrance/update.js, organization/update.js, rigging/update.js, etc.
const updatedFields = {
  reviewer: req.token.id,
  ...
};
```

The `change_massif` database trigger (and equivalents for all other entities) uses `id_reviewer IS NULL` to distinguish a row that has never been edited from one that has:

```sql
elsif NEW.is_deleted = false AND NEW.id_reviewer is null then
    type_change := 'create';   -- row has never been updated
elsif NEW.is_deleted = false then
    type_change := 'update';   -- row has been updated at least once (reviewer is set)
```

This is intentional: the first time someone updates a freshly-created entity, it still shows as "created" in the feed until the `reviewer` field is populated. All other entities correctly transition to `update` after their first edit because their update controller sets `reviewer`. Massif never does, so `id_reviewer` stays NULL permanently, and every update — first or hundredth — is classified as `create`.

## Evidence

Massif id=6481 (Luberon), created 2026-07-01, has 8 `create` entries in `t_last_change` all dated 2026-08-07. Azure Log Analytics confirms 8 legitimate `PUT /api/v1/massifs/6481` calls by user id=60181 (Paul AUBERTIN) at 00:21 UTC, all returning 200. The `h_massif` history table confirms the polygon was unchanged across all 8 snapshots — `is_sensitive_locked` was the only toggling field, confirming these were real edits, not phantom writes.

`t_massif` for id=6481 has `id_reviewer = NULL` despite multiple edits, which would never happen for any other entity type.

## Fix

Add `reviewer: req.token.id` to the update payload in `api/controllers/v1/massif/update.js`, consistent with every other entity:

```js
const cleanedData = {
  reviewer: req.token.id,
  ...MassifService.getConvertedDataFromClientRequest(req),
};
```

ClemRz:
Fixes massif edits always appearing as "created" in the recent-changes feed.

## Why

The `change_massif` DB trigger uses `id_reviewer IS NULL` to decide whether a row has ever been edited:

```sql
elsif NEW.is_deleted = false AND NEW.id_reviewer is null then
    type_change := 'create';
elsif NEW.is_deleted = false then
    type_change := 'update';
```

`massif/update.js` never set `reviewer`, so `id_reviewer` stayed `NULL` permanently. Every edit — first or hundredth — was classified as `create` in the feed. All other entity update controllers (cave, entrance, organization, rigging, etc.) already set `reviewer: req.token.id`.

## What

- Add `reviewer: req.token.id` to the update payload in `api/controllers/v1/massif/update.js`
- Add a reviewer assertion to the existing update integration test

## Related

- Closes #1769

### DIFF
--- a/api/controllers/v1/massif/update.js
+++ b/api/controllers/v1/massif/update.js
@@ -12,7 +12,10 @@ module.exports = async (req, res) => {
     return res.notFound({ message: `Massif of id ${massifId} not found.` });
   }
 
-  const cleanedData = MassifService.getConvertedDataFromClientRequest(req);
+  const cleanedData = {
+    reviewer: req.token.id,
+    ...MassifService.getConvertedDataFromClientRequest(req),
+  };
 
   // conversion of geoJson into PostGis Geom
   if (cleanedData.geogPolygon) {
--- a/api/controllers/v1/massif/update.js
+++ b/api/controllers/v1/massif/update.js
@@ -13,8 +13,9 @@ module.exports = async (req, res) => {
   }
 
   const cleanedData = {
-    reviewer: req.token.id,
+    // dateReviewed will be updated automatically by the SQL historisation trigger
     ...MassifService.getConvertedDataFromClientRequest(req),
+    reviewer: req.token.id,
   };
 
   // conversion of geoJson into PostGis Geom
@@ -63,21 +64,19 @@ module.exports = async (req, res) => {
 
   // Handle name update inline (consistent with entrance and cave update)
   if (req.body.name) {
-    const nameUpdate = {};
+    const nameUpdate = { reviewer: req.token.id };
     if (nameText !== undefined) {
       nameUpdate.name = nameText;
     }
     if (nameLanguage !== undefined) {
       nameUpdate.language = nameLanguage;
     }
-    if (Object.keys(nameUpdate).length > 0) {
-      const updatedName = await TName.updateOne({
-        massif: massifId,
-        isMain: true,
-      }).set(nameUpdate);
-      if (!updatedName) {
-        sails.log.warn(`Massif ${massifId} has no main name record to update.`);
-      }
+    const updatedName = await TName.updateOne({
+      massif: massifId,
+      isMain: true,
+    }).set(nameUpdate);
+    if (!updatedName) {
+      sails.log.warn(`Massif ${massifId} has no main name record to update.`);
     }
   }
 
--- a/test/integration/4_routes/Massifs/update.test.js
+++ b/test/integration/4_routes/Massifs/update.test.js
@@ -118,6 +118,56 @@ describe('Massif features', () => {
         });
     });
 
+    it('should appear as update (not create) in the recent-changes feed after a second PUT (#1769)', async () => {
+      // Regression: before the fix, reviewer was never set so id_reviewer stayed NULL,
+      // and the change_massif trigger always classified edits as 'create'.
+      //
+      // This test creates a massif with reviewer=null (simulating a freshly created
+      // entity that has never been edited), issues two consecutive PUTs, and asserts
+      // that reviewer is set after the first PUT and remains set after the second.
+      // The change_massif DB trigger uses id_reviewer IS NULL to decide 'create' vs
+      // 'update' — so a populated reviewer is both necessary and sufficient for
+      // subsequent edits to be classified as 'update' in the feed.
+      const stub = sinon.stub(EntranceService, 'updateInSearch').resolves();
+      let nullReviewerMassifId;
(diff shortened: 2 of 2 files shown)


Paul-AUB (dismissed):
Verified the fix against the actual trigger definition and the surrounding code paths:

- `sql/0_triggers.sql:222-227` — `change_massif()` classifies as `create` exactly when `id_reviewer IS NULL`, so populating `reviewer` is the right fix, and it matches the `entrance/update.js:19-23` pattern precisely.
- `sql/0_triggers.sql:203` — `histo_update_massif()` sets `NEW.date_reviewed := now()` itself, so not setting `dateReviewed` in the controller is correct.
- `MassifService.getConvertedDataFromClientRequest` (`api/services/MassifService.js:224-232`) returns an explicit whitelist with no `reviewer` key, so the spread cannot clobber the new field.
- `config/policies.js:240` — `v1/massif/update` is behind `tokenAuth`, so `req.token.id` is always defined.
- `converters.js` uses `convertIfObject(source.reviewer, …)`, so massif responses now returning a populated reviewer instead of `null` is a safe, intended change.
- Other massif write paths are already correct: `MassifService.setSensitivity` (`api/services/MassifService.js:436-440`) sets `reviewer` + `dateReviewed`, and `create.js` correctly leaves `reviewer` null so the first feed entry stays `create`.

The test fixture creates the massif with `reviewer: 2` while the acting user is `user1@user1.com`, so the new assertion is not vacuous — it genuinely proves the reviewer was reassigned by the update.

Nice, minimal, well-targeted fix. A few non-blocking notes below.

1. **[api/controllers/v1/massif/update.js:74-81]** The inline `TName.updateOne(...)` block does not set `reviewer`, whereas the equivalent block in `entrance/update.js:97-101` does. `histo_update_name()` copies `NEW.id_reviewer` into `h_name` (`sql/0_triggers.sql:908,923`), so renaming a massif currently writes a history row attributed to the previous reviewer (or `NULL`). There is no `last_change_name` trigger, so this has no effect on the recent-changes feed and is not a regression from this PR — but it is the same attribution gap the PR is closing, and it might be worth fixing in the same pass:

   ```js
   const nameUpdate = { reviewer: req.token.id };
   ```

2. **[test/integration/4_routes/Massifs/update.test.js:112-115]** The assertion proves the column is populated, which is a good proxy, but it does not cover the behaviour the issue actually describes — that a *second* edit is classified as `update` rather than `create` in the feed. Consider a regression test that issues two consecutive `PUT`s and asserts the resulting `t_last_change` rows (or the `/api/v1/changes/recent` payload) contain an `update` entry for the massif. `test/integration/4_routes/Changes/get-recent.test.js` already exercises that endpoint and could serve as a starting point. This would lock in the fix even if someone later refactors the payload construction.

### Nitpicks (Optional)

3. **[api/controllers/v1/massif/update.js:15-18]** `cave/update.js:26-28`, `entrance/update.js:20-21` and `rigging/update.js:23-24` all carry a short comment next to `reviewer` explaining that `dateReviewed` is handled by the SQL historisation trigger. Adding the same line here would keep the four controllers visually consistent and preempt the "shouldn't we also set dateReviewed?" question on a future read:

   ```js
   const cleanedData = {
     reviewer: req.token.id,
     // dateReviewed will be updated automatically by the SQL historisation trigger
     ...MassifService.getConvertedDataFromClientRequest(req),
   };
   ```

4. **[api/controllers/v1/massif/update.js:15-18]** Placing `reviewer` before the spread matches `entrance/update.js`, so the current ordering is defensible on consistency grounds. That said, `organization/update.js:18-22` puts it after the spread, which makes the field structurally impossible to shadow if `getConvertedDataFromClientRequest` ever gains a `reviewer` key. Purely defensive — there is no client-controlled override path today since the service returns a fixed whitelist.

Issue #1772, opened 2026-08-16
Reordering a newly created comment corrupts the recent changes feed


Paul-AUB:
**Describe the bug**

Creating a comment on an entrance, moving it above existing comments, and then editing it can corrupt the semantics of the recent changes feed:

- the contributor who created the new comment is reported as having `changed` a comment instead of having created it;
- displaced existing comments are reported as newly `created` by their original authors at the time of the reorder, even though those users took no action.

The frontend does not infer these actions: it renders `mainAction` / `subAction` returned by `GET /api/v1/changes/recent`.

<img width="807" height="167" alt="Image" src="https://github.com/user-attachments/assets/50d933c8-2f84-45c3-9bf4-fd2f33f4d757" />

**Validated production example**

This was reproduced/observed on entrance [6099 - Guiers Mort (Grotte-exsurgence du)](https://fr.wikicaves.org/ui/entrances/6099) on 2026-08-16.

The public recent-changes endpoint returned:

```json
[
  {
    "date": "2026-08-16T15:00:52.196Z",
    "author": "Paul AUBERTIN",
    "mainEntityId": 6099,
    "subEntityTypes": ["comment"],
    "subAction": "change"
  },
  {
    "date": "2026-08-16T14:57:29.114Z",
    "author": "Soufflet Benjamin",
    "mainEntityId": 6099,
    "subEntityTypes": ["comment"],
    "subAction": "create"
  },
  {
    "date": "2026-08-16T14:57:27.300Z",
    "author": "BTH",
    "mainEntityId": 6099,
    "subEntityTypes": ["comment"],
    "subAction": "create"
  }
]
```

However, `GET /api/v1/entrances/6099` shows that:

- comment `4639` is the actual new comment (`dateInscription: 2026-08-16T14:57:11.835Z`), authored by Paul AUBERTIN, moved to `relevance: 1`, then edited at `15:00:52.196Z`;
- comment `2685`, authored by Soufflet Benjamin, was actually created in 2015;
- comment `3133`, authored by BTH, was actually created in 2017.

The `14:57:27` and `14:57:29` events match the relevance swaps of the existing comments, not their creation dates.

**To reproduce**

1. Use an entrance with at least two existing comments authored by other users and with no reviewer.
2. Create a new comment.
3. Move the new comment upward across the existing comments using the relevance controls.
4. Edit the new comment within the six-hour recent-change grouping window.
[...]

PR #1776, opened 2026-08-18
fix(recentChanges): relevance swap corrupts recent changes feed

ClemRz:
Fixes the three bugs described in #1772 that caused relevance swaps to corrupt the recent changes feed.

## Why

Reordering a sub-entity (comment, description, rigging, etc.) via the relevance controls produced two symptoms visible in `GET /api/v1/changes/recent`:

- Displaced old sub-entities appeared as `create` events attributed to their original authors, even though those authors took no action.
- The real creation of the newly-reordered sub-entity was obscured and surfaced as `change` instead of `create`.

## What

**`sql/1_2026_08_17_fix_change_triggers.sql`** — incremental migration replacing all ten `change_*()` trigger functions with two fixes:

1. Replace `NEW.id_reviewer IS NULL` with `TG_OP = 'INSERT'` as the `create` guard. An UPDATE caused by a relevance swap is never an INSERT, so it no longer misclassifies as `create`. Mirrors the already-correct `change_guideline()` pattern.
2. Add a `current_setting('app.relevance_swap_skip_log', true)` guard at the top of each function. `RelevanceService` sets this flag via `SET LOCAL` for the neighbor-only update inside the swap transaction, so the involuntary side-effect produces no `t_last_change` entry at all. Same pattern as the pre-existing `app.is_enrichment` flag in `EnrichmentQueueService`.

**`api/services/RelevanceService.js`** — the neighbor relevance update now uses raw SQL with `SET LOCAL app.relevance_swap_skip_log = 'true'` so the trigger skips logging for it. The target entity's update still goes through Waterline and is logged normally.

**`api/services/RecentChangeService.js`** — `groupChanges()` gains same-entity priority logic for sub-actions: when the same sub-entity id produced both a `create` and a later `update` within the grouping window (i.e. created then edited), `create` takes priority over `update`. Mixed actions across different sub-entity ids still correctly resolve to `change`.

**`api/models/TLastChange.js`** — fixed `entityRelatedType` from `type: 'number'` to `type: 'string'` / `columnType: 'varchar(30)'` to match the production schema.

**Test files** — regression test covering the full scenario (create → move × 2 → edit → assert recent changes), fixture comments for entrance 999, and `CREATE_COMMENT_TRIGGERS` in `customSQL.js` to install `change_comment` / `last_change_comment` in the test DB.

## Related

- Issue: https://github.com/GrottoCenter/grottocenter-api/issues/1772

### DIFF
--- a/api/models/TLastChange.js
+++ b/api/models/TLastChange.js
@@ -32,9 +32,9 @@ module.exports = {
     },
 
     entityRelatedType: {
-      type: 'number',
+      type: 'string',
       columnName: 'type_related_entity',
-      columnType: 'int4',
+      columnType: 'varchar(30)',
     },
 
     entityRelatedId: {
--- a/api/services/RecentChangeService.js
+++ b/api/services/RecentChangeService.js
@@ -63,6 +63,10 @@ function groupChanges(changes) {
     mainAction: c.id_related_entity ? null : c.type_change, // Null in case it is a change on a sub entity
     subEntityTypes: c.id_related_entity ? [c.type_entity] : [],
     subAction: c.id_related_entity ? c.type_change : null, // Null when no sub entities
+    // Tracks the entity id that last set subAction, so we can apply
+    // create/restore priority correctly when the same entity is later
+    // seen with a higher-priority action (e.g. create after update).
+    subActionEntityId: c.id_related_entity ? c.id_entity : null,
     name: c.name, // Can be the real entity name or the related entity name
   });
 
@@ -75,8 +79,23 @@ function groupChanges(changes) {
       g.mainAction = c.type_change; // eslint-disable-line no-param-reassign
     if (c.id_related_entity && !g.subEntityTypes.includes(c.type_entity))
       g.subEntityTypes.push(c.type_entity);
-    if (c.id_related_entity && g.subAction !== c.type_change)
-      g.subAction = 'change'; // eslint-disable-line no-param-reassign
+    if (c.id_related_entity && g.subAction !== c.type_change) {
+      // 'create' and 'restore' take priority over 'update' only when it is the
+      // same entity that produced both events (e.g. a sub-entity was created
+      // then edited within the grouping window).  When different entity ids are
+      // involved the author genuinely performed mixed actions → 'change'.
+      if (
+        g.subAction === 'update' &&
+        ['create', 'restore'].includes(c.type_change) &&
+        g.subActionEntityId === c.id_entity
+      ) {
+        g.subAction = c.type_change; // eslint-disable-line no-param-reassign
+        g.subActionEntityId = c.id_entity; // eslint-disable-line no-param-reassign
+      } else {
+        g.subAction = 'change'; // eslint-disable-line no-param-reassign
+        g.subActionEntityId = null; // eslint-disable-line no-param-reassign
+      }
+    }
   };
 
   for (const change of changes) {
@@ -106,7 +125,9 @@ function groupChanges(changes) {
   }
 
   const allGroups = Object.values(authorChanges).flat();
-  return allGroups.sort((a, b) => b.date - a.date);
+  return allGroups
+    .sort((a, b) => b.date - a.date)
+    .map(({ subActionEntityId, ...group }) => group); // strip internal tracking field
 }
 
 async function getRecent() {
--- a/api/services/RecentChangeService.js
+++ b/api/services/RecentChangeService.js
@@ -90,7 +90,15 @@ function groupChanges(changes) {
         g.subActionEntityId === c.id_entity
       ) {
         g.subAction = c.type_change; // eslint-disable-line no-param-reassign
-        g.subActionEntityId = c.id_entity; // eslint-disable-line no-param-reassign
+        // subActionEntityId is already equal to c.id_entity (checked above).
+      } else if (
+        ['create', 'restore'].includes(g.subAction) &&
+        c.type_change === 'update' &&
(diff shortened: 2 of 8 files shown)


Paul-AUB (commented):
The change adds .autoOrient() to the sharp pipeline before resize, switches variant selection to the oriented
  width via metadata.autoOrient, and adds a --fix-orientation mode to backfill-thumbnails.js that re-downloads
  all thumbnailed files, skips ones already upright (orientation absent or 1), and regenerates the rest.

  Findings

  - [P2] Fix-orientation regenerates leave orphaned blobs when variant set shrinks
    - Location: scripts/backfill-thumbnails.js:228-245
    - Why it matters: For a file whose stored width crossed a variant threshold but whose oriented width does
  not (e.g. stored 4000×1300 landscape but EXIF-6 makes upright 1300×4000, so originalWidth drops from 4000 to
  1300 and large [1920] is no longer applicable — or the more common case of stored 1500×… downgrading to <1280
  oriented), the new generate() call will produce thumbnailMedium/Large: null. The DB row is then set to null
  for that variant, but the old, wrong-oriented blob (e.g. thumbnails/large/foo.webp) is not deleted — it
  becomes an orphan taking storage indefinitely. Only the variant paths that survive re-generation get
  overwritten. This is a new orphan class introduced by this PR's regeneration flow; unlike the pre-existing
  "partial upload failure" case documented in ThumbnailService.generate, this one happens on the success path
  so it will occur every time the oriented width crosses down a variant threshold.
    - Confidence: 0.8
    - Suggested fix: Before updateOne, diff the old vs new thumbnail paths for the record and delete any old
  blob whose new value is null (or whose path changed). Alternatively, document this as an accepted trade-off
  and add a cleanup pass to the storage-orphan job you already reference in ThumbnailService.js.
  - [P2] No test coverage for the two behavior changes
    - Location: test/integration/1_services/ThumbnailService.test.js (unchanged),
  api/services/ThumbnailService.js:80-86,105-112
    - Why it matters: The two shipped behaviors — (a) EXIF orientation is applied before resize, and (b)
  variant selection uses the oriented width — are silent behaviors that would regress invisibly if
  .autoOrient() were ever removed or reordered (e.g. some future author swapping in .rotate() for
  size-invariant use, or the metadata.autoOrient accessor stubbed away). There is an existing resize/generate
  suite next door, so the marginal cost of a fixture with EXIF orientation ≠ 1 and an assertion on the output
  dimensions is low. Without a regression test the fix will drift back.
    - Confidence: 0.9
    - Suggested fix: Add a test that feeds a landscape-stored buffer with EXIF orientation 6 to resize() and
  asserts the output height > width (i.e. rotation applied), and a generate() test that asserts the applicable
  variants are chosen against the oriented width.
  - [P3] Comment about metadata.autoOrient availability is slightly misleading
    - Location: api/services/ThumbnailService.js:109-112
    - Why it matters: The comment says metadata.autoOrient "falls back to the stored dimensions when no
  orientation tag is present." That is what the code does via || metadata; the library itself populates
  metadata.autoOrient (with the same values as metadata.width/height) whenever it can, regardless of whether
  the tag is present. Future readers may think the || metadata fallback is exercised on every non-EXIF image
  when in practice it's dead defensive code for sharp ≥ 0.34.
    - Confidence: 0.7
    - Suggested fix: Reword to "always present in sharp 0.34+; the fallback is defensive for older versions."
  - [P3] --fix-orientation dry-run count is misleading
    - Location: scripts/backfill-thumbnails.js:130-143
    - Why it matters: The dry-run prints "Would process X files" where X includes every file with any
  thumbnail; but in a real run most of those get short-circuited by the orientation === 1 check (line 221)
  without any thumbnail regeneration. Operators trying to size the impact of the migration will over-estimate
  work by orders of magnitude. The orientation check requires downloading the blob, so it cannot be moved
  earlier without changing the design — but the dry-run output can at least warn.
    - Confidence: 0.7
    - Suggested fix: In --fix-orientation --dry-run, log that the count is an upper bound because orientation-1
  files will be skipped only at process time.
  - [P3] Redundant metadata decode in fix-orientation path
    - Location: scripts/backfill-thumbnails.js:218-226
    - Why it matters: For every file that needs regeneration, sharp(imageBuffer).metadata() is called here and
  then ThumbnailService.generate() calls sharp(imageBuffer, {limitInputPixels: ...}).metadata() again on the
  same buffer. Metadata parses headers only, so cost is small, but the second call also enforces a 100 MP limit
  that this pre-check does not — so a malformed file that would be rejected by generate() passes the pre-check
  silently.
    - Confidence: 0.5
    - Suggested fix: Either pass {limitInputPixels: 100 * 1024 * 1024} to the pre-check sharp() call for
  consistency, or expose an isAlreadyUpright(buffer) helper on ThumbnailService to keep the guard next to the
  constant.
    - Confidence: 0.5
    - Suggested fix: Either pass {limitInputPixels: 100 * 1024 * 1024} to the pre-check sharp() call for consistency, or expose an isAlreadyUpright(buffer) helper on
  ThumbnailService to keep the guard next to the constant.

  Final Summary

  - Overall explanation: The core fix — .autoOrient() + metadata.autoOrient.width — is applied in the right order and works with the installed sharp version. The main concern is
  the migration script's regeneration flow: on the success path, when the oriented width causes a smaller applicable-variant set than the pre-fix generation, the DB is nulled but
  old wrong-oriented blobs are left behind. There is also no new test coverage locking in either shipped behavior.
  - Overall confidence: 0.8

Paul-AUB (dismissed):
see comment above, ignore my first (hidden now) comment

PR #1777, opened 2026-08-18
fix(observations): sync v_measurement_wide with production definition

ClemRz:
Brings the `v_measurement_wide` migration in sync with the production view definition.

## Why

The `9_12` migration recreated the view to introduce per-unit conversion factors (`t_unit`), but in doing so dropped four columns that were already live in production. Any environment rebuilt from migrations would be missing those columns, causing BI queries and API consumers that rely on them to fail.

## What

- **`observation_name`** — the main, non-deleted name for the observation (via `LEFT JOIN t_name`)
- **`sensor_label`** — the label from `t_sensor_configuration`
- **`device_id`** / **`device_name`** — device identity via `t_sensor_configuration → t_device`

Single new migration file: `sql/9_19_2026_08_17_v_measurement_wide_add_missing_columns.sql`

View creation was validated against the local dev database — all 34 columns match the live schema.

## Testing

SQL-only change (view recreation). Verified the `CREATE VIEW` executes without errors and the resulting column list matches production.

## Related

No linked issue — discovered via DDL comparison between production and the migration history.

### DIFF
--- a/sql/9_19_2026_08_17_v_measurement_wide_add_missing_columns.sql
+++ b/sql/9_19_2026_08_17_v_measurement_wide_add_missing_columns.sql
@@ -1,0 +1,82 @@
+\c grottoce;
+
+-- ============================================================
+-- Migration: Sync v_measurement_wide with production definition
+--
+-- The last migration (9_12) is missing three additions that were
+-- applied directly in production:
+--   - observation_name  (from t_name, main non-deleted name)
+--   - sensor_label      (from t_sensor_configuration)
+--   - device_id / device_name (from t_device via t_sensor_configuration)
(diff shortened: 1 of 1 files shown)


Paul-AUB (approved):
The PR does exactly what it advertises: realign the versioned definition with the live view so environments rebuilt from migrations end up with the same column set.

  - Overall explanation: The diff achieves its stated goal (1:1 sync with the production view, 34 columns verified). The risks I flagged (t_name duplication, missing index,
  is_deleted filters) all exist in production today — fixing them here would break the very invariant this PR establishes.
  - Overall confidence: 0.9

PR #1779, opened 2026-08-18
fix(document): prevent cyclic parent relationships

### LINKED ISSUE #1778: Prevent cyclic document parents from hanging organization details
**Describe the bug**

Document writes currently allow invalid parent hierarchies, including a document referencing itself. If a cyclic document is included in an organization's authored or edited documents, loading the organization details can hang indefinitely.

`GrottoService.getPopulatedOrganization()` collects the organization's documents and calls `DocumentService.getCollectionAncestors()`. The latter uses a recursive CTE with `UNION ALL` and no visited-path/cycle detection, so cyclic data never reaches a terminating condition.

Document update/validation paths also do not reject direct or indirect cycles. A parent can remain persisted after changing to a document type that should not have one when the parent field is omitted rather than explicitly cleared.

**To Reproduce**

1. In test data, create two documents that form a parent cycle (a direct self-reference is the smallest case).
2. Associate one cyclic document with an organization as editor or organization author.
3. Request the organization details endpoint.
4. Observe that hierarchy traversal does not terminate.

**Expected behavior**

- Writes reject self-parenting and indirect parent cycles with a 400 response.
- Document types that cannot have a parent clear or reject an existing parent explicitly.
- `getCollectionAncestors()` always terminates even when legacy/corrupt cyclic data exists.
- Organization details return a bounded response instead of hanging.

**Suggested hardening**

- Validate parent relationships in every document create, update, and validation path.
- Add a database constraint preventing direct self-references; indirect cycles still require application or trigger validation.
- Make the recursive CTE cycle-safe using PostgreSQL `CYCLE`, a visited-id path, or equivalent, with a defensive depth limit.
- Add regression tests for direct and indirect cycles, type changes that remove parent eligibility, and organization detail traversal over pre-existing corrupt data.
- Audit and repair existing cyclic document hierarchies separately through the operational maintenance process.

ClemRz:
## Summary

Fixes the root causes of the cyclic document parent hang described in #1778. The data fix (clearing id 234218) is already applied in production; this PR prevents the same corruption from happening again and makes existing corrupt data survivable.

## Why

- **`getCollectionAncestors()` could hang indefinitely** when a document's `id_parent` chain formed a cycle (A→A or A→B→A). The recursive CTE used `UNION ALL` with no termination guard, so it looped forever. This caused any org details request that included such a document to time out.
- **No write-time validation** prevented a document from being set as its own parent or creating an indirect cycle. The corruption path was: type change → front-end omits `parent` field → back-end preserved the stale FK.
- **No DB constraint** blocked self-references at the storage level.

## What

- `getCollectionAncestors()` — added `CYCLE id SET is_cycle USING path` (PG 14+) so cyclic rows are detected and skipped instead of looped forever.
- `DocumentService.checkParentCycle()` — new helper that walks the proposed parent's ancestor chain to detect indirect cycles before any write.
- `update.js` and `update-with-new-entities.js` — call `checkParentCycle()` before persisting; return 400 on a cycle.
- `getConvertedDataFromClient()` — when the resolved document type does not allow a parent (`TYPES_ALLOWING_PARENT`), the `parent` field is explicitly set to `null` rather than left `undefined`, so a type change always clears a stale parent regardless of whether the client sends the field.
- `sql/9_20_2026_08_18_document_no_self_parent_constraint.sql` — adds `CHECK (id <> id_parent)` to `t_document`; `test/customSQL.js` mirrors it for the test DB.
- `TYPES_ALLOWING_PARENT` / `TYPES_REQUIRING_PARENT` constants replace the local `MANDATORY_PARENT_TYPES` string array.
- 14 new integration tests covering direct cycles, indirect cycles, unrelated pairs, and the parent-clearing behaviour.

## Related

- Closes #1778

### DIFF
--- a/api/controllers/v1/document/multiple-validate.js
+++ b/api/controllers/v1/document/multiple-validate.js
@@ -70,6 +70,41 @@ async function validateAndUpdateDocument(
   const { scalarData, collectionData } =
     normalizeAndSplitDocumentData(documentData);
 
+  // Re-validate the parent assignment against the current hierarchy.
+  // The hierarchy may have changed since the modification was submitted
+  // (e.g. a sibling was re-parented while this modification was pending),
+  // so the stored parent could now create a cycle even though it passed
+  // validation at submission time.
+  // Also re-apply the type-vs-parent policy in case the stored type no
+  // longer allows a parent (uses the stored type with a fallback to the
+  // current document type, matching the update-with-new-entities.js path).
+  const effectiveTypeId = scalarData.type ?? document.type;
+  scalarData.parent = DocumentService.clearParentIfTypeDisallows(
+    effectiveTypeId,
+    scalarData.parent
+  );
+  if (scalarData.parent != null) {
+    const parentError = await DocumentService.validateParentAssignment(
+      document.id,
+      scalarData.parent
+    );
+    if (parentError) {
+      // Reject the modification rather than persisting a corrupt hierarchy.
+      // Clear the pending modification so the document returns to its last
+      // validated state and the moderator is informed of why it was rejected.
+      await TDocument.updateOne(document.id).set({
+        modifiedDocJson: null,
+        validationComment: `Auto-rejected: ${parentError}`,
+        validator: validationAuthor,
+        dateValidation: new Date(),
+      });
+      sails.log.warn(
+        `Document ${document.id} modification auto-rejected: ${parentError}`
+      );
+      return { rejected: true, reason: `Auto-rejected: ${parentError}` };
+    }
+  }
+
   await sails.getDatastore().transaction(async (db) => {
     // Update associated data not handled by TDocument manually
     // Updated before the TDocument update so the last_change_document DB trigger will fetch the last updated name
@@ -120,6 +155,7 @@ async function validateAndUpdateDocument(
     }
     await Promise.all(filePromises);
   });
+  return { rejected: false };
 }
 
 async function updateSearchAndNotify(req, documentId, userId) {
@@ -204,11 +240,34 @@ module.exports = async (req, res) => {
 
     if (isAModifiedDoc) {
       // eslint-disable-next-line no-await-in-loop
-      await validateAndUpdateDocument(
+      const result = await validateAndUpdateDocument(
         document,
         change.validationComment,
         req.token.id
       );
+      if (result.rejected) {
+        // The pending modification was auto-rejected due to a parent cycle.
+        // Send a REJECT author notification with the auto-rejection reason,
+        // matching the behaviour of the manual-rejection path above.
+        // eslint-disable-next-line no-await-in-loop
+        const rejectedDoc = await DocumentService.getPopulatedDocument(
+          document.id
+        );
+        // eslint-disable-next-line no-await-in-loop
+        await NotificationService.notifyAuthor(
+          rejectedDoc,
+          req.token.id,
+          NotificationService.NOTIFICATION_TYPES.REJECT,
+          result.reason
+        ).catch((err) =>
+          sails.log.error(
+            'Document multiple-validate notifyAuthor error',
+            document,
+            err
+          )
+        );
+        continue; // eslint-disable-line no-continue
+      }
     } else {
       // Likely a document creation
       // eslint-disable-next-line no-await-in-loop
--- a/api/controllers/v1/document/update-with-new-entities.js
+++ b/api/controllers/v1/document/update-with-new-entities.js
@@ -42,20 +42,27 @@ module.exports = async (req, res) => {
   // Reject cyclic parent assignments before touching any data.
   // scalarData.parent may be a raw ID or an object with an id property.
   const rawParent = scalarData.parent;
-  const proposedParentId =
+  const resolvedParent =
     rawParent != null && typeof rawParent === 'object'
       ? rawParent.id
       : rawParent;
+
+  // Apply the same type-vs-parent policy as getConvertedDataFromClient:
+  // if the type does not allow a parent, clear it now so the ORM write
+  // does not preserve a stale id_parent from a previous type state.
+  const effectiveTypeId = scalarData.type ?? currentDocument.type;
+  scalarData.parent = DocumentService.clearParentIfTypeDisallows(
+    effectiveTypeId,
+    resolvedParent
+  );
+
+  const proposedParentId = scalarData.parent;
   if (proposedParentId != null) {
-    const hasCycle = await DocumentService.checkParentCycle(
-      Number(documentId),
-      Number(proposedParentId)
+    const parentError = await DocumentService.validateParentAssignment(
+      documentId,
+      proposedParentId
     );
-    if (hasCycle) {
-      return res.badRequest(
-        'The proposed parent would create a cycle in the document hierarchy.'
-      );
-    }
+    if (parentError) return res.badRequest(parentError);
   }
 
   // Wrap the scalar update, entity creation, and all replaceCollection calls in
--- a/api/controllers/v1/document/update-with-new-entities.js
+++ b/api/controllers/v1/document/update-with-new-entities.js
@@ -39,6 +39,25 @@ module.exports = async (req, res) => {
     }
   }
 
+  // Reject cyclic parent assignments before touching any data.
+  // scalarData.parent may be a raw ID or an object with an id property.
+  const rawParent = scalarData.parent;
+  const proposedParentId =
+    rawParent != null && typeof rawParent === 'object'
+      ? rawParent.id
+      : rawParent;
+  if (proposedParentId != null) {
+    const hasCycle = await DocumentService.checkParentCycle(
+      Number(documentId),
+      Number(proposedParentId)
+    );
+    if (hasCycle) {
+      return res.badRequest(
+        'The proposed parent would create a cycle in the document hierarchy.'
+      );
+    }
+  }
+
   // Wrap the scalar update, entity creation, and all replaceCollection calls in
   // a single transaction so the document is never left in a partially-updated state.
   let updatedDocument;
--- a/api/controllers/v1/document/update.js
+++ b/api/controllers/v1/document/update.js
@@ -45,18 +45,18 @@ module.exports = async (req, res) => {
   }
 
   // Reject cyclic parent assignments before any file upload side-effects.
+  documentDataForValidation.parent = DocumentService.clearParentIfTypeDisallows(
+    effectiveType,
+    documentDataForValidation.parent
+  );
   const proposedParentId = documentDataForValidation.parent;
   if (proposedParentId != null) {
     const documentId = req.param('id');
-    const hasCycle = await DocumentService.checkParentCycle(
-      Number(documentId),
-      Number(proposedParentId)
+    const parentError = await DocumentService.validateParentAssignment(
+      documentId,
+      proposedParentId
     );
-    if (hasCycle) {
-      return res.badRequest(
-        'The proposed parent would create a cycle in the document hierarchy.'
-      );
-    }
+    if (parentError) return res.badRequest(parentError);
   }
 
   // Add new files
--- a/api/controllers/v1/document/update.js
+++ b/api/controllers/v1/document/update.js
@@ -44,6 +44,21 @@ module.exports = async (req, res) => {
     );
   }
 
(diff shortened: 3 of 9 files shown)


Paul-AUB (changes requested):
# Code Review — PR #1779 (`fix/document-cyclic-parent`)

Review target: branch `origin/fix/document-cyclic-parent` vs `origin/develop` (source: override)
Diff files: 7 (`api/controllers/v1/document/update-with-new-entities.js`,
`api/controllers/v1/document/update.js`, `api/services/DocumentService.js`,
`config/constants/document.js`, `sql/9_20_2026_08_18_document_no_self_parent_constraint.sql`,
`test/customSQL.js`, `test/integration/1_services/DocumentService.test.js`).

Single commit: `cdc7ac16 fix(document): prevent cyclic parent relationships`.

---

## Findings

### [P1] `update-with-new-entities` does not clear `parent` on type change — the endpoint can persist a Collection with a parent

- **Location:** `api/controllers/v1/document/update-with-new-entities.js:23-60`, contrasted with `api/services/DocumentService.js:190-210`
- **Why it matters:** The PR adds the "explicitly clear parent when the type does not allow one" rule inside `getConvertedDataFromClient`, which is invoked by `update.js`. `update-with-new-entities.js` never routes through that helper — it takes `req.body.document` verbatim into `scalarData`, then straight into `TDocument.updateOne(documentId).set(scalarData)`. So a client that switches a document from Article/Issue to Collection through this endpoint (front-end omits `parent`, or worse sends the old one) will end up with `type=Collection` and a non-null `id_parent`. That is the exact broken state the type-clear branch was introduced to prevent, and the cycle-check added a few lines above does not catch it (a stale-but-valid ancestor is not a cycle). The new behavioural divergence between the two update paths is introduced by this diff.
- **Confidence:** 0.85
- **Suggested fix:** Extract the type-vs-parent policy into a small helper (e.g. `DocumentService.normalizeParentForType(scalarData)`) and call it from both `update.js` (via the existing conversion) and `update-with-new-entities.js` before the cycle check.

### [P2] Cycle detection has a TOCTOU window between check and update

- **Location:** `api/controllers/v1/document/update.js:47-60`, `api/controllers/v1/document/update-with-new-entities.js:43-56`, `api/services/DocumentService.js:588-628`
- **Why it matters:** `checkParentCycle` runs *before* the transaction opens. Under default Read Committed isolation, two concurrent requests can both pass the check and then commit updates that together form a cycle — e.g. Req1 sets `A.parent = B`, Req2 sets `B.parent = A`. Each in isolation is acyclic; together they are not. The new DB constraint only blocks direct self-reference (`id <> id_parent`), so indirect cycles produced by a race have no safety net. Given how the doc tree is edited (moderator UI, individual actions), this is not high-frequency, but the failure mode is silently-corrupted state that `getCollectionAncestors` had to be hardened against precisely because it existed in the wild.
- **Confidence:** 0.8
- **Suggested fix:** Run `checkParentCycle` *inside* the update transaction after taking `SELECT … FOR UPDATE` on the target row (and ideally the proposed parent), or switch that transaction to `SERIALIZABLE`. Either narrows the window to zero.

### [P2] Migration will fail on production data that already contains a self-parented row

- **Location:** `sql/9_20_2026_08_18_document_no_self_parent_constraint.sql:1-8`
- **Why it matters:** The migration issues an unconditional `ADD CONSTRAINT … CHECK (id <> id_parent)`. Postgres validates the constraint against every existing row at ALTER time — if any legacy row has `id = id_parent`, the migration aborts and the deploy fails. The commentary in `getCollectionAncestors` (previously suggesting `SELECT id, id_parent FROM t_document WHERE id = id_parent`) explicitly acknowledges this case has been observed. Test setup mirror in `test/customSQL.js` is safe (fresh DB), so this is only visible when applied to real data.
- **Confidence:** 0.75
- **Suggested fix:** Prepend an `UPDATE t_document SET id_parent = NULL WHERE id = id_parent;` (log rows first) or add the constraint as `NOT VALID` and validate in a follow-up step after cleanup.

### [P2] `Number()` coercion silently swallows non-numeric ids and produces a false-negative cycle check

- **Location:** `api/controllers/v1/document/update.js:50-55`, `api/controllers/v1/document/update-with-new-entities.js:47-52`
- **Why it matters:** `Number(documentId)` returns `NaN` if the path parameter is not numeric (also true for `Number(proposedParentId)` if the client sends a non-numeric object). The short-circuit `documentId === proposedParentId` then evaluates `NaN === NaN` → `false`, and the recursive CTE is called with `NaN` as a parameter — which `sendNativeQuery` will stringify to `'NaN'`, matching no rows, so `checkParentCycle` returns `false` and the update proceeds. Whether route middleware guarantees a numeric id upstream is not visible from the diff, and the two endpoints handle validation differently.
- **Confidence:** 0.7
- **Suggested fix:** Guard `if (!Number.isInteger(documentId) || !Number.isInteger(proposedParentId))` and either treat as `badRequest` or bail out with the cycle-detected message.

### [P3] `TYPES_ALLOWING_PARENT` and `TYPES_REQUIRING_PARENT` are literal-identical arrays

- **Location:** `config/constants/document.js:17-28`
- **Why it matters:** Both constants are `[ISSUE, ARTICLE]` and are documented with the same criterion ("Issue under Collection, Article under Issue"). Currently every type that *may* have a parent also *must* have one, so the two names carry a semantic distinction that has no material effect. Fine for now, but if the two sets ever diverge (e.g. a future type that optionally has a parent) the coupling by copy-paste will silently drift. Consider deriving one from the other, or picking one name and documenting the equivalence.
- **Confidence:** 0.6
- **Suggested fix:** Drop `TYPES_REQUIRING_PARENT` and reuse `TYPES_ALLOWING_PARENT` in `updateDocument`, or vice-versa — comment the equivalence at the definition site.

### [P3] Test title "should leave parent undefined when no type is given" asserts `equal(1)`, not undefined

- **Location:** `test/integration/1_services/DocumentService.test.js:817-824`
- **Why it matters:** The `it()` label reads "leave parent undefined" but the assertion is `should(result.parent).equal(1)` — the actual behaviour is "return the id as-is." Not a functional defect; just misleading when a future contributor greps for the constraint. Rename to "should return the client-sent parent id when no type is given" to keep the test discoverable.
- **Confidence:** 0.9
- **Suggested fix:** Rename the `it()` block.

---

## Final Summary

- **Overall correctness:** correct
- **Overall explanation:** The core fix is sound: the recursive CTE with `CYCLE … USING path` correctly hardens `getCollectionAncestors` against corrupt data, `checkParentCycle` correctly detects both direct and indirect cycles, and the new DB `CHECK` gives a last-line defence. The residual risks are (1) a real behavioural divergence introduced between `update` and `update-with-new-entities` because the parent-clearing rule only ships in the `getConvertedDataFromClient` path, and (2) the classic TOCTOU race that the new check does not close.
- **Overall confidence:** 0.8

Paul-AUB (changes requested):
Findings

  [P1] Migration VALIDATE CONSTRAINT aborts if any self-parented row exists

  - Location: sql/9_20_2026_08_18_document_no_self_parent_constraint.sql:14-16
  - Why it matters: The comment claims the migration "never fails on pre-existing corrupt rows," but VALIDATE CONSTRAINT explicitly re-scans every existing row and raises an error on the first
    violation. Since the file itself calls VALIDATE CONSTRAINT t_document_no_self_parent immediately after NOT VALID, the whole migration will abort in a production database that still holds any id =
    id_parent row — exactly the case the file's follow-up note ("Any remaining self-parented rows should be audited and cleared separately") acknowledges could exist. The NOT VALID step alone would
    have made the migration truly non-blocking; combined with an eager VALIDATE it doesn't.
  - Confidence: 0.9
  - Suggested fix: Either (a) UPDATE t_document SET id_parent = NULL WHERE id = id_parent; before validating, or (b) drop the VALIDATE CONSTRAINT line here and validate in a follow-up migration once
    ops has confirmed the table is clean. Also update the comment so it doesn't overpromise safety.

  [P2] update.js skips the type-vs-parent clearing when the client omits type

  - Location: api/services/DocumentService.js:197-204, api/controllers/v1/document/update.js:47-65
  - Why it matters: getConvertedDataFromClient only enforces "clear parent when type disallows it" when typeFound is truthy — i.e. when the request body includes a type. In
    update-with-new-entities.js the new code uses effectiveTypeId = scalarData.type ?? currentDocument.type, so the same guard falls back to the stored type. update.js has no such fallback, so a
    PATCH-style request that changes only parent on a Collection document will pass through with the parent set, staging a modifiedDocJson that reintroduces the exact "stale id_parent after a
    Collection" state the diff sets out to prevent. The two update paths now diverge on the very policy this diff was meant to unify.
  - Confidence: 0.75
  - Suggested fix: In update.js, after computing documentDataForValidation, apply DocumentService.clearParentIfTypeDisallows(effectiveType, documentDataForValidation.parent) using the same
    effectiveType already computed for the issue check. That keeps the single-source policy in clearParentIfTypeDisallows and makes both controllers behave identically.

  [P3] TYPES_REQUIRING_PARENT = TYPES_ALLOWING_PARENT shares the array reference

  - Location: config/constants/document.js:30
  - Why it matters: Any future mutation (push, splice) on one constant will silently update the other. The inline comment warns against divergence, but relying on discipline rather than a fresh array
    is a small foot-gun given how easy this is to sidestep.
  - Confidence: 0.5
  - Suggested fix: const TYPES_REQUIRING_PARENT = [...TYPES_ALLOWING_PARENT]; — cheap, and future edits stay independent.

  ---

  Final Summary

  - Overall explanation: The cycle-detection service, the two controller guards, and the extracted type-vs-parent helper are logically sound and adequately tested. The main risks are operational (the
    migration's VALIDATE CONSTRAINT step against dirty prod data) and a symmetry gap between the two update controllers when the client omits type.
  - Overall confidence: 0.8

Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[api/controllers/v1/document/multiple-validate.js:70] Revalidate pending parent assignments when applying `modifiedDocJson`.** `update.js` checks the proposed parent only when it creates the pending modification; `validateAndUpdateDocument()` later writes that stored `scalarData` without checking it against the hierarchy as it exists at approval time. For example, two sibling documents A and B can start under the same parent, a user can submit `A.parent = B` (which passes), and B can subsequently be changed to point to A while A's modification is still pending. Approving A then persists the indirect cycle A → B → A. The database constraint only prevents direct self-references, so it does not catch this. Please normalize the pending parent/type policy and call `validateParentAssignment()` against the current hierarchy immediately before applying the modification (ideally inside the same transaction with the appropriate locking/isolation), and add a regression test for a hierarchy that changes between submission and moderator approval.

Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[api/controllers/v1/document/multiple-validate.js:104] Stop the normal validation flow after auto-rejecting a cyclic pending modification.** `validateAndUpdateDocument()` currently returns without telling its caller that the modification was rejected, so the loop continues at line 258 into `updateSearchAndNotify()` and then calls `notifyAuthor(..., NOTIFICATION_TYPES.VALIDATE, ...)` at line 273. As a result, the author and subscribers receive a successful-validation notification even though the pending edit was auto-rejected, and the author-facing call uses the moderator's original (possibly empty) comment rather than the auto-rejection reason stored on the document. Please return an explicit outcome and branch like the existing manual-rejection path (including a `REJECT` author notification), then extend the regression test to assert that no `VALIDATE` notification is created for this case.

2. **[api/services/DocumentService.js:637] Enforce the integer contract before executing the recursive query.** The helper's documentation says both IDs must be finite integers, but `Number.isFinite()` accepts values such as `1.5`. PostgreSQL will then try to bind that value against the integer `id` column and can raise an invalid-input error, turning a malformed parent ID into a 500 instead of the intended 400 response. Consider using `Number.isInteger()` (and, if appropriate for this schema, requiring positive IDs) and adding a decimal-ID regression test.

Paul-AUB (dismissed):
Reviewed the full 510-line diff, prioritizing the document write paths, cycle-detection service logic, and database migration.

1. **[test/integration/4_routes/Documents/multiple-validate.test.js:423] Assert that the auto-rejection creates the expected `REJECT` notification.** The regression test currently proves that no `VALIDATE` notification is emitted, but it would also pass if the auto-rejection emitted no author notification at all. Consider querying the new notifications for the document, author, and `REJECT` type (as the existing manual-rejection tests do) so both halves of the notification contract are protected.

PR #1780, opened 2026-08-21
fix(organization): return 400 instead of 500 on over-long postalCode

### LINKED ISSUE #1774: bug(api): [E_INVALID_VALUES_TO_SET] tgrotto - postalCode exceeds maxLength(10) returns 500 on update
**Describe the bug**
`PUT /api/v1/organizations/:id` returns a 500 error when the `postalCode` field in the request body exceeds 10 characters. The `TGrotto` model defines `maxLength: 10` for `postalCode`, but the update controller passes the value directly to `TGrotto.updateOne()` without prior length validation. Waterline catches the violation and throws an `E_INVALID_VALUES_TO_SET` error, which bubbles up as an unhandled 500 instead of a 400 Bad Request.

The actual input `"4400 Flémalle"` is 13 characters, 3 over the limit.

**To Reproduce**
Based on the trace logs:
1. Authenticate as a user with Moderator/Administrator role
2. `PUT /api/v1/organizations/1788`
3. Include a `postalCode` value longer than 10 characters (e.g., `"4400 Flémalle"`)

**Input Data**

- **Endpoint:** `PUT /api/v1/organizations/1788`
- **Request Body:**
```json
{
  "id": 1788,
  "address": "128, rue de la Grotte",
  "city": "Liège",
  "country": { "id": "BE" },
  "customMessage": "-",
  "latitude": null,
  "longitude": null,
  "mail": "info@leschercheursdelawallonie.be",
  "postalCode": "4400 Flémalle",
  "url": "https://leschercheursdelawallonie.be/"
}
```
- **User:** id=60181, nickname="Paul AUBERTIN", groups=["Administrator","Moderator","Leader"]

**Expected behavior**
The API should return `400 Bad Request` with a descriptive message when `postalCode` exceeds 10 characters, rather than propagating a Waterline `E_INVALID_VALUES_TO_SET` error as a 500. The `organization/update` controller should validate `postalCode` length before calling `TGrotto.updateOne()`, consistent with how `nameText` is validated via `validateNameLength`.

**Error Details**
- Error Type: `OperationalError [UsageError]` — `E_INVALID_VALUES_TO_SET`
- Occurrences: 1
- Time Range: 2026-08-14T21:14:51Z to 2026-08-14T21:14:51Z
- Representative Trace ID: `01a00220-af10-74d9-9531-7ed7350b4ce7`
- Stack Trace:
```
Details:
  Could not use specified `postalCode`.  Violated one or more validation rules:
  • Value was -3 characters longer than the configured maximum length (10)

    at flaverr (.../node_modules/flaverr/index.js:94:15)
    at _.extend._WLModel [as _handleExec] (.../waterline/lib/waterline/methods/update-one.js:138:15)
  code: 'E_INVALID_VALUES_TO_SET'
```

**Additional context**
- Service: `api/services/GrottoService.js` — `getConvertedDataFromClientRequest` passes `postalCode` through `trimIfString` but does not enforce length
- Controller: `api/controllers/v1/organization/update.js` — no `postalCode` length guard before `TGrotto.updateOne()`
- Model: `TGrotto` (`tgrotto`) — `postalCode: { type: 'string', allowNull: true, maxLength: 10 }`
- DB Error Code: `E_INVALID_VALUES_TO_SET` (Waterline validation)
- Fix: add a `postalCode` length check (max 10) in `organization/update.js` (and potentially `organization/create.js`) before the ORM call, returning `res.badRequest()` on violation

tristan0x:
## 🤔 What

Add validatePostalCodeLength (built on the existing validateStringLength util, mirroring nameValidation) and guard both the create and the update controllers before any write. The check runs on the trimmed value from GrottoService.getConvertedDataFromClientRequest so a postal code that fits once trimmed is still accepted.

## 🤷‍♂️ Why

PUT /api/v1/organizations/:id passed postalCode straight to TGrotto.updateOne(); a value longer than the column limit (varchar(10)) made Waterline throw E_INVALID_VALUES_TO_SET, surfacing as an unhandled 500 instead of a client error.

## Related

Closes #1774

### DIFF
--- a/api/controllers/v1/organization/create.js
+++ b/api/controllers/v1/organization/create.js
@@ -2,6 +2,9 @@ const ControllerService = require('../../../services/ControllerService');
 const GrottoService = require('../../../services/GrottoService');
 const { toOrganization } = require('../../../services/mapping/converters');
 const { validateNameLength } = require('../../../utils/nameValidation');
+const {
+  validatePostalCodeLength,
+} = require('../../../utils/postalCodeValidation');
 
 module.exports = async (req, res) => {
   // Check params
@@ -22,6 +25,12 @@ module.exports = async (req, res) => {
     author: req.token.id,
     dateInscription: new Date(),
   };
+
+  // Validate postalCode length (checked on the trimmed value)
+  const postalCodeError = validatePostalCodeLength(cleanedData.postalCode);
+  if (postalCodeError) {
+    return res.badRequest(postalCodeError);
+  }
 
   const nameData = {
     author: req.token.id,
--- a/api/controllers/v1/organization/update.js
+++ b/api/controllers/v1/organization/update.js
@@ -4,6 +4,9 @@ const NotificationService = require('../../../services/NotificationService');
 const EnrichmentQueueService = require('../../../services/EnrichmentQueueService');
 const { toOrganization } = require('../../../services/mapping/converters');
 const { validateNameLength } = require('../../../utils/nameValidation');
+const {
+  validatePostalCodeLength,
+} = require('../../../utils/postalCodeValidation');
 
 module.exports = async (req, res) => {
   // Check if organization exists
@@ -34,6 +37,13 @@ module.exports = async (req, res) => {
     // Clear stale enrichment field — it'll be repopulated by the async job
     cleanedData.iso_3166_2 = null;
     coordinatesChanged = true;
+  }
+
+  // Validate postalCode length before any write (checked on the trimmed value,
+  // so a value that fits once trimmed is not rejected)
+  const postalCodeError = validatePostalCodeLength(cleanedData.postalCode);
+  if (postalCodeError) {
+    return res.badRequest(postalCodeError);
   }
 
   // Validate name length
--- a/api/utils/postalCodeValidation.js
+++ b/api/utils/postalCodeValidation.js
@@ -1,0 +1,21 @@
+const { validateStringLength } = require('./stringLengthValidation');
+
+const POSTAL_CODE_MAX_LENGTH = 10; // TGrotto.postalCode maxLength
+
+/**
+ * Validates that a postal code does not exceed the DB column limit.
+ * Returns an error message if invalid, or null if valid.
+ *
+ * @param {string} postalCode - The postal code to validate
+ * @returns {string|null} Error message or null
+ */
+const validatePostalCodeLength = (postalCode) => {
+  const error = validateStringLength(
+    'Postal code',
+    postalCode,
+    POSTAL_CODE_MAX_LENGTH
+  );
+  return error ? error.message : null;
+};
+
+module.exports = { POSTAL_CODE_MAX_LENGTH, validatePostalCodeLength };
--- a/assets/swaggerV1.yaml
+++ b/assets/swaggerV1.yaml
@@ -5931,7 +5931,8 @@ paths:
         - name: postalCode
           in: query
           schema:
-            type: number
+            type: string
+            maxLength: 10
         - name: region
           in: query
           schema:
@@ -5954,7 +5955,7 @@ paths:
                 items:
                   $ref: '#/components/schemas/Organization'
         '400':
-          description: You must provide at least a name to create an organization.
+          description: You must provide at least a name to create an organization, or the postal code is too long.
         '403':
           description: You are not authorized to create an organization.
 
@@ -6040,6 +6041,8 @@ paths:
                   type: string
                 postalCode:
                   type: string
+                  description: Postal code (max 10 characters)
+                  maxLength: 10
                 region:
                   type: string
                 url:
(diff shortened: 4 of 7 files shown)


ClemRz (dismissed):

- **[assets/swaggerV1.yaml, ~line 5931 & 6041]** The `postalCode` field is typed as `number` in the `POST /organizations` query parameters section, but it is a `string` in the model, the service, and the update request body. This is a pre-existing inconsistency, but since this PR adds validation against that field it's a good moment to align it. Additionally, now that `postalCode` is enforced to `maxLength: 10`, it's worth documenting that constraint in the spec the same way `name.text` is documented in the PUT body:

  ```yaml
  # POST /organizations — query params (line ~5931)
  - name: postalCode
    in: query
    schema:
      type: string        # was: number
      maxLength: 10

  # PUT /organizations/{id} — request body (line ~6041)
  postalCode:
    type: string
    maxLength: 10         # add this
  ```

- **[test/integration/4_routes/Organization/update.test.js, ~line 231]** The `should return 200 when postalCode is exactly 10 characters` test mutates org 1's `postalCode` to `'1234567890'` without resetting it after the test. Since the parallel runner isolates shards at the DB level this is less risky than it looks, but within the same shard later tests that assert on org 1's `postalCode` value could be surprised. Consider adding a teardown consistent with how other tests in this file reset state:

  ```js
  .end(async (err, res) => {
    if (err) return done(err);
    try {
      should(res.body.postalCode).equal('1234567890');
      // Reset to avoid affecting subsequent tests
      await TGrotto.updateOne({ id: 1 }).set({ postalCode: '84000' });
      return done();
    } catch (testErr) {
      return done(testErr);
    }
  });
  ```

### Nitpicks (Optional)

- **[test/integration/2_utils/postalCodeValidation.test.js, ~line 29]** The empty string case is implicitly covered by `validateStringLength` (length 0 ≤ 10 → null), but there is no explicit test for it. The `stringLengthValidation.test.js` file tests `''` directly. A single line alongside the existing null/undefined case would make the boundary explicit:

  ```js
  it('should return null for null, undefined, empty string and non-string values', () => {
    should(validatePostalCodeLength(null)).be.null();
    should(validatePostalCodeLength(undefined)).be.null();
    should(validatePostalCodeLength('')).be.null();
    should(validatePostalCodeLength(84000)).be.null();
  });
  ```

ClemRz (approved):
Clean, well-scoped fix. The new utility follows the `nameValidation` pattern exactly, the validation fires on the already-trimmed value (so whitespace-padded values that fit once trimmed are still accepted), and the Swagger spec is correctly updated. The tests added in the second commit (addressing my earlier review) close the gaps I flagged.

- **[api/controllers/v1/organization/create.js:26–31 / update.js:39–45 — and the rest of TGrotto's bounded fields]** `postalCode` is the only one of TGrotto's nine `maxLength`-constrained string fields that is now guarded. Fields like `address` (200), `mail` (50), `city` / `county` / `region` (100 each), and `url` (200) are still unguarded and would surface the same `E_INVALID_VALUES_TO_SET` 500 if a client sends an over-long value. This PR fixes the reported bug, but it's worth opening a follow-up to apply `validateStringLength` to the remaining bounded fields in one pass — `validateStringLengths` (the multi-field variant in `stringLengthValidation.js`) is already designed for exactly that.

- **[test/integration/4_routes/Organization/create.test.js]** The `should return 400 when postalCode is too long` test doesn't assert that no organization was actually created in the database. For the update controller this is checked explicitly (`should(organization.postalCode).not.equal('4400 Flémalle')`), which is nice. For create, because the guard fires before `GrottoService.createGrotto` is called this is not a real risk, but a short assertion like the one below would make the guarantee explicit and symmetric with the update test:

  ```js
  .end(async (err, res) => {
    if (err) return done(err);
    should(res.body.message).containEql('Postal code');
    // Guard fires before createGrotto, so nothing should have been persisted
    const created = await TGrotto.find({ postalCode: '4400 Flémalle' });
    should(created).have.length(0);
    return done();
  });
  ```

### Nitpicks (Optional)

- **[api/utils/postalCodeValidation.js:1–19]** The `validatePostalCodeLength` wrapper is a faithful copy of `validateNameLength` — same single-field delegation pattern, same JSDoc shape, same `error ? error.message : null` unwrap. If the project accumulates more of these per-field wrappers (one per bounded column), the boilerplate will grow linearly. No action needed now, but when the follow-up for the other bounded fields lands it might be worth evaluating whether a single `makeFieldLengthValidator(fieldLabel, maxLength)` factory would reduce duplication, similar to how `validateStringLengths` already handles the multi-field case.

- **[assets/swaggerV1.yaml, ~line 9239 — Organization schema component]** The `postalCode` property in the `Organization` response schema (`#/components/schemas/Organization`) is typed as `string` and is correct, but it doesn't carry a `maxLength: 10` annotation. The request body and query param schemas now document the constraint, so adding it to the response schema too would make it complete — though response schemas are conventionally less strict than request schemas, so this is truly optional.

PR #1786, opened 2026-08-25
fix(rateLimit): log rate limit rejections

ClemRz:
## 🤔 What

Rate limit rejections now leave a log line. Previously a `429` was invisible in application logs.

- Adds a shared `handler` to all four limiters in `config/rateLimit/rateLimiter.js` (`general`, `delete`, `auth`, `adminAuth`)
- Logs limiter name, IP, method, path, authenticated user, effective limit, hit count and window
- Adds 7 tests covering log contents and response integrity

## 🤷‍♂️ Why

- **A rejected request left no trace.** `generalRateLimit`, `deleteRateLimit` and `authRateLimit` all run *before* `responseTimeLogger` and `requestLogger` in the middleware order, and `express-rate-limit`'s default handler responds without calling `next()`. Neither the `Req ::` nor the `Res ::` line was ever emitted — so there was no way to tell from application logs that a client had been limited, let alone which limiter fired or against which IP.
- **This is a live blind spot, not a hypothetical one.** Verified against production: App Insights recorded **63 × `429`** in the last 30 days, all within the last day — `GET /api/v1/geoloc/entrances` across 4 paths from a single client walking map tiles. The application log for that day contains **1478 `Res ::` lines, none of them `429`**. The burst App Insights captured at 16:37:50 produced **zero** application log lines; the next entry is a health check 8 seconds later.
- **Brute-force attempts were unobservable.** `authRateLimit` and `adminAuthRateLimit` protect login and password-reset. Those are the rejections most worth seeing, and they were silent.
- **There was no other source for the client IP.** Azure is not a usable fallback here:
  - There are **no W3C HTTP logs** on this App Service. `httpLogs.fileSystem` reports `enabled: true`, but `LogFiles/http/` returns 404 — that directory is Windows-only. On Linux the only file log is the application's own stdout.
  - **App Insights anonymizes the client IP.** `client_IP` is `0.0.0.0` on **all 1483 requests** in the window, not only the rejected ones. Only geo-derived `client_City` / `client_CountryOrRegion` survive, and those cannot distinguish several clients behind one regional egress.

  So before this change a rate-limited client's IP was unavailable from any source. The `ip` field in the new log line is the only place it becomes visible.
- Only `adminAuthRateLimit` was incidentally visible, because it sits after the loggers in the chain.

## 🔍 How

`limitExceededHandler(limiterName)` is a factory returning an `express-rate-limit` `handler`, attached to each limiter. Two details drove the implementation, both verified at runtime rather than assumed:

- **`options.limit` is not the effective limit.** It holds the `max` function unresolved, because `options` is computed once when the limiter is created. The resolved per-request values live on `req.rateLimit` (`{limit, used, remaining, resetTime, key}`). Logging `options.limit` would have written a function into the log line — a test pins this.
- **Overriding `handler` replaces the library's default responder entirely**, so the handler must send the response itself. The `429` status, message body, `RateLimit-*` and `Retry-After` headers are all preserved.

The `sails` global is guarded: a throw inside the handler would convert the `429` into a `500` and let the caller past the limit. Verified with the global absent.

Log lines are prefixed with the trace ID (`sails.log` is patched by `api/utils/logger.js`, and `traceId` runs first in the middleware order), so they correlate with the `X-Trace-Id` header the client received.

The middleware `order` in `config/http.js` is untouched — moving the loggers above the limiters would have produced `Res :: … 429` lines but no IP or limiter name.

Sample output:

```
[<traceId>] Rate limit exceeded: {"limiter":"auth","ip":"…","method":"POST",
"path":"/api/v1/login","userId":42,"nickname":"Bob","limit":10,"used":11,"windowMs":900000}
```

The `ip` value carries the port suffix Azure forwards (`<addr>:<port>`), confirmed in production logs — this is the behaviour the existing `keyGenerator` comment documents.

## 🧪 Testing

Rate limiting is disabled in `test` and `development` (`isTestOrDev`), so the tests set `NODE_ENV=production` and exercise the limiters on a bare express app, as the existing suite already did.

```shell
```

- Targeted `rateLimiter.test.js`: **23 passing** (was 16)
- `admin-rate-limit.test.js`: **4 passing**
- Full suite: **3322 passing, 0 failing**
- `npm run lint`: clean

New tests cover: field contents; the resolved-numeric-limit trap; authenticated user attribution; delete-limiter naming; response status/body/headers surviving the handler override; silence while under the limit; and `429` still returned when the `sails` global is absent.

## 📸 Previews

n/a — log output only, sample shown above.

### DIFF
--- a/config/rateLimit/rateLimiter.js
+++ b/config/rateLimit/rateLimiter.js
@@ -49,6 +49,53 @@ const isTestOrDev = () =>
   process.env.NODE_ENV === 'test' || process.env.NODE_ENV === 'development';
 
 /**
+ * Builds the `handler` that express-rate-limit invokes when a limit is exceeded.
+ *
+ * Without this, rejections are invisible in the logs. `generalRateLimit`,
+ * `deleteRateLimit` and `authRateLimit` all run *before* `responseTimeLogger`
+ * and `requestLogger` in the middleware order (see config/http.js), and the
+ * library's default handler responds without calling `next()`. Neither the
+ * `Req ::` nor the `Res ::` line is ever emitted, so a 429 leaves no trace
+ * beyond Azure's own HTTP logs — which carry no trace ID to correlate with.
+ *
+ * Overriding `handler` replaces the default responder entirely, so this must
+ * also send the response itself.
+ *
+ * `options.limit` is NOT the effective limit: it holds the `max` function
+ * unresolved, because `options` is computed once when the limiter is created.
+ * The resolved per-request numbers live on `req.rateLimit`.
+ *
+ * sails.log is patched by api/utils/logger.js to prefix the trace ID, and
+ * `traceId` runs first in the middleware order, so these lines correlate with
+ * the X-Trace-Id header the client received.
+ */
+const limitExceededHandler = (limiterName) => (req, res, next, options) => {
+  // Guarded: a throw here would turn the 429 into a 500 and let the request
+  // through unlabelled. Tests exercise these limiters on a bare express app
+  // where the sails global may be absent.
+  if (typeof sails !== 'undefined' && sails.log && sails.log.warn) {
+    sails.log.warn(
+      'Rate limit exceeded:',
+      JSON.stringify({
+        limiter: limiterName,
+        ip: req.ip,
+        method: req.method,
+        path: req.path,
+        userId: req.token ? req.token.id : undefined,
+        nickname: req.token ? req.token.nickname : undefined,
+        limit: req.rateLimit ? req.rateLimit.limit : undefined,
+        used: req.rateLimit ? req.rateLimit.used : undefined,
+        windowMs: options.windowMs,
+      })
+    );
+  }
+
+  // Mirrors the library's default handler.
+  res.status(options.statusCode);
+  if (!res.writableEnded) res.send(options.message);
+};
+
+/**
  * Returns the general rate limit for a request based on the caller's role.
  * Unauthenticated visitors get the lowest cap; higher roles get progressively
  * more headroom. Administrators are skipped entirely (see `skip`).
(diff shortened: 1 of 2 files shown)


ClemRz:
Verified against production before merge. Two corrections to the PR description, and one confirmation.

**Confirmed — the blind spot is live, not hypothetical.**

- App Insights recorded **63 × `429`** in the last 30 days, all within the last day: `GET /api/v1/geoloc/entrances` across 4 paths from a single client (Fort Smith, Arkansas) walking map tiles.
- Today's Docker stdout log contains **1478 `Res ::` lines, none of them `429`**.
- The 16:37:50 burst that App Insights captured produced **zero application log lines** — the next entry is a health check 8 seconds later.

So `generalRateLimit` is rejecting real traffic today and leaving no trace.

**Correction 1 — there are no W3C HTTP logs on this App Service.** `httpLogs.fileSystem` reports `enabled: true`, but `LogFiles/http/` returns 404; that directory is Windows-only. On Linux App Service the only file log is Docker stdout, which is the app's own output.

**Correction 2 — App Insights anonymizes the client IP.** `client_IP` is `0.0.0.0` on **all 1483 requests** in the window, not only the rejected ones — Azure strips it by default. Only geo-derived `client_City` / `client_CountryOrRegion` survive.

Together these mean that before this change there was **no way to obtain a rate-limited client's IP from any source**. The PR description characterised Azure's logs as a lossy fallback that "records the status code but carries no trace ID"; in fact they are not a fallback for the IP question at all. The `ip` field in the new log line is the only place that information becomes available.

Incidentally this also confirms the port-suffix behaviour the `keyGenerator` comment documents — production log lines carry `"ip":"81.220.109.216:64462"`.

Paul-AUB (dismissed):

1. **[test/integration/2_utils/rateLimiter.test.js:104]** Consider adding one rejection assertion each for `authRateLimit` and `adminAuthRateLimit`. The shared handler behavior is well covered through the general and delete limiters, but directly exercising these two configurations would also pin their handler wiring and logged limiter names.

ClemRz (commented):
Done — added `should name the auth limiter on authRateLimit rejections` and `should name the adminAuth limiter on adminAuthRateLimit rejections` to the rejection logging block. Both exercise the limiter past its limit and assert the expected `limiter` name in the log payload. 25 passing (was 23).

PR #1793, opened 2026-08-25
feat(search): allow sorting document search results by author

### LINKED ISSUE #1789: Allow sorting document search results by author
Split out of [grottocenter-front#1472](https://github.com/GrottoCenter/grottocenter-front/issues/1472) §1/§2, per that issue's own request that *"chaque demande devra faire l'objet d'un ticket spécifique"*.

## Problem

Bibliography search results cannot be ordered by author. The front-end's "sort by author" column is disabled with a comment stating *"the search index exposes no sortable author field"*, and the card view has no author entry in its sort options.

Typesense cannot sort on array fields. A document's authors are stored as `string[]` and split across two relations — persons (`j_document_caver_author` → `authors.nickname`) and organizations (`j_document_grotto_author` → `authorsOrganization.name`). There is no scalar field to sort on.

## Request

Expose a sortable author field on the `documents` collection, usable through the existing generic `sort` parameter of the advanced-search endpoint:

```
POST /api/v1/advanced-search  { "entity": "documents", "sort": "authorsSort:asc" }
```

Requirements:

- Must cover **both** person authors and organization authors — a user sorting "by author" does not care which kind it is.
- Must be scoped to `entity=documents` and return a clean `400` elsewhere, like the existing `dataQuality` guard for `entrances`.
- Must stay consistent between the full reindex and incremental single-document upserts.

## Notes

The existing `dataQuality` computed-sort-field on entrances is the pattern to follow.

Any denormalized-key approach is an approximation: the author join tables have no ordinal column, so true first-listed-author order cannot be recovered. That trade-off should be documented wherever the key is computed.

## Front-end readiness

The UI exists and only needs the field. See front#1472 §1 (table view Author column) and §2 (card view sort options).

ClemRz:
Adds a sortable author key to the `documents` Typesense collection, so bibliography results can be ordered A→Z by author.

Closes #1789

## Why

- **The front-end column is disabled today.** The documents table has a "sort by author" column turned off with a comment saying the search index exposes no sortable author field. This adds that field.
- **Typesense cannot sort on arrays.** A document's authors are `string[]` split across persons (`authors.nickname`) and organizations (`authorsOrganization.name`), so a denormalized scalar key is the only way to sort on them.
- **One "author" concept for users.** The key pools persons and organizations, so sorting by author doesn't require the user to know which kind an author is.

## What

- `api/utils/computeDocumentAuthorsSort.js` — computes the alphabetically-first author name across both pools, normalized (NFKD diacritic stripping, lowercase, whitespace collapse) so byte-order sort matches human A→Z. Authorless documents get a `~` sentinel that sorts last ascending.
- `api/dbSync/entities/document.js` — new `authorsSort` schema field, computed per row during full reindex.
- `api/services/DocumentService.js` — `updateInSearch` recomputes the key on every upsert, so incremental edits match the reindexed baseline.
- `advanced-search.js` — `authorsSort` scoped to `entity=documents`, returning a clean `400` elsewhere. Sorting flows through the existing generic `sort` param; no new wiring.

Follows the existing `dataQuality` computed-sort-field pattern.

## Production actions required

> [!IMPORTANT]
> This changes the Typesense schema. `authorsSort` is **not populated until a full search reindex runs in each environment** — until then, sorting by author returns unsorted results. No DB migration is needed; this is search-index-only.

## Known approximation

This is an **alphabetical-first-author** key, not true bibliographic order. The author join tables have no ordinal column, so first-listed-author order isn't stored and can't be recovered — a work by "Zola & Adam" sorts under **A**. Co-authors don't tie-break, persons and organizations are pooled without preference, non-Latin scripts aren't romanized, and ordering is raw code-point rather than locale-aware. All documented in the helper.

*Upgrade path:* add an ordinal column to the join tables and key on the ranked-first name. Only the helper's name selection changes — schema, query wiring and the front-end contract stay identical.

## Testing

`npm test` → **3357 passing, 0 failing**. Includes unit tests, property-based tests for the helper invariants (normalization idempotency, sentinel ordering, pool commutativity), and route-level tests pinning the `400` entity guard.

## Provenance

Extracted from #1742, authored by @dawoldo during their internship; original commits and authorship preserved. That PR bundled two independent features — this is the search half. The citation half is in the companion PR. Verified the two branches together reproduce #1742 byte-for-byte, plus a fix for the `authorsGrotto` → `authorsOrganization` rename that landed on `develop` after the original work.

### DIFF
--- a/api/controllers/v1/search/advanced-search.js
+++ b/api/controllers/v1/search/advanced-search.js
@@ -9,6 +9,7 @@ const normalizeDataQualityFilter = require('../../../utils/normalizeDataQualityF
 // Sort fields that are only valid for specific entities
 const ENTITY_SPECIFIC_SORT_FIELDS = {
   dataQuality: ['entrances'],
+  authorsSort: ['documents'],
 };
 
 module.exports = async (req, res) => {
--- a/api/dbSync/entities/document.js
+++ b/api/dbSync/entities/document.js
@@ -1,4 +1,7 @@
 const exportUtils = require('../utils');
+const {
+  computeDocumentAuthorsSort,
+} = require('../../utils/computeDocumentAuthorsSort');
 
 const query = `
     SELECT
@@ -192,6 +195,11 @@ async function* processRows(source) {
       ];
       row.cave = row.cave?.[0] ?? null;
 
+      row.authorsSort = computeDocumentAuthorsSort(
+        row.authors?.map((e) => e.nickname),
+        row.authorsOrganization?.map((e) => e.name)
+      );
+
       yield row;
     }
   }
@@ -298,6 +306,11 @@ module.exports = {
           facet: true,
           optional: true,
         },
+        // Denormalized scalar key for sorting biblio results by author.
+        // Typesense cannot sort on the `authors.nickname` array field, so this
+        // holds the alphabetical-first author name (persons + organizations).
+        // See api/utils/computeDocumentAuthorsSort.js for method & limitations.
+        { name: 'authorsSort', type: 'string', optional: true, sort: true },
         { name: 'cave.name', type: 'string', optional: true, sort: true },
         { name: 'entrances.name', type: 'string[]', optional: true },
         { name: 'massifs.name', type: 'string[]', optional: true },
--- a/api/services/DocumentService.js
+++ b/api/services/DocumentService.js
@@ -13,6 +13,9 @@ const {
   TYPES_ALLOWING_PARENT,
   TYPES_REQUIRING_PARENT,
 } = require('../../config/constants/document');
+const {
+  computeDocumentAuthorsSort,
+} = require('../utils/computeDocumentAuthorsSort');
 
 // Normalize a collection value to a plain ID (handles both raw IDs and objects).
 const normalizeToId = (item) =>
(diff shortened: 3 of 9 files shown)


Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[api/services/DocumentService.js:103] Incremental indexing can compute the key from a secondary organization name.** `getPopulatedDocument()` fills `authorsOrganization` through `NameService.setNames()`, which puts every name in `organization.names` without an ordering guarantee and exposes the designated main name separately as `organization.name`. The full reindex explicitly selects `n.is_main = true`, but this path reads `names[0]`; after an organization changes its main name, a document upsert can therefore produce a different `authorsSort` key from a full reindex. Please compute the incremental key from `e.name` (and add a regression case with multiple names whose main name is not the first array element) so the two indexing paths stay consistent.

2. **[api/utils/computeDocumentAuthorsSort.js:18] The `~` sentinel does not sort after all valid normalized author names.** Normalization deliberately retains some non-ASCII Latin letters as well as Cyrillic and CJK characters; those values compare after `~`, so ascending order places an authorless document before authors whose keys begin with those characters. The property test currently avoids this by restricting generated names to characters below `~`, but documents are not restricted to that alphabet. Please use an empty-author representation that Typesense will place after the full supported character range and cover non-Latin names in the ordering regression tests.

Paul-AUB (approved):
Reviewed with priority on the business logic, search-index consistency, and public API surface because this diff exceeds 500 changed lines.

The previously reported incremental organization-name mismatch and authorless ordering defect are both addressed at the current tip. The focused author-sort helper suite passes (24 tests), and the PR's build/test, static analysis, CodeQL, and coverage checks are green. I found no remaining blocking issues.

PR #1794, opened 2026-08-25
feat(documents): expose bibliographic citation fields on embedded document lists

### LINKED ISSUE #1790: Expose bibliographic citation fields on embedded document lists (ISO 690)
Split out of [grottocenter-front#1472](https://github.com/GrottoCenter/grottocenter-front/issues/1472) §3, per that issue's own request that *"chaque demande devra faire l'objet d'un ticket spécifique"*.

## Problem

Wherever a document appears **embedded in another entity** — a caver's publications, a cave's or entrance's or massif's document list — it is serialized by `toSimpleDocument`, which carries only `{ id, type, title, description, files, dates, isValidated, isDeleted }`.

That is not enough to render a bibliographic reference. The front-end therefore shows the bare title, and cannot build the ISO 690 display requested in front#1472 §3:

- **Articles:** `LASTNAME, Firstname, Year. Article title. Journal name. Volume, number, pages`
- **Books:** `LASTNAME, Firstname, Year. Book title. Edition. City: Publisher. Collection. ISBN`

The data all exists server-side; it is simply dropped when a document is embedded.

## Request

Expose the fields a citation needs on embedded document lists: `authors`, `authorsOrganization`, `editor`, `library`, `issue`, `pages`, `identifier` / `identifierType`, the parent title, and the legacy `oldBBS` block.

Affected endpoints — every one that embeds a document list:

- `GET /api/v1/cavers/{id}` and the caver delete response
- `GET /api/v1/caves/{id}`
- `GET /api/v1/entrances/{id}`
- `GET /api/v1/massifs/{id}`

Constraints:

- The extra joins must not be paid by callers that only need a title and a type — keep the light path available.
- The advertised OpenAPI shape must match what is actually populated on every route that returns it. A converter promising citation fields on a route that does not populate them is a contract violation.
- Organization names live in a separate `TName` collection, and a document's title lives in its `descriptions` — neither can be reached through a plain association populate.

## Related

- #1746 — exposing the person/organization `type` on these same embedded author references is the natural next increment.

ClemRz:
Embedded document lists now carry the fields needed to render a full bibliographic reference, instead of just a title.

Closes #1790

## Why

- **ISO 690 display is impossible today.** Wherever a document is embedded in another entity, `toSimpleDocument` emits only `{ id, type, title, description, files, dates, flags }`. The front-end can't build `LASTNAME, Firstname, Year. Title. Journal. Volume, pages` from that, so it shows a bare title.
- **The data already exists server-side** — it's simply dropped when a document is embedded in a caver, cave, entrance or massif payload.
- **No extra round-trips.** The alternative is one API call per document to build a reference list, which doesn't scale on a collection page.

## What

- `DocumentService.getDocumentsForCitation()` — like `getDocuments()`, but also populates `identifierType`, `authors`, `authorsOrganization`, `editor`, `library`, and resolves parent descriptions plus organization names (neither reachable through a plain association populate). Kept separate so callers needing only a title and type don't pay the extra joins.
- `converters.toCitationDocument` — spreads `toSimpleDocument` and adds the citation fields. Purely additive, so existing consumers are unaffected.
- Wired into the caver, cave, entrance and massif detail paths, plus the caver delete response.
- OpenAPI: new `CitationDocument` and `SimpleOrganization` schemas; `name` marked `nullable` since `getMainName` can return `null`.

The parent is deliberately flattened to `{ id, title }` rather than nested as a document — a citation only needs the parent title, and nesting would let a card drag in a whole parent chain.

## Deployment

No DB migration, no reindex, no config. Backward-compatible.

## Testing

`npm test` → **3339 passing, 0 failing**. Includes a route test pinning the citation fields on the caver delete response.

## Provenance

Extracted from #1742, authored by @dawoldo during their internship; original commits and authorship preserved. That PR bundled two independent features — this is the citation half, which needs no Typesense reindex and so shouldn't be gated behind the search half (companion PR #1793). Verified the two branches together reproduce #1742 byte-for-byte.

One fix on top of the original: the citation path populated `authorsGrotto`, which `develop` renamed to `authorsOrganization` after the work was written — populating the old name silently returned no organization authors.


Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[assets/swaggerV1.yaml:8276]** The new citation schemas reject values that the converter legitimately serializes as `null`. `toCitationDocument` passes through nullable `datePublication`, `dateReviewed`, `issue`, `pages`, and legacy BBS columns, while `convertIfObject` returns `null` for an absent `editor`, `library`, or `parent`. `SimpleOrganization.language` can likewise be `null` whenever `getMainName` is `null`, because `getMainLanguage` uses the same missing main-name record and `TGrotto` has no fallback language field. Only `SimpleOrganization.name` is currently marked nullable. This makes valid endpoint responses violate the newly advertised OpenAPI contract and can break generated clients or response validators. Please either mark all nullable response properties appropriately (using an OpenAPI-compatible nullable wrapper for `$ref` values) or omit those properties when their values are null.

2. **[test/integration/4_routes/Cavers/delete.test.js:105]** Consider pinning the citation payload on each affected GET route (caver, cave, entrance, and massif), including an organization name, flattened parent title, and legacy BBS data. The current added assertion covers only the caver delete response and two fields, so a missed populate on any of the primary routes could still pass the suite.

Paul-AUB (changes requested):
Review prioritized the business logic, public API contract, and integration coverage because this diff changes 507 lines.

### Issues (Must Fix)

1. **[api/services/mapping/converters.js:248]** The flattened parent still does not provide enough data for the ISO 690 article citation this PR is meant to enable. The domain explicitly models `Article -> Issue -> Collection` and forbids storing `issue` on an Article (`config/constants/document.js:13-25`, `api/controllers/v1/document/create.js:28-35`), but this converter exposes only the direct Issue's `{ id, title }`; `getDocumentsForCitation()` likewise fetches only that parent's descriptions. Consequently an embedded Article loses the Issue's number/publication data and the Collection title needed as the journal name. This is also the gap called out in the latest upstream `grottocenter-front#1472` discussion: the issue title and periodical title are distinct. Please expose enough of the parent chain (or normalized citation fields) to render the complete article reference, and cover an Article -> Issue -> Collection case end to end; the current test only exercises Issue -> Collection and therefore cannot catch this regression.

Paul-AUB (approved):
Reviewed the latest commit (`3874e2c2`), prioritizing the citation-resolution logic, public API contract, and integration coverage because this diff changes 619 lines.

No blocking issues remain. The previously requested nullable schema fields, route-level payload coverage, and full Article -> Issue -> Collection citation chain are now addressed, and the current CI checks pass.

Issue #1796, opened 2026-08-27
Inconsistent role gating on delete/restore: Moderators can permanently delete core content


ClemRz:
**Describe the bug**

Three related inconsistencies in how `RightService` group checks are applied across delete/restore controllers. Roles are **not hierarchical** (`PERMISSION_SYSTEM.md`), so which group a controller checks decides exactly who can act — and the checks were not applied uniformly.

Grouping these together because they share one root cause and one fix decision: pick the intended rule per operation, then make every controller match.

---

### 1. Permanent deletion of core content is gated on Moderator, not Administrator

For 10 of 13 soft-deletable entities the `isPermanent` branch is reached after **only** a `MODERATOR` check, with no Administrator re-check before the irreversible hard delete.

`api/controllers/v1/cave/delete.js` is representative — `MODERATOR` at line 11, then `deletePermanently` read at line 47 and acted on at line 55, no further gate:

| Gate before hard delete | Controllers |
|---|---|
| `MODERATOR` only | `cave`, `comment`, `description`, `document`, `entrance`, `history`, `location`, `massif`, `organization`, `rigging` |
| Re-checks `ADMINISTRATOR` | `device` (line 13), `sensor-configuration`, `guideline` (line 38) |

`device/delete.js` shows the intended shape: Moderator to soft-delete, Administrator required to escalate to permanent.

Consequence: a Moderator can irreversibly destroy caves, entrances and documents. `PERMISSION_SYSTEM.md` claimed permanent deletion was Administrator-only until this audit; the doc has been corrected to match the code, but the code is likely what's wrong.

### 2. Restore is Moderator-only for every entity except guidelines

12 of 13 restore controllers check `MODERATOR` alone. `guideline/restore.js` (lines 10-14) accepts Moderator **or** Administrator.

So guideline is the outlier here — either it should drop to `MODERATOR` only, or the other 12 should accept both.

### 3. `device`/`sensor-configuration` and `guideline` disagree on restore

A narrower instance of #2 between entities modelled almost identically: `device/restore.js` (line 9) and `sensor-configuration/restore.js` check `MODERATOR` only, while `guideline/restore.js` accepts either role. Their `delete.js` counterparts *all three* accept either role — so within the same entity, delete and restore use different rules.

---

**To Reproduce** (case 1)

1. Authenticate as a user holding **only** the Moderator group.
2. `DELETE /api/v1/caves/<id>?isPermanent=true`
[...]

PR #1797, opened 2026-08-27
docs(permissions): document guidelines, org associations and role gating

ClemRz:
## 🤔 What

Brings `PERMISSION_SYSTEM.md` back in line with the code. Documentation only — no behaviour changes.

- Documents **legislation guidelines** (#1452) across the role sections, the permission matrix and a new Special Permission Cases subsection
- Documents **responsible organization associations** (#1598) the same way
- Corrects **six matrix rows** that credited Administrators with powers the controllers gate on `MODERATOR`
- Splits the delete/restore rows by entity class, because no single rule covers them
- Adds a **Known Permission Inconsistencies** section separating the defects tracked in #1796 from behaviour that is accepted
- Records that the open `v1/observation/import` endpoint is intentionally unrestricted
- Documents the `['tokenAuth', 'validateId']` ordering and the two guideline routes that skip `validateId` on purpose

## 🤷‍♂️ Why

Both features shipped **after** the doc's last update (`7f7a4db8`, 15 June) yet appeared nowhere in it: guidelines landed 4 June (`2fb9fa36`), organization associations 25 June (`7902a330`). A grep for `guideline`, `legislation`, `regulation` or `responsible` returned nothing.

Two claims had also become wrong rather than merely incomplete:

- *"Users can modify any content except other's comments"* — guidelines are a second exception
- *"Rollback to a previous version of any content"* — guidelines are own-only, and are the **only** entity with a rollback route, so the single thing this claim described was the thing it got wrong

Auditing the rest of the matrix then surfaced six rows resting on the assumption that Administrator is a superset of Moderator — which the doc's own preamble explicitly denies. Being wrong in the *permissive* direction is the risk here: a reader planning a change may conclude a destructive operation is admin-gated when it is not.

## 🔍 How

Every row was checked against the controller that enforces it, not against the surrounding prose.

The corrected rows are soft delete, restore, permanent delete, view soft-deleted content, validate documents, and manage document duplicates. All six had ✅ for Administrator where the controller checks `MODERATOR` alone.

Delete and restore are now split by entity class, since one row cannot express the real rule:

| Operation | Moderator-only | Either role | Administrator-only |
|---|---|---|---|
| Soft delete | 10 core entities | caver, device, sensor-config, guideline | — |
| Restore | 12 entities | guideline | — |
| Permanent delete | 10 core entities | — | device, sensor-config, guideline |

The `Known Permission Inconsistencies` section then draws the distinction the matrix cannot:

- **Tracked as defects** (#1796) — Moderators can permanently delete core content; the two restore-gating outliers
- **Accepted** — Administrators lacking moderation powers, since in practice they hold every group and cumulate them. The matrix still marks these ❌ because it records what each role grants *on its own*

The observation-import note is deliberately emphatic ("Do not 'fix' the missing role check here"). Its absent role check is indistinguishable from an oversight when reading the code, and it had already been flagged as a gap once during this audit before being confirmed intentional.

## 🧪 Testing

No tests apply — `npm run lint` is scoped to `**/*.js`, so this file is outside it, and there is no markdown linter configured.

Verified manually instead:

- Every matrix row traced to its enforcing controller
- Table integrity checked with an `awk` column-count pass across all rows
- All referenced issue numbers resolved via `gh issue view`

One correction worth flagging for review: I initially attributed the guideline permanent-delete work to `#1208`, taken from the `Fix/1208` branch name. In this repo `#1208` is an unrelated Dependabot PR — the number refers to an issue elsewhere — so that reference was dropped rather than left as a misleading cross-link.

## 📸 Previews

N/A — markdown only.

## Related

- #1796 — the authorization gaps this doc now records (opened alongside this work)
- #1452 — legislation guidelines
- #1598 — organization responsible for geographic entity

### DIFF
--- a/PERMISSION_SYSTEM.md
+++ b/PERMISSION_SYSTEM.md
@@ -30,8 +30,8 @@ Roles are **not hierarchical**; for instance, an Administrator does not automati
     - Edit caves, entrances, documents, massifs, organisations, descriptions, locations, riggings, histories
     - Edit own comments
     - Edit own legislation guidelines
-    - Unlink documents from entities
-    - Rollback to a previous version of any content, except guidelines (own only)
+    - Rollback own legislation guidelines to a previous version (guidelines are the only entity exposing a rollback
+      route)
     - Associate/dissociate responsible organizations for any country, region, or massif
     - Add/remove own explored entrances
     - Set main name for entities
@@ -77,6 +77,7 @@ Roles are **not hierarchical**; for instance, an Administrator does not automati
       see "Caver Deletion")
     - Update any user's comments
     - Update/rollback any user's legislation guidelines
+    - Unlink documents from caves, entrances, and massifs
     - Update documents that have modifications pending moderator approval
     - Validate documents
     - Manage duplicates (documents and entrances)
@@ -159,7 +160,8 @@ Roles are **not hierarchical**; for instance, an Administrator does not automati
 | **Document Management**                                       |
 | Validate documents                                            | ❌ | ❌ | ❌ | ✅ | ❌ |
 | Manage document duplicates                                    | ❌ | ❌ | ❌ | ✅ | ❌ |
-| Unlink documents from entities                                | ❌ | ✅ | ✅ | ✅ | ✅ |
+| Link a document to a cave/entrance/massif                     | ❌ | ✅ | ✅ | ✅ | ✅ |
+| Unlink documents from entities                                | ❌ | ❌ | ❌ | ✅ | ❌ |
 | Update a document with modifications pending approval         | ❌ | ❌ | ❌ | ✅ | ❌ |
 | **Scientific Domain (Devices, Sensors, Observations)**        |
 | View device/sensor configuration details                      | ✅ | ✅ | ✅ | ✅ | ✅ |
@@ -266,6 +268,14 @@ Countries, regions, and massifs can be linked to the organizations in charge of 
 - **Comment Updates**: Users can update their own comments; Moderators can update any comments
 - **Guideline Updates**: Users can update and roll back only their own guidelines; Moderators and Administrators can
   update and roll back any guideline
+
+### Document Linking
+Linking and unlinking are **not** gated symmetrically for caves, entrances, and massifs:
+- **Link** (`add-document`): any authenticated user — the controllers perform no role check at all
+- **Unlink** (`unlink-document`): `MODERATOR` only, and `ADMINISTRATOR` is *not* accepted as an alternative
+
+So an ordinary user can attach a document to a cave but cannot detach it again, and a non-Moderator Administrator
+cannot detach it either.
 
 ### Snapshot and History Access
 - **Public History**: Available to all users for non-sensitive content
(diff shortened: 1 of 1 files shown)


Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[PERMISSION_SYSTEM.md:327]** The caver entry does not match `api/controllers/v1/caver/delete.js`. Caver deletion is a permanent account deletion rather than the soft-delete/restore flow described by this section, and the controller does not accept either role interchangeably: deleting a `CAVER` requires `ADMINISTRATOR`, while deleting an `AUTHOR` requires `MODERATOR` (lines 27-30). Please remove cavers from this soft-delete grouping or document their separate type-dependent deletion rules so readers are not told that either role can perform both operations.

Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[PERMISSION_SYSTEM.md:34]** This still grants Users rollback access to "any content," but the only rollback route in `config/routes.js` is `v1/guideline/rollback`; the new guideline section also explicitly says guidelines are the only entity exposing rollback. Consider replacing this with a guideline-specific permission such as "Rollback own legislation guidelines" so the role summary no longer contradicts both the code and line 242.
2. **[PERMISSION_SYSTEM.md:33,162]** The User summary and matrix still say every authenticated role can unlink documents, but `api/controllers/v1/cave/unlink-document.js`, `entrance/unlink-document.js`, and `massif/unlink-document.js` all require `MODERATOR` and do not accept `ADMINISTRATOR` as an alternative. Since this PR says every matrix row was checked against its controller, please remove this capability from User/Leader/Administrator and mark the matrix row as Moderator-only (or document any narrower unlink path that supports the current entries).

3. **[PERMISSION_SYSTEM.md:409]** Consider separating the reasons the two guideline routes omit `validateId`. The ISO `:snapshotId` incompatibility applies to `rollback`, but `get-snapshots` has only the numeric `:id` parameter and could use `validateId`; it currently omits the policy because its controller performs its own numeric check. As written, the shared explanation can make it sound as though both routes contain an ISO snapshot parameter.

Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[PERMISSION_SYSTEM.md:266-267]** The updated ownership summary is still broader than the controller rules and now contradicts the matrix immediately above it. A user also cannot update a document while `modifiedDocJson` is present (`api/controllers/v1/document/update.js:19-28`), and an Administrator who does not hold `MODERATOR` cannot update another user's comment or such a pending document (`api/controllers/v1/comment/update.js:14-21`). Please narrow these two bullets or list those exceptions so the summary preserves the role distinctions this PR documents elsewhere.

Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[PERMISSION_SYSTEM.md:377]** The statement that passing `entityId` merges the caver's "contributions and associations" is broader than the behavior in `api/controllers/v1/caver/delete.js`. That controller only reassigns a fixed set of legacy models and collections; it does not reassign `TGuideline`/`HGuideline`, the scientific-data models, or the `JOrganizationCountry`/`JOrganizationRegion`/`JOrganizationMassif` `author` and `reviewer` references introduced by the features documented here. Those tables have foreign keys to `t_caver` without an `ON DELETE` action, so a caver referenced by one of them cannot necessarily be deleted or merged as this paragraph promises. Please scope this description to the entities the controller actually handles and record the unsupported references as a known limitation, or update the deletion path before retaining the blanket claim.

2. **[PERMISSION_SYSTEM.md:344-349]** Consider changing the devices/sensor-configurations/guidelines bullet to say explicitly that they accept either role for **soft deletion**. Under the current "delete/restore" heading, line 348 reads as if both operations accept either role, while the next line correctly says device and sensor-configuration restore are Moderator-only.
3. **[PERMISSION_SYSTEM.md:246-250]** Consider saying that an author must ask a Moderator **or Administrator** to remove a guideline. The preceding permission rule allows either role, so naming only Moderator makes the summary narrower than the documented behavior.

Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[PERMISSION_SYSTEM.md:385-394]** The new limitation still understates and partly misdescribes the failure mode. `api/controllers/v1/caver/delete.js` performs all reference updates and collection cleanup before `TCaver.destroyOne` at line 168, without a transaction, so an untouched foreign key can reject the final delete only after the earlier mutations have committed. The caver can therefore be partially merged/reassigned rather than being unable to be "merged at all." The untouched-reference list is also incomplete: `TTimeSeriesQualityLog.changedBy` (`sql/9_04_2026_06_03_scientific_observations_ddl.sql:369`) and `TJobBatch.initiator` (`sql/9_16_2026_06_26_job_batch_table.sql:15`) both reference `t_caver` without an `ON DELETE` action and are not handled by this controller, so either row can block the final delete too. Please include those references and document the partial-write behavior so this section matches the actual consequences of the code path.

Paul-AUB (approved):

Issue #1782, opened 2026-08-23
Add a public guideline detail endpoint


Paul-AUB:
# Add a public guideline detail endpoint

## Context

The frontend needs a dedicated public page at `/ui/guidelines/:id`. The API
currently exposes the paginated guideline list, but no GET endpoint for one
guideline. Loading successive list pages to find an ID would be unreliable and
inefficient.

## Endpoint

```http
GET /api/v1/guidelines/{id}
```

- Public: no authentication required.
- Return only non-deleted guidelines.
- Return `404` when the guideline does not exist or is soft-deleted.

## Required response

```json
{
  "id": 123,
  "title": "Cave access restrictions",
  "description": "Visitors must contact the local organization.",
  "language": {
    "id": "fra",
    "refName": "French"
  },
  "countries": [
    { "id": "FR", "name": "France" }
  ],
  "regions": [
    { "id": "FR-01", "name": "Ain", "countryId": "FR" }
  ],
  "massifs": [
    { "id": 42, "name": "Vercors" }
  ],
  "author": {
    "id": 7,
    "nickname": "Paul"
  },
  "reviewer": null,
  "dateInscription": "2026-05-12T09:30:00.000Z",
  "dateReviewed": null,
  "isDeleted": false
}
```

Geographical relations must be hydrated with readable names. Regions must also
include `countryId`, which is required to build the frontend route
`/ui/countries/:countryId/regions/:regionId`.

## Acceptance criteria

- `GET /api/v1/guidelines/{id}` is available without a JWT.
- A non-deleted guideline returns `200` with the shape above.
- Countries, regions and massifs contain IDs and readable names; regions contain
  their country ID.
- A missing or soft-deleted guideline returns `404` without exposing its data.
- The endpoint does not introduce N+1 queries for geographical relations.
- The OpenAPI specification documents the endpoint, response schema and errors.
- Automated tests cover public access, hydrated relations, missing IDs and
  soft-deleted guidelines.

PR #1798, opened 2026-08-31
fix(guideline): allow guidelines without geography, relax edit rights, add public detail endpoint

### LINKED ISSUE #1775: Autoriser une directive sans association géographique et les droits d'édition
## Comportement proposé pour permettre de supprimer une association à une directive

L’API doit autoriser qu’une directive ne soit associée à aucun pays, aucune région et aucun massif.

En particulier, le retrait de la dernière association via PATCH /api/v1/guidelines/{id} doit réussir sans supprimer la directive. Celle-ci doit rester disponible afin de pouvoir être associée ultérieurement à une autre entité.

De plus, la modification (PATCH et/ou PUT, y compris la restauration d'une version précédente) doit être disponible à tous les utilisateurs, pas juste à l'auteur.

ClemRz:
Closes #1775 and #1782.

## 🤔 What

- **Guidelines without geography are now valid** — the empty-geo guard is gone from `PATCH /api/v1/guidelines/:id` (#1775)
- **Editing is open to any authenticated user** — the author/moderator/admin check is gone from `update` and `rollback` (#1775)
- **New public `GET /api/v1/guidelines/:id`** returning a detail view with readable names for `language`, `countries`, `regions` and `massifs` (#1782)

## 🤷‍♂️ Why

The two issues are really one story: the front end needs a standalone guideline page, and the rules that stopped it from existing had to go first.

- **A guideline is a document, not a map pin.** Requiring at least one country, region or massif on update meant an editor could never remove the last association, even when the guideline was genuinely global in scope. #1775 asks for that constraint to be lifted on update — it stays on create, where it still reflects a sensible default.
- **The author check was blocking the wiki model.** Guidelines are collaboratively edited; gating updates on authorship contradicted how the rest of the contribution flow works. Authentication (`tokenAuth`) remains the gate.
- **The detail page had no endpoint to call.** The list and by-entity endpoints return bare FK codes (`"fra"`, `"FR"`, `"FR-01"`), which is fine for a list but forces the front end into a fan-out of follow-up requests just to render one page. `GET /:id` hydrates those relations server-side instead.

## 🔍 How

The detail response is served by a **new** `toGuideline` converter rather than by enriching the shared `toSimpleGuideline`. That distinction is the main design decision here:

- `toSimpleGuideline` is shared by `find-all`, `find-for-entity`, `get-snapshots`, `toMassif`, `toRegion` and `toCountry`. Turning its `regions: ["FR-01"]` into `regions: [{ id, name, countryId }]` would be a **response type change on six endpoints** that neither issue asked to touch, so it keeps its existing shape.
- `toGuideline` spreads it and overrides `countries` / `regions`, and is used only by `GET /:id`.
- Likewise `GuidelineService.getGuidelineDetail` is separate from `getGuideline`, so the create / update / rollback / restore / delete responses are untouched.

Detail-only hydration: `.populate('language')` for the readable `refName`, and massif names via `NameService.setNames` — they live in `t_name`, not on the massif row. That's **one batched query regardless of massif count**, matching the existing `CaveService` idiom; I verified it with a `TName.find` call counter (1 call for 4 massifs).

Swagger gets a new `GuidelineDetail` schema so the shared `Guideline` schema stays accurate for the leaner endpoints, and the now-unreachable `403` responses were dropped from the `patch` and `rollback` operations.

## 🧪 Testing

Full suite **3357 passing / 0 failing**; `npm run lint` clean.

`find.test.js` creates its own guideline in `before()` rather than asserting on seeded guideline 1. This is deliberate, not incidental: `update.test.js`'s new empty-geo test clears guideline 1's associations, and both files run in the same shard — asserting on the seeded row failed under `mocha update.test.js find.test.js`. Same workaround already documented in `rollback.test.js`. Coverage includes the hydrated `language.refName` / `countries[].name` / `regions[].countryId` / `massifs[].name`, an empty-geo guideline (#1775), 404 for soft-deleted and non-existent ids, no-auth 200, plus a guard asserting `by-entity` still returns bare ISO strings.

## 🔗 Companion front-end change

The detail endpoint returns `language` as an object, whereas `Contribution.jsx` previously assumed a string. That is **already handled** in GrottoCenter/grottocenter-front#1534, so no further front-end work is needed for this to ship.

The list and by-entity endpoints are unaffected, since `toSimpleGuideline` is unchanged — the new object shape is scoped to consumers of the detail endpoint.

## 📸 Previews

`GET /api/v1/guidelines/:id` (verified against fixtures):

```json
{
  "id": 1,
  "language": { "id": "fra", "refName": "French" },
  "countries": [{ "id": "FR", "name": "France" }],
  "regions": [{ "id": "FR-01", "name": "Ain", "countryId": "FR" }],
  "massifs": [{ "id": 1, "name": "The massif with name 9" }]
}
```

### DIFF
--- a/PERMISSION_SYSTEM.md
+++ b/PERMISSION_SYSTEM.md
@@ -13,7 +13,7 @@ Roles are **not hierarchical**; for instance, an Administrator does not automati
 - **Permissions**:
   - View public cave/entrance/document/organisation/massif/person data
   - Search through cave/entrance/document/organisation/massif/person/device data
-  - View legislation guidelines (list, by geographic entity, and snapshots)
+  - View legislation guidelines (list, single guideline by id, by geographic entity, and snapshots)
   - View the organizations responsible for a country/region/massif
   - View complete entity history
   - View statistics
@@ -29,9 +29,9 @@ Roles are **not hierarchical**; for instance, an Administrator does not automati
     - Create legislation guidelines
     - Edit caves, entrances, documents, massifs, organisations, descriptions, locations, riggings, histories
     - Edit own comments
-    - Edit own legislation guidelines
-    - Rollback own legislation guidelines to a previous version (guidelines are the only entity exposing a rollback
-      route)
+    - Edit any legislation guideline, including other users' (no ownership or role check)
+    - Rollback any legislation guideline to a previous version, including other users' (guidelines are the only entity
+      exposing a rollback route)
     - Associate/dissociate responsible organizations for any country, region, or massif
     - Add/remove own explored entrances
     - Set main name for entities
@@ -76,7 +76,6 @@ Roles are **not hierarchical**; for instance, an Administrator does not automati
     - Permanently delete author records (`type: AUTHOR`; deleting a `CAVER` account requires Administrator instead —
       see "Caver Deletion")
     - Update any user's comments
-    - Update/rollback any user's legislation guidelines
     - Unlink documents from caves, entrances, and massifs
     - Update documents that have modifications pending moderator approval
     - Validate documents
@@ -106,9 +105,9 @@ Roles are **not hierarchical**; for instance, an Administrator does not automati
   - **System Operations**:
     - Import data from CSV files (documents and entrances)
     - System configuration and maintenance
-  - **Content Moderation** (guidelines only; other content moderation is Moderator-only):
-    - Update/rollback any user's legislation guidelines
-    - Delete/restore legislation guidelines
+  - **Content Moderation** (guideline deletion only; other content moderation is Moderator-only):
+    - Delete/restore legislation guidelines — the only guideline operation an Administrator gates, since updating and
+      rolling back them is open to any authenticated user
   - **Sensitive Data Management**:
     - View coordinates of sensitive entrances
     - Remove sensitive flag from entrances
@@ -134,7 +133,7 @@ Roles are **not hierarchical**; for instance, an Administrator does not automati
 | Search content (including devices)                            | ✅ | ✅ | ✅ | ✅ | ✅ |
 | View statistics                                               | ✅ | ✅ | ✅ | ✅ | ✅ |
 | View history/snapshots                                        | ✅ | ✅ | ✅ | ✅ | ✅ |
-| View guidelines (list/by-entity/snapshots)                    | ✅ | ✅ | ✅ | ✅ | ✅ |
+| View guidelines (list/by-id/by-entity/snapshots)              | ✅ | ✅ | ✅ | ✅ | ✅ |
 | View responsible organizations of country/region/massif       | ✅ | ✅ | ✅ | ✅ | ✅ |
 | **Content Creation**                                          |
 | Create caves/entrances/documents/organisations/massifs        | ❌ | ✅ | ✅ | ✅ | ✅ |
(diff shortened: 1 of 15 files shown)


Paul-AUB (changes requested):
Reviewed all 11 changed files. Since this diff exceeds 500 changed lines, I prioritized the authorization changes, service/converter behavior, public API and OpenAPI contracts, and integration coverage.

### Issues (Must Fix)

1. **[assets/swaggerV1.yaml:8744]** The new detail response references the shared `Language` schema, but the endpoint actually serializes the readable name as `language.refName` (`toLanguage` in `api/services/mapping/converters.js`), while that schema declares `ref_name` at line 8827. As a result, the newly documented endpoint advertises a property it does not return and omits the `refName` property required by issue #1782. Please make the OpenAPI schema match the response (for example, by correcting the shared schema if that is safe or by defining the detail endpoint's language object explicitly).

2. **[test/integration/4_routes/Guidelines/rollback.test.js:92]** The rollback success case uses `user1` (ID 3) against a guideline whose author is also ID 3, so it would still pass with the authorization check removed by this PR restored. Consider adding a success case with an authenticated non-author who is neither moderator nor administrator, as `update.test.js` does, so the newly relaxed rollback permission is protected against regression while the existing 401 case continues to verify authentication.

### Nitpicks (Optional)

3. **[test/integration/1_services/GuidelineService.property.test.js:67]** The two `validateMassifIds` properties cover unchanged behavior and are unrelated to either linked issue. Consider keeping this PR focused by moving that coverage to a separate change or a dedicated validator test file.

Paul-AUB (dismissed):
Reviewed the current head (`ceac9fd1`). The previously reported OpenAPI mismatch and rollback-RBAC coverage gap are resolved. Because this diff exceeds 500 changed lines, I prioritized the authorization changes, detail service/converter behavior, public API and OpenAPI contracts, and their integration coverage.

The latest CI run reached 3,397 passing tests. Its single failure is the unrelated timing-sensitive `Cavers/login-banned.test.js` metadata assertion; the new guideline detail, hydration, empty-geography, update-RBAC, and rollback-RBAC cases all passed.

1. **[test/integration/2_utils/computeDocumentAuthorsSort.property.test.js:27]** The Unicode-generator and normalization-precondition fixes are useful, but they are unrelated to either guideline issue and arrived as separate `test(search)` commits. Consider extracting them into a dedicated PR so this change stays focused and the search-test stabilization can be reviewed and tracked independently.

### Nitpicks (Optional)

2. **[test/integration/4_routes/Guidelines/update.test.js:103]** The two new regression-test names still say the endpoint “currently returns” the old error, while both cases now pass on this branch. Consider describing them as regression cases (and trimming the “WILL FAIL on unfixed code” commentary) so future test output reflects the current expected behavior.

Paul-AUB (dismissed):

Paul-AUB (dismissed):
Reviewed the current head (`4c4ab7fc`). The previously reported OpenAPI mismatch, rollback-RBAC coverage gap, and stale update-test names are resolved. Because this diff exceeds 500 changed lines, I prioritized authorization behavior, the detail service/converter, public API and OpenAPI contracts, and their integration coverage; I also reviewed the new permission-documentation reconciliation. The current Analyze, CodeQL, build-test, and coverage checks all pass.

1. **[PERMISSION_SYSTEM.md:239]** The new read-policy explanation does not match the implemented behavior in two places. `validateId` responds through `res.notFound`, so a malformed ID currently yields `404`, not `400`. Also, although the detail controller hides a soft-deleted guideline, the public `/guidelines/:id/snapshots` controller does not check the live row or its deletion state, and the update trigger records the pre-delete title/description/language in `h_guideline`; callers can therefore still read that data through snapshots. Consider describing the `404` behavior accurately and narrowing "there is no authenticated way to read one back" to the direct detail endpoint (or explicitly documenting the snapshot exception).

### Nitpicks (Optional)

2. **[PERMISSION_SYSTEM.md:439]** The heading says guidelines are "deletable only by moderators," but the sentence immediately below it (and the delete controller) also allow Administrators. Consider changing the heading to "deletable only by Moderators or Administrators" (or "privileged users") so the summary agrees with the documented matrix.

Paul-AUB (dismissed):
Issues (Should Fix)
1. [assets/swaggerV1.yaml:8855] GuidelineDetail.reviewer places nullable: true beside $ref. Under OpenAPI 3.0, Reference Object siblings are ignored, so the schema still rejects reviewer: null ([OpenAPI 3.0 specification](https://spec.openapis.org/oas/v3.0.0.html#reference-object)). This is a normal response: guideline creation leaves reviewer unset and the converter preserves null. Please use an OpenAPI-3.0-compatible nullable reference and correct the copied pattern in Guideline.reviewer at line 8775 as well. A detail-endpoint test covering a null reviewer would protect the contract.

PR #1805, opened 2026-09-08
feat(organization): surface authored and published documents separately

ClemRz:
Organization pages could not show the organization's documents, and cavers showed a truncated document count as if it were the total. Part of #1729 (item F).

## Why

- **The organization payload returned neither list.** `getPopulatedOrganization` merged documents the organization *authored* (`j_document_grotto_author`) with documents it *published* (`t_document.id_editor`), then replaced the merged set with the Collections *above* those documents. Clan des Tritons (695) authored 76 documents and published 1954; the API returned 5 unrelated Collections. G.S. Vulcain: 34 / 563, 2 returned.
- **Authoring and publishing are different editorial relations.** An organization that wrote a topo is not the same as one that published someone else's, so merging them loses meaning even once the list is correct.
- **Unbounded lists don't scale.** The FFS publishes ~2000 documents. Each list is now a 10-item preview paired with an unbounded count, so the UI can render "10 of 1954" and link out to a filtered advanced search — both filters (`authorsOrganization.name`, `editor.name`) are already live in the documents index.
- **Cavers had the same counting problem.** `documents` is capped at 10, so the frontend badge derived from `documents.length` under-reported anyone with more.

## What

- `documents` → `authoredDocuments` + `publishedDocuments`, each capped at 10 and ordered most-recent-first, hydrated with citation fields.
- New `authoredCount` / `publishedCount` on organizations, `authoredCount` on cavers. Cavers get no published counterpart — `TDocument.editor` is organization-only.
- Deleted documents are filtered in the query rather than after the cap, so they can't consume a preview slot.
- Swagger schemas for both entities.

Two fixes found while verifying, both pre-existing:

- `GrottoService.updateInSearch` spreads leftover fields into the Typesense payload; the old rollup already leaked there and the new fields would too. Now destructured out, with a regression test.
- `organization/delete.js` permanent-delete merge read the populated `documents`, which held Collection ancestors — it handed the survivor the wrong document set and then wiped the real `j_document_grotto_author` rows. Now reads the association uncapped.

`getCollectionAncestors` is dropped as unused.

## Breaking change

`documents` is gone from the organization payload. **Must ship with the companion `grottocenter-front` PR in the same release** — that PR splits the single `Collections` section into authored/published, sources counts from the server, and switches the person badge to `authoredCount`.

## Related

- Issue: #1729 (item F)

### DIFF
--- a/api/controllers/v1/organization/delete.js
+++ b/api/controllers/v1/organization/delete.js
@@ -112,10 +112,16 @@ module.exports = async (req, res) => {
       await JGrottoCaveExplorer.destroy({ grotto: organizationId });
     }
 
-    if (organization.documents.length > 0) {
+    // Authorship links must be read uncapped and straight from the association:
+    // the populated organization only carries a preview page of the documents it
+    // authored, which would silently drop the rest of the authorship rows.
+    const { documents: authoredDocuments } =
+      await TGrotto.findOne(organizationId).populate('documents');
+
+    if (authoredDocuments.length > 0) {
       if (shouldMergeInto) {
         const existingDocuments = mergeIntoEntity.documents.map((e) => e.id);
-        const documentsToAdd = organization.documents
+        const documentsToAdd = authoredDocuments
           .map((e) => e.id)
           .filter((e) => !existingDocuments.includes(e));
         await TGrotto.addToCollection(mergeIntoId, 'documents', documentsToAdd);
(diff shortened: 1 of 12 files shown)


Paul-AUB (dismissed):

1. **[api/controllers/v1/organization/delete.js:118]** Consider adding a permanent-delete merge regression test where the source organization authored more than 10 documents and the survivor already shares at least one of them. This path now correctly bypasses the capped preview and filters collisions, but that subtle data-loss fix is not currently exercised by the changed tests.

PR #1807, opened 2026-09-09
fix(guideline): allow privileged users to retrieve soft-deleted guidelines

### LINKED ISSUE #1804: Allow privileged users to retrieve soft-deleted guidelines
## Problem

`DELETE /api/v1/guidelines/{id}` performs a reversible soft delete and returns
the guideline with `isDeleted: true`. The frontend then invalidates the active
detail query, as it does for entrances, massifs, documents, and organizations.

However, `GET /api/v1/guidelines/{id}` currently returns `404` for every
soft-deleted guideline, including requests made by a Moderator or
Administrator. This behavior was specified in #1782, but it prevents the
standard deleted-entity workflow:

1. The soft delete succeeds with `200`.
2. The detail query is refetched.
3. The refetch returns `404`.
4. The frontend can no longer reliably display the deleted state or expose the
   restore action, especially after a page reload.

The current frontend works around this with an `isDeleted=true` UI query
parameter and stale cached data. The URL parameter is not sent to the API, so a
reload still receives `404` and cannot reload the guideline details.

This differs from the established behavior of other soft-deletable entities,
whose detail endpoints remain readable by privileged users after deletion.

## Proposed API behavior

Keep `GET /api/v1/guidelines/{id}` public for active guidelines, but make its
soft-deleted behavior depend on the authenticated user's permissions:

- Active guideline: return `200` as today, with or without authentication.
- Soft-deleted guideline requested by a Moderator or Administrator: return
  `200` with the hydrated `GuidelineDetail` response and `isDeleted: true`.
- Soft-deleted guideline requested anonymously or by an unprivileged user:
  keep returning `404`, preserving the current public data-exposure contract.
- Missing guideline: return `404` for everyone.

No `includeDeleted` query parameter should be required: the endpoint can infer
the allowed representation from the optional authentication token, consistently
with the detail endpoints for entrances and other soft-deletable entities.

## Acceptance criteria

- [ ] An active guideline remains publicly readable without a JWT.
- [ ] A Moderator can retrieve a soft-deleted guideline and receives `200` with
      hydrated relations and `isDeleted: true`.
- [ ] An Administrator can retrieve a soft-deleted guideline and receives the
      same response.
- [ ] An anonymous or otherwise unprivileged caller receives `404` for a
      soft-deleted guideline.
- [ ] A genuinely missing guideline still returns `404` for every caller.
- [ ] A delete followed by a privileged detail GET and restore completes
      without an intermediate `404`.
- [ ] The OpenAPI description documents the role-dependent behavior.
- [ ] Integration tests cover active, soft-deleted, privileged, unprivileged,
      and missing cases.

## Frontend impact

Once this contract is available, the frontend can remove its `isDeleted=true`
URL workaround, remain on the canonical guideline detail URL, refetch after a
soft delete, and render the same deleted-state/restore UI pattern used for an
entrance.

ClemRz:
## 🤔 What

Makes `GET /api/v1/guidelines/{id}` reveal soft-deleted guidelines to privileged callers instead of 404ing everyone.

- Active guideline: `200`, with or without a token (unchanged)
- Soft-deleted, caller is Moderator **or** Administrator: `200` with the hydrated `GuidelineDetail` and `isDeleted: true`
- Soft-deleted, anonymous or unprivileged: `404` (unchanged)
- Missing: `404` for everyone

Closes #1804.

## 🤷‍♂️ Why

The endpoint returned `404` for every soft-deleted guideline — including to the roles that had just deleted it. That broke the standard deleted-entity flow: the soft delete succeeds with `200`, the front end invalidates the detail query, the refetch `404`s, and after a reload there's no way to render the deleted state or expose the restore action.

The front end currently works around this with an `isDeleted=true` UI query parameter plus stale cached data. The parameter never reaches the API, so a reload still `404`s.

Every other soft-deletable entity already keeps its detail endpoint readable by privileged users after deletion. This aligns guidelines with that.

## 🔍 How

`guideline/find.js` reads the optional `req.token` — `parseAuthToken` already populates it on every route when a valid bearer is sent — and falls through to the normal response when the caller holds either role.

Two deliberate departures from the core-content `find` controllers:

- **Checks Moderator and Administrator independently**, not `MODERATOR` alone. Roles aren't hierarchical here, and guideline delete/restore both accept either role, so a Moderator-only gate would let an Administrator soft-delete a guideline and then be unable to read it back — the same bug this fixes, one role over.
- **Returns the full hydrated representation**, not the reduced `toDeletedEntity` shape, following `device/find.js` and `sensor-configuration/find.js`. The front end needs the whole record to render the deleted state and restore action after a reload.

Unprivileged `404`s reuse the missing-id message, so the deleted row's existence isn't leaked.

Scope was kept tight — no policy, service or converter changes were needed. Adding `tokenAuth` would break the public read; `getGuidelineDetail` never filtered `isDeleted`; `toSimpleGuideline` already emits it. No `includeDeleted` parameter.

**Not changed, deliberately:** `PATCH` and `rollback` still `404` on a deleted row for every role. `change_guideline()` only emits an `update` row in `t_last_change` when `is_deleted = false`, so an in-place edit of a deleted guideline would be silently unaudited — and `rollback` has no role check at all, so relaxing it would let a plain user mutate content they can't read. Recorded in `PERMISSION_SYSTEM.md` rather than changed.

## 🧪 Testing

`npm test` — Guidelines folder is **66 passing, 0 failing** (10 new cases).

Full suite shows 4 failures in `Changes/get-recent-comment-relevance-swap.test.js`. **These pre-date this branch**: with the changes stashed, shard 3 is 354 pass / 4 fail; with them, 363 pass / 4 fail — same file, same four tests. That file passes in isolation on clean `develop`, so it's a shard-ordering flake unrelated to this work.

New coverage: anonymous / plain User / Leader all `404`; Moderator and Administrator each `200` with full relations; missing id `404`s for privileged callers too; malformed id still `404`s for a Moderator (pinning that `validateId` runs ahead of the new branch); and the hidden-deleted and missing `404` bodies carry the same message shape, so a refactor can't start leaking existence.

Two notes on test construction:

- The deleted fixture is soft-deleted via `TGuideline.destroyOne` rather than created with `isDeleted: true`, so `histo_delete` runs for real. That makes the test prove the junction rows survive a soft delete — the implicit contract `delete.js` relies on when it echoes pre-delete associations back, previously untestable through the API.
- A round-trip block runs delete → privileged GET → anonymous GET → restore → public GET over HTTP, so the DB trigger and `RecentChangeService` run for real. Step two is the request that used to `404`.

## 📸 Previews

n/a — API only.

## Follow-ups

- **Front end:** can now drop the `isDeleted=true` URL workaround and stay on the canonical detail URL. One caveat: `delete.js` and `restore.js` respond with `toSimpleGuideline` (bare ID strings for countries/regions) while the detail GET returns `toGuideline` (hydrated objects) — so the client must *invalidate* the detail query, not merge the mutation response into cache, or it will clobber the hydrated relations.
- **Docs debt found while auditing, not fixed here:** inconsistency item #3 in `PERMISSION_SYSTEM.md` says `cave/find.js` "forces `isDeleted = false` on list queries" in a way that hides the row. It doesn't — `getPopulatedCave(caveId, subEntitiesWhere)` applies that filter to *sub-entities*, so an anonymous caller gets `200` with a reduced stub, not a `404`. Verified by probe. The wording pre-dates this branch; I dropped the claim from my own additions rather than widen scope. Worth its own issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/PERMISSION_SYSTEM.md
+++ b/PERMISSION_SYSTEM.md
@@ -13,7 +13,8 @@ Roles are **not hierarchical**; for instance, an Administrator does not automati
 - **Permissions**:
   - View public cave/entrance/document/organisation/massif/person data
   - Search through cave/entrance/document/organisation/massif/person/device data
-  - View legislation guidelines (list, single guideline by id, by geographic entity, and snapshots)
+  - View active legislation guidelines (list, single guideline by id, by geographic entity, and snapshots) — a
+    soft-deleted guideline yields 404 by id, though its text stays readable through snapshots
   - View the organizations responsible for a country/region/massif
   - View complete entity history
   - View statistics
@@ -71,6 +72,7 @@ Roles are **not hierarchical**; for instance, an Administrator does not automati
     - **Permanently** delete those same entities — the permanent path reuses the Moderator check with no additional
       Administrator gate (see "Known Permission Inconsistencies")
     - Delete/restore legislation guidelines (permanent deletion of these requires Administrator)
+    - View a soft-deleted legislation guideline by id, with the full detail representation
     - Delete/restore devices and sensor configurations (permanent deletion of these requires Administrator)
     - Update any device or sensor configuration (regardless of ownership)
     - Permanently delete author records (`type: AUTHOR`; deleting a `CAVER` account requires Administrator instead —
@@ -108,6 +110,8 @@ Roles are **not hierarchical**; for instance, an Administrator does not automati
   - **Content Moderation** (guideline deletion only; other content moderation is Moderator-only):
     - Delete/restore legislation guidelines — the only guideline operation an Administrator gates, since updating and
       rolling back them is open to any authenticated user
+    - View a soft-deleted legislation guideline by id — the only soft-deleted content an Administrator can read without
+      also holding the Moderator role (see "Known Permission Inconsistencies" #3)
   - **Sensitive Data Management**:
     - View coordinates of sensitive entrances
     - Remove sensitive flag from entrances
@@ -133,7 +137,7 @@ Roles are **not hierarchical**; for instance, an Administrator does not automati
 | Search content (including devices)                            | ✅ | ✅ | ✅ | ✅ | ✅ |
 | View statistics                                               | ✅ | ✅ | ✅ | ✅ | ✅ |
 | View history/snapshots                                        | ✅ | ✅ | ✅ | ✅ | ✅ |
-| View guidelines (list/by-id/by-entity/snapshots)              | ✅ | ✅ | ✅ | ✅ | ✅ |
+| View active guidelines (list/by-id/by-entity/snapshots)       | ✅ | ✅ | ✅ | ✅ | ✅ |
 | View responsible organizations of country/region/massif       | ✅ | ✅ | ✅ | ✅ | ✅ |
 | **Content Creation**                                          |
 | Create caves/entrances/documents/organisations/massifs        | ❌ | ✅ | ✅ | ✅ | ✅ |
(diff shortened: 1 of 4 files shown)


Paul-AUB (dismissed):

1. **[assets/swaggerV1.yaml:3544]** Consider declaring bearer authentication as optional for this operation (for example, with an empty security requirement alongside `bearerAuth`). The description now explains that a Moderator or Administrator can retrieve a soft-deleted guideline, but without an operation-level security requirement OpenAPI tooling treats this GET as unauthenticated and Swagger UI/generated clients may omit the configured bearer token. Expressing both anonymous and bearer-authenticated alternatives would make the machine-readable contract match the documented behavior.

PR #1813, opened 2026-09-21
fix(geoloc): cap the bounding box area on GET /geoloc/entrances

### LINKED ISSUE #1810: bug(api): unbounded bbox on GET /api/v1/geoloc/entrances blocks the event loop and degrades every route
**Describe the bug**

`GET /api/v1/geoloc/entrances` accepts an unbounded bounding box. `GeoLocService.checkAndGetCoordinatesParams` (`api/services/GeoLocService.js:300`) validates only that latitudes fall in `[-90, 90]` and longitudes in `[-180, 180]`, so a whole-planet request is considered valid:

```
GET /api/v1/geoloc/entrances?sw_lat=-90&sw_lng=-180&ne_lat=90&ne_lng=180
```

The controller then runs `PUBLIC_ENTRANCES_IN_BOUNDS` with `LIMIT 100000` (`api/controllers/v1/geoloc/find-entrances.js`), which currently matches **133,967** entrances. Postgres returns that in ~1.3s, but the API takes 5–20s: the remaining time is Node marshalling and JSON-serializing ~134k rows × 25 columns. That work is CPU-bound and synchronous, and the App Service is configured with `numberOfWorkers = 1`, so there is a single event loop for the entire site.

While one of these requests serializes, **every other request queues behind it**. This is head-of-line blocking, not a slow-query problem.

**To Reproduce**

1. `GET /api/v1/geoloc/entrances?sw_lat=-90&sw_lng=-180&ne_lat=90&ne_lng=180`
2. Concurrently issue any trivial request, e.g. `GET /api/v1/languages`
3. Observe the trivial request's latency inflate by seconds

**Expected behavior**

Requests with an unreasonably large bounding box should be rejected with `400` (as the massif polygon path already does — see commit `c40da25f`, "feat(massif): enforce polygon area limit"), or the response should be capped/paginated so that no single request can monopolise the event loop.

**Evidence**

Latency by bounding-box area, `GET /api/v1/geoloc/entrances`, last 3 days (Azure Log Analytics, `AppServiceConsoleLogs`):

| bbox area | n | p50 | p95 | p99 |
|---|---|---|---|---|
| tiny <0.01 | 6482 | 43ms | 424ms | 1036ms |
| small <1 | 2808 | 41ms | 174ms | 310ms |
| med <10 | 323 | 56ms | 1534ms | 2117ms |
| large <100 | 468 | 384ms | 5243ms | 5805ms |
| xl <1000 | 478 | **5097ms** | 9569ms | 11630ms |
| world >=1000 | 444 | 1437ms | 12507ms | 15156ms |

Impact on **unrelated** endpoints, bucketed by how many large-bbox (`area >= 10`) geoloc requests occurred in the same minute. Big-geoloc requests are excluded from these percentiles entirely:

| large-bbox req/min | minutes | median p50 | median p95 | p90 of p95 | % over 2s |
|---|---|---|---|---|---|
| 0 | 4944 | 137ms | 501ms | 713ms | 0.11% |
| 1–2 | 653 | 127ms | 476ms | 1054ms | 0.78% |
| 3–5 | 152 | 112ms | 486ms | **4441ms** | **2.47%** |

Medians are unchanged while the tail moves 6× and 2s-violations move 22× — the signature of event-loop blocking rather than database contention. Consistently, *every* route's p99 degraded, including `GET /api/v1/languages` (358ms → 423ms), a trivial lookup that can only slow down if the process itself is busy.

Confirmed with `EXPLAIN (ANALYZE, BUFFERS)` on production for the world bbox: **1309ms** execution, all buffers `shared hit`. The database is not the bottleneck.

**Knock-on effect: false Sev1 alerts**

Slow responses cause clients to give up. Application Insights records the aborted requests with `ResultCode 0` and:

```
"error": "The request has been aborted and the network socket has closed."
```

193 such aborts in 4 days, 86 of them on `/geoloc/entrances`. The App Service front end counts aborted connections toward the `Http5xx` platform metric, which trips the `grottocenter api 500 errors alert` rule (`Http5xx >= 1` over 15 min, Sev1) — even though the application never returns a 500 (`AppServiceConsoleLogs` contains **zero** `Sending 500` lines over the same window, and `AppExceptions` is empty).

Example burst, 2026-09-21T18:13 (3 × `Http5xx`, one alert email):

```
18:13:11.393  Req GET /geoloc/entrances?sw_lat=45.1875... (okhttp/5.1.0, 176.146.228.0:63428)  -> no response
18:13:11.426  Req  (identical URL, same client:port)                                            -> no response
18:13:11.452  Req  (identical URL, same client:port)                                            -> 200 in 40ms
```

**Additional context**

- Service: `api/services/GeoLocService.js` (`checkAndGetCoordinatesParams` at :300, `PUBLIC_ENTRANCES_IN_BOUNDS` at :30)
- Controller: `api/controllers/v1/geoloc/find-entrances.js`
- Not a regression from a deploy: no commits between 2026-09-15 and 2026-09-21, and `GeoLocService.js` was last modified 2026-06-28. The trigger was a traffic-mix change — world-scale bbox requests rose from ~15 per 12h to 113–192 starting 2026-09-19T12:00, almost entirely from a single client.
- The sibling `LEFT JOIN LATERAL` on `v_data_quality_compute_entrance` is **not** the main cost: it is a materialized view with a usable index (0.002ms/row), contributing ~550ms of the 1.3s.
- Related: the same query's sort spills to disk (separate issue).

ClemRz:
Rejects bounding boxes larger than 35 000 km² on `GET /api/v1/geoloc/entrances` with a `400 BBOX_AREA_EXCEEDED`, before any query runs.

## Why

- **One request could stall the whole API.** An unbounded bounding box matches ~134 000 entrances. Postgres returns that set in ~1.3s, but the endpoint took 5–40s — the rest is Node marshalling and JSON-serialising 134k rows × 25 columns synchronously, on a single event loop (`numberOfWorkers = 1`).
- **The blocking is measurable on unrelated routes.** Bucketing other endpoints by how many large geoloc requests shared the same minute: 0 large → p95 713ms; 3–5 large → p95 4441ms, with medians unchanged. Tail moves 6×, medians flat — head-of-line blocking, not database contention.
- **It was firing the Sev1 alert.** Slow responses make clients give up; App Service counts aborted connections toward the `Http5xx` platform metric. 193 aborts in 4 days, 86 on this endpoint, while the application returned zero 500s.
- **Rate limiting does not cover it.** `rateLimiter.js` allows 200 requests per 10 minutes per IP; the observed 3–5 world-scale requests per minute sits far inside that budget.
- **No production client is affected.** All three callers verified in `grottocenter-front`: the web map fetches one z12 tile per request (67–95 km², ~370–520× headroom), `MapMassif` is gated on zoom ≥ 13 (~1 300 km²), and duplicate detection uses a 1 km radius (~4 km²). Server-side, `grottocenter.org` (12 530 requests) and the mobile app (1 224) have zero requests over the cap.

## What

- `api/utils/computeBoundingBoxAreaKm2.js` — exact spherical rectangle area, pure and synchronous.
- `GeoLocService.validateBoundingBoxArea()` — returns `null | { code, message }`, mirroring `MassifService.validatePolygon`, and logs one `warn` on rejection (`log-response.js` would otherwise record a 400 only at `verbose`).
- One call in `find-entrances.js`, placed before the massif lookup so an oversized box never pays for a database round trip first. Consequence: an oversized box plus an unknown `massif` id now returns 400 rather than 404 — pinned by a test.
- Swagger: the `400` response, and a note that inverted longitudes are normalised rather than wrapped.
- Tests: a unit suite and 7 fast-check properties for the helper, a service block, 10 route tests, and a regression block asserting the other 11 geoloc route names still accept world bounds.

Two details worth a reviewer's attention:

**The cap is deliberately not in `checkAndGetCoordinatesParams`.** That function is the first statement of all eight geoloc controllers, and the web app requests world bounds from four of the others on every map page load. Putting the check there would break the production map.

**The area is not wrap-aware, on purpose.** The query filters with planar `ST_Within(point_geom, ST_MakeEnvelope(...))`. Verified on PostGIS 3.4.3: `ST_MakeEnvelope(170,-10,-170,10,4326)` yields `POLYGON((170 -10,170 10,-170 10,-170 -10,170 -10))`, and `ST_Within` accepts `POINT(0 0)` while rejecting `POINT(175 0)` — a 340°-wide box of ~83.6M km², not a 20° strip. A wrap-aware `Δλ` would understate the most expensive requests by 17× and let them through. The area is also in km² rather than degrees², because the polar `useNearbyEntrances` box is 6.5 degrees² but only ~3 km² of real surface and must keep working.

## Related

- Closes #1810
- First commit is an unrelated one-line docs correction (`POST /massifs` still cited the pre-`d33eb56b` limit); it is separated so it can be read on its own.

## Follow-ups, not in this PR

- `MapMassif.jsx` calls `.wrap()` per corner, so panning a massif map across ±180 produces `|Δλ| ≈ 359.6°` → 400. Degrades gracefully (markers vanish, no crash) and needs a massif straddling the antimeridian. Frontend issue.
- Coordinate params are never validated as numeric, and `?sw_lat=` coerces to 0. Affects all eight geoloc endpoints.
- `work_mem` sort spill on `PUBLIC_ENTRANCES_IN_BOUNDS` — this cap shrinks the sort input and may remove it without server tuning.


Paul-AUB (approved):
Reviewed the critical request path, bounding-box calculation, API contract, and focused test coverage. No must-fix issues found.

PR #1814, opened 2026-09-22
Feature/check-authentification

YohannSchatt:
## 🤔 What

Add endpoint 
/api/v1/auth/check

## 🤷‍♂️ Why

For the AI server, verify whether the user is connected and able to use it.

## 🔍 How

New endpoint, with rule of AuthToken


ClemRz:
Why not using `POST /api/v1/login` instead?

